PR #1, opened 2020-07-21
Dynamic script icons

invalidCards:
Puts down a framework for easily creating icons, based on [Remix Icons](https://github.com/Remix-Design/RemixIcon). The CSS isn't my proudest work ever, and I don't expect this to be pulled in its current form, but I am open to suggestions 😄


docs/framework.md
```diff
@@ -30,7 +30,7 @@ This path is automatically fed into `browser.runtime.getURL()`, so should start
 
 ## Metadata files
 
-Each module should be accompanied by a `.json` file of matching name; i.e. `example.js` should be accompanied by `example.json` in the same directory level. Right now, only two keys are supported; expect more to be added in future.
+Each module should be accompanied by a `.json` file of matching name; i.e. `example.js` should be accompanied by `example.json` in the same directory level. Right now, only five keys are supported; expect more to be added in future.
```
AprilSylph, line 33:
```suggestion
Each module should be accompanied by a `.json` file of matching name, i.e. `example.js` should be accompanied by `example.json` in the same directory level. Right now, four keys are supported.
```
(i know the `"preferences"` documentation is missing but the number may as well be accurate in the meantime)

src/browser/action.js
```diff
@@ -49,17 +49,31 @@ async function renderScripts() {
...
+      iconInner.classList.add(icon, 'ri-fw');
+      iconInner.style.color = iconForeground;
+      icondiv.appendChild(iconInner);
+      fieldset.appendChild(icondiv);
+    }
+
     if (description) {
       const p = document.createElement('p');
       p.textContent = description;
+      if (icon) {
```
AprilSylph, line 74:
```suggestion
      if (icon.class_name !== undefined) {
```

AprilSylph:
is RemixIcons available as an npm package? if it is, it would be great to add it as a dependency and add it to `dev/copy-libs.sh`

invalidCards:
[Yes it is](https://www.npmjs.com/package/remixicon)

AprilSylph:
everything LGTM save the CSS, i'll try playing with it soon to see how it can be cleaned up!

PR #2, opened 2020-08-10
New script: Painter

invalidCards:
Make nice coloured edges at the top of your post to see quickly what kinda thing you're looking at.

### DIFF
--- a/src/browser/action.js
+++ b/src/browser/action.js
@@ -144,6 +144,11 @@ async function renderScripts() {
             preferenceListItem.appendChild(preferenceLabel);
             preferenceListItem.appendChild(preferenceInput);
             break;
+          case 'color':
+            preferenceInput.value = savedPreference;
+            preferenceListItem.appendChild(preferenceInput);
+            preferenceListItem.appendChild(preferenceLabel);
+            break;
           case 'select':
             for (const [value, text] of Object.entries(preference.options)) {
               const option = document.createElement('option');
--- a/src/browser/action.js
+++ b/src/browser/action.js
@@ -176,9 +176,7 @@ async function renderScripts() {
       showButtons: false,
       allowEmpty: true
     });
-    $makeSpectrum.on('change.spectrum', (e) => {
-      writePreference(e);
-    });
+    $makeSpectrum.on('change.spectrum', writePreference);
   });
 }
 
--- a/src/browser/action.js
+++ b/src/browser/action.js
@@ -168,15 +168,17 @@ async function renderScripts() {
 
     scriptsSection.appendChild(fieldset);
 
-    $(fieldset).find('.makeSpectrum').spectrum({
-      preferredFormat: "hex",
+    const $makeSpectrum = $(fieldset).find('.makeSpectrum');
+    
+    $makeSpectrum.spectrum({
+      preferredFormat: 'hex',
       showInput: true,
       showButtons: false,
       allowEmpty: true
     });
-    $(fieldset).find('.makeSpectrum').on('change.spectrum', (e) => {
+    $makeSpectrum.on('change.spectrum', (e) => {
       writePreference(e);
-    })
+    });
   });
 }
 
--- a/src/browser/action.js
+++ b/src/browser/action.js
@@ -120,7 +120,12 @@ async function renderScripts() {
         preferenceInput.addEventListener('input', writePreference);
 
         if (inputType === 'input') {
-          preferenceInput.type = preference.type;
+          if (preference.type === 'color') {
+            preferenceInput.type = 'text';
+            preferenceInput.classList.add('makeSpectrum');
+          } else {
+            preferenceInput.type = preference.type;
+          }
         }
 
         const preferenceLabel = document.createElement('label');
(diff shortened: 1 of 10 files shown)


src/scripts/painter.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const paint = async function() {
+    const { timelineObject } = await fakeImport('/src/util/react-props.js');
+
+    [...document.querySelectorAll('[data-id]:not(.xkit_painter_painted)')]
+    .forEach(async postElement => {
+      postElement.classList.add('xkit_painter_painted');
+
+      const post_id = postElement.dataset.id;
+      const { canDelete, liked, rebloggedFromId } = await timelineObject(post_id);
```
AprilSylph, line 15: nitpick, but for regular object destructuring i'm preferring no spaces around the identifiers; i'm using spacing around to specifically denote module syntax

```suggestion
      const {canDelete, liked, rebloggedFromId} = await timelineObject(post_id);
```

src/scripts/painter.js
```diff
@@ -0,0 +1,74 @@
...
+
+      if (!coloursToApply.length) {
+        return;
+      }
+
+      const step = 100/coloursToApply.length;
+      let borderImage = 'linear-gradient(to right';
+      for (let i = 0; i < coloursToApply.length; i++) {
+        borderImage += `, ${coloursToApply[i]} ${step*i}% ${step*(i+1)}%`;
+      }
```
AprilSylph, line 42: i think this would be nicer as a `.forEach()`

```suggestion
      coloursToApply.forEach((colour, i) => 
        borderImage += `, ${colour} ${step*i}% ${step*(i+1)}%`
      );
```

src/scripts/painter.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const { postListener } = await fakeImport('/src/util/mutations.js');
+    postListener.addListener(paint);
+    paint();
+  }
+
+  const clean = async function() {
+    const { postListener } = await fakeImport('/src/util/mutations.js');
+    postListener.removeListener(paint);
+    $('.xkit_painter_painted article').css('border-top', '').css('border-image-source', '').css('border-image-slice', '');
```
AprilSylph, line 67: i'd prefer this to be split across multiple lines

```suggestion
    $('.xkit_painter_painted article')
    .css('border-top', '')
    .css('border-image-source', '')
    .css('border-image-slice', '');
```

src/scripts/painter.js
```diff
@@ -0,0 +1,74 @@
...
+    $('.xkit_painter_painted').removeClass('xkit_painter_painted');
+  }
+
+  const stylesheet = '/src/scripts/painter.css';
+
+  return { main, clean, stylesheet };
```
AprilSylph, line 73: no need to add an empty stylesheet; if a script doesn't specify a stylesheet at all, `xkit.js` doesn't attempt to find it
(this is also a request to delete `src/scripts/painter.css`)

```suggestion
  return { main, clean };
```

src/scripts/painter.json
```diff
@@ -0,0 +1,27 @@
...
+      "label": "Original post colour"
+    },
+    "reblog": {
+      "type": "text",
+      "label": "Reblogged post colour"
+    },
+    "liked": {
+      "type": "text",
+      "label": "Liked post colour"
+    }
```
AprilSylph, line 25: just tested this out, and all these preference values default to the string `"undefined"` for... probably obvious reasons

it [is required that preferences have a `"default"` value](https://github.com/AprilSylph/XKit-Rewritten/blob/master/docs/framework.md#preferences-preference-name-default), so it should probably just be an empty string for each preference here

```suggestion
    "own": {
      "type": "text",
      "label": "Own post colour",
      "default": ""
    },
    "original": {
      "type": "text",
      "label": "Original post colour",
      "default": ""
    },
    "reblog": {
      "type": "text",
      "label": "Reblogged post colour",
      "default": ""
    },
    "liked": {
      "type": "text",
      "label": "Liked post colour",
      "default": ""
    }
```

invalidCards:
i'm gonna die trying

src/browser/action.js
```diff
@@ -156,6 +167,16 @@ async function renderScripts() {
...
+
+    $(fieldset).find('.makeSpectrum').spectrum({
+      preferredFormat: "hex",
+      showInput: true,
+      showButtons: false,
+      allowEmpty: true
+    });
+    $(fieldset).find('.makeSpectrum').on('change.spectrum', (e) => {
+      writePreference(e);
+    })
```
AprilSylph, line 179: would be nice to save `.makeSpectrum` instead of calling the jQuery constructor twice

```suggestion
    const $makeSpectrum = $(fieldset).find('.makeSpectrum');
    
    $makeSpectrum.spectrum({
      preferredFormat: 'hex',
      showInput: true,
      showButtons: false,
      allowEmpty: true
    });
    $makeSpectrum.on('change.spectrum', (e) => {
      writePreference(e);
    });
```

also, can't you just pass in `writePreference` directly here?

PR #4, opened 2020-08-15
New script: Quick Queue

invalidCards:
Adds a button to each post, queueing the post if possible, and drafting if not.

### DIFF
--- a/src/scripts/quick_queue.js
+++ b/src/scripts/quick_queue.js
@@ -21,12 +21,10 @@
 
     const {uuid: defaultBlogUuid = ''} = await fetchDefaultBlog();
 
-    const {blog, content, layout, reblogKey} = await timelineObject(post_id);
+    const {blog, reblogKey} = await timelineObject(post_id);
     try {
       const response = await apiFetch(`/v2/blog/${defaultBlogUuid}/posts`, {method: 'POST',
         body: {
-          content,
-          layout,
           state: 'queue',
           parent_tumblelog_uuid: blog.uuid,
           parent_post_id: post_id,
@@ -44,8 +42,6 @@
         try {
           const draftResponse = await apiFetch(`/v2/blog/${defaultBlogUuid}/posts`, {method: 'POST',
             body: {
-              content,
-              layout,
               state: 'draft',
               parent_tumblelog_uuid: blog.uuid,
               parent_post_id: post_id,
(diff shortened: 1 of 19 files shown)


src/scripts/quick_queue.js
```diff
@@ -0,0 +1,125 @@
+(function() {
+  let controlsSelector;
+
+  const quickQueue = async function(event) {
+    const postElement = event.target.closest('div[data-id]');
+    if (!postElement) {
+      event.target.innerHTML = '<path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636z"/>';
```
AprilSylph, line 7: i think all these paths would be better as named constants available to the entire module, both for source-labelling purposes and for potential reduced repetition (it looks like this path in particular is reused at least once?)

src/util/user-blogs.js
```diff
@@ -0,0 +1,36 @@
...
+    const blogs = await fetchUserBlogs();
+    return blogs.map(blog => blog.name);
+  };
+
+  /**
+   * @return {String} - The default ("main") blog for the user
+   */
+  const fetchDefaultBlog = async function() {
+    const blogs = await fetchUserBlogs();
+    return blogs.filter(blog => blog.primary === true)[0];
```
AprilSylph, line 32:
```suggestion
    return blogs.find(blog => blog.primary === true);
```

AprilSylph (changes requested):

src/scripts/quick_queue.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const { apiFetch } = await fakeImport('/src/util/tumblr-helpers.js');
+    const { timelineObject } = await fakeImport('/src/util/react-props.js');
+    const { fetchDefaultBlog } = await fakeImport('/src/util/user-blogs.js');
+
+    const {name: defaultBlog = ''} = await fetchDefaultBlog();
+
+    const {blog, content, layout, reblogKey} = await timelineObject(post_id);
+    try {
+      const response = await apiFetch(`/v2/blog/${defaultBlog}.tumblr.com/posts`, {method: 'POST',
```
AprilSylph, line 22: does `/v2/user/info` not give us UUIDs? it'd be nice to use those instead of .tumblr.com identifiers
(also, even if it doesn't, can't we just do `/v2/blog/${defaultBlog}/posts`? i thought non-URL identifiers were always treated as usernames)

AprilSylph (changes requested):
it would be super nice (and entirely possible, i'm sure) to have the Quick Queue button between the reply and reblog buttons, like the "Fast Queue" button is on the mobile app

invalidCards:
Closing this in favour of two PRs so that the user blog utilities can be merged without needing to wait on Tumblr documentation.

PR #5, opened 2020-08-24
Re-add documentation on the color preference type

invalidCards:
I *think* this is all it needs in terms of documentation, but then again I implemented the thing, so it may not be as obvious to outsiders.

### DIFF
--- a/docs/framework.md
+++ b/docs/framework.md
@@ -102,7 +102,7 @@ For `"select"`-type preferences, an object of value/label entries. Unused for ot
 Default value of the preference to display to the user. This does not automatically set the default value of the preference in storage; scripts should ensure to use the same defaults as they display.
 
 If the preference `type` is `"checkbox"`, this value should be a boolean.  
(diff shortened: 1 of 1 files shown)


docs/framework.md
```diff
@@ -103,6 +103,7 @@ Default value of the preference to display to the user. This does not automatica
 
 If the preference `type` is `"checkbox"`, this value should be a boolean.  
 If the preference `type` is `"text"`, this value should be a string.  
+If the preference `type` is `"color"`, this value should either be a string representing a hexadecimal colour code (`#1a2b3c`) or an empty string.
```
AprilSylph, line 106:
```suggestion
If the preference `type` is `"color"`, this value should either be a string representing a hexadecimal colour code (i.e. `"#1a2b3c"`) or an empty string.  
```

AprilSylph (changes requested):

AprilSylph (approved):

PR #10, opened 2020-09-15
new script: Search My Stuff

marcustyphoon:
This ports my Search Likes PR from XKit 7. 

It loads posts via endless scrolling and hides all but the desired results. Because this can quickly result in unreasonably large amounts of loaded content (and because I wanted to learn this kind of thing anyway), I made an effort to take performance and responsiveness into account, hence things like rendering posts to the dom in chunks. Yes, this makes the code huge. Sorry :P

It also uses a manual scrape of the react props to get searchable text, which probably isn't worth it. (I was gonna just use mark.js and live with the oddities of finding the note count and etc.)

It's supposed to run on likes, channel, queue, and drafts, but right now it just runs everywhere. (It's sort of hilarious on the dashboard, but this obviously should never be given to the public :D)

I would personally probably close this to reduce PR clutter until more dependencies are implemented, but I'll submit it for the record and that can be up to you.

- [ ] don't run literally everywhere lmao (handle soft refresh/navigation)
- [ ] mark.js
- [ ] use mark.js instead of actually searching manually?
- [ ] use sidebar function when available
- [ ] remove unnecessary jquery
- [ ] double check all asynchronous logic

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -8,6 +8,7 @@
   "one_click_postage",
   "open_in_tabs",
   "painter",
+  "search_boxes",
   "timestamps",
   "tweaks"
 ]
--- a/src/scripts/search_boxes.css
+++ b/src/scripts/search_boxes.css
@@ -1,8 +1,8 @@
-#search-likes-box {
+#xkit-search-box {
 	padding: 10px 10px 28px 10px;
 }
 
-#search-likes-input {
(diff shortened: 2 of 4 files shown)


AprilSylph:
yeah, this should be closed until the core has the features that this script wants to use. nice concept though! being able to search not just likes but your own posts/queue/drafts sounds very useful

PR #21, opened 2020-10-02
[Painter] Add tag highlighting

invalidCards:
#### User-facing changes
- Users may now select a colour for highlighting user-input tags using the Painter extension.
- There is also a functionality for highlighting tags from the source post, in a similar way to how Tumblr's tag filtering works.

#### Technical explanation
If the tag colour is set, the script loops through every tag on the post to see if any in the textarea match it, and applies the colour to the post if so. If this initial search doesn't get a result, and the checkbox for searching the source post is ticked, it will fetch the source post and look through its tags in a similar way, colouring the post if a match is found.

#### Issues this closes
None.

### DIFF
--- a/src/scripts/painter.js
+++ b/src/scripts/painter.js
@@ -3,29 +3,61 @@
   let originalColour;
   let reblogColour;
   let likedColour;
+  let tagColour;
+  let colouredTags;
+  let colourSourceTags;
 
   const excludeClass = 'xkit-painter-done';
 
   const paint = async function () {
     const { getPostElements } = await fakeImport('/src/util/interface.js');
     const { timelineObject } = await fakeImport('/src/util/react_props.js');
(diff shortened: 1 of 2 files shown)


src/scripts/painter.js
```diff
@@ -28,6 +35,31 @@
...
+      if (tagColour) {
+        let tagColourFound = false;
+        for (const tag of tags) {
+          if (tagArray.includes(tag.toLowerCase())) {
+            coloursToApply.push(tagColour);
+            tagColourFound = true;
+            break;
+          }
+        }
+        if (!tagColourFound && colourSourceTags) {
```
AprilSylph, line 47: shouldn't this check for `rebloggedRootId` and `rebloggedRootUuid` before making the request? so that it doesn't make a broken request on original posts

AprilSylph (approved):

PR #22, opened 2020-10-03
Debounce preference input

invalidCards:
#### User-facing changes
- None

#### Technical explanation
All preference writes are debounced by half a second, which means that changing preferences (especially of `text` and `textarea` types) will no longer create restart spam among preferences.

#### Issues this closes
None.

### DIFF
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -36,7 +36,7 @@ const debounce = (callback, ms) => {
   };
 };
 
-const writePreference = debounce(async function (event) {
+const writePreference = async function (event) {
   const { id } = event.target;
   const [scriptName, preferenceType, preferenceName] = id.split('.');
   const storageKey = `${scriptName}.preferences.${preferenceName}`;
@@ -52,7 +52,7 @@ const writePreference = debounce(async function (event) {
       browser.storage.local.set({ [storageKey]: event.target.value });
       break;
   }
-}, 500);
+};
 
 const renderScripts = async function () {
   const installedScripts = await getInstalledScripts();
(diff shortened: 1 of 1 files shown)


AprilSylph (changes requested):
debouncing the entire `writePreference` function will mean that changing multiple preferences rapidly also has a chance of triggering debouncing, thus leading to discarding user input - `writePreference` shouldn't be modified, but instead, each preference should have its own debounced version of it, so that changing only the same preference rapidly results in debouncing.

```diff
-      preferenceInput.addEventListener('input', writePreference);
+      preferenceInput.addEventListener('input', debounce(writePreference));
```

src/browser_action/render_scripts.js
```diff
@@ -110,7 +118,7 @@ const renderScripts = async function () {
 
       const preferenceInput = preferenceTemplateClone.querySelector(inputType);
       preferenceInput.id = `${scriptName}.${preference.type}.${key}`;
-      preferenceInput.addEventListener('input', writePreference);
+      preferenceInput.addEventListener('input', debounce(writePreference, 500));
```
AprilSylph, line 121:
```suggestion
      preferenceInput.addEventListener('input', ['text', 'textarea'].includes(preference.type) ? debounce(writePreference, 500) : writePreference);
```

AprilSylph (changes requested):
works great for text/textarea preferences! but... debouncing other preference types basically just adds a delay, which doesn't feel great to use. i think debouncing should be limited to only preference types which we reasonably expect to receive rapid inputs

AprilSylph (approved):

PR #25, opened 2020-10-09
new script: Mutual Checker

marcustyphoon:
#### User-facing changes
Ports the Mutual Checker extension from XKit 7.

#### Technical explanation
This finds the author of each post, checks if that author is following the user's main blog or a selected one of their sideblogs via an api call, and prepends an icon to the author's name if so.

Posts by mutuals get a `from-mutual` class that other extensions could make use of.

#### Issues this closes
n/a


AprilSylph (changes requested):
my original XKit sin has come back to haunt me...

the filename (and thus also entry in the scripts index) should be snake_case for consistency
(`mutualchecker` &rarr; `mutual_checker`)

src/scripts/mutual_checker.js
```diff
@@ -0,0 +1,111 @@
...
+      const allBlogs = await fetchUserBlogNames();
+      if (allBlogs.includes(selectedBlogChanges.newValue)) {
+        myBlog = selectedBlogChanges.newValue;
+      } else {
+        myBlog = (await fetchDefaultBlog()).name;
+      }
+      removeIcons();
+      addIcons();
+    }
+  };
```
AprilSylph, line 71:
```suggestion
```

src/scripts/mutual_checker.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const main = async function () {
+    browser.storage.onChanged.addListener(onStorageChanged);
+
```
AprilSylph, line 75:
```suggestion
```

src/scripts/mutual_checker.js
```diff
@@ -0,0 +1,111 @@
...
+      icon = aprilFoolsIcon;
+    }
+
+    const { onNewPosts } = await fakeImport('/src/util/mutations.js');
+    onNewPosts.addListener(addIcons);
+    addIcons();
+  };
+
+  const clean = async function () {
+    browser.storage.onChanged.removeListener(onStorageChanged);
```
AprilSylph, line 103:
```suggestion
```

src/scripts/mutual_checker.js
```diff
@@ -0,0 +1,111 @@
...
+    const { onNewPosts } = await fakeImport('/src/util/mutations.js');
+    onNewPosts.removeListener(addIcons);
+
+    removeIcons();
+  };
+
+  return { main, clean, stylesheet: true };
```
AprilSylph, line 110:
```suggestion
  return { main, clean, stylesheet: true, autoRestart: true };
```

AprilSylph (changes requested):
since Mutual Checker only does the one thing, and you're already clearing `mutuals` in `main()`, you could cut out `onStorageChanged()` and just use [`autoRestart: true`](https://github.com/AprilSylph/XKit-Rewritten/blob/master/docs/framework.md#autorestart) here

src/scripts/mutual_checker.json
```diff
@@ -0,0 +1,16 @@
...
+    "color": "white",
+    "background_color": "#777777"
+  },
+  "preferences": {
+    "selectedBlog": {
+      "type": "text",
+      "label": "Blog to check, if not your main",
+      "default": ""
+    }
+  }
```
AprilSylph, line 15: this makes me feel uneasy... i see that you've included some verification logic, but i think simply not obeying a preference is something to be avoided

marcustyphoon: Yeah, 100% agreed, I really don't like it either. Given that the preferences page has (if I understand correctly) no ability to get information from the main extension or the tumblr window, any ideas how to make this work elegantly? Removing the ability to select no-primary blogs seems quite limiting but is definitely an option, making it a sidebar toggle seems a bit silly, and adding buttons to some part of the tumblr UI that lists every one of your blogs... could work maybe?

AprilSylph: for now i think i'd prefer to go with only checking the primary blog. i'm fairly sure i would put sideblog-checking back in eventually when i figure out the most elegant way to do it, but for now it seems more important to get the basic feature in without any attached mess

src/scripts/mutual_checker.js
```diff
@@ -0,0 +1,76 @@
...
+  const aprilFoolsIcon = '<svg class="xkit-mutual-icon" fill="#00b8ff"  xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 1000"><path d="M858 352q-6-14-8-35-2-12-4-38-3-38-6-54-7-28-22-43t-43-22q-16-3-54-6-26-2-38-4-21-2-34.5-8T619 124q-9-7-28-24-29-25-44-34-24-16-47-16t-47 16q-15 9-44 34-19 17-28 24-16 12-29.5 18t-34.5 8q-12 2-38 4-38 3-54 6-28 7-43 22t-22 43q-3 16-6 54-2 26-4 38-2 21-8 34.5T124 381q-7 9-24 28-25 29-34 44-16 24-16 47t16 47q9 15 34 44 17 19 24 28 12 16 18 29.5t8 34.5q2 12 4 38 3 38 6 54 7 28 22 43t43 22q16 3 54 6 26 2 38 4 21 2 34.5 8t29.5 18q9 7 28 24 29 25 44 34 24 16 47 16t47-16q15-9 44-34 19-17 28-24 16-12 29.5-18t34.5-8q12-2 38-4 38-3 54-6 28-7 43-22t22-43q3-16 6-54 2-26 4-38 2-21 8-34.5t18-29.5q7-9 24-28 25-29 34-44 16-24 16-47t-16-47q-9-15-34-44-17-19-24-28-12-16-18-29zm-119 62L550 706q-10 17-26.5 27T488 745l-11 1q-34 0-59-24L271 584q-26-25-27-60.5t23.5-61.5 60.5-27.5 62 23.5l71 67 132-204q20-30 55-38t65 11.5 37.5 54.5-11.5 65z"/></svg>';
+
+  const addIcons = async function () {
+    const { getPostElements } = await fakeImport('/src/util/interface.js');
+    getPostElements({ excludeClass, noPeepr: true }).forEach(async postElement => {
```
AprilSylph, line 11: i think it makes more sense to process filtered posts since the header is visible at first
(this isn't perfect since our changes to the header are lost upon expanding, but hopefully i'll figure something out to deal with that)

```suggestion
    getPostElements({ excludeClass, noPeepr: true, includeFiltered: true }).forEach(async postElement => {
```

src/scripts/mutual_checker.css
```diff
@@ -0,0 +1,5 @@
+.xkit-mutual-icon {
+	height: 1.23em;
+	vertical-align: text-bottom;
```
AprilSylph, line 3: > **vertical-align** has no effect on this element since it’s not an inline or table-cell element.

since this ends up as a flex item, you'd need to instead use `align-self` to vertically position it. i think it looks best as it is, though
```suggestion
```

src/scripts/mutual_checker.css
```diff
@@ -0,0 +1,5 @@
...
+	padding-right: 0.3em;
```
AprilSylph, line 4: personal nitpick
```suggestion
	margin-right: 0.5ch;
```

AprilSylph (changes requested):
this is looking great, and it works beautifully! i think these are my last nitpicks before merging - i'm very happy to have this feature back. thank you again!

marcustyphoon:
Hm, not sure if filtered posts are actually working at the moment, but I haven't looked into why yet!

AprilSylph:
they're working for me
image

marcustyphoon:
Ah never mind, pulled the wrong repo, whoops :S

AprilSylph (approved):

PR #29, opened 2020-10-15
Popup: use background color to indicate browser focus

marcustyphoon:
#### User-facing changes
- Use the blue highlight color on the focused script preference title element, rather than on any open script preference title element. I think this is less visually confusing and makes tab navigation more clear.
- Make the grey highlight on hover only active when the title element is hovered, not when the dropdown contents are hovered. (Not sure if this is better or not? It feels a bit more natural to me. Revert 1f5b422 to see the difference.) 

#### Technical explanation
The blue highlight css has priority over the grey highlight css (only) because it comes second.

#### Issues this closes
n/a

### DIFF
--- a/src/browser_action/configuration.css
+++ b/src/browser_action/configuration.css
@@ -61,7 +61,7 @@
   color: #444444;
 }
 
-.script:hover summary {
+.script summary:hover {
   background-color: #efefef;
 }
 
--- a/src/browser_action/configuration.css
+++ b/src/browser_action/configuration.css
@@ -61,11 +61,11 @@
   color: #444444;
 }
 
-.script:hover:not([open]) summary {
+.script:hover summary {
   background-color: #efefef;
 }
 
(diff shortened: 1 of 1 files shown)


AprilSylph:
> Not sure if this is better or not? It feels a bit more natural to me.

Given that the difference is
```diff
- .script:hover:not([open]) summary {
+ .script summary:hover {
```
then it's essentially the same behaviour as before, just done in a better way!

I'll test this tomorrow to make double-sure that this feels as good as it sounds, but I'm pretty confident that it does.
Thanks for the contribution!

marcustyphoon:
Aren't you supposed to be asleeeeep? :D

Anyways, what I'm referring to there is `.script summary:hover {` vs `.script:hover summary {`, the second of which would be exactly the same as before. I think this is better, but wanted to note the (technically a) change!

AprilSylph (approved):

PR #174, opened 2021-06-30
Render scripts synchronously

marcustyphoon:
#### User-facing changes
- When the popup is opened to the default "configuration" tab, all scripts pop in at once rather than one at a time. This is quite a bit faster in Chrome and Edge (tested on MacOS), which re-render the popup at a different height each time a script is added in a fairly janky animated process, but - rather annoyingly - a bit slower in Firefox (though the animation is still cleaner).

#### Technical explanation
- `scriptTemplateClone`s are appended synchronously after all are generated. Preferences are still asynchronous, which is fine since they don't affect the popup size.
- If you wanted to, you could probably get rid of some of the ~8 frame delay on Firefox by rendering them all as blank and fetching the JSON asynchronously afterward? but yeah it's 8 frames

#### Issues this closes
- n/a

(Fun aside: apparently if you open and close a webextension popup enough times in Firefox, Stylus and uBlock Origin break!)

### DIFF
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -115,6 +115,7 @@ const renderPreferences = async function ({ scriptName, preferences, preferenceL
 };
 
 const renderScripts = async function () {
+  const scriptTemplateClones = [];
   const installedScripts = await getInstalledScripts();
   const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
@@ -160,8 +161,9 @@ const renderScripts = async function () {
       renderPreferences({ scriptName, preferences, preferenceList });
     }
 
-    scriptsDiv.appendChild(scriptTemplateClone);
+    scriptTemplateClones.push(scriptTemplateClone);
   }
+  scriptTemplateClones.forEach(scriptTemplateClone => scriptsDiv.appendChild(scriptTemplateClone));
 
   const $makeSpectrum = $(scriptsDiv).find('.makeSpectrum');
 
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -126,7 +126,7 @@ const renderPreferences = async function ({ scriptName, preferences, preferenceL
 };
 
 const renderScripts = async function () {
-  const scriptTemplateClones = [];
+  scriptsDiv.style.display = 'none';
   const installedScripts = await getInstalledScripts();
   const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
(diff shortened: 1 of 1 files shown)


src/browser_action/render_scripts.js
```diff
@@ -160,8 +161,9 @@ const renderScripts = async function () {
       renderPreferences({ scriptName, preferences, preferenceList });
     }
 
-    scriptsDiv.appendChild(scriptTemplateClone);
+    scriptTemplateClones.push(scriptTemplateClone);
   }
+  scriptTemplateClones.forEach(scriptTemplateClone => scriptsDiv.appendChild(scriptTemplateClone));
```
AprilSylph, line 166: is this the best solution? this is still adding one script `<details>` at a time, even if it's done very fast. if showing them all simultaneously is the goal, wouldn't it be better to hide `scriptsDiv` by default and then only show it once it's full?

marcustyphoon: From the perspective of the DOM they do the same thing; the whole synchronous Javascript loop completes before the browser gets to touch anything. I just tried it with display:none and display:block and it seems to have the same result in both browsers as the synchronous loop, and I think it's more obvious what the code is for that way, so I agree!

src/browser_action/render_scripts.js
```diff
@@ -115,6 +115,7 @@ const renderPreferences = async function ({ scriptName, preferences, preferenceL
 };
 
 const renderScripts = async function () {
+  const scriptTemplateClones = [];
```
AprilSylph, line 118:
```suggestion
  scriptsDiv.style.display = 'none';
```

src/browser_action/render_scripts.js
```diff
@@ -160,8 +161,9 @@ const renderScripts = async function () {
       renderPreferences({ scriptName, preferences, preferenceList });
     }
 
-    scriptsDiv.appendChild(scriptTemplateClone);
+    scriptTemplateClones.push(scriptTemplateClone);
```
AprilSylph, line 164:
```suggestion
    scriptsDiv.appendChild(scriptTemplateClone);
```

src/browser_action/render_scripts.js
AprilSylph:
```suggestion

  scriptsDiv.style.display = '';
```

AprilSylph (changes requested):

AprilSylph:
reviewing this PR made me realise there's a potential race condition wrt spectrum, so i fixed that. if my push introduces merge conflicts to this PR i'll be so mad

AprilSylph:
god dammit

AprilSylph (approved):

PR #176, opened 2021-07-02
 Shorten Posts: fix avatars in Chrome, part 3

marcustyphoon:
#### User-facing changes
- Shorten posts doesn't break sticky avatars in Chrome

#### Technical explanation
- Goes back to (only using) `overflow-y: clip`, which now that [https://github.com/AprilSylph/XKit-Rewritten/commit/f6ffb57776c129d611327328af70e8211bb8a611](https://github.com/AprilSylph/XKit-Rewritten/commit/f6ffb57776c129d611327328af70e8211bb8a611) is implemented, seems to work on both browsers.

### DIFF
--- a/src/scripts/shorten_posts.css
+++ b/src/scripts/shorten_posts.css
@@ -6,8 +6,7 @@
 
 .xkit-shorten-posts-shortened article {
   position: static;
-
-  overflow: hidden;
+  overflow-y: clip;
   max-height: var(--shortened-post-height);
 }
 
--- a/src/scripts/shorten_posts.css
+++ b/src/scripts/shorten_posts.css
@@ -6,6 +6,8 @@
 
 .xkit-shorten-posts-shortened article {
   position: static;
+
+  overflow-y: hidden;
   overflow-y: clip;
   max-height: var(--shortened-post-height);
 }


src/scripts/shorten_posts.css
```diff
@@ -6,8 +6,7 @@
 
 .xkit-shorten-posts-shortened article {
   position: static;
-
```
AprilSylph, line 9: this is my fault for not updating the style guide, but i'm trying to use [Idiomatic CSS](https://github.com/necolas/idiomatic-css) across stylesheets now, meaning there should be blank lines between positioning and box model rules. (not a request for changes, just letting you know!)

AprilSylph (changes requested):
[According to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y#browser_compatibility), `overflow-y` `clip` value is only supported beginning with Chrome 90, meaning that `minimum_chrome_version` in `src/manifest.json` should be bumped accordingly.

https://github.com/AprilSylph/XKit-Rewritten/blob/a703c69397c50944f7ac1d5b82ef5410569978ad/src/manifest.json#L54

marcustyphoon:
Oh, good point. It looks like leaving in the `hidden` rule as a backup has no effect in Firefox 89/Edge 91, and causes the not-great-looking-but-functional effect in Chromium 80:



It looks like there's some graphical issues in the popup in that browser, so maybe it's worth bumping the minimum version in any case, but it does function just fine!

src/scripts/shorten_posts.css
marcustyphoon: Oh right! I did see that commit when I was scrolling through to catch up, but forgot to look into it - and, I'll be honest, I don't have a great foundational grasp of CSS yet and don't instinctively know which properties would go in which category. Good motivation to look into that more!

AprilSylph (approved):

PR #194, opened 2021-07-27
[AccessKit] New script: display alt text below images

### LINKED ISSUE #192: Display alt text below images
I think it'd be nice to have an option to make alt text visible to users not using screenreaders by inserting it as captions below the images, sort of like the "visible captions" option that used to exist back when tumblr image captions were a thing: https://github.com/new-xkit/XKit/blob/master/Extensions/accesskit.js#L407

I've been poking around at a rough draft of this, but I wanted to check first whether it sounds like a feature you'd be interested in adding atm (and also whether you're currently open for random new feature contributions at all, or if it'd be better to wait until the rewrite is further along/out of beta?)

turtleseason:
#### User-facing changes
Adds an option to AccessKit that displays any (non-default) image alt text in posts as a caption below the image.

#### Technical explanation
It checks for any not-yet-processed imageBlocks in onPostsMutated (skipping blocks where the image hasn't loaded yet); if the alt text isn't the default "Image" text, it inserts a caption into the block (I used figcaption since imageBlock is a figure, but div would probably work too)

Like I mentioned in #192, the styling probably needs some work, haha. I also haven't spent much time testing in Chrome yet, though a quick check in tumblr.com/search/alt+text seemed OK.

#### Issues this closes
closes #192

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -37,3 +37,24 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
 body.accesskit-no-user-colours article span[style^="color"] {
   color: inherit !important;
 }
+
+
+figure.accesskit-visible-alt-text {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
+figure.accesskit-visible-alt-text figcaption {
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-top: 2px;
+  padding: 5px;
+
+  line-height: 1;
+  text-align: center;
+  background: #ddd;
+  color: #111;
+}
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -50,9 +50,9 @@ figure.accesskit-visible-alt-text figcaption {
   display: flex;
   justify-content: center;
   align-items: center;
-  padding: 5px;
+  padding: .75ch;
 
-  line-height: 1;
+  line-height: 1.25;
   text-align: center;
   background-color: rgb(var(--secondary-accent));
   color: rgb(var(--black));
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -50,9 +50,10 @@ figure.accesskit-visible-alt-text figcaption {
   display: flex;
   justify-content: center;
   align-items: center;
-  padding: .75ch;
+  padding: .75ch 1ch;
 
-  line-height: 1.25;
+  font-size: .875rem;
+  line-height: 1.5;
   text-align: center;
   background-color: rgb(var(--secondary-accent));
   color: rgb(var(--black));
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -50,11 +50,10 @@ figure.accesskit-visible-alt-text figcaption {
   display: flex;
   justify-content: center;
   align-items: center;
-  margin-top: 2px;
   padding: 5px;
 
   line-height: 1;
   text-align: center;
-  background: #ddd;
-  color: #111;
+  background-color: rgb(var(--secondary-accent));
+  color: rgb(var(--black));
 }
--- a/src/scripts/accesskit/visible_alt_text.js
+++ b/src/scripts/accesskit/visible_alt_text.js
@@ -1,0 +1,46 @@
+import { descendantSelector, keyToCss } from '../../util/css_map.js';
+import { addStyle, removeStyle } from '../../util/interface.js';
+import { onPostsMutated } from '../../util/mutations.js';
+
+let imageBlockSelector;
+let css;
+
+const className = 'accesskit-visible-alt-text';
+
+const processImages = function () {
+  [...document.querySelectorAll(`${imageBlockSelector}:not(.${className})`)]
+    .forEach(imageBlock => {
+      const image = imageBlock.querySelector('img');
+      if (image) {
+        imageBlock.classList.add(className);
+
+        if (image.alt && image.alt.toLowerCase() !== 'image') {
+          const caption = document.createElement('figcaption');
+          caption.textContent = image.alt;
+          imageBlock.appendChild(caption);
+        }
+      }
+    });
+};
+
+export const main = async function () {
+  imageBlockSelector = await keyToCss('imageBlock');
+
+  const imageBlockButtonInnerSelector = await descendantSelector('imageBlockButton', 'buttonInner');
+
+  // Setting this for all images ensures side-by-side images align vertically even if one has a caption and the other doesn't
+  css = `${imageBlockButtonInnerSelector} { height: 100%; }`;
+  addStyle(css);
+
+  onPostsMutated.addListener(processImages);
+  processImages();
+};
+
+export const clean = async function () {
+  onPostsMutated.removeListener(processImages);
+
+  removeStyle(css);
+
+  document.querySelectorAll(`.${className} figcaption`).forEach(caption => caption.remove());
+  document.querySelectorAll(`.${className}`).forEach(element => element.classList.remove(className));
+};
--- a/src/scripts/accesskit/visible_alt_text.js
+++ b/src/scripts/accesskit/visible_alt_text.js
@@ -5,15 +5,15 @@ import { onPostsMutated } from '../../util/mutations.js';
 let imageBlockSelector;
 let css;
 
-const className = 'accesskit-visible-alt-text';
+const processedClass = 'accesskit-visible-alt-text';
 
 const processImages = function () {
   [...document.querySelectorAll(imageBlockSelector)]
-    .filter(imageBlock => imageBlock.classList.contains(className) === false)
+    .filter(imageBlock => imageBlock.classList.contains(processedClass) === false)
     .forEach(imageBlock => {
       const image = imageBlock.querySelector('img');
       if (image) {
-        imageBlock.classList.add(className);
+        imageBlock.classList.add(processedClass);
 
         if (image.alt) {
           const caption = document.createElement('figcaption');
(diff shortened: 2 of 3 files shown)


src/scripts/accesskit.css
```diff
@@ -37,3 +37,24 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-top: 2px;
+  padding: 5px;
+
+  line-height: 1;
+  text-align: center;
+  background: #ddd;
+  color: #111;
```
AprilSylph, line 59: Rule no.1 of inserting elements into the webpage: remember that different palettes exist!

I get that it's not easy to figure out how to use all the different variables, though. Personally, for this sort of complimentary in-post element, I would use "black" text on a "secondary accent" background.

| True Blue | Dark Mode | Low-Contrast Classic |
|-----------|------------|----------------------|
| Screen Shot 2021-07-29 at 14 57 21 | Screen Shot 2021-07-29 at 14 57 35 | Screen Shot 2021-07-29 at 14 59 29 |



```suggestion
  background-color: rgb(var(--secondary-accent));
  color: rgb(var(--black));
```

src/scripts/accesskit.css
```diff
@@ -37,3 +37,24 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
...
+  display: flex;
+  flex-direction: column;
+}
+
+figure.accesskit-visible-alt-text figcaption {
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-top: 2px;
```
AprilSylph, line 53: Personally not a fan of having a margin between the image and the alt text container. Without margin, it feels more like a captioned photo on Wikipedia, just without the outer frame.

```suggestion
```

AprilSylph (changes requested):

src/scripts/accesskit/visible_alt_text.js
```diff
@@ -0,0 +1,46 @@
...
+const className = 'accesskit-visible-alt-text';
+
+const processImages = function () {
+  [...document.querySelectorAll(`${imageBlockSelector}:not(.${className})`)]
+    .forEach(imageBlock => {
+      const image = imageBlock.querySelector('img');
+      if (image) {
+        imageBlock.classList.add(className);
+
+        if (image.alt && image.alt.toLowerCase() !== 'image') {
```
AprilSylph, line 17: This won't work for images posted by non-English users. I feel like it would be easier to remove this check altogether rather than use the localised string though; I for one would definitely prefer to explicitly see that an image has no user-set alt text, even if only to be sure that the feature is working.

AprilSylph:
A very impressive first contribution! I hope you found the framework relatively easy to write for.

I did have a thought about simplifying this feature down to duplicating an image's alt text into its `title` attribute so that it's shown on hover, but just as I was about to suggest it, I realised it would make the feature completely unusable on devices without a cursor input...

src/scripts/accesskit.css
turtleseason: LOL, here I was wondering if there was some magical way to detect the current theme colors, and the variables were right in front of me the whole time... noob mistakes 😆

turtleseason:
Thanks so much for your help and feedback (and of course for all of your work on XKit)!

It did feel easy to work with (and it helped that there was a similar processing-all-images script in disable-gifs to ~~steal~~ learn from haha). I feel like I'm too much of a newbie to really judge, but the scripts seemed really clean and easy to read, I didn't have trouble finding what I needed!

src/scripts/accesskit.css
```diff
@@ -37,3 +37,23 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
...
+}
+
+figure.accesskit-visible-alt-text figcaption {
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 5px;
+
+  line-height: 1;
```
turtleseason, line 55: By the way, looking at this again, the text feels a tiny bit cramped to me - what do you think of bumping the line-height to 1.25 (and maaaybe padding to .75ch)?

AprilSylph: Tumblr uses `line-height: 1.5;` in posts so I think it's best to just copy that. I'd have to take another look for the padding.

src/scripts/accesskit/visible_alt_text.js
AprilSylph: > though on the other hand, maybe it's good that they'll get a taste of how screenreader users probably feel, haha [...] maybe it'll help motivate more people to (prod other people to) start actually using alt text in their posts

I think this is justification enough. I would love for this feature to become an educational tool on the importance of setting alt text.

> I feel like it could be irritating for people who just want to see extra information where it exists

If anyone finds it annoying enough to complain about in a discussion or issue, we can address it there and maybe change the behaviour if there's a stronger argument for ignoring the default alt text.

src/scripts/accesskit.css
turtleseason: Fair point! I was worried that using the full line spacing makes the caption look too large/take up too much space, but on second thought I'm probably just overthinking it x)

AprilSylph: If you feel that's the case (and to be honest, I did too - I tried reducing the font weight only to realise it was at its minimum already) maybe we could reduce the `font-size` to `.875rem` (to match the post header)?

turtleseason: I think I like that! I just made the side padding a little larger, too - `padding: .75ch 1ch`

turtleseason:
One more question btw - once all the changes are ironed out, I'm assuming I should rebase to some extent - should I flatten everything into one commit, or keep most of the history and just clean it up a little (e.g. my messing around with the spacing)?

AprilSylph:
No need! I squash every PR upon merge, so unless there are merge conflicts displayed, the finished PR will just be one clean commit at the top of the history.

src/scripts/accesskit/visible_alt_text.js
```diff
@@ -0,0 +1,46 @@
...
+import { addStyle, removeStyle } from '../../util/interface.js';
+import { onPostsMutated } from '../../util/mutations.js';
+
+let imageBlockSelector;
+let css;
+
+const className = 'accesskit-visible-alt-text';
+
+const processImages = function () {
+  [...document.querySelectorAll(`${imageBlockSelector}:not(.${className})`)]
```
AprilSylph, line 11: `keyToCss` can't be relied on to produce just a single class selector. If Tumblr adds a new element with a different `imageBlock` class, this will end up as `.nonsenseOne, .nonsenseTwo:not(.accesskit-visible-alt-text)`, which won't work as intended.

Since there's no way to group selectors in CSS to achieve the desired effect with `:not()`, we can either use `keyToClasses` to only grab one classname to use, or we can do the more robust thing: filter the array instead of trying to construct a perfect selector.

```suggestion
  [...document.querySelectorAll(imageBlockSelector)]
    .filter(imageBlock => imageBlock.classList.contains(className) === false)
```

I believe it is technically more performant to use a perfect selector (browsers' CSS engines are *very* fast), but robustness has to come first.

src/scripts/accesskit/visible_alt_text.js
```diff
@@ -0,0 +1,46 @@
...
+  processImages();
+};
+
+export const clean = async function () {
+  onPostsMutated.removeListener(processImages);
+
+  removeStyle(css);
+
+  document.querySelectorAll(`.${className} figcaption`).forEach(caption => caption.remove());
+  document.querySelectorAll(`.${className}`).forEach(element => element.classList.remove(className));
```
AprilSylph, line 45: Since `clean()` is only called once upon user input, we can sacrifice a bit of performance for the sake of code simplicity, and instead use jQuery to remove these modifications.

```suggestion
  $(`.${className} figcaption`).remove();
  $(`.${className}`).removeClass(className);
```

AprilSylph (changes requested):
Looking good! I was about to approve this with a note that I'd make a few tweaks before releasing the feature, but in case you end up submitting further contributions, I thought I should instead take the time to be educational.

---

Another comment I couldn't include: `className` for a custom class used across a script is maybe a little too generic. When using a classname to exclude already-processed elements I typically call it `excludeClass`, although there might be something more fitting here given that we're also using it to style elements.

src/scripts/accesskit/visible_alt_text.js
turtleseason: Ooh makes sense! I guess if we really wanted to stick with CSS selectors, maybe generating a string with keyToClasses and 
```imageBlockSelector = imageBlockClasses.foreach(imageBlockClass => `${imageBlockClass}:not(.${className})`).join(', ')``` would work? Though idk if the tradeoff in readability/simplicity is worth it.

turtleseason:
> I typically call it `excludeClass`, although there might be something more fitting here given that we're also using it to style elements.

Just to throw out a couple possibilities before changing it - maybe processedClass or styledClass? (Or processed[Image]BlockClass to be a bit more descriptive)

Thanks again for taking the time to review this 🙏 Your comments have been really helpful!

src/scripts/accesskit/visible_alt_text.js
AprilSylph: Readability comes first, especially when the loss in performance is negligible in real-life measures. `onPostsMutated` listener triggering is debounced, after all.

I admit I do want to write some kind of function that makes creating selectors like this more feasible, but my ideas for that are still cooking.

AprilSylph:
`processedClass` looks best to me! It's not the most descriptive option, but being more verbose doesn't seem useful enough in this case to justify having such a painfully long constant name.

AprilSylph (approved):
A great feature idea, and beautifully executed! It's been a joy to review this. Thank you for your contribution!

marcustyphoon:
Wonderful! I love this.

src/scripts/accesskit/visible_alt_text.js
marcustyphoon: I had the same thoughts as discussed here. I did think, additionally, about whether the localized version of "Image" should have some kind of special handling - something like being faded, being italic, being slightly smaller, being replaced by "no alt text," etc. Probably not worth the extra lines of code, but I figured I'd mention the idea!

PR #203, opened 2021-08-11
No Recommended: Hide blogs in sidebar on tag pages

turtleseason:
#### User-facing changes
Extends No Recommended's "Hide recommended blogs in the sidebar" setting to also hide the sidebar blogs widget on /tagged pages (XKit 7 currently does this with the Tweaks > Sidebar > Hide Recommended blogs option).

#### Technical explanation
Checking for the widget is a little different on tag pages than on the dashboard (the label includes the tag name so checking it requires a regex, and I also used a cssmap class selector to avoid something like `aside > div > div > div > div`). I was torn between trying to merge the two blocks in checkForRecommendedBlogs together vs. duplicating some code to try and keep the code more straightforward(?); went with the latter, but could definitely change that if needed.

Was also wondering about renaming recommendedBlogsLabel and recommendedBlogsPattern to something that indicates one is used on the dashboard and the other on tag pages, but I wanted to err on the side of not messing with existing variables (and also most of my ideas were excruciatingly long lol).

(In hindsight I'm wondering if I should have opened an issue to discuss this first - if so, sorry about that ^^ Definitely open to discussion here, and of course don't hesitate to close this if it's not needed atm~)

### DIFF
--- a/src/scripts/no_recommended/hide_recommended_blogs.js
+++ b/src/scripts/no_recommended/hide_recommended_blogs.js
@@ -1,11 +1,15 @@
 import { onBaseContainerMutated } from '../../util/mutations.js';
 import { translate } from '../../util/language_data.js';
 import { addStyle, removeStyle } from '../../util/interface.js';
+import { descendantSelector } from '../../util/css_map.js';
 
 const excludeClass = 'xkit-no-recommended-blogs-done';
 const hiddenClass = 'xkit-no-recommended-blogs-hidden';
 
 let recommendedBlogsLabel;
+let recommendedBlogsPattern;
+
+let tagPageTitleSelector;
 
 const css = `.${hiddenClass} { display: none; }`;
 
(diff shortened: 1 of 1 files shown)


AprilSylph (changes requested):
the variable naming could be better but that's my fault for not updating the name of `recommendedBlogsLabel` when the text changed. i think the ideal would be...

- `recommendedBlogsLabel` -> `checkOutTheseBlogsLabel`
- `tagPageLabel` -> `topBlogsLabel`
- `recommendedBlogsPattern` -> `topBlogsPattern`

but i'd just as happily do this myself after merging.

i think we could do away with `tagPageTitleSelector` entirely though: it looks like the recommended blogs lists themselves are both `aside ul[aria-label][style]`, with the aria label matching the title text we're currently testing for, so the two blocks *could* be merged by targeting+testing those instead of the titles.

turtleseason:
> i think we could do away with `tagPageTitleSelector` entirely though: it looks like the recommended blogs lists themselves are both `aside ul[aria-label][style]`, with the aria label matching the title text we're currently testing for, so the two blocks _could_ be merged by targeting+testing those instead of the titles.

Ooh this is much cleaner, thank you!!
Would be even simpler if `[aria-label^=""]` matched any string, but unfortunately it looks like we need to check and remove the selector entirely if the tag name appears at the beginning or end of the label. But even then it looks much better than before imo!

AprilSylph (changes requested):
ah, something i failed to consider when suggesting testing the `ul` elements... the regular recommended blogs don't get an `ul` until the blogs load, which delays the sidebar item's disappearance noticeably.

this is painful but it looks like it would be best to revert your last commit, then redo the variable name changes and re-implement the `topBlogsSelector` selector code to simplify the second loop (since that widget loads all at once).

AprilSylph:
if only Tumblr used `h1` for the "Top %1$s blogs" title so we could fit the logic into one loop without introducing extra wait time...

also, sorry for not reviewing this as quickly as i would have liked. i know how frustrating it can be as a contributor to outpace your reviewer, so i hope that my off-days don't ever leave me lagging annoyingly far behind.

turtleseason:
Well that explains why I suddenly started noticing the delay more while testing this - just didn't connect the dots that it was my changing stuff that caused it 😅

And no worries at all about the pace of reviews! There's no rush on my end - if anything, I worry about taking up too much of your time with reviews, I wouldn't want there to be any pressure to get to them on off days or when it's inconvenient.

AprilSylph (approved):
performant behaviour and easy-to-follow code! thanks for taking the time to refine this feature!

i'll aim to include this in a patch release on Monday.

PR #218, opened 2021-09-01
Control buttons util: Cache remix icon JSON

marcustyphoon:
#### User-facing changes
- None

#### Technical explanation
This caches the remix icon JSON so it doesn't have to be fetched and parsed again if multiple scripts use it. This is, practically speaking, fairly negligible, so it might not even be worth it.

I really don't like this syntax, but I can't think of a way to make it... not, you know, kind of a mess.

### DIFF
--- a/src/util/control_buttons.js
+++ b/src/util/control_buttons.js
@@ -1,3 +1,5 @@
+let iconsPromise;
+
 /**
  * Fetch the SVG belonging to a specific RemixIcon.
  *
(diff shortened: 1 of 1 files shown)


AprilSylph:
I've got a different overhaul planned for RemixIcon usage, so I probably won't be taking a look at this, but I'll leave it open until I can close it via commit.

marcustyphoon:
Sounds good! I would personally just close it then, it's irrelevant levels of minor.

PR #219, opened 2021-09-01
Inject util: use a single listener+callback

marcustyphoon:
#### User-facing changes
- None

#### Technical explanation
Rather than creating a new callback function and a new event listener for each concurrent use of the inject function, this uses a single callback with a list of all the currently pending injected functions.

This should reduce the time spent processing each message if there are a lot of injected functions used at once (i.e. timestamps querying the API for reblog timestamp data). (It seemed to go from about 5ms to about 1.5 if I'm reading the profiler correctly.)

This is... almost possibly a barely perceptible performance gain! Such benefit, much wow. I think if you did an a/b test on me it'd be like a 50/50 tossup if I could tell which was which, maybe a bit less.

### DIFF
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -1,4 +1,23 @@
 let nonce;
+const pending = new Map();
+
+const callback = event => {
+  const { result, exception, xkitCallbackNonce } = event.data;
+
+  if (event.origin === `${location.protocol}//${location.host}` && pending.has(xkitCallbackNonce)) {
+    const [resolve, reject] = pending.get(xkitCallbackNonce);
+    pending.delete(xkitCallbackNonce);
+    if (pending.size === 0) {
+      window.removeEventListener('message', callback);
+    }
+
+    if (exception === undefined) {
+      resolve(JSON.parse(result || 'null'));
+    } else {
+      reject(Object.assign(new Error(), JSON.parse(exception)));
+    }
+  }
+};
 
 /**
  * @param {Function} asyncFunc - Asynchronous function to run in the page context
@@ -11,21 +30,11 @@ export const inject = (asyncFunc, args = []) => new Promise((resolve, reject) =>
     nonce = scriptWithNonce.nonce || scriptWithNonce.getAttribute('nonce');
   }
 
+  if (pending.size === 0) {
+    window.addEventListener('message', callback);
+  }
   const callbackNonce = Math.random();
-
-  const callback = event => {
-    if (event.origin === `${location.protocol}//${location.host}` && event.data.xkitCallbackNonce === callbackNonce) {
-      window.removeEventListener('message', callback);
-
-      const { result, exception } = event.data;
-
-      if (exception === undefined) {
-        resolve(JSON.parse(result || 'null'));
-      } else {
-        reject(Object.assign(new Error(), JSON.parse(exception)));
-      }
-    }
-  };
+  pending.set(callbackNonce, [resolve, reject]);
 
   const script = document.createElement('script');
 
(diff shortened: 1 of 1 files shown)


src/util/inject.js
```diff
@@ -1,4 +1,23 @@
...
+
+const callback = event => {
+  const { result, exception, xkitCallbackNonce } = event.data;
+
+  if (event.origin === `${location.protocol}//${location.host}` && pending.has(xkitCallbackNonce)) {
+    const [resolve, reject] = pending.get(xkitCallbackNonce);
+    pending.delete(xkitCallbackNonce);
+    if (pending.size === 0) {
+      window.removeEventListener('message', callback);
+    }
```
AprilSylph, line 12: is this necessary? the code would be a lot simpler if the event was just always attached, but i can see that being a problem if Tumblr uses `postMessage` for anything too

src/util/inject.js
```diff
@@ -1,4 +1,23 @@
 let nonce;
+const pending = new Map();
+
+const callback = event => {
```
AprilSylph, line 4: I feel like the naming could be better here...

personally I think I would use `callbacks` for the Map and `messageHandler` for the Function.

also, unrelated, just realised how much of a rush I wrote this module in! now that i'm looking more carefully, i'm asking myself why i didn't just destructure `data` and `origin` here.

src/util/inject.js
marcustyphoon: A quick test indicates that Tumblr does indeed use postMessage, though funnily enough I think we essentially have the "opposite" problem, where Tumblr fires a whole bunch of callbacks on *our* messages. 40ms of overhead for 5ms of code, fun times.

image

I can't imagine doing one extra string comparison every time Tumblr code posts a message is a big deal by comparison. Might even be a performance gain if adding/removing an event listener is relatively expensive (not sure, but this is in the <1ms range so it doesn't matter either way).

(I started testing things with 250 posts loaded and the potential performance improvements in those scenarios are actually quite noticeable! Will write about some today or tomorrow.)

src/util/inject.js
marcustyphoon: Totally agreed, I like those names!

src/util/inject.js
```diff
@@ -1,4 +1,24 @@
...
+    callbacks.delete(xkitCallbackNonce);
+
+    if (exception === undefined) {
+      resolve(JSON.parse(result || 'null'));
+    } else {
+      reject(Object.assign(new Error(), JSON.parse(exception)));
+    }
+  }
+};
+window.addEventListener('message', messageHandler);
```
AprilSylph, line 21: spacing nitpick

```suggestion
};

window.addEventListener('message', messageHandler);
```

AprilSylph (changes requested):

src/util/inject.js
```diff
@@ -1,4 +1,24 @@
 let nonce;
+const callbacks = new Map();
+
+const messageHandler = event => {
+  const {
+    origin,
+    data: { result, exception, xkitCallbackNonce }
+  } = event;
```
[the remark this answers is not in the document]
AprilSylph: ah, i failed to account for the proceeding blank line. sorry about that

src/util/inject.js
marcustyphoon: Go for it! We can always reopen closed PRs, right?

AprilSylph: It wouldn't be a replacement to this PR! It would be easier to implement it in the version of the module you're proposing... I think. I always have `git revert` if I'm wrong.

marcustyphoon: > feel free to close this **conversation**

Ahh, right, reading comprehension is a thing. That's what I get for looking at github notifications on my phone :D

AprilSylph (approved):

PR #221, opened 2021-09-02
New script: Shrink posts in queue/drafts

marcustyphoon:
Okay so first off, I have no idea what you would call this script.

#### User-facing changes
Adds a script to [shrink/collapse/condense/whatever] posts in your queue, and optionally drafts, between the header and footer. Makes it easier to rearrange your queue. Is an excuse to use CSS box shadows.

Previously a function of Enhanced Queue/Queue+. (Shuffling your queue is now built in to Tumblr; clearing your queue will presumably be in mass deleter.)

The collapsed area is scrollable, so users can still see the entire post content.

image

#### Technical explanation
Had the code lying around from learning the Rewritten framework. Figured I'd update it to learn the changes to the framework since then.

In order to place a shadow within (over) the shrunk post content, an absolutely positioned element with a positioned parent of the correct size is needed. I don't know how to do this with less than 3 DOM elements. Pseudocode:

```diff
<article>
  <div></div>
  <header></header>
+ <div class="xkit-queue-shadow-outer" style="position: relative">
+   <div class="xkit-queue-shadow" style="position: absolute; z-index: 1; width: 100%; height: 100%;"></div>
+   <div class="xkit-collapsed-queue-container" style="max-height: 200px; overflow-y: scroll">
      (post content goes here)
+   </div>
+ </div>
  <div>
    <footer></footer>
  </div>
</article>
```

#### Issues this closes
None, but I think somebody requested it back when XK7 wasn't abandonware.


lihatt:
I used to use this feature all the time and would be delighted to have this back in XKit.

AprilSylph:
> In order to place a shadow within (over) the shrunk post content, an absolutely positioned element with a positioned parent of the correct size is needed. I don't know how to do this with less than 3 DOM elements.

What you can (probably) do is use [CSS pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) to be less invasive with DOM manipulation. Any `::before` or `::after` pseudo-elements count as children of their targets, so you can make the target `position: relative;` and the pseudo-element `position: absolute;` to get a fully-stylable absolutely-positioned new child.

```diff
- .xkit-queue-shadow-outer { position: relative; }
- .xkit-queue-shadow { position: absolute; }
+ .xkit-queue-shadow { position: relative; }
+ .xkit-queue-shadow::before { position: absolute; }
```

Not sure if you could put the pseudo-element on the scroll container, it might use the container height instead of the content height, but then again it might not scroll with the content because of its positioning...

P.S. Sorry for taking so long to get to reviewing this!

marcustyphoon:
Oh, riiiight, those things! They've always confused me but seemed really useful to learn.

src/scripts/collapsed_queue.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+};
+
+export const main = async function () {
+  const { runInQueue, runInDrafts } = await getPreferences('collapsed_queue');
+  const regexGroup = [
+    ...runInQueue ? ['queue'] : [],
+    ...runInDrafts ? ['draft'] : []
+  ].join('|');
+  timelineRegex = new RegExp(String.raw`\/v2\/blog\/[^/]+\/posts\/(${regexGroup})`);
```
AprilSylph, line 38: Since we're using the `RegExp` constructor rather than literal notation, these forward-slashes don't need to be escaped. They have no special meaning in regular expressions themselves.

```suggestion
  timelineRegex = new RegExp(String.raw`/v2/blog/[^/]+/posts/(${regexGroup})`);
```

AprilSylph (changes requested):
Otherwise LGTM!

src/scripts/collapsed_queue.js
marcustyphoon: Oh, man, I'm still not used to wrapping my brain around the difference between "what you have to type to create a regex" and "the characters that make up the regex" :D The `String.raw` call should be unnecessary as well, then, if I'm understanding correctly.

AprilSylph (approved):

PR #253, opened 2021-09-09
Quick Reblog: Remember last selected blog

### LINKED ISSUE #220: Quick Reblog: Option to remember last selected blog
Quick Reblog does not currently remember the last selected blog. In the case where almost all reblogs are not made to the primary blog, selecting the correct blog every time becomes tedious.

turtleseason:
#### User-facing changes
- Remembers the last selected blog in the Quick Reblogs popup. The selection persists across navigation in the React page, but not across full page reloads - I admittedly haven't looked much into how saving user data in XKit works yet, but I'd be down to look into storing the selection more persistently if anyone thinks it'd be a good idea.
(Edit: the more I think about this, the more I think it probably should persist across reloads if possible, so I'll keep this as a draft until I've looked into that more - open to any feedback in the meantime though!)

#### Technical explanation
- Probably worth noting that the event listener on blogSelector sticks around when this option is disabled and could theoretically get attached multiple times if you toggle the setting a lot (doesn't affect behavior, though) - I doubt it's worth the extra lines of code to clean up properly, but it could be done.

#### Issues this closes
Resolves #220

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -33,12 +33,12 @@ let timeoutID;
 
 let popupPosition;
 let showBlogSelector;
+let rememberLastBlog;
 let showCommentInput;
 let quickTagsIntegration;
 let showTagsInput;
 let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
-let rememberLastBlog;
 
 const storageKey = 'quick_reblog.alreadyRebloggedList';
 const excludeClass = 'xkit-quick-reblog-alreadyreblogged-done';
@@ -179,12 +179,12 @@ export const main = async function () {
   ({
     popupPosition,
     showBlogSelector,
+    rememberLastBlog,
     showCommentInput,
     quickTagsIntegration,
     showTagsInput,
     alreadyRebloggedEnabled,
-    alreadyRebloggedLimit,
-    rememberLastBlog
+    alreadyRebloggedLimit
   } = await getPreferences('quick_reblog'));
 
   popupElement.className = popupPosition;
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -38,6 +38,7 @@ let quickTagsIntegration;
 let showTagsInput;
 let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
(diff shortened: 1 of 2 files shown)


AprilSylph:
>  The selection persists across navigation in the React page, but not across full page reloads

This is the only way it can work. Full page reloads can't be expected to belong to the same account; any behaviour which is broken merely by using container tabs in Firefox should not be implemented.

src/scripts/quick_reblog.js
```diff
@@ -52,7 +54,7 @@ const showPopupOnHover = ({ currentTarget }) => {
 
   const thisPostID = currentTarget.closest('[data-id]').dataset.id;
   if (thisPostID !== lastPostID) {
-    blogSelector.value = blogSelector.options[0].value;
+    blogSelector.value = rememberLastBlog ? lastBlogID : blogSelector.options[0].value;
```
AprilSylph, line 57: I don't see the point of storing `lastBlogID`? I would just put a condition here to leave `blogSelector.value` alone if the preference is enabled.

turtleseason:
> Full page reloads can't be expected to belong to the same account

I see, makes sense (and I'm guessing storing data on a per-tumblr-account basis would be overly complicated (assuming it's even possible) and/or something we don't want to do on principle?)

> I don't see the point of storing lastBlogID?

Oh, wow, yeah you're totally right lol. Guess that ends up making this a pretty trivial contribution overall 😅

AprilSylph (changes requested):
nitpick, but i highly prefer preferences being listed in the script in the same order they're specified in the manifest

L41 -> L36
L187 -> L182

AprilSylph:
> and I'm guessing storing data on a per-tumblr-account basis would be overly complicated (assuming it's even possible) and/or something we don't want to do on principle?

That, and it would require saving the primary blogs' UUIDs in the addon storage. I don't think addon configs should be able to expose who you are on a website, even if that information is a complete zero-risk to most people.

> Guess that ends up making this a pretty trivial contribution overall 😅

No worries! It's great to get to do some collaborative programming again for once, no matter the scale. Certainly beats doing every little thing myself.

AprilSylph (approved):

PR #274, opened 2021-09-17
ESLint: Lint for incorrect import paths

marcustyphoon:
#### User-facing changes
- None

#### Technical explanation
This enables eslint-plugin-import's recommended options. This includes import/no-unresolved, which will check for import paths that are mistyped.

Apparently the Standard JS ESLint config doesn't enable this rule, despite including the plugin. Not sure why, but I could imagine a reason.

#### Issues this closes
n/a

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,7 +7,8 @@
   },
   "extends": [
     "semistandard",
-    "plugin:jsdoc/recommended"
+    "plugin:jsdoc/recommended",
+    "plugin:import/recommended"
   ],
   "parserOptions": {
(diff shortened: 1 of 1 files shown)


AprilSylph:
It looks like it would be pretty easy to add that rule to Standard by contributing to [`eslint-config-standard`](https://github.com/standard/eslint-config-standard), and if your PR there gets shot down then you should get an explanation in the process!

I'd also like to know why it's not included already... it seems like such a basic "don't write broken code" check. If you do open an issue or PR, please link to it here so I can follow it.

AprilSylph (approved):

PR #275, opened 2021-09-18
Test: Debounce mutations only to the next DOM render

marcustyphoon:
#### User-facing changes
- Potential for slightly reduced UI flickering

#### Technical explanation
Playing around with this. This implements a "debounce" function that runs all callbacks before the next DOM update/render, ensuring that (the synchronous parts of) the callbacks won't immediately cause a second rerender.

This means, for example, that with notificationblock enabled, the notification drawer immediately opens with blocked notifications missing, rather than appearing with all notifications visible and then popping the blocked ones away.

This has the potential to lag the UI, though, if callbacks put lots of expensive calculation in their synchronous portions. (Note that the `async` keyword doesn't make a function asynchronous automatically; only the portions of the function after an await statement are, and only if that await statement resolves after the microtask queue is empty!)

On the other hand, the "flickering" process is in itself a source of (a different kind of) lag, as rerendering the DOM twice is often more expensive than the code. So, it depends!

In theory this technique could be used on only some of the callbacks or something, etc.

#### Issues this closes
n/a

### DIFF
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -37,10 +37,10 @@ const debounce = (func, ms) => {
 };
 
 const debounceDOM = (func) => {
-  let timeoutID;
+  let requestID;
   return (...args) => {
-    cancelAnimationFrame(timeoutID);
-    timeoutID = requestAnimationFrame(() => func(...args));
+    cancelAnimationFrame(requestID);
+    requestID = requestAnimationFrame(() => func(...args));
   };
 };
 
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -36,10 +36,22 @@ const debounce = (func, ms) => {
   };
 };
 
-const runOnNewPosts = debounce(() => onNewPosts.trigger(), 10);
-const runOnPostsMutated = debounce(() => onPostsMutated.trigger(), 100);
-const runOnBaseContainerMutated = debounce(() => onBaseContainerMutated.trigger(), 100);
-const runOnGlassContainerMutated = debounce(() => onGlassContainerMutated.trigger(), 100);
+const debounceDOM = (func) => {
+  let queued = false;
+  return (...args) => {
+    if (queued) return;
+    queued = true;
+    requestAnimationFrame(() => {
+      queued = false;
+      func(...args);
+    });
+  };
+};
+
+const runOnNewPosts = debounceDOM(() => onNewPosts.trigger());
+const runOnPostsMutated = debounceDOM(() => onPostsMutated.trigger());
+const runOnBaseContainerMutated = debounceDOM(() => onBaseContainerMutated.trigger());
+const runOnGlassContainerMutated = debounceDOM(() => onGlassContainerMutated.trigger());
 
 const observer = new MutationObserver(mutations => {
   if (onNewPosts.listeners.length !== 0 || onPostsMutated.listeners.length !== 0) {
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -37,14 +37,10 @@ const debounce = (func, ms) => {
 };
 
 const debounceDOM = (func) => {
-  let queued = false;
+  let timeoutID;
   return (...args) => {
-    if (queued) return;
-    queued = true;
-    requestAnimationFrame(() => {
-      queued = false;
-      func(...args);
-    });
+    cancelAnimationFrame(timeoutID);
+    timeoutID = requestAnimationFrame(() => func(...args));
   };
 };
 


src/util/mutations.js
```diff
@@ -36,10 +36,22 @@ const debounce = (func, ms) => {
...
-const runOnPostsMutated = debounce(() => onPostsMutated.trigger(), 100);
-const runOnBaseContainerMutated = debounce(() => onBaseContainerMutated.trigger(), 100);
-const runOnGlassContainerMutated = debounce(() => onGlassContainerMutated.trigger(), 100);
+const debounceDOM = (func) => {
+  let queued = false;
+  return (...args) => {
+    if (queued) return;
+    queued = true;
+    requestAnimationFrame(() => {
+      queued = false;
```
AprilSylph, line 45: is this queue setup better than using `cancelAnimationFrame` like the regular debounce function uses `clearTimeout`?

AprilSylph (commented):
VERY exciting concept

marcustyphoon:
Note that, exciting as this sounds, there are limitations to this concept besides the need to carefully vet callbacks for expensive synchronous or async-but-in-a-microtask code:

- If you *do* want callbacks to take advantage of this and do their dom mutations before the render to avoid flickering, now they *can't* use another event loop. In particular, inject() callbacks come back later down the line, so this has no effect on anything that uses `apiFetch` or a fresh `timelineObject` call (or `exposeTimelines`, except I have an improvement for that that I'm about to PR). This limits the benefit of this technique.
- Callbacks get run more frequently this way, even than with a debounce of zero. Debouncing, by definition, waits until a "clump" of mutations ends before firing callbacks, while this flushes all queued callbacks whenever the DOM renders, and sometimes (often) Tumblr's React renders the DOM multiple times in a row! So in theory, this is a tradeoff: less repaints, but more `document.querySelectorAll` calls. I haven't managed to determine whether this is a net improvement or net regression of CPU cost, yet! Especially given the "QSAs get extremely expensive if you scroll down a lot" thing, one must be a bit careful here.

Then again, not flickering is neat :D

src/util/mutations.js
marcustyphoon: Performance-wise, my guess is that this is technically faster but easily within margin of error/not measurable. rAF/cAF, like setTimeout, do show up in the profiler, while setting a variable in a closure certainly doesn't. I wouldn't worry about that and do whichever seems more clear to understand.

Possibly more interesting is that this architecture is more like a trailing throttle (callback will definitely fire "soon," just not back to back) than a debounce (callback will fire once there are no updates for a while). Here's the trailing throttle I wrote for my CSS map ~~monstrosity~~ uh, userscript:

```js
const throttle = (func, ms) => {
  let running = false;
  return (...args) => {
    if (!running) {
      running = true;
      setTimeout(() => {
        running = false;
        func(...args);
      }, ms);
    }
  };
};
```
Cancelling and re-starting a timer, as a debounce does, serves the purpose of moving the end of the delay forward in time (well, in a debounce of greater than zero milliseconds, anyway). Cancelling and instantly re-starting an rAF queues the task for the same moment in time - right before the next DOM render - so it serves the same purpose as doing nothing!

Which is to say, of course, that both methods have the same result.

AprilSylph: I see... I feel like there might be a way to better communicate this with the code itself, but I'll give that task to myself after it's merged.

marcustyphoon: I'm perfectly happy with using `clearAnimationFrame` for the same syntax, too! Should do the same thing (I assume - haven't tried it yet, actually.)

AprilSylph:
> Callbacks get run more frequently this way, even than with a debounce of zero. [...] So in theory, this is a tradeoff: less repaints, but more `document.querySelectorAll` calls.

What I'm hearing here is that `onNewPosts` could definitely use this while the others might not see any performance gain, and that the way the listeners work in the first place could be wildly improved (which I've already expressed before).

Even if no performance improvements can be gleamed from this straight away, I think improving everything to the point that it would be is worth it. So keep this PR open!!

marcustyphoon:
> What I'm hearing here is that `onNewPosts` could definitely use this while the others might not see any performance gain

Sorta! The one extension I've found that definitely does have a perceptible improvement (no flicker) is NotificationBlock, and that one is run on `onBaseContainerMutated`!

AprilSylph:
Hmm... thinking how to improve the listeners...

Is `Element.querySelectorAll()` just as bad as `document.querySelectorAll()` when the document contains many nodes? My gut says it shouldn't be; if mutations' sizes are consistent while the document bloats, only querying the mutations should be easier on the browser, right?

marcustyphoon:
According to my notes,

> Element.children? (have not tested this yet; I tried element.QSA and it's worse)

Apparently it's slightly worse! No idea why that would be. Not sure where I put that code (how was I getting the element with which to QSA? I don't remember.)

I'm not actually sure if I tested Element.children on the timeline, either. What's really fast, though, is - given that we can assume all the posts we want are `listTimelineObject`s:

Screen Shot 2021-09-04 at 1 46 43 AM

...using good old getElementsByClassName, like it's 2009 and we're debating whether JQuery is performant enough.

src/util/mutations.js
```diff
@@ -37,14 +37,10 @@ const debounce = (func, ms) => {
 };
 
 const debounceDOM = (func) => {
-  let queued = false;
+  let timeoutID;
```
AprilSylph, line 40:
```suggestion
  let requestID;
```

src/util/mutations.js
```diff
@@ -37,14 +37,10 @@ const debounce = (func, ms) => {
...
   return (...args) => {
-    if (queued) return;
-    queued = true;
-    requestAnimationFrame(() => {
-      queued = false;
-      func(...args);
-    });
+    cancelAnimationFrame(timeoutID);
+    timeoutID = requestAnimationFrame(() => func(...args));
```
AprilSylph, line 43:
```suggestion
    cancelAnimationFrame(requestID);
    requestID = requestAnimationFrame(() => func(...args));
```

AprilSylph (changes requested):
MDN calls `requestAnimationFrame()`'s return value `requestID` in https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame#parameters

marcustyphoon:
> (I'll get around to drafting a PR with some tests of this kind of thing at some point, probably.)

Not draft PRing this yet because a) this API would be awful, just seeing if it works and b) I have enough PRs already, but trying out the performance of deduplicating the getPostElements QSA. Does work I think... but there's still a lot of other QSAs from other things and we're still much less expensive than Tumblr's own JS, so I think it's only be worth doing if we come up with a clean API. If it makes the code convoluted, I personally don't think it's worth it.

https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon:replace-QSA-experiment-3?expand=1

AprilSylph:
Yeah, I was thinking more along the lines of doing something different than `onNewPosts`; so far all I've got is a concept for a new module that
1. Listens for base container mutations to detect timeline elements
2. Automatically adds `data-timeline` to those which it finds
3. Collects the posts from those timelines (via `[...Element.children].filter()` and feeds them to callbacks
4. Adds a new MutationObserver to the timelines which also report new posts to the callbacks

But of course, whether this is an improvement or not hinges on whether directly-attached, `{ childList: true, subtree: false }` MutationObservers are more performant than what we're doing right now, which... I assume it is.

marcustyphoon:
I've been thinking about that kind of thing too. Was hoping I'd have some flash of insight as to how to use a MutationObserver in a way that guarantees bug-free behavior in weird/future edge cases, but haven't thought of anything on that front yet. Definitely the cleaner/more performant option if we can do it.

AprilSylph:
@marcustyphoon I've been thinking hard about how to implement this technology and I've got something working that I'm very happy with. I'm not going to push it to the main branch straight away, and in fact I'd like you to review it so I might better understand the performance implications.

Given your active dedication to the project and admirable priorities, I'd like to give you "Collaborator" status for the repository. I've been thinking about this for a while now (you definitely deserve more badges on your issue/discussion comments) and if I trust you to review my own code, I trust you to review others' too.

(You'll also gain the ability to publish releases, but I would really prefer to retain exclusive control over those, so please don't unless I drop off the face of the internet a la Atesh.)

marcustyphoon:
Oh, thanks, I appreciate it! (Yay, presumably I can actually review request if there has only been a comment-review on a PR, now :D) I'll do my best!

PR #276, opened 2021-09-18
Refactor exposeTimelines and NotificationBlock's injected code

marcustyphoon:
#### User-facing changes
- None (slight potential for reducing flicker along with other changes; insignificant performance gain)

#### Technical explanation
This adjusts a pattern that gets used twice in the codebase: once in the exposeTimelines util, and once in NotificationBlock's custom inject callback.

Right now, both callbacks always inject a function that get the CSS map and use a querySelectorAll to see if they actually have to do anything. Most of the time, though, these functions should do nothing (new timelines are infrequent; most of the time mutations are not new notification drawer elements).

This:

- Generates the CSS selector outside of the inject call. (This isn't a performance gain - tumblr.getCssMap is ~free - but it's nice, and an excuse to use the previously-unused ability to give an injected function arguments.)

- Skips the inject() if the selector doesn't match anything. Now that we improved inject callback performance, this only saves like a millisecond, which is irrelevant. It does, though, make exposeTimelines potentially synchronous if there's nothing for it to do.

  Sadly, this currently removes on-load flickering from exactly zero extensions: the Hide My Posts tweak and Show Originals use `timelineObjectMemoized`, so they're still async on fresh posts, and Seen Posts queries browser storage, which if removed would bring back potential race condition bugs with having XKit Rewritten in multiple tabs.

- Names the injected functions, I suppose, so you get a nicer stack trace in the off chance one breaks. (Not sure what to name these, actually - you tend to be much better at that than I.)

#### Issues this closes
 - n/a

### DIFF
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -16,7 +16,7 @@ let blockedPostTargetIDs;
 
 const buildStyles = () => blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const notificationBlockInjected = async (notificationSelector) => {
+const unburyTargetPostIds = async (notificationSelector) => {
   [...document.querySelectorAll(notificationSelector)].forEach(async notificationElement => {
     const reactKey = Object.keys(notificationElement).find(key => key.startsWith('__reactInternalInstance'));
     let fiber = notificationElement[reactKey];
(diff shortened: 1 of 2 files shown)


AprilSylph:
HOW did I never think of doing this...

as for naming, how about...
- `notificationBlockInjected` -> `unburyTargetPostIds`
- `exposeTimelinesInjected` -> `unburyGivenPaths`

marcustyphoon:
👍 

...also, why did I call this branch "boop?" O.o

AprilSylph:
\*reads "boop" as in "bo-op" (like "noop") and imagines my friend Bo going to the hospital for a surgery\*

AprilSylph (approved):

PR #282, opened 2021-09-22
No Recommended: CSS optimization

marcustyphoon:
#### User-facing changes
- Dashboard should generally perform better if the "hide blog carousels" option is enabled

#### Technical explanation

This slightly changes the way No Recommended's option to hide blog carousels works. The current CSS causes Tumblr's code to get a bit confused: the recommended blogs carousel has infinite scrolling, and certain CSS manipulations confuse its "distance to the right edge" calculation into loading new posts in the background constantly. This fixes this behavior, which reduces the total amount of content loaded on the dash by about half.

image
(image shows results of loading page with fix, then reverting fix and waiting for about a minute)

#### Issues this closes
- n/a

### DIFF
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -4,7 +4,7 @@ import { onNewPosts } from '../../util/mutations.js';
 
 const hiddenClass = 'xkit-no-recommended-blog-carousels-hidden';
 
-const css = `.${hiddenClass} > div { display: none; }`;
+const css = `.${hiddenClass} > div { visibility: hidden; position: absolute }`;
 
 let blogCarouselSelector;
 let listTimelineObjectSelector;
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -4,7 +4,10 @@ import { onNewPosts } from '../../util/mutations.js';
 
 const hiddenClass = 'xkit-no-recommended-blog-carousels-hidden';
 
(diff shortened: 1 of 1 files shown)


AprilSylph (dismissed):
Causes massive horizontal overflow for me. `.pDYwz` ends up 4878px wide.

src/scripts/no_recommended/hide_blog_carousels.js
```diff
@@ -4,7 +4,7 @@ import { onNewPosts } from '../../util/mutations.js';
 
 const hiddenClass = 'xkit-no-recommended-blog-carousels-hidden';
 
-const css = `.${hiddenClass} > div { display: none; }`;
+const css = `.${hiddenClass} > div { visibility: hidden; position: absolute }`;
```
AprilSylph, line 7:
```suggestion
const css = `
  .${hiddenClass} { position: relative; }
  .${hiddenClass} > div { visibility: hidden; position: absolute; max-width: 100%; }
`;
```

marcustyphoon:
Oh yeah, not sure how I didn't notice that. Looking at exactly what causes it layout-wise - probably fixable by just setting 540px width, but it doesn't feel particularly clean.

AprilSylph (changes requested):
I believe the correct solution here is to limit the widths of the hidden elements to those of their parents.

src/scripts/no_recommended/hide_blog_carousels.js
marcustyphoon: Perfect, looks good to me! (I wish there was a way to checkout code suggestions like PRs :D Or guess maybe Github CLI has that, but I doubt it.)

AprilSylph (approved):

Issue #251, opened 2021-09-08
Option to hide "You're caught up." banner entirely?


amconners:
AccessKit lets you de-animate the banner, but I can't see any way to hide it and its suggestions entirely. Perhaps an option under No Recommended would fit this option.

AprilSylph:
The way I'd personally like to implement this is a little more complex than simply hiding the carousel, since the header for it is indisputably useful; I'd like to be able to modify the header text, across all languages, to remove the "Check these out:" part. The all languages part is difficult, but I think it's worth doing.

marcustyphoon:
It looks like the grammar in some of these will make it quite difficult indeed:
```
Alles gesehen? Wir haben noch viel mehr:
Seen everything? We have a lot more:
```
```
Acha que já viu de tudo? Que nada. Temos mais para você:
Do you think you've seen it all? That's nothing. We have more for you:
```
```
Ti vedo interessato. Dai un'occhiata a questi:
I see you're interested. Check out these: 
```
https://gist.github.com/marcustyphoon/e92c3fcac661fe8b6582fa0c9f01ac09

AprilSylph:
Yeah, looks like a plain separator line is the best (see: easiest) solution.

PR #295, opened 2021-10-02
Add line separator tweak

invalidCards:
#### User-facing changes
- Adds a tweak to turn the "You're caught up" section into a simple line separator.
image

#### Technical explanation
Completely hides the "You're caught up" title and turns the carousel into a 0px-high element with a 4 pixel top border that is coloured with the `white-on-dark` variable (which makes it visible in every default theme, at least).

#### Issues this closes
None necessarily, but it might be a (temporary) alternative to the proposed solution in #251.


src/scripts/tweaks.json
```diff
@@ -51,6 +51,11 @@
...
       "default": false
+    },
+    "caught_up_line": {
+      "type": "checkbox",
+      "label": "Turn \"You're caught up\" into a separating line",
+      "default": false
     }
   }
-}
+}
```
AprilSylph, line 61: Nitpick: newline at end of file is preferred

```suggestion
}

```

I recommend setting up EditorConfig!

AprilSylph (changes requested):

src/scripts/tweaks/caught_up_line.js
```diff
@@ -0,0 +1,43 @@
...
+let tagChicletCarouselItemSelector;
+let carouselWrapperSelector;
+let listTimelineObjectSelector;
+
+const createCaughtUpLine = async function () {
+  [...document.querySelectorAll(tagChicletCarouselItemSelector)]
+    .map(tagChicletCarouselItem => tagChicletCarouselItem.closest(carouselWrapperSelector))
+    .forEach(carouselWrapper => {
+      carouselWrapper?.classList.add(borderClass);
+      carouselWrapper?.closest(listTimelineObjectSelector)?.previousElementSibling.classList.add(hiddenClass);
```
AprilSylph, line 22: do we need to action on `carouselWrapper`? couldn't we just go straight to the closest `listTimelineObject`, eliminating the need to use `closest()` twice?

src/scripts/tweaks/caught_up_line.js
```diff
@@ -0,0 +1,43 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+import { onNewPosts } from '../../util/mutations.js';
+
+const hiddenClass = 'xkit-tweaks-caught-up-line-title';
+const borderClass = 'xkit-tweaks-caught-up-line-border';
+
+const styleElement = buildStyle(`
+  .${hiddenClass} { display: none; }
```
AprilSylph, line 9: I believe this would reintroduce the K-to-top bug. The recommended tag carousel option in No Recommended uses `.${hiddenClass} > div { display: none; }`.

src/scripts/tweaks/caught_up_line.js
AprilSylph: looking at https://github.com/AprilSylph/XKit-Rewritten/blob/f41443311d0fe02c6935993f073ac89ff9f7a15d/src/scripts/no_recommended/hide_tag_carousels.js#L13-L21 more closely, i think we do want to use `carouselWrapper` but the other way round.

targeting every `tagChicletCarouselItem` and then mapping them to their closest `carouselWrapper` will give us an array full of duplicate items, while targeting every `carouselWrapper` and then filtering only to those with `tagChicletCarouselItem` descendants will give us an array of only unique items we need to action on.

the filtering does use `querySelector` though so repeat actioning per tag chiclet might actually be actually more performant? maybe @marcustyphoon has the answer.

AprilSylph: i'm going to go all-in with the assumption that filtering the array down to unique items is going to be more performant, but i'm fine handling that myself.

src/scripts/tweaks/caught_up_line.js
```diff
@@ -0,0 +1,41 @@
...
+  .${hiddenClass} > div { display: none; }
+  .${borderClass} > div > div { border-top: 4px solid rgb(var(--white-on-dark)); height: 0px; }
+`);
+
+let tagChicletCarouselItemSelector;
+let listTimelineObjectSelector;
+
+const createCaughtUpLine = async function () {
+  [...document.querySelectorAll(tagChicletCarouselItemSelector)]
+    .map(tagChicletCarouselItem => tagChicletCarouselItem.closest(listTimelineObjectSelector))
```
AprilSylph, line 18:
```suggestion
    .map(tagChicletCarouselItem => tagChicletCarouselItem.closest(listTimelineObjectSelector))
    .filter((element, index, array) => array.indexOf(element) === index)
```

Does this work? This seems to be the easiest way to remove duplicate items from an array without interrupting the chain, but I'm not at liberty to test its implementation today.

AprilSylph (commented):
Otherwise seems fine, will test and hopefully approve on Monday

src/scripts/tweaks/caught_up_line.js
marcustyphoon: I don't think it'll make a performance difference (anything run once in onNewPosts will be 0ms) but it's a fun little Javascript problem to solve!

`new Set(items.map(/* */).forEach(/* */)` is probably the clearest way semantically, given that the reader knows what an ES6 Set does.

src/scripts/tweaks/caught_up_line.js
```diff
@@ -0,0 +1,41 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+import { onNewPosts } from '../../util/mutations.js';
+
+const hiddenClass = 'xkit-tweaks-caught-up-line-title';
+const borderClass = 'xkit-tweaks-caught-up-line-border';
+
+const styleElement = buildStyle(`
+  .${hiddenClass} > div { display: none; }
+  .${borderClass} > div > div { border-top: 4px solid rgb(var(--white-on-dark)); height: 0px; }
```
AprilSylph, line 10: i think it would be cleaner to just target the first `div` and set `overflow-y: hidden;` explicitly rather than relying on Tumblr's carousel CSS
```suggestion
  .${borderClass} > div {
    box-sizing: content-box;
    height: 0px;
    overflow-y: hidden;
    border-top: 4px solid rgb(var(--white-on-dark));
  }
```

AprilSylph (approved):

PR #307, opened 2021-10-09
Show Originals: Ensure user can scroll

marcustyphoon:
#### User-facing changes

In rare cases with large displays, Show Originals could result in a user's timeline rendering as short enough vertically that it fits in the browser viewport, resulting in nowhere to scroll to trigger infinite scrolling. This ensures this does not happen.

#### Technical explanation

The timeline is set to a minimum height of 100vh, or 100% of the browser viewport height, if it does not contain pagination controls. It is not right at the top of the page, so this always results in the page height being at least a bit taller than the viewport.

Duplicating `/v2/timeline/dashboard` as both string and regex doesn't feel very DRY. `getPostElements` could be overloaded to take a string for its `timeline` prop but I remember you preferring to avoid `typeof`.

#### Issues this closes
None, but discussed in discussion #107

### DIFF
--- a/src/scripts/show_originals.js
+++ b/src/scripts/show_originals.js
@@ -2,9 +2,27 @@ import { getPostElements } from '../util/interface.js';
 import { timelineObjectMemoized, exposeTimelines } from '../util/react_props.js';
 import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
+import { keyToCss } from '../util/css_map.js';
 
 const excludeClass = 'xkit-show-originals-done';
 const hiddenClass = 'xkit-show-originals-hidden';
+const lengthenedClass = 'xkit-show-originals-lengthened';
+
+const dashboardTimeline = '/v2/timeline/dashboard';
+
+const lengthenTimelines = async () => {
+  const timeline =
+    document.querySelector(`[data-timeline="${dashboardTimeline}"]:not(.${excludeClass})`);
+
+  if (timeline) {
+    timeline.classList.add(excludeClass);
+    const paginationCss = await keyToCss('manualPaginatorButtons');
+
+    if (!timeline.querySelector(paginationCss)) {
+      timeline.classList.add(lengthenedClass);
+    }
+  }
+};
 
 let showOwnReblogs;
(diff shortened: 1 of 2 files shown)


src/scripts/show_originals.js
```diff
@@ -2,9 +2,27 @@ import { getPostElements } from '../util/interface.js';
 import { timelineObjectMemoized, exposeTimelines } from '../util/react_props.js';
 import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
+import { keyToCss } from '../util/css_map.js';
 
 const excludeClass = 'xkit-show-originals-done';
 const hiddenClass = 'xkit-show-originals-hidden';
+const lengthenedClass = 'xkit-show-originals-lengthened';
+
+const dashboardTimeline = '/v2/timeline/dashboard';
```
AprilSylph, line 11:
```suggestion
const dashboardTimeline = '/v2/timeline/dashboard';
const dashboardTimelineRegExp = new RegExp(dashboardTimeline);
```

src/scripts/show_originals.js
```diff
@@ -14,6 +32,7 @@ const processPosts = async function () {
   const whitelist = whitelistedUsernames.split(',').map(username => username.trim());
 
   await exposeTimelines();
+  lengthenTimelines();
 
   getPostElements({ excludeClass, timeline: /\/v2\/timeline\/dashboard/, includeFiltered: true }).forEach(async postElement => {
```
AprilSylph, line 37:
```suggestion
  getPostElements({ excludeClass, timeline: dashboardTimelineRegExp, includeFiltered: true }).forEach(async postElement => {
```

AprilSylph (changes requested):
> Duplicating `/v2/timeline/dashboard` as both string and regex doesn't feel very DRY. `getPostElements` could be overloaded to take a string for its `timeline` prop but I remember you preferring to avoid `typeof`.

i believe you could reduce repetition by constructing the regex from the string, at least

marcustyphoon:
I... cannot believe I assumed that would not work because of the required escape characters :D

AprilSylph (approved):

PR #349, opened 2021-12-05
Show Originals: Run in additional locations

marcustyphoon:
- [x] Decide on easily understandable option labels
 
#### User-facing changes
- Adds additional locations for Show Originals to run, toggled by preferences: Peepr and the secret "Your Blog Subscriptions" dashboard (https://www.tumblr.com/timeline/blog_subscriptions)

#### Technical explanation
As per the below comment, this currently implements the logic of where to activate the extension by adding a class to the timeline elements whose contents should be examined. I can think of some other ways this could potentially work.

That I can think of, changing getPostElements to take advantage of the new pageModifications functionality via a passthrough in onNewPosts (if this is something you intend to try) should not break this.

#### Issues this closes
Discussion #337


marcustyphoon:
Discussion: do you think it makes sense to move the logic of when Show Originals should or should not run from a regex in `getPostElements` to a class that is conditionally added to the timeline?

What makes it difficult to clearly specify the intent, with the current system, is that the `timeline` attribute `/v2/blog/blognamegoeshere/posts` is essentially the same when viewing other people's blogs in Peepr and when viewing your own blog's posts. Hence, one regex does not suffice to target "only Peepr."

The two options I can think of to make this clear are
a) adding an "onlyPeepr" flag to `getPostElements` and combining the results of multiple `getPostElements` calls together per the user's settings, or 
b) not using the filtering options in `getPostElements`, and simply determining once per timeline whether Show Originals should be active or not (e.g. `timelineElement.closest('[role="dialog"]') !== null` means we're in Peepr).

The commit above implements the second one, but I'm honestly not sure which makes more sense/is more clear.

marcustyphoon:
If this topic engenders more than a cursory discussion, I'm glad to make it a github discussion or separate PR once this is merged, but: any opinion on adding some kind of visual indicator to timelines that this script affects? We do have "Original Posts" in `window.tumblr.translations`, which isn't terrible UI copy. 

(I would have made it a separate PR already, but it would obviously merge conflict with this one, and rebasing it would decouple the comment history from the comment thread, at which point I figured it was cleaner not to.)

src/scripts/show_originals.json
```diff
@@ -22,6 +22,21 @@
...
       "default": ""
+    },
+    "runOnDashboard": {
+      "type": "checkbox",
+      "label": "Run on the dashboard",
+      "default": true
+    },
+    "runOnPeepr": {
+      "type": "checkbox",
+      "label": "Run in [user friendly description of Peepr goes here]",
```
AprilSylph, line 33: hopefully the phrase "blog view" will clearly refer to `/blog/view`

```suggestion
      "label": "Run on the blog view drawer",
```

src/scripts/show_originals.js
```diff
@@ -6,36 +6,57 @@ import { keyToCss } from '../util/css_map.js';
...
+let whitelistedUsernames;
+let runOnDashboard;
+let runOnPeepr;
+let runOnBlogSubscriptions;
 
-  if (timeline) {
-    timeline.classList.add(excludeClass);
-    const paginationCss = await keyToCss('manualPaginatorButtons');
+const lengthenTimeline = async (timeline) => {
+  const paginationCss = await keyToCss('manualPaginatorButtons');
```
AprilSylph, line 20:
```suggestion
  const paginatorSelector = await keyToCss('manualPaginatorButtons');
```

this makes me realise how poorly-named `keyToCss` is. will have to do a repo-wide search and replace sometime

src/scripts/show_originals.js
```diff
@@ -6,36 +6,57 @@ import { keyToCss } from '../util/css_map.js';
...
 
-    if (!timeline.querySelector(paginationCss)) {
-      timeline.classList.add(lengthenedClass);
-    }
+  if (!timeline.querySelector(paginationCss)) {
```
AprilSylph, line 22:
```suggestion
  if (!timeline.querySelector(paginatorSelector)) {
```

src/scripts/show_originals.js
```diff
@@ -6,36 +6,57 @@ import { keyToCss } from '../util/css_map.js';
...
+const processTimelines = () => {
+  [...document.querySelectorAll(`[data-timeline]:not(.${excludeClass})`)]
+    .forEach(timeline => {
+      timeline.classList.add(excludeClass);
+
+      const onDashboard = timeline.dataset.timeline === '/v2/timeline/dashboard';
+      const isSinglePostPeepr = timeline.dataset.timeline.includes('permalink');
+      const onPeepr = timeline.closest('[role="dialog"]') !== null && isSinglePostPeepr === false;
+      const onBlogSubscriptions = timeline.dataset.timeline === '/v2/timeline' &&
+        timeline.dataset.which === 'blog_subscriptions';
```
AprilSylph, line 36: I would like multiple conditions to follow the philosophy of our array/object newline rules; either nothing gets its own line, or everything does

```suggestion
      const onBlogSubscriptions =
        timeline.dataset.timeline === '/v2/timeline' &&
        timeline.dataset.which === 'blog_subscriptions';
```

AprilSylph (changes requested):

src/scripts/show_originals.js
marcustyphoon: Yeah, agreed. I was trying to think of something extremely short that would be sufficiently clear because of how it now gets used inline with resolveExpressions so much.

AprilSylph: Oh, that's a factor I hadn't considered. I was thinking `keyToSelector`, but that would be clunky written out multiple times in the same line.

Considering what `keyToCss` returns now, how about `keyToIs`? Not sure if that's too esoteric.

src/scripts/show_originals.js
```diff
@@ -6,36 +6,58 @@ import { keyToCss } from '../util/css_map.js';
...
 
 const processPosts = async function () {
   const whitelist = whitelistedUsernames.split(',').map(username => username.trim());
 
   await exposeTimelines();
-  lengthenTimelines();
+  processTimelines();
+
+  const postElements = getPostElements({ excludeClass, includeFiltered: true })
+    .filter(postElement => postElement.closest('[data-timeline]').classList.contains(activeTimelineClass));
```
AprilSylph, line 58: Would there be a performance draw from changing this to

```suggestion
    .filter(postElement => postElement.matches(`[data-timeline].${activeTimelineClass} div');
```

?

src/scripts/show_originals.js
marcustyphoon: Hmm... yeah, that feels a bit too esoteric imo, not that I have any great ideas. I would assume that didn't do the translation step, just converted an array of classes to an is selector.

`import { cssMap } from '../util/css_map.js';` - kinda cute; kinda weird to have to rename the storage variable inside the module
`import { obfuscate } from '../util/css_map.js';` - fairly obscure
`import { map } from '../util/css_map.js';` - nice if you only ever use it inline in a template string that is obviously going to be a stylesheet; very strange in other contexts

I guess there's always
```js
export const keyToSelector = /* */;
export const map = keyToSelector;
```

or making a pattern of using `import { keyToSelector as map } from '../util/css_map.js';` where relevant, but... I dunno

src/scripts/show_originals.js
marcustyphoon: Shouldn't make any difference :) Anything that gets run like 10 times can be done however; it's the querySelectors that get run in some fashion on like 30,000 DOM nodes where performance becomes at all significant!

marcustyphoon: Oh, the other thought I had that's sort of related to this is that, in theory, you could skip this check entirely, run the extension on every timeline, and just change the selector to `${activeClass} ${hiddenClass} article { display: none; }` or whatever. Can't really think of a reason to do it that way, but it's probably a possibility for all of the exposeTimelines scripts.

AprilSylph (approved):

PR #407, opened 2022-01-24
add pageModifications engine

AprilSylph:
#### User-facing changes
- TimeFormat no longer waits for the page to stop mutating before formatting Tumblr's timestamps. This eliminates the brief time where Tumblr's unformatted timestamps are visible, but may make post rendering technically slower. Overall, this is visually interpretable as a performance improvement.

#### Technical explanation
Adds a new export, `pageModifications`, to `util/mutations.js`. This new object exposes two functions:
- `register(selector, modifierFunction)`: Adds the callback to the listener list and associates the selector with it, then runs the callback, passing all matching elements on the page as an Array argument.
- `unregister(modifierFunction)`: Removes the callback from the listener list.

Once a modifier function is registered, the ever-active MutationObserver will compile an array of every mutation between frames, and queue `onBeforeRepaint()` for before the next frame using [`Window.requestAnimationFrame()`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) (if it's not queued already). This function creates a flat array of every added node since the last frame, and then passes those added nodes (using `Array.prototype.filter()`) and their descendants which match the associated selector (using `Element.querySelectorAll()`) to the modifier functions.

I am aware of the mild performance drawback of `Element.querySelectorAll()`, but right now the code merely matches the concept: to run a function on every matching element as they're added. It also eliminates the need for building exclusions into selectors for some functions (where running the modification again, i.e. from the addon being reloaded in Firefox, has no effect on the user experience).

I don't know if this will end up replacing `onBaseContainerMutated` or any other of the current exports, but for now, I think this is a good way for features to opt into between-frame code execution.

#### Issues this closes
- closes #275

### DIFF
--- a/src/scripts/timeformat.js
+++ b/src/scripts/timeformat.js
@@ -1,5 +1,5 @@
 import moment from '../lib/moment.js';
-import { onBaseContainerMutated } from '../util/mutations.js';
+import { pageModifications } from '../util/mutations.js';
 import { getPreferences } from '../util/preferences.js';
 
 let format;
@@ -31,8 +31,8 @@ const constructRelativeTimeString = function (unixTime) {
   return relativeTimeFormat.format(-0, 'second');
 };
 
-const formatTimeElements = function () {
-  [...document.querySelectorAll('time[datetime]:not([data-formatted-time]')].forEach(timeElement => {
+const formatTimeElements = function (timeElements) {
+  timeElements.forEach(timeElement => {
     const momentDate = moment(timeElement.dateTime, moment.ISO_8601);
     timeElement.dataset.formattedTime = momentDate.format(format);
     if (displayRelative) timeElement.dataset.formattedTime += `\u2002\u00B7\u2002${constructRelativeTimeString(momentDate.unix())}`;
@@ -41,12 +41,12 @@ const formatTimeElements = function () {
 
 export const main = async function () {
   ({ format, displayRelative } = await getPreferences('timeformat'));
-  onBaseContainerMutated.addListener(formatTimeElements);
+  pageModifications.register('time[datetime]:not([data-formatted-time]', formatTimeElements);
   formatTimeElements();
 };
 
 export const clean = async function () {
-  onBaseContainerMutated.removeListener(formatTimeElements);
+  pageModifications.unregister(formatTimeElements);
   $('[data-formatted-time]').removeAttr('data-formatted-time');
 };
 
--- a/src/scripts/timeformat.js
+++ b/src/scripts/timeformat.js
@@ -41,7 +41,7 @@ const formatTimeElements = function (timeElements) {
 
 export const main = async function () {
   ({ format, displayRelative } = await getPreferences('timeformat'));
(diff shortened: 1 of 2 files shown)


AprilSylph:
(Darn, looks like closing a PR by merging another PR isn't something GitHub supports anymore...)

AprilSylph:
@marcustyphoon I meant to include you as a coauthor for the first commit since using `requestAnimationFrame` was your idea in the first place, but since squash-merges are the only merge type I have enabled here, you can just include the Co-Authored-By line when merging (or I'll try to remember to, if you approve without merging).

marcustyphoon:
Oh man, is that ever fast :D Love this!

I played around a bit with adding removedElements behavior to this engine in the same way that addedElements works (calling the callback with `{ added, removed }`), purely since I wanted to test with the most expensive selector I had onhand (scroll-to-bottom's behavior before the most recent patch), but it doesn't seem particularly useful that I can think of, since a removed node doesn't have its place in the DOM anymore, so it's difficult to identify if it was really the one you were looking for, and this API doesn't really seem suited for reliably checking the lack of something, anyway. Still a fun stress test (it's a 25x speedup or something nuts). Can you think of a use case for this?

I think the cool thing about these methods is that the callbacks don't have to be entirely synchronous; I could register 
```
pageModifications.register(someSelector, (elements) => {
  doSynchronousThing(elements);
  await doAsynchronousThing(elements);
  // or setTimeout(() => doSomethingDelayed()) if the thing to delay is not inherently async
})
```
and still get the benefit of the selectors only running on added nodes and the ability to sequence code at will. For example—race conditions aside; maybe we wouldn't actually do this one, but to illustrate the point—seen posts could dim posts you have seen in the rAF callback, since it's quick, then do the slower step of updating the browser storage asynchronously later on.

marcustyphoon (approved):

AprilSylph:
I think there might be a usecase for detecting removed nodes, and the current filtering algorithm would work on it: `Element.matches()` and `Element.querySelectorAll()` don't require the element to be in the DOM, just to be in memory!

```js
const elementNotInDOM = document.createElement('div');
const childNotInDOM = document.createElement('p');
elementNotInDOM.append(childNotInDOM);
childNotInDOM.textContent = 'Hello!';

elementNotInDOM.matches('div');                 // true
elementNotInDOM.querySelectorAll('p');          // NodeList [ p ]
elementNotInDOM.querySelector('p').textContent; // "Hello!"
```

But of course, something without an immediate usecase isn't worth implementing. Introducing extra complexity for only a potential/theoretical future payoff isn't how I like to do things.

marcustyphoon:
Yeah, the filtering works, but descendent selectors didn't seem to... which makes sense, if the elements in question have been past-tense-removed from the DOM :D That means you can tell what element it is, but not where it came from... which, in the case of the knightRiderLoader example, was pretty important! :D

AprilSylph:
Ooooh right. Yeah, no way to filter them based on where they were in the DOM.

PR #416, opened 2022-01-26
Experiment: Synchronous/0-QSA timelineObjectMemoized

marcustyphoon:
#### User-facing changes
- Even less "flickering" effect/performance impact when loading new posts (besides initial page load).

#### Technical explanation

Here's the basic idea:

Reading react props isn't, in itself, asynchronous. The only thing making timelineObject asynchronous (and this requiring a double browser repaint each time a script that needs timelineObject data wants to make a DOM mutation) is that the communication between the XKit extension context and the Tumblr page context is asynchronous. The only way to communicate synchronously between these two contexts that I know of is via DOM mutation.

Thus, this test injects a constantly running script that communicates timelineObject data synchronously via DOM mutation. Oh, and because it's a constantly running script, it uses the #407 method to avoid page-wide QSA calls, so synchronous/asynchronous aside, it's a decent performance gain.

I had three ideas for this:
- Add the timelineObject data as a data attribute on the post itself. Oh, hey, we're back in pre-redpop. Drawback: devtools element view becomes basically unusable.
- Add a dummy DOM element, put a mutationObserver on it from both sides, and communicate messages bidirectionally by writing/reading it instead of using MessageChannel. I think this could work, but it's sort of ridiculous.
- Screw it; add a dummy DOM element and just write all of the timelineObject data to it. XKit can just read the relevant data attribute whenever it needs it. Drawback: must elegantly handle race conditions or failure of the injected script on XKit's side by waiting/re-injecting it/etc. ~~_(I have not observed this being necessary, yet, actually, but still)_~~

This implements the third (without the race/failure handling logic, at the moment, and only for timelineObjectMemoized). Non-memoized timelineObject is... probably doable, though synchronously informing the page script it needs to re-query the props seems nontrivial (more like option 2).

Anyways, this all might not be worthwhile. Sure makes Show Originals feel darn fast, though, when combined with patching out the callback debounce!

### DIFF
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -1,6 +1,6 @@
 const baseContainerNode = document.getElementById('base-container');
 const dialogSelector = '[role="dialog"][aria-modal="true"]';
-const postSelector = '[tabindex="-1"][data-id]';
+export const postSelector = '[tabindex="-1"][data-id]';
 
 const ListenerTracker = function () {
   this.listeners = [];
--- a/src/util/react_props.js
+++ b/src/util/react_props.js
@@ -1,21 +1,84 @@
 import { inject } from './inject.js';
 import { keyToClasses } from './css_map.js';
+import { postSelector } from './mutations.js';
 
 const cache = {};
+
+const init = () => inject((postSelector) => {
+  const dataElement = document.createElement('div');
+  dataElement.classList.add('xkit-data');
+  document.body.appendChild(dataElement);
+
+  const writeTimelineObject = (postElement) => {
+    const id = postElement.dataset.id;
+    if (!id) return;
+
+    const reactKey = Object.keys(postElement).find(key => key.startsWith('__reactInternalInstance') || key.startsWith('__reactFiber'));
+    let fiber = postElement[reactKey];
+
+    while (fiber !== null) {
+      const { timelineObject } = fiber.memoizedProps || {};
+      if (timelineObject !== undefined) {
+        // postElement.dataset.timelineobject = JSON.stringify(timelineObject);
+        dataElement.dataset[id] = JSON.stringify(timelineObject);
+        return;
+      } else {
+        fiber = fiber.return;
+      }
+    }
+  };
+
+  [...document.querySelectorAll(postSelector)].forEach(writeTimelineObject);
+
+  const observer = new MutationObserver(function timelineObjectNew (mutations) {
+    const newPosts = mutations.flatMap(({ addedNodes }) => [...addedNodes])
+      .filter(addedNode => addedNode instanceof HTMLElement)
+      .filter(addedNode => addedNode.matches(postSelector) || addedNode.matches(`${postSelector} > div`) || addedNode.matches(`${postSelector} article`) || addedNode.querySelector(postSelector) !== null);
+
+    newPosts.forEach(writeTimelineObject);
+  });
+
+  observer.observe(document.body, {
+    childList: true,
+    subtree: true
+  });
+}, [postSelector]);
+
+init();
+let dataElement;
+
+const getTimelineObject = postID => {
+  dataElement = dataElement || [...document.body.children].find(element => element.classList.contains('xkit-data'));
+  return dataElement?.dataset?.[postID];
+};
+
+const timelineObjectMemoSync = postID => {
+  cache[postID] = new Promise(resolve => {
+    const data = getTimelineObject(postID);
+    if (data) {
+      const timelineObject = JSON.parse(data);
+      console.log('sick codes yo!');
+      resolve(timelineObject);
+    } else {
+      console.log('well this is unfortunate, you need to implement queueMicroTask/rAF fallback (which you should do anyway tbh)');
+    }
+  }).catch((error) => console.log('Error in timelineObjectMemoized!', error));
+  return cache[postID];
+};
 
 /**
  * @param {string} postID - The post ID of an on-screen post
  * @returns {Promise<object>} - The post's buried timelineObject property (cached; use
  *  timelineObject if you need up-to-date properties that may have changed)
  */
-export const timelineObjectMemoized = async postID => cache[postID] || timelineObject(postID);
+export const timelineObjectMemoized = async postID => cache[postID] || timelineObjectMemoSync(postID);
 
 /**
  * @param {string} postID - The post ID of an on-screen post
  * @returns {Promise<object>} - The post's buried timelineObject property
  */
 export const timelineObject = async function (postID) {
-  cache[postID] = inject(async id => {
+  return inject(async function timelineObjectOld (id) {
     const postElement = document.querySelector(`[tabindex="-1"][data-id="${id}"]`);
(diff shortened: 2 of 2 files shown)


AprilSylph:
I would feel uneasy about relying so heavily on a dummy DOM element. It opens up the process to potential interference from page scripts and other extension scripts.

I have thought about this approach plenty and I think sticking to MessageChannel is the best because it's the most secure; the handshake key is random enough and once the handshake is complete, nothing can interfere with it.

marcustyphoon:
Yeah, it would take a lot of fallbacks and checks for me to feel like it would be good enough to rely on, and that could get quite complicated to maintain and extend! I don't think of it as much of a security drawback, but it definitely seems like a reliability one.

I may test out mirroring #407 on the page context side just for the performance gain using the current messaging system in place later. It's super minor, though, of course, so probably not worth doing either (unless it has a surprising effect or unlocks something else interesting.)

marcustyphoon:
🤔

AprilSylph:
`document.currentScript`! Yes! It exists!!! oh my *god* how did I not find this myself

AprilSylph:
need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it need it

AprilSylph:
wait wait WAIT WAIT this means we could do element-based timelineObject reading too (instead of ID-based) by appending the script to the post container and reading react props from `document.currentScript.parentNode` couldn't we?

marcustyphoon:
wait WHAT

...huh. I... can't think of a reason why that wouldn't work, omg

AprilSylph:
> It's important to note that this will not reference the [`<script>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) element if the code in the script is being called as a callback or event handler; it will only reference the element while it's initially being processed.

I wonder if we can just do `const { currentScript } = document;` at the top and keep the reference for later? We could use this for async stuff as well if we can. Fuck a MessageChannel, just slap an attribute-listening MutationObserver on there

marcustyphoon:
Only question there is, is it possible for the script element to get deleted while we wait for an apiFetch? Probably not - React only touches things inside `root`, I would assume.

AprilSylph:
We've so far not had any problems with adding stylesheets to the `documentElement` or style elements to `document.head`, so I'd say that it's safe.

marcustyphoon:
`inject` / `injectSync`
`injectAsync` / `inject`
`injectAsync` / `injectSync`
`inject(func, { sync: true })`
`inject(func, { async: true })`

Could imagine a bunch of different forms for this.

Anyways, it sounds like you have plenty of ideas here, and this is pretty much all I've got at the moment*, so I think it makes more sense for you to play with it/work on it without me making a PR! I'll fix up my PRs if they need any changes.

*exception: this, uh, beauty:

<details>
<summary>Synchronous function coalescing work</summary>

```js
/**
 * @returns {Promise<void>} Resolves within the same event loop it is called in.
 */
const waitWithinEventLoop = async () => {
  await Promise.resolve();
  await Promise.resolve();
  await Promise.resolve();
  // css map functions mean you would need an absurd number of these at the moment - this is somewhat problematic
};

/**
 * @returns {Promise<void>} Resolves as soon as possible after the event loop it is called in.
 */
const waitAfterEventLoop = () =>
  Promise.race([
    new Promise(resolve => setTimeout(resolve)),
    new Promise(resolve => requestAnimationFrame(resolve))
  ]);

/**
 * Attempts to combine multiple identical function calls within the same event loop into one call.
 *
 * @param {Function} func - Input function to be performance optimized.
 * @param {Function} waitFunc -
 * @returns {Function} A coalesced form of the input function.
 */
const coalesce = (func, waitFunc) => {
  let cache;
  return (...args) => {
    const key = JSON.stringify(args);
    if (cache === undefined) {
      cache = {};
      waitFunc().then(() => { cache = undefined; });
    }
    if (cache[key] === undefined) {
      cache[key] = func(...args);
    }
    return cache[key];
  };
};
```
</details>

marcustyphoon:
> wait wait WAIT WAIT this means we could do element-based timelineObject reading too (instead of ID-based) by appending the script to the post container and reading react props from `document.currentScript.parentNode` couldn't we?

omfg this could finally be an opportunity to use a WeakMap for something (for the timelineObjectMemoized cache)

I have been waiting for this since the dawn of time

(...of course, it's possible that it's not worth memoizing them anymore but WHATEVER I CAN DREAM)

AprilSylph:
I think it's still worth caching them. If anything, using elements instead of post IDs means there's no reason for a cacheless version, since the cache would be invalidated by that post element being removed from the DOM.

AprilSylph:
Thinking long-term, not sure if this is worth it. Chrome's Manifest V3 will force us to switch to some arcane system where we have a service worker and are forced to inject stuff via `scripting.executeScript` (direct injection via content script is blocked by CSP), which is asynchronous (and we'll have to communicate with the service worker asynchronously, too). So while we could make a bunch of stuff synchronous now, we'll have to either revert it or drop off the Web Store come January 2023.

marcustyphoon:
Oh yeah, that is true! Darn. And I guess you can't send DOM node references via structured clone, so there's no way to do do element-based timelineObject reading in that case either.

(Hm, now I want to try an MV3-based version of this codebase :D I have some performance enhancement ideas that *should* work with that system, but I definitely don't know its quirks!)

AprilSylph:
Hmm, I might be wrong. If we rely on `document.currentScript` *heavily*, we might be able to pass MV3's CSP by injecting static scripts instead of inline ones. We could pass arguments as `data-*` attributes. And outside of arguments, every call to `inject()` is essentially static already.

I'll have to test if this works as I expect it to.

AprilSylph:
Okay, the good news is that the CSP works as I expected. We pass the CSP for Chrome extensions by specifying a script `src`, and we pass Tumblr's CSP by setting the `nonce` as we usually do.

The bad news is that adding the script is not synchronous, probably *because* it's not an inline script. The result handling has to go inside a `load` event listener.

[Here's my code](https://gist.github.com/AprilSylph/b84392939f92a5df4e4743d026ca3074). I guess what I'll test next is keeping the script inline but static, and specifying a hash in the manifest CSP.

AprilSylph:
Damn. Chrome considers `sha256-*` CSP values to be insecure and refuses to load the extension. I guess MV3 really does block inline scripts outright.

marcustyphoon:
Darn. Hmm, sounds like outside of moving nontrivial amounts of code into the page context/main world, writing code that assumes that injection can be a guaranteed synchronous process or that we can place the script element wherever we want in order to skip a querySelector call certainly has the potential to be wasted work. (Which is a bit sad, but it's not like the upside was huge anyways, in my mind. Still having to spam `await` is a bit unfortunate, of course, but hey, it's not promise chains.)

AprilSylph:
We should be able to still skip `querySelector` calls - `document.currentScript` still works within bundled scripts, so we can still use the script's position in the DOM.

marcustyphoon:
Ah, using the `script.src` thing - right, okay. Yep, I'm with you.

PR #422, opened 2022-01-27
Notifications framework: tiny optimization

marcustyphoon:
I don't know that I would actually make this change! It is a bit silly, but it works.

#### User-facing changes
Plus: Completely insignificant performance gain (<1ms) when Tumblr does any DOM mutations.
Minus: Changing to/from mobile view/peepr while a notification popup is active does not transfer the notification popup to the new UI. Anyone who finds the toast container with dev tools when it's not in the right place will be confused.

#### Technical explanation
This defers moving the toast container to the right place until it actually gets used to display a notification.

#### Issues this closes
n/a

### DIFF
--- a/src/util/notifications.js
+++ b/src/util/notifications.js
@@ -1,4 +1,3 @@
-import { pageModifications } from './mutations.js';
 import { keyToCss } from './css_map.js';
 
 const toastContainer = Object.assign(document.createElement('div'), { id: 'xkit-toasts' });
@@ -21,14 +20,13 @@ const addToastContainerToPage = async () => {
   }
 };
 
-pageModifications.register('*', addToastContainerToPage);
-
 const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
 
 /**
  * @param {string} textContent - Text to display to the user as a notification
  */
 export const notify = async textContent => {
+  await addToastContainerToPage();
   const toast = Object.assign(document.createElement('div'), { textContent, className: 'visible' });
   toastContainer.append(toast);
   await sleep(4000);


marcustyphoon:
I wasn't sure if this sort of "discussion, but in the form of a proposed code change, but I'm not necessarily saying I think you should do it" was more appropriate as a pull request or a discussion (or something else, or just post it on tumblr :P). It's something I've been meaning to ask about. I would completely understand wanting PRs to be, you know, actual requests to have code pulled. Happy to do whichever you prefer!

AprilSylph:
Hmm, maybe not too valuable as a performance gain, but I can see this making sense from a user experience standpoint - if I close peepr, it's natural for any notifications within it to also disappear, rather than move to the sidebar only to start disappearing nigh-immediately. If that comes with a tiny performance gain as a side-effect, I think I'd be happy about this!

I'll need to test this to see if I'm right about it being a UX improvement, but I might not get around to it today. I can imagine it being jarring for in-sidebar notifications to disappear when moving between regular pages.

AprilSylph:
Also, I'm perfectly happy to use pull requests as a type of discussion with code attached, so long as you don't spend so much time on preparing the PR that it delays discussion on the subject significantly (which by my current standards is a week or more).

AprilSylph (approved):
I'm happy with this as a UX improvement. Toasts in the sidebar don't follow you to peepr or vice-versa, which more clearly separates peepr from the main timeline.

I wouldn't say that notifications disappearing between page switches is an improvement, but it's not exactly a drawback either. Tumblr has no illusion of the sidebar being the same between pages (just going from the dashboard to /blog/april rerolls the Radar), and since they're volatile elements anyway, there's only a 5-second window available per toast to even observe this behaviour.

marcustyphoon:
Interesting, I hadn't considered that this behavior could actually be an improvement until you brought it up, but I agree! 

> so long as you don't spend so much time on preparing the PR that it delays discussion on the subject significantly

(Is this referring to #349? I can't tell!)

AprilSylph:
Nope, that doesn't count because that was spawned from a discussion itself (#337).

I just want to avoid anyone working too hard on an idea just for the idea itself to be shot down.

Issue #412, opened 2022-01-25
Quick reblog does not remember last blog selected between pages


irradiatedsnakes:
### Platform

Windows 10

### Browser

Firefox 96.0.2 (64 bit)

### Addon version

0.14.1

### Details

When the "remember the last blog selected in the popup" option for quick reblogs is selected, the selection only lasts on the current page- when a new page is opened or the page is reloaded, it defaults to the primary blog and the correct blog has to be selected again.

AprilSylph:
That's correct. XKit Rewritten is account-neutral and does not store your blog name in its storage anywhere.

PR #429, opened 2022-01-31
Quick Reblog: Add default blog text preference

marcustyphoon:
I recognize the desire not to silently store information about the user's account within extension storage, but I also think having to remember to toggle Quick Reblog's blog selection every time you open a new tab is quite inconvenient for people who do not use their main blogs. Here is a sort of "please all and you will please none" solution, in which the user explicitly opts in to the extension storing their data by manually typing in their preferred default blog... or blogs, if they have multiple accounts.

#### User-facing changes
- Adds a textbox preference to Quick Reblog allowing the default-selected blog to be customized.

#### Technical explanation
The script searches the user's blog list for a matching preference entry in main(), doing nothing if no entry matches.

I looked at `Intl.Collator.prototype.compare` as a fancier version of `a.toUpperCase() === b.toUpperCase()` with better internationalization/special character apparently, though because Tumblr usernames are URL-safe, I think it makes no difference whatsoever. Could be nice to know about for some future use case though,

#### Issues this closes
Addresses core request of #220, #317, #412.

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -249,10 +249,12 @@ export const main = async function () {
   if (defaultBlogs.length) {
     for (const defaultBlog of defaultBlogs.split(/[\s,]+/).filter(Boolean)) {
       activeDefaultBlog = userBlogs.find(({ name }) => name.toUpperCase() === defaultBlog.toUpperCase())?.uuid;
-      if (activeDefaultBlog) break;
+      if (activeDefaultBlog) {
+        blogSelector.value = activeDefaultBlog;
+        break;
+      }
     }
   }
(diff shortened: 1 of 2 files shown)


AprilSylph:
I'm questioning whether I'm adamant enough about not storing blogs' UUIDs to merge this instead of going for a cleaner solution that makes the blog remembering option work like #412 wants it to. Quick Reblog has too many preferences as it is 😫

marcustyphoon:
Amen to that! We could do some funny things like running the UUIDs through SHA-256, even. Can SubtleCrypto.digest() be run in a web extension context? Might be fun just to get to know we did it.

PR #430, opened 2022-01-31
Quick Reblog: Fix console error on drafts page

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
This fixes a console error when alreadyReblogged tries to process the reblog link on posts that don't have one, like when viewing the drafts page.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -189,7 +189,7 @@ const processPosts = async function () {
 
     if (alreadyRebloggedList.includes(rootID)) {
       const reblogLink = postElement.querySelector('footer a[href*="/reblog/"]');
-      if (!reblogLink) { return; }
+      if (!reblogLink) { continue; }
       const buttonDiv = reblogLink.parentNode;
       makeButtonReblogged({ buttonDiv, state: 'published' });
     }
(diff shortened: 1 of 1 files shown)


src/scripts/quick_reblog.js
```diff
@@ -189,6 +189,7 @@ const processPosts = async function () {
 
     if (alreadyRebloggedList.includes(rootID)) {
       const reblogLink = postElement.querySelector('footer a[href*="/reblog/"]');
+      if (!reblogLink) { return; }
```
AprilSylph, line 192: This is inside a `for` loop rather than a `forEach` callback.

```suggestion
      if (!reblogLink) { continue; }
```

AprilSylph (changes requested):

src/scripts/quick_reblog.js
AprilSylph: It may seem reasonably safe to see one post without a reblog link and assume all unprocessed posts also don't have one, but I'm sure there's a way to break this assumption.

marcustyphoon: ...Ah. Well. Now I know how it feels :D And I have to say, it doesn't feel great!

AprilSylph (approved):

PR #432, opened 2022-01-31
Experiment: add array override to descendentSelector

marcustyphoon:
- [ ] update jsdoc

#### User-facing changes
- none

#### Technical explanation
This adds functionality to the descendantSelector utility function allowing developers to pass arrays of strings, which will not be translated. This is primarily to facilitate a concise way to use element name selectors in descendantSelector i.e. `await descendantSelector(['article'], ['footer'], 'controls')`.

(It also has the side benefit of allowing descendantSelector to be used to compose intermediate-complexity selector arrays with plain translated selectors in a somewhat more readable way than directly invoking `cartesian`, if that ever becomes necessary. This complexity is obviously to be avoided where possible, of course.)

I also thought of and played with e.g. `await descendantSelector('element:article', 'element:footer', 'controls')` / `await descendantSelector('untranslated:article', 'untranslated:footer', 'controls')`, which is arguably easier to comprehend, though it's less clean to type and doesn't allow for the bonus composition behavior.

_Yes, it's a slippery slope to creating 15 utility functions for CSS selector composition that wind up being substantially more complex than just composing strings manually or giving up and writing translated classes into the dom directly with a mutationObserver, but this by itself feels okay to me._ 

#### Issues this closes
n/a

### DIFF
--- a/src/util/css_map.js
+++ b/src/util/css_map.js
@@ -48,18 +48,3 @@ export const descendantSelector = async function (...keys) {
     .map(selectors => selectors.join(' '))
     .join(', ');
 };
-
-// tests
-(async () => {
-  const test1 = await descendantSelector('post', 'footer', 'controls');
-  console.log(test1);
-  console.log([...document.querySelectorAll(test1)]);
-  const test2 = await descendantSelector(['article'], ['footer'], 'controls');
-  console.log(test2);
-  console.log([...document.querySelectorAll(test2)]);
-
-  const specificTimelineClasses = (await keyToClasses('timeline'))
-    .map(className => `.${className}:not[data-timeline="whatever"]`);
-  const test3 = await descendantSelector(specificTimelineClasses, 'listTimelineObject');
-  console.log(test3);
-})();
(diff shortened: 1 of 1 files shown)


src/util/css_map.js
```diff
@@ -36,8 +36,12 @@ export const descendantSelector = async function (...keys) {
   const sets = [];
 
   for (const key of keys) {
-    const set = await keyToClasses(key);
-    sets.push(set.map(className => `.${className}`));
+    if (Array.isArray(key)) {
+      sets.push(key);
```
AprilSylph, line 40: This means we'd be able to put more than one selector part per array to essentially emulate `:any()`, wouldn't it? I like this idea.

AprilSylph: Hold the phone, `:is()` has actual browser support now. Perhaps we should be looking at removing cartesian entirely.

marcustyphoon: In the 3 minutes between these two comments, I clicked the relevant MDN and my eyes bugged out.

AprilSylph:
One thing I've been playing around with in my head for a while is a completely different function that accepts a template literal and transforms the passed promises (i.e. `keyToCss` calls) and leaves everything else alone, but I was having trouble figuring out how to do that with cartesian products. With CSS `:is()` support across the three main browser engines, I think I'd like to pursue using that instead.

I'll keep this open for now, but if I'm successful in implementing such a thing, I'll close this via that commit.

PR #437, opened 2022-02-01
Add tweak to hide note count type indicators

marcustyphoon:
#### User-facing changes
- Adds a tweak to hide the colorful note count type indicators next to the note count on posts.

#### Technical explanation
Are you supposed to add a co-author credit to your merge commit if 99% of your PR is copy-pasted? It seems polite, right? :D

Anyways, I've been running `.📝noteCountTypes { display: none !important; }` in Stylus with my translation script pretty much ever since I got this a/b test.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -13,6 +13,11 @@
       "label": "Attempt to revert the post footer redesign",
       "default": false
     },
+    "hide_note_count_types": {
+      "type": "checkbox",
+      "label": "Hide the type indicators next to the note count",
+      "default": false
+    },
     "slim_filtered_screens": {
       "type": "checkbox",
       "label": "Use a slim layout for filtered posts",
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -15,7 +15,7 @@
     },
     "hide_note_count_types": {
       "type": "checkbox",
-      "label": "Hide the type indicators next to the note count",
+      "label": "Hide the note type badges next to the note count",
       "default": false
     },
     "slim_filtered_screens": {
--- a/src/scripts/tweaks/hide_note_count_types.js
+++ b/src/scripts/tweaks/hide_note_count_types.js
@@ -4,5 +4,5 @@ import { buildStyle } from '../../util/interface.js';
 const styleElement = buildStyle();
 keyToCss('noteCountTypes').then(selector => { styleElement.textContent = `${selector} { display: none !important; }`; });
 
-export const main = async ()  => document.head.append(styleElement);
+export const main = async () => document.head.append(styleElement);
 export const clean = async () => styleElement.remove();
--- a/src/scripts/tweaks/hide_note_count_types.js
+++ b/src/scripts/tweaks/hide_note_count_types.js
@@ -1,0 +1,14 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const styleElement = buildStyle();
+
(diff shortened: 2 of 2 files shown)


src/scripts/tweaks/hide_note_count_types.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const main = async function () {
+  const selector = await keyToCss('noteCountTypes');
+  styleElement.textContent = `${selector} { display: none !important; }`;
+  document.head.append(styleElement);
+};
+
+export const clean = async function () {
+  styleElement.remove();
+};
```
AprilSylph, line 14: lack of top-level await support in eslint 7 makes april a sad girl, but until eslint-config-standard gets updated we can at least do shit like this

```suggestion
const styleElement = buildStyle();
keyToCss('noteCountTypes').then(selector => { styleElement.textContent = `${selector} { display: none !important; }`; });

export const main = async ()  => document.head.append(styleElement);
export const clean = async () => styleElement.remove();
```

(if this looks worse to you feel free to just resolve this conversation without committing it)

src/scripts/tweaks.json
```diff
@@ -13,6 +13,11 @@
       "label": "Attempt to revert the post footer redesign",
       "default": false
     },
+    "hide_note_count_types": {
+      "type": "checkbox",
+      "label": "Hide the type indicators next to the note count",
```
AprilSylph, line 18: not sure "indicators" is the best descriptor here

```suggestion
      "label": "Hide the note type badges next to the note count",
```

AprilSylph (changes requested):

src/scripts/tweaks/hide_note_count_types.js
marcustyphoon: I'm for it! Glad to see that rather than co-author for writing the script I copy-pasted, you can get co-author for improving the parts that I copy-pasted :P

src/scripts/tweaks.json
marcustyphoon: 👍

src/scripts/tweaks/hide_note_count_types.js
```diff
@@ -0,0 +1,8 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const styleElement = buildStyle();
+keyToCss('noteCountTypes').then(selector => { styleElement.textContent = `${selector} { display: none !important; }`; });
+
+export const main = async ()  => document.head.append(styleElement);
```
AprilSylph: OH NO I FUCKED IT

```suggestion
export const main = async () => document.head.append(styleElement);
```

AprilSylph: See, this is why I want a new keyboard.

marcustyphoon: I always figured my struggles with my full-height desktop keyboard were a me thing... but they're presumably primarily because I can't touch-type, in my case. I think since I grew up exclusively using a laptop, it was never inconvenient enough to stop just looking down at the keyboard, and at this point... okay, yeah, I should probably force myself to learn even after all these years.

AprilSylph (approved):
I have no way of testing this (still don't have the badges update, nor the new notes UI) but the code looks good. I'll let you merge this yourself after testing that it still works after the edits (if you haven't already).

marcustyphoon:
...Wait, the new notes UI isn't universal? Wow, okay.

PR #438, opened 2022-02-01
Revert "Scroll to Bottom: Query performance optimization (#431)"

marcustyphoon:
#### User-facing changes
- Scroll to Bottom is 0.02% slower. I think it'll survive.

#### Technical explanation
This reverts the performance-optimization-at-the-expense-of-clean-code of #431, as the (single) selector created by the new resolveExpressions function makes it unnecessary. (I haven't done what I would consider exhaustive tests of `:is()` performance yet, but the original 4-level cartesian took 500ms, the 3-level cartesian took 130ms, and both the single selector in #431 and the new singly nested :is() take less than 1.5ms, so... yeah.)

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -3,16 +3,13 @@ import { translate } from '../util/language_data.js';
 import { pageModifications } from '../util/mutations.js';
 
 let knightRiderLoaderSelector;
-let loaderSelector;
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
BEAUTIFUL. I was hoping using `:is()` would be a performance improvement over cartesian-product selector lists, and I guess I was right!

PR #442, opened 2022-02-02
Inject utility: Ensure functions have names

marcustyphoon:
#### User-facing changes
- None. Like, really, really none.

#### Technical explanation
This is super minor; feel free to close or modify. This names some of the common injected functions ~~by converting them from arrow functions (whose ES6 "names" don't make it through `toString()`, unfortunately).~~ This is exclusively to make them not show up as "anonymous" in devtools and has no other benefit.

Edit: This now queries injected functions' names before inserting them into the script element, then ensures that in the page context they will have an xkit-specific name based on their original name, with a fallback for anonymous injected functions.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const unburyTargetPostIds = async (notificationSelector) => {
+const unburyTargetPostIds = async function unburyTargetPostIds (notificationSelector) {
   [...document.querySelectorAll(notificationSelector)]
     .filter(({ dataset: { targetPostId } }) => targetPostId === undefined)
     .forEach(async notificationElement => {
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const unburyTargetPostIds = async function xkitUnburyTargetPostIds (notificationSelector) {
+const unburyTargetPostIds = async (notificationSelector) => {
   [...document.querySelectorAll(notificationSelector)]
     .filter(({ dataset: { targetPostId } }) => targetPostId === undefined)
     .forEach(async notificationElement => {
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const xkitUnburyTargetPostIds = async (notificationSelector) => {
+const unburyTargetPostIds = async (notificationSelector) => {
   [...document.querySelectorAll(notificationSelector)]
     .filter(({ dataset: { targetPostId } }) => targetPostId === undefined)
     .forEach(async notificationElement => {
@@ -39,7 +39,7 @@ const xkitUnburyTargetPostIds = async (notificationSelector) => {
     });
 };
 
-const processNotifications = () => inject(xkitUnburyTargetPostIds, [notificationSelector]);
+const processNotifications = () => inject(unburyTargetPostIds, [notificationSelector]);
 
 const onButtonClicked = async function ({ currentTarget }) {
   const postElement = currentTarget.closest('[data-id]');
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const unburyTargetPostIds = async (notificationSelector) => {
+const xkitUnburyTargetPostIds = async (notificationSelector) => {
   [...document.querySelectorAll(notificationSelector)]
     .filter(({ dataset: { targetPostId } }) => targetPostId === undefined)
     .forEach(async notificationElement => {
@@ -39,7 +39,7 @@ const unburyTargetPostIds = async (notificationSelector) => {
     });
 };
 
-const processNotifications = () => inject(unburyTargetPostIds, [notificationSelector]);
+const processNotifications = () => inject(xkitUnburyTargetPostIds, [notificationSelector]);
 
 const onButtonClicked = async function ({ currentTarget }) {
   const postElement = currentTarget.closest('[data-id]');
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const unburyTargetPostIds = async function unburyTargetPostIds (notificationSelector) {
+const unburyTargetPostIds = async function xkitUnburyTargetPostIds (notificationSelector) {
   [...document.querySelectorAll(notificationSelector)]
     .filter(({ dataset: { targetPostId } }) => targetPostId === undefined)
     .forEach(async notificationElement => {
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -64,7 +64,7 @@ export const inject = (asyncFunc, args = []) => new Promise((resolve, reject) =>
   callbacks.set(callbackId, [resolve, reject]);
 
   const script = document.createElement('script');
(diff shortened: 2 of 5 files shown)


AprilSylph (changes requested):
Kinda cringe of JavaScript to make us type function names twice!

I'm okay with this idea, but if we're going to do this at all, I would rather have *all* injected functions named.

src/scripts/notificationblock.js
```diff
@@ -19,7 +19,7 @@ const buildCss = () => blockedPostTargetIDs.length === 0
   ? ''
   : blockedPostTargetIDs.map(id => `[data-target-post-id="${id}"]`).join(', ').concat(' { display: none; }');
 
-const unburyTargetPostIds = async (notificationSelector) => {
+const unburyTargetPostIds = async function unburyTargetPostIds (notificationSelector) {
```
AprilSylph, line 22:
```suggestion
const unburyTargetPostIds = async function xkitUnburyTargetPostIds (notificationSelector) {
```

src/util/react_props.js
```diff
@@ -15,7 +15,7 @@ export const timelineObjectMemoized = async postID => cache[postID] || timelineO
  * @returns {Promise<object>} - The post's buried timelineObject property
  */
 export const timelineObject = async function (postID) {
-  cache[postID] = inject(async id => {
+  cache[postID] = inject(async function unburyTimelineObject (id) {
```
AprilSylph, line 18:
```suggestion
  cache[postID] = inject(async function xkitTimelineObject (id) {
```

src/util/react_props.js
```diff
@@ -32,7 +32,7 @@ export const timelineObject = async function (postID) {
   return cache[postID];
 };
 
-const unburyGivenPaths = async (selector) => {
+const unburyGivenPaths = async function unburyGivenPaths (selector) {
```
AprilSylph, line 35:
```suggestion
const unburyGivenPaths = async function xkitUnburyGivenPaths (selector) {
```

src/util/tumblr_helpers.js
```diff
@@ -6,10 +6,9 @@ import { inject } from './inject.js';
  * @returns {Promise<Response|Error>} Resolves or rejects with result of window.tumblr.apiFetch()
  */
 export const apiFetch = async function (...args) {
-  return inject(
-    async (resource, init) => window.tumblr.apiFetch(resource, init),
-    args
-  );
+  return inject(async function callApiFetch (resource, init) {
```
AprilSylph, line 9:
```suggestion
  return inject(async function xkitApiFetch (resource, init) {
```

AprilSylph (changes requested):
I also think there would be a benefit to a naming scheme for these functions to make them clearly identifiable as XKit-produced. My suggestion is simply `xkit` plus the name of the constant or containing function.

marcustyphoon:
Yeah, I definitely don't love the syntax, even when writing the injected function inline. Feels pretty clunky. And I assume the only one looking at this code in devtools is probably me, and I can just manually name them when testing things (and at this point I can identify most of them in the profiler from what they look like :P) I guess it is good to have a name in the stack trace if an injected function throws, though.

...actually, wait, does the catch() not work?

```js
inject(async function testFunctionName () {
  JSON.parse('fkghdkg');
}).then(console.log);
```

Screen Shot 2022-02-02 at 1 10 00 PM

Well, we can discuss that elsewhere.

Anyway, I guess a different option would be adding a third parameter to `inject()` to request a name, and modify the injected script skeleton to like, 
```js
`const ${name} = {${asyncFunc.toString()}};
${name}(...${JSON.stringify(args)}).then(/* */).catch(/* */)`
```
Kind of feels cleaner, actually? Makes it more explicit whether you're supplying a function name or not. (Of course, passing a named non-arrow function and a different name parameter would ignore the parameter in favor of the actual function name in that case, but a) as long as it's not anonymous, whatever and b) not using arrow functions? blasphemy).

AprilSylph:
It looks like, actually, you should be able to get the constant name in `inject` via the `arguments` object?

```js
const func = () => {};
const args = [];
(function (f, a) {
  console.log(arguments[0].name)
})(func, args);
// => 'func'
```

marcustyphoon:
Trouble is, we're stringifying the function and (effectively) `eval()`ing it.
```js
const func = () => {};

const stringified = func.toString()
console.log(stringified); 
// "() => {}"

eval(`
console.log((${stringified}).name)
`); 
// ""
```

My idea of manually passing a function name as a parameter to inject() was totally silly, though, I just realized: inject() has the real function, and thus knows the real function name.

```diff
export const inject = (asyncFunc, args = []) => new Promise((resolve, reject) => {
  const callbackId = Math.random();
  callbacks.set(callbackId, [resolve, reject]);

  const script = document.createElement('script');
+ const name = asyncFunc.name || 'xkitInjected';

  script.setAttribute('nonce', nonce);
  script.textContent = `{
-   (${asyncFunc.toString()})(...${JSON.stringify(args)})
+   const ${name} = ${asyncFunc.toString()};
+   ${name}(...${JSON.stringify(args)})
    .then(result => window.xkit$${injectKey}.messagePort.postMessage({
```
Didn't test it yet, but it should work and require almost no changes to scripts.

marcustyphoon:
Not like it matters, but what would you do with the mega editor function as far as name, extract-or-inline, and extract-to-local-or-module? Technically it's just a generic fetch, right?

src/util/inject.js
```diff
@@ -64,10 +64,12 @@ export const inject = (asyncFunc, args = []) => new Promise((resolve, reject) =>
   callbacks.set(callbackId, [resolve, reject]);
 
   const script = document.createElement('script');
+  const name = asyncFunc.name || 'xkitInjected';
```
AprilSylph, line 67:
```suggestion
  const name = `xkit$${asyncFunc.name || 'injected'};
```

AprilSylph (changes requested):
If we're handling the name passing in `inject()` anyway, we can put an identifying prefix there and not have to change anything else.

src/util/inject.js
marcustyphoon: Ooh, good point.

src/util/inject.js
```diff
@@ -64,10 +64,12 @@ export const inject = (asyncFunc, args = []) => new Promise((resolve, reject) =>
   callbacks.set(callbackId, [resolve, reject]);
 
   const script = document.createElement('script');
+  const name = `xkit$${asyncFunc.name || 'injected'};
```
AprilSylph, line 67: I LOVE TYPOS I LOVE TYPOS I LOVE TYPOS
```suggestion
  const name = `xkit$${asyncFunc.name || 'injected'}`;
```

AprilSylph: I blame VS Code for automatically creating closing quotes/brackets/parens when you type an opening one. I've gotten lazy.

marcustyphoon:
...this does, though, beg the question: if you don't name the inner function as a named inline non-arrow function... then what other name do you give the injected getCssMap function, which you are injecting in getCssMap, in order to call Tumblr's function getCssMap?

AprilSylph:
...`getCssMapFromPage`? `pageContextGetCssMap`? 😅

AprilSylph:
Maybe the inline injections can stay inline if they're only called once per pageload.

marcustyphoon:
"Right, good, yes. Now I'll just name the apiFetch injection, uh, wait, hm."

marcustyphoon:
I mean, honestly... all of the one-liners can't really fail and cause a stack trace that would be confusing anyways. Shall we just make the change to inject(), make a habit of extracting any injected function big enough to make sense to extract, and just leave the simple ones that would naturally be inline as-is?

AprilSylph:
Yeah, I think that's best.

I was workshopping a comment saying you could do something like
```js
export const apiFetch = async (...args) => inject(
  async function apiFetch (resource, init) { return window.tumblr.apiFetch(resource, init); },
  args
);
```
but it might not be worth doing. Maybe though?

marcustyphoon:
Yeah, that's what I had in the original PR commit :D

AprilSylph:
I think we're gonna have to go with a step-by-step review here. Let's see how things look with *just* the `inject()` change; I'm happy with that code, but would you be happy with the results?

AprilSylph:
> I think I am losing my mind

Me too, just a little bit. I like increasing debuggability though!

marcustyphoon:
Alright, this is exclusively:

- change to inject() that uses function names where applicable, and `xkit$injected` otherwise
- Make timelineObject use the same structure as exposeTimelines

I think those both make sense. `apiFetch`: shrug! We could leave it as is, name it inline, or name it something else.

AprilSylph (approved):
I'm happy with this, at least. It definitely makes sense to extract the injected part of `timelineObject` and makes the logic easier to follow.

The next step would be to extract more large injected functions, if they can be given sane names within their modules. Do you want to do that here, or in a subsequent PR?

marcustyphoon:
I think the only other one in the repository is editPostFormTags! If you have a reasonable name idea for that one... I do not :D

AprilSylph:
Inputs handled by React are referred to as "controlled inputs", right? How about `controlTagsInput`?

marcustyphoon:
Ooh, I like it! And indeed they are!

AprilSylph (approved):
I think I'd like to merge this as-is, then think about naming injected function expressions later. Do you want to rename this PR to make the commit title more accurate? :D

marcustyphoon:
Hold on, now we have to come up with a clear and concise name for a PR too? Wild 😆

AprilSylph:
I know right? What a nightmare!

(But it IS part of the contributing guidelines...)

PR #446, opened 2022-02-03
Page Modifications utility: Fix rare theoretical lag in inactive tabs(?)

marcustyphoon:
So... I have absolutely no idea how to test this.

#### User-facing changes
- Fixes a potential bug where page modifications that occur when the window is not focused might not be processed by XKit until the next mutation. (I think this would only occur when Tumblr background-fetches to do things like add notification badges on the menu icons or fetches ads, neither of which XKit Rewritten currently does anything with... so, practically, nothing.)

#### Technical explanation

I noticed this happening with my personal "hide the activity badge on the notification icon" userstyle/userscript combination, which works essentially the same way as XKit Rewritten's page modification engine.

`window.requestAnimationFrame` is not fired on inactive tabs, since it's... not exactly designed for this use case. So page modification updates could be skipped until the next mutation that happens when the tab is active again. XKit Rewritten's mutationsPool system ensures that no mutations get lost in the wash, unlike my personal script, but this could still have a slight visual effect in theory.

This fixes the problem by falling back to a setTimeout if rAF hasn't been called ~~by the time the event loop empties~~ after a delay.

#### Issues this closes
n/a

### DIFF
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -86,7 +86,10 @@ const onBeforeRepaint = () => {
 const observer = new MutationObserver(mutations => {
   mutationsPool.push(...mutations);
   if (repaintQueued === false) {
-    window.requestAnimationFrame(onBeforeRepaint);
+    Promise.race([
+      new Promise(resolve => window.requestAnimationFrame(resolve)),
+      new Promise(resolve => setTimeout(resolve, 0))
+    ]).then(onBeforeRepaint);
     repaintQueued = true;
   }
 });
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -88,7 +88,7 @@ const observer = new MutationObserver(mutations => {
   if (repaintQueued === false) {
     Promise.race([
       new Promise(resolve => window.requestAnimationFrame(resolve)),
-      new Promise(resolve => setTimeout(resolve, 0))
+      new Promise(resolve => setTimeout(resolve, 500))
     ]).then(onBeforeRepaint);
     repaintQueued = true;
   }
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -83,12 +83,30 @@ const onBeforeRepaint = () => {
   }
 };
 
+const test = {};
+
 const observer = new MutationObserver(mutations => {
   mutationsPool.push(...mutations);
   if (repaintQueued === false) {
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Whoops, I didn't realize that the event loop gets emptied multiple times between repaints, so `setTimeout(callback, 0)` is much too fast of a fallback and can result in the page modification handler firing multiple times before the repaint, which defeats the whole point. 500ms is enough for this to be fairly rare; 1000ms or more seems perfectly reasonable as well.

AprilSylph:
do `requestAnimationFrame` callbacks not fire immediately upon re-focusing a tab which has had mutations? I would have assumed the next repaint would be "when the tab needs to be shown", rather than happening silently without triggering `requestAnimationFrame` callbacks.

marcustyphoon:
Trouble is, I'm not really sure how to test it! This is my best guess at the cause of what I have occasionally observed, but again, that's with a totally different script.

I guess I can run a custom XKit Rewritten extension that does something noticeable like turns the activity badges pink, then see if I can get it to fail without this PR, then see if this PR fixes it.

AprilSylph:
If your script doesn't have a mutationsPool mechanism, are you sure they're not somehow being discarded before the tab is refocused?

marcustyphoon:
Mine is stateless; it runs `querySelectorAll('*')` on the whole page every time, ignoring the mutationObserver output. (Could maybe be optimized, but hey, it's dev-only at the moment :D) So I don't thiiiiink so?

marcustyphoon:
...actually, now that I think about it, maybe Tumblr doesn't add any nodes when the relevant thing occurs, and it's just replacing the classes... and I'm filtering by addedNodes...

Er. Actually. Uh. Lemme just um. Test something real quick.

marcustyphoon:
Yeah, I think I just had a bug in my code and I misinterpreted this as being the cause!

marcustyphoon:
> 'time until rAF:' 1206146.099999994

Yep, you're right: the animation frame gets called on page focus. Certainly this means we have to be a bit careful assuming in the callbacks that the elements they are being supplied by pageModifications still exist, since they might in rare cases have been added ages ago, but that's not hard and we have to do that anyway.

(Also, aw, do closed PRs not update with commits to their source branches? That's sad.)

PR #450, opened 2022-02-04
Quick reblog: Save last selected blog in local storage

marcustyphoon:
Honestly, I'm not sure if this was something you wanted to work on yourself, and if so, I am 100% fine with that! I just wanted to learn about the crypto functions available in javascript and the way XKit Rewritten interfaces with storage... and then wound up learning about attaching event listeners and how to convert an ArrayBuffer to base64.*

*hope you're in Node; if not, reject all of the solutions as awful and do something entirely different.

Pretty sure there are ways to do this in less code. (Honestly, if you want to use any of this, feel free to copy-paste/I could make it a branch in this repo if you want.)

#### User-facing changes
- Quick reblog's "remember the last selected blog in the popup" option persists through page reloads, yet the local storage contains no record of who the user is.

#### Technical explanation
Instead of saving the UUID of the most recently selected blog (and the UUID of your main, so your preferences are remembered on a per-account basis), this saves the SHA256 hash of both UUIDs instead. This identifies them without storing them (unless you can break SHA256, in which case you have better things to do).

I guess if the UUID of your main blog changes, this forgets your preference? Can that happen? Oh well, pretty minor if it does.

This adds an onChange listener to the blog selector so that it can update local storage immediately when you select a blog.

I spent way too long trying to think of an elegant, easily-comprehendable-without-a-comment syntax to search through the values of an object to get the matching key before just saying screw it and also storing the object in reverse ¯\\\_(ツ)\_/¯ 

#### Issues this closes
#429; #220, #317, #412.

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -222,7 +222,7 @@ const updateQuickTags = (changes, areaName) => {
 };
 
 /**
- * adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#basic_example
+ * adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string
  *
  * @param {string} data - String to hash
  * @returns {Promise<string>} Hexadecimal string of a unique hash of the input
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -270,11 +270,8 @@ export const main = async function () {
   );
 
   if (rememberLastBlog) {
-    const hashToUuid = {};
     for (const { uuid } of userBlogs) {
-      const hash = await sha256(uuid);
-      uuidToHash[uuid] = hash;
-      hashToUuid[hash] = uuid;
+      uuidToHash[uuid] = await sha256(uuid);
     }
 
     const mainBlog = userBlogs.find(blog => blog.primary === true);
@@ -284,10 +281,8 @@ export const main = async function () {
       await browser.storage.local.get(rememberedBlogStorageKey);
 
     const savedBlogHash = rememberedBlogs[accountKey];
-
-    if (savedBlogHash && hashToUuid[savedBlogHash]) {
-      blogSelector.value = hashToUuid[savedBlogHash];
-    }
+    const savedBlogUuid = Object.keys(uuidToHash).find(uuid => uuidToHash[uuid] === savedBlogHash);
+    if (savedBlogUuid) blogSelector.value = savedBlogUuid;
 
     blogSelector.addEventListener('change', updateRememberedBlog);
   }
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -235,13 +235,13 @@ const sha256 = async (data) => {
   return hashHex;
 };
 
-const updateRememberedBlog = async event => {
+const updateRememberedBlog = async ({ currentTarget: { value: selectedBlog } }) => {
   if (!rememberLastBlog) return;
 
-  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
-    await browser.storage.local.get(rememberedBlogStorageKey);
+  const {
+    [rememberedBlogStorageKey]: rememberedBlogs = {}
+  } = await browser.storage.local.get(rememberedBlogStorageKey);
 
-  const selectedBlog = event.target.value;
   const selectedBlogHash = blogHashes[selectedBlog];
 
   rememberedBlogs[accountKey] = selectedBlogHash;
@@ -274,11 +274,12 @@ export const main = async function () {
       blogHashes[uuid] = await sha256(uuid);
     }
 
-    const mainBlog = userBlogs.find(blog => blog.primary === true);
-    accountKey = blogHashes[mainBlog.uuid];
+    const { uuid: primaryUuid } = userBlogs.find(({ primary }) => primary === true);
+    accountKey = blogHashes[primaryUuid];
 
-    const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
-      await browser.storage.local.get(rememberedBlogStorageKey);
+    const {
+      [rememberedBlogStorageKey]: rememberedBlogs = {}
+    } = await browser.storage.local.get(rememberedBlogStorageKey);
 
     const savedBlogHash = rememberedBlogs[accountKey];
     const savedBlogUuid = Object.keys(blogHashes).find(uuid => blogHashes[uuid] === savedBlogHash);
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -48,7 +48,7 @@ let queueTag;
 let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
 
-const storageKey = 'quick_reblog.alreadyRebloggedList';
+const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
 
 const renderTagSuggestions = () => {
@@ -155,13 +155,13 @@ const reblogPost = async function ({ currentTarget }) {
       notify(response.displayText);
 
       if (alreadyRebloggedEnabled) {
-        const { [storageKey]: alreadyRebloggedList = [] } = await browser.storage.local.get(storageKey);
+        const { [alreadyRebloggedStorageKey]: alreadyRebloggedList = [] } = await browser.storage.local.get(alreadyRebloggedStorageKey);
         const rootID = rebloggedRootId || postID;
 
         if (alreadyRebloggedList.includes(rootID) === false) {
           alreadyRebloggedList.push(rootID);
           alreadyRebloggedList.splice(0, alreadyRebloggedList.length - alreadyRebloggedLimit);
-          await browser.storage.local.set({ [storageKey]: alreadyRebloggedList });
+          await browser.storage.local.set({ [alreadyRebloggedStorageKey]: alreadyRebloggedList });
         }
       }
     }
@@ -178,7 +178,7 @@ const reblogPost = async function ({ currentTarget }) {
 });
 
 const processPosts = async function (postElements) {
-  const { [storageKey]: alreadyRebloggedList = [] } = await browser.storage.local.get(storageKey);
+  const { [alreadyRebloggedStorageKey]: alreadyRebloggedList = [] } = await browser.storage.local.get(alreadyRebloggedStorageKey);
   filterPostElements(postElements).forEach(async postElement => {
     const { id } = postElement.dataset;
     const { rebloggedRootId } = await timelineObjectMemoized(id);
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -283,8 +283,8 @@ export const main = async function () {
       hashToUuid[hash] = uuid;
     }
 
-    const mainBlog = userBlogs[0].uuid;
-    accountKey = uuidToHash[mainBlog];
+    const mainBlog = userBlogs.find(blog => blog.primary === true);
+    accountKey = uuidToHash[mainBlog.uuid];
 
     const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
       await browser.storage.local.get(rememberedBlogStorageKey);
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -40,6 +40,8 @@ let suggestableTags;
 let popupPosition;
 let showBlogSelector;
 let rememberLastBlog;
+let uuidToHash;
+let accountKey;
 let showCommentInput;
 let quickTagsIntegration;
 let showTagsInput;
@@ -49,6 +51,7 @@ let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
 
 const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
+const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
 
 const renderTagSuggestions = () => {
@@ -218,6 +221,34 @@ const updateQuickTags = (changes, areaName) => {
   }
 };
 
+/**
+ * adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
+ *
+ * @param {string} data - string to hash
+ * @returns {Promise<string>} hash - hexadecimal string of a unique hash of the input
+ */
+const sha256 = async (data) => {
+  const msgUint8 = new TextEncoder().encode(data);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+};
+
+const updateRememberedBlog = async event => {
+  if (!rememberLastBlog) return;
+
+  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+    await browser.storage.local.get(rememberedBlogStorageKey);
+
+  const selectedBlog = event.target.value;
+  const selectedBlogHash = uuidToHash[selectedBlog];
+
+  rememberedBlogs[accountKey] = selectedBlogHash;
+  console.log('setting', rememberedBlogs);
+  browser.storage.local.set({ [rememberedBlogStorageKey]: rememberedBlogs });
+};
+
 export const main = async function () {
   ({
     popupPosition,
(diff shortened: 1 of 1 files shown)


src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,34 @@ const updateQuickTags = (changes, areaName) => {
   }
 };
 
+/**
+ * adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
+ *
+ * @param {string} data - string to hash
+ * @returns {Promise<string>} hash - hexadecimal string of a unique hash of the input
```
AprilSylph, line 228: jsdoc nitpicks

```suggestion
 * adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#basic_example
 *
 * @param {string} data - String to hash
 * @returns {Promise<string>} Hexadecimal string of a unique hash of the input
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +270,42 @@ export const main = async function () {
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
 
+  if (rememberLastBlog) {
+    // const userBlogHashEntries = await Promise.all(userBlogs.map(async ({ uuid }) => [uuid, await sha256(uuid)]));
+    // userBlogHashes = Object.fromEntries(userBlogHashEntries);
+    // accountKey = userBlogHashEntries[0][1];
+
```
AprilSylph, line 277:
```suggestion
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +270,42 @@ export const main = async function () {
...
+    const savedBlogHash = rememberedBlogs[accountKey];
+
+    console.log(userBlogs.map(({ name, uuid }) => ({ name, uuid })));
+    console.log('userBlogHashes', uuidToHash);
+    console.log('mainBlogHash', accountKey);
+    console.log('savedBlogHash', savedBlogHash);
+
+    // const [savedBlog] = userBlogHashEntries.find(([key, value]) => value === savedBlogHash) || [];
+
```
AprilSylph, line 300:
```suggestion
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +270,42 @@ export const main = async function () {
...
+    if (savedBlogHash && hashToUuid[savedBlogHash]) {
+      blogSelector.value = hashToUuid[savedBlogHash];
+      console.log('userblogUUIDs[savedBlogHash]', hashToUuid[savedBlogHash]);
```
AprilSylph, line 303:
```suggestion
```

AprilSylph (changes requested):
First (and perhaps most obviously), this commented-out code and debug `console.log` calls don't need to be here.

I know this was originally just an experiment, but commented-out code and debug logs have no reason for being once the feature actually works.

src/scripts/quick_reblog.js
```diff
@@ -40,6 +40,8 @@ let suggestableTags;
 let popupPosition;
 let showBlogSelector;
 let rememberLastBlog;
+let uuidToHash;
+let accountKey;
```
AprilSylph, line 44: GitHub won't let me suggest this properly because where these should go is untouched, but this collection of `let` statements is for preferences - these should be moved to the collection above, e.g.

```js
let lastPostID;
let timeoutID;
let suggestableTags;
let uuidToHash;
let accountKey;
```

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,34 @@ const updateQuickTags = (changes, areaName) => {
...
+  if (!rememberLastBlog) return;
+
+  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+    await browser.storage.local.get(rememberedBlogStorageKey);
+
+  const selectedBlog = event.target.value;
+  const selectedBlogHash = uuidToHash[selectedBlog];
+
+  rememberedBlogs[accountKey] = selectedBlogHash;
+  console.log('setting', rememberedBlogs);
```
AprilSylph, line 248: Missed this one in my first review

```suggestion
```

AprilSylph (changes requested):

src/scripts/quick_reblog.js
```diff
@@ -36,6 +36,8 @@ draftButton.dataset.state = 'draft';
 let lastPostID;
 let timeoutID;
 let suggestableTags;
+let uuidToHash;
```
AprilSylph, line 39: We actually have no reason to ever reassign this, since user blogs are cached anyway. So let's declare it as a constant further down...

```suggestion
```

src/scripts/quick_reblog.js
```diff
@@ -48,7 +50,8 @@ let queueTag;
 let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
 
-const storageKey = 'quick_reblog.alreadyRebloggedList';
+const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
+const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
```
AprilSylph, line 56: ...here!
```suggestion
const uuidToHash = {};
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,30 @@ export const main = async function () {
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
 
+  if (rememberLastBlog) {
+    uuidToHash = {};
```
AprilSylph, line 273:
```suggestion
```

AprilSylph (changes requested):

src/scripts/quick_reblog.js
marcustyphoon: Wait, but then -- oh! Because we're mutating an object!

I am so used to immutable JS from Redux that I forgot that updating data stored in an object doesn't necessarily mean reassigning the variable :D

src/scripts/quick_reblog.js
```diff
@@ -239,6 +268,29 @@ export const main = async function () {
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
 
+  if (rememberLastBlog) {
+    const hashToUuid = {};
```
AprilSylph, line 273: I see what this is getting at, but this is only used in one place, and it's easy enough to do a sort of reverse-search of `uuidToHash` in one line.

```suggestion
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +268,29 @@ export const main = async function () {
...
+    for (const { uuid } of userBlogs) {
+      const hash = await sha256(uuid);
+      uuidToHash[uuid] = hash;
+      hashToUuid[hash] = uuid;
```
AprilSylph, line 277:
```suggestion
      uuidToHash[uuid] = await sha256(uuid);
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,29 @@ export const main = async function () {
...
+    accountKey = uuidToHash[mainBlog.uuid];
+
+    const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+      await browser.storage.local.get(rememberedBlogStorageKey);
+
+    const savedBlogHash = rememberedBlogs[accountKey];
+
+    if (savedBlogHash && hashToUuid[savedBlogHash]) {
+      blogSelector.value = hashToUuid[savedBlogHash];
+    }
```
AprilSylph, line 290: I personally find this easier to read, too; we're trying to find a matching hash, and there's not a much more direct way of doing that than `.find()` with a strict equality check.

```suggestion
    const savedBlogUuid = Object.keys(uuidToHash).find(uuid => uuidToHash[uuid] === savedBlogHash);
    if (savedBlogUuid) blogSelector.value = savedBlogUuid;
```

AprilSylph (changes requested):

src/scripts/quick_reblog.js
marcustyphoon: Ahh, there we go. My oneliner was really awkward because I was using `entries` instead of `keys`, and... well, you can see what I did in the commented code in the initial commit; it was bad :D

AprilSylph: Hey, you weren't off the mark by much!

src/scripts/quick_reblog.js
```diff
@@ -48,8 +49,10 @@ let queueTag;
 let alreadyRebloggedEnabled;
 let alreadyRebloggedLimit;
 
-const storageKey = 'quick_reblog.alreadyRebloggedList';
+const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
+const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
+const uuidToHash = {};
```
AprilSylph, line 55:
```suggestion
const blogHashes = {};
```

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,33 @@ const updateQuickTags = (changes, areaName) => {
...
+};
+
+const updateRememberedBlog = async event => {
+  if (!rememberLastBlog) return;
+
+  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+    await browser.storage.local.get(rememberedBlogStorageKey);
+
+  const selectedBlog = event.target.value;
+  const selectedBlogHash = uuidToHash[selectedBlog];
```
AprilSylph, line 245:
```suggestion
  const selectedBlogHash = blogHashes[selectedBlog];
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,24 @@ export const main = async function () {
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
 
+  if (rememberLastBlog) {
+    for (const { uuid } of userBlogs) {
+      uuidToHash[uuid] = await sha256(uuid);
```
AprilSylph, line 274:
```suggestion
      blogHashes[uuid] = await sha256(uuid);
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,24 @@ export const main = async function () {
...
+    }
+
+    const mainBlog = userBlogs.find(blog => blog.primary === true);
+    accountKey = uuidToHash[mainBlog.uuid];
```
AprilSylph, line 278:
```suggestion
    accountKey = blogHashes[mainBlog.uuid];
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,24 @@ export const main = async function () {
...
+
+    const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+      await browser.storage.local.get(rememberedBlogStorageKey);
+
+    const savedBlogHash = rememberedBlogs[accountKey];
+    const savedBlogUuid = Object.keys(uuidToHash).find(uuid => uuidToHash[uuid] === savedBlogHash);
```
AprilSylph, line 284:
```suggestion
    const savedBlogUuid = Object.keys(blogHashes).find(uuid => blogHashes[uuid] === savedBlogHash);
```

AprilSylph (changes requested):
Looking good! But now that there's only one hash map, I think a cleaner name for `uuidToHash` is in order...

src/scripts/quick_reblog.js
marcustyphoon: It's interesting to analyze why the pattern here doesn't tend to occur to me! I think on some level it's how much of my process of learning JS was doing performance optimizations, and writing out uuidToHash twice doesn't intuitively feel like a reduction in "things that the code does" even though that's a totally meaningless way to characterize code.

AprilSylph: programmers can have little a repeat yourself, as a treat

src/scripts/quick_reblog.js
marcustyphoon: My brain immediately goes, "wait, what do the keys of blogHashes represent?" But while this makes sense, I'm pretty meh on it:

```diff
- const savedBlogUuid = Object.keys(blogHashes).find(uuid => blogHashes[uuid] === savedBlogHash);
+ const savedBlogUuid = userBlogs.map(blog => blog.uuid).find(uuid => blogHashes[uuid] === savedBlogHash);
```

Aside: another thing my brain does is go, "hold on, can't you do the `.then(console.log)` trick to shorten the "obj => obj.prop" pattern?" Which, you know... no.

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,33 @@ const updateQuickTags = (changes, areaName) => {
...
+ */
+const sha256 = async (data) => {
+  const msgUint8 = new TextEncoder().encode(data);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+};
+
+const updateRememberedBlog = async event => {
```
AprilSylph, line 238: In most cases I find it good practice to use `currentTarget` over `target`, as it ensures you're getting the data from the element you're attaching the event listener to (assuming you're attaching the function directly, as we are here).

```suggestion
const updateRememberedBlog = async ({ currentTarget: { value: selectedBlog } }) => {
```

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,33 @@ const updateQuickTags = (changes, areaName) => {
...
+  if (!rememberLastBlog) return;
+
+  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+    await browser.storage.local.get(rememberedBlogStorageKey);
+
+  const selectedBlog = event.target.value;
```
AprilSylph, line 244:
```suggestion
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,24 @@ export const main = async function () {
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
 
+  if (rememberLastBlog) {
+    for (const { uuid } of userBlogs) {
+      blogHashes[uuid] = await sha256(uuid);
+    }
+
+    const mainBlog = userBlogs.find(blog => blog.primary === true);
+    accountKey = blogHashes[mainBlog.uuid];
```
AprilSylph, line 278:
```suggestion
    const { uuid: primaryUuid } = userBlogs.find(({ primary }) => primary === true);
    accountKey = blogHashes[primaryUuid];
```

src/scripts/quick_reblog.js
```diff
@@ -239,6 +269,24 @@ export const main = async function () {
...
+
+    const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+      await browser.storage.local.get(rememberedBlogStorageKey);
```
AprilSylph, line 281:
```suggestion
    const {
      [rememberedBlogStorageKey]: rememberedBlogs = {}
    } = await browser.storage.local.get(rememberedBlogStorageKey);
```

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,33 @@ const updateQuickTags = (changes, areaName) => {
...
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+};
+
+const updateRememberedBlog = async event => {
+  if (!rememberLastBlog) return;
+
+  const { [rememberedBlogStorageKey]: rememberedBlogs = {} } =
+    await browser.storage.local.get(rememberedBlogStorageKey);
```
AprilSylph, line 242:
```suggestion
  const {
    [rememberedBlogStorageKey]: rememberedBlogs = {}
  } = await browser.storage.local.get(rememberedBlogStorageKey);
```

AprilSylph (changes requested):
And now, some destructuring. Also, nitpicks for existing destructuring.

Applying this batch of suggestions will make the PR identical to my local version.

src/scripts/quick_reblog.js
```diff
@@ -218,6 +221,33 @@ const updateQuickTags = (changes, areaName) => {
...
+  const msgUint8 = new TextEncoder().encode(data);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+};
+
+const updateRememberedBlog = async ({ currentTarget: { value: selectedBlog } }) => {
+  if (!rememberLastBlog) return;
+
```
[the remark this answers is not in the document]
marcustyphoon: Nope, it shouldn't! At least, I can't think of a crazy race condition for it to matter.

AprilSylph (approved):
Good work! Thanks for following along with the review process. Hopefully it was enlightening in some small way.

marcustyphoon:
Indeed; thanks for typing it all up so I could follow through your thought process!

PR #451, opened 2022-02-06
Improve behavior when extension is initialized twice

### LINKED ISSUE #449: Vanilla Video Not Removing Original Tumblr Video
### Platform

Windows 10

### Browser

Firefox 96.0.3

### Addon version

0.14.6

### Details

Vanilla Video is failing to remove Tumblr video embed.
image
image

marcustyphoon:
#### User-facing changes
- Displays an informational modal if XKit Rewritten is loaded on a page more than once, usually due to an automatic background update or the user toggling the extension off and on in Firefox.

#### Technical explanation
This injects a script that sets a(nother) property on the `window` object when XKit Rewritten initializes, and checks if it has been previously set. If so, the extension has been previously run on this page, so we may as well inform the user (or remind the developer) that surprising behavior may occur.

(I'm not actually familiar with how extension updates work, but I assume this method should work. Not sure of the best UI copy to describe this or the best place to put code of this kind.)

#### Issues this closes
#449

### DIFF
--- a/src/util/check_repeat_init.js
+++ b/src/util/check_repeat_init.js
@@ -17,14 +17,15 @@ const getRunningXkit = async (currentVersion) => {
 export const checkRepeatInit = inject(getRunningXkit, [currentVersion])
   .then(result => {
     if (result) {
+      const updateMessage = result !== currentVersion
+        ? [`The extension appears to have been auto-updated to ${currentVersion}!`, document.createElement('br')]
+        : [];
+
       showModal({
         title: 'XKit Rewritten has been initialized multiple times',
         message: [
-          result === currentVersion
-            ? 'Not sure how this happened! Are you a developer?'
-            : `XKit Rewritten appears to been auto-updated to ${currentVersion}!`,
-          document.createElement('br'),
-          'Hard refreshing Tumblr via the refresh button or f5 key is recommended if you observe duplicate effects.'
+          ...updateMessage,
+          'Hard refreshing this browser tab via the refresh button or f5 key is recommended.'
         ],
         buttons: [modalCompleteButton]
       });
--- a/src/util/check_repeat_init.js
+++ b/src/util/check_repeat_init.js
@@ -1,5 +1,5 @@
 import { inject } from './inject.js';
-import { showModal, modalCompleteButton } from './modals.js';
+import { showModal, modalCancelButton } from './modals.js';
 
 const currentVersion = browser.runtime.getManifest().version;
 
@@ -25,9 +25,16 @@ export const checkRepeatInit = inject(getRunningXkit, [currentVersion])
         title: 'XKit Rewritten has been initialized multiple times',
         message: [
           ...updateMessage,
-          'Hard refreshing this browser tab via the refresh button or f5 key is recommended.'
+          'Refreshing this browser tab is recommended to avoid unexpected behavior.'
         ],
-        buttons: [modalCompleteButton]
+        buttons: [modalCancelButton,
+          Object.assign(document.createElement('button'), {
+            textContent: 'Reload',
+            className: 'blue',
+            onclick: () => {
+              location.reload(true);
+            }
+          })]
       });
     }
   });
--- a/src/util/check_repeat_init.js
+++ b/src/util/check_repeat_init.js
@@ -9,7 +9,7 @@ const getRunningXkit = async (currentVersion) => {
     value: currentVersion,
     writable: false,
     enumerable: false,
-    configurable: false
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Oh, actually, would this get triggered if you disabled and re-enabled the extension in Firefox? If so, the "this happened without a version update" copy would obviously need to be more normal.

src/util/check_repeat_init.js
```diff
@@ -0,0 +1,33 @@
...
+        ? [`The extension appears to have been auto-updated to ${currentVersion}!`, document.createElement('br')]
+        : [];
+
+      showModal({
+        title: 'XKit Rewritten has been initialized multiple times',
+        message: [
+          ...updateMessage,
+          'Hard refreshing this browser tab via the refresh button or f5 key is recommended.'
+        ],
+        buttons: [modalCompleteButton]
```
AprilSylph, line 30: you could use `modalCancelButton` here and a second custom button that calls `location.reload(true)`. Seems odd to direct the user to refresh without giving them a button to do so easily

marcustyphoon: Oh, good idea!

AprilSylph:
I think what I'd rather look at is if more features and utilities can clean themselves up safely from across instances. Right now, `util/sidebar.js` cleans itself up, and I just pushed 9a9631a47cfd2546472d935bf6232fa92b70c78f which makes `util/meatballs.js` clean itself up.

I was actually hesitant to move to using `Element.remove()` to remove `<style>` elements because it would impede this goal, but since all our `<style>` elements have an `xkit` class, we could just put `$('style.xkit').remove()` in `content_scripts/main.js` right before every feature is initialised.

Of course, there's a lot more work involved in this approach, but I am very interested in creating a seamless experience.

marcustyphoon:
Hmmm. Yeah, I think cleaning up the vast majority of plausibly noticeable effects should be possible. (I mean, in practice there don't seem to be much of any in the first place. I think stuff like event listeners that are lying around that try to read/write local storage just error out and then don't do anything problematic, at the moment.)

AprilSylph:
> I think stuff like event listeners that are lying around that try to read/write local storage just error out and then don't do anything problematic

As far as I can tell, they just get garbage-collected. Anything defined from the unloaded content script just ceases to be if it wasn't added to the DOM (even if it was *attached* to a DOM element).

marcustyphoon:
I tried patching up a few of these (scroll to bottom, shorten posts, and quick tags/trim reblogs, as well as classic search which will take a bit of refactoring) and I can see two primary methods of doing it: either removing the outdated elements and exclude classes (effectively, running most of `clean()` in `main()`) or re-attaching new event listeners to the outdated elements and skipping adding new ones.

Ex:

```js
const cleanDOM = () => {
  $(`.${excludeClass}`).removeClass(excludeClass);
  $(`.${shortenClass}`).removeClass(shortenClass);
  $(`.${tagsClass}, .${buttonClass}`).remove();
};

  cleanDOM();
  // ...
}
```

```js
  const outdatedButtons = document.querySelectorAll(`.${buttonClass}`);
  outdatedButtons.forEach(expandButtonClone => expandButtonClone.addEventListener('click', unshortenOnClick));
  // ...
```

The second should be a bit more performant, but the first seems easier to understand, and is of course more robust against the updated code not being the same as the original, which is kind of the whole point of a version update that would trigger this. Of course, changes to clean() will mean this cleanup doesn't always work.

A third option is just running `clean().then(() => main())` in 'runScript` in `content_scripts/main.js`, which is using a very big and very poorly performing hammer on a pretty small nail.

AprilSylph:
I wasn't thinking of modifying `runScript` at all; I was thinking of modifying `init` to detect if XKit has already run, then calling `destroyScript` for each enabled script if it has (and calling `$('style.xkit').remove()`) before continuing with the usual `.forEach(runScript)`.

To achieve the desired effect from this, each feature's `clean()` method will need to be slightly adjusted to not rely on references to elements in memory, and tested for if it errors out before its `main()` is run (I'm looking at you, Old Sidebar in XKit 7). Beyond that, it's just a matter of making the utility modules clean themselves up like the sidebar/meatballs utils or repetition-prevent themselves like the remixicon util.

marcustyphoon:
Yep, makes sense; that should be pretty simple!

AprilSylph:
Finally gonna decline this one instead of letting it rot in review hell.

Any weird stuff that's caused by an extension update/reload should be fixed individually. Stuff like #449 is pretty rare and isn't worth displaying a warning about with each patch.

Issue #452, opened 2022-02-06
Hide Filtered Posts tweak: Possible to desync UI / caching discussion


marcustyphoon:
### Platform

MacOS 10.14.6

### Browser

Edge 97.0.1072.76

### Addon version

Master c1b577f

### Details

Because the filtered posts tweak uses timelineObjectMemoized, and because some of the Tumblr settings page is now part of the new web platform (React/Redpop), it is possible to desync its UI from the current state by navigating to settings, filtering or unfiltering a word or tag, then navigating back to a page with a post containing the relevant word or tag. The hidden/unhidden state will have been cached from before the preference change.

This is obviously nearly impossible to run into or notice and may not be worth solving, but it does serve as a good example of the principle that we should be somewhat careful using timelineObjectMemoized as more of the site migrates to soft refreshes. I see no instances in the current codebase where this would be a significant issue, but there are a few minor ones and it's something to keep in mind for the future.

I had some ideas about caches that conditionally clear themselves, which could be a way to solve things like this in the future while still taking advantage of a cache. For example, a cache like timelineObject or userBlogs could be instructed to clear itself if you navigate off of a settings page, ensuring that whatever settings you may have changed cannot be the cause of unintended behavior.

AprilSylph:
I think a better solution in this case would simply to skip reading the `timelineObject` at all. The tweak could instead simply find filtered screens within posts and hide the ancestor `article`.

I think I relied on it a lot more than I really needed to because of the pain of constructing good selectors, which is all but trivialised by `resolveExpressions`.

PR #453, opened 2022-02-06
Experiment: Create navigation utility

marcustyphoon:
```js
import { onNavigation } from '../util/navigation.js';
onNavigation.addListener((currentLocation, previousLocation) =>
  console.log(`whoa, you soft navigated to ${currentLocation} from ${previousLocation}`)
);
```
This implements a (somewhat clunky, implementation-wise) wrapper around window.tumblr.on('navigation') so that XKit Rewritten can subscribe to soft navigation events.

It's somewhat clunky because we don't currently have a framework for injecting functions and giving them a way to call a repeatable callback, something that I have tried writing and given up on various times (it worked great, but the resulting code was nigh-incomprehensible.) Of course, this isn't that comprehensible either ~~- maybe recursion with then() would work better than a looping await - but I digress. (Let me know if you want me to write the recursive version; should be quite easy.)~~

More interestingly, what would you use this for? I thought of using it to clear caches (see #452). Now that more Tumblr pages like some of the settings pages are React, and thus use soft navigation, it becomes harder to ensure that caches' contents continue to make sense as they navigate the UI.

As discussed, for example, Mutual Checker can display slightly out of date results right after you follow/unfollow someone. That feels fine, and in line with user expectations... but if they keep that tab open, navigate a bit, refresh a few times by clicking the Tumblr logo, and things are still not up to date, maybe that doesn't feel as fine.

### DIFF
--- a/src/util/navigation.js
+++ b/src/util/navigation.js
@@ -1,0 +1,38 @@
+import { inject } from './inject.js';
+
+export const onNavigation = Object.freeze({
+  listeners: new Set(),
+  addListener (callback) {
+    if (this.listeners.has(callback) === false) {
+      this.listeners.add(callback);
+    }
+  },
+  removeListener (callback) {
+    this.listeners.delete(callback);
+  }
+});
+
+const waitForNavigation = async () => new Promise(resolve => {
+  const navigationHandler = () => {
+    window.tumblr.off('navigation', navigationHandler);
+    resolve();
+  };
+  window.tumblr.on('navigation', navigationHandler);
+});
+
+let previousLocation = new URL(location);
+(async () => {
+  while (true) {
+    try {
+      await inject(waitForNavigation);
+    } catch (e) {
+      console.log(e);
+      break;
+    }
+    const currentLocation = new URL(location);
+    for (const callback of onNavigation.listeners) {
+      callback(currentLocation, previousLocation);
+    }
+    previousLocation = currentLocation;
+  }
+})();
--- a/src/util/navigation.js
+++ b/src/util/navigation.js
@@ -21,18 +21,15 @@ const waitForNavigation = async () => new Promise(resolve => {
 });
 
 let previousLocation = new URL(location);
-(async () => {
-  while (true) {
-    try {
-      await inject(waitForNavigation);
-    } catch (e) {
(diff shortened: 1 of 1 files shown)


marcustyphoon:
For example, #452 is solved by:

```js
import { onNavigation } from '../util/navigation.js';

let cache = {};
onNavigation.addListener((currentLocation, previousLocation) => {
  if (previousLocation.pathname.startsWith('/settings')) {
  cache = {};
  }
});
```
in `react_props.js`.

marcustyphoon:
Another thing this could be useful for is avoiding the looping apiFetch in tag tracking plus. Whether you're following a tag is obviously not that volatile, but is volatile enough that I would want to re-query that on navigation.

On the other hand, this one is perhaps easier to solve in other ways.

marcustyphoon:
Closing this to reduce clutter, but it's still a concept I find interesting.

marcustyphoon:
Less hacky version, but actually more lines of code.

Like a third of this can be removed if you just skip the MessageChannel and just use postMessage. Performance impact should be irrelevant.

```js
import { getRandomHexString } from './crypto.js';

const injectKey = getRandomHexString();

let previousLocation = new URL(location);

const onNavigation = () => {
  const currentLocation = new URL(location);
  for (const callback of navigation.callbacks) {
    callback(currentLocation, previousLocation);
  }
  previousLocation = currentLocation;
};

await new Promise(resolve => {
  const receivePort = ({ origin, data, ports: [receivedPort] }) => {
    if (origin === location.origin && data === injectKey) {
      window.removeEventListener('message', receivePort);
      receivedPort.addEventListener('message', onNavigation);
      receivedPort.start();
      resolve();
    }
  };
  window.addEventListener('message', receivePort);

  inject(injectKey => {
    const channel = new MessageChannel();
    window.postMessage(injectKey, location.origin, [channel.port2]);
    window.tumblr.on('navigation', () => channel.port1.postMessage(true));
  }, [injectKey]);
});

export const navigation = Object.freeze({
  callbacks: new Set(),
  register (callback) {
    this.callbacks.add(callback);
  },
  unregister (callback) {
    this.callbacks.delete(callback);
  }
});
```

PR #463, opened 2022-02-09
Show Originals: New in-dash toggle UI 

### LINKED ISSUE #499: Show Originals: Peepr option does not always detect Peepr beta
### Platform

MacOS 10.14.6

### Browser

Edge 98

### Addon version

0.14.6

### Details

Show Originals' option to run in the blog view drawer works with the new beta version if it is opened by clicking within the Tumblr UI, as it opens in a glass container over the screen just like old Peepr in that case, but it does not work if you directly navigate to `https://www.tumblr.com/blog/view/blogname` with the beta, as the blog view container is directly embedded as the main page content now.

marcustyphoon:
### User-facing changes
- Replaces the preference checkboxes specifying where Show Originals will run with an in-dashboard controls UI that allows the user to toggle the extension, where applicable. The user's selection is remembered. On blogs where the user's settings would show all posts anyway (those you have whitelisted; your own if you always show your own posts), the controls UI appears for consistency but is disabled.

#### Technical explanation
The user's toggle selection is saved silently in local storage under a non-preference key similar to the Quick Reblog feature added in #450.

Rather than implementing complex conditional/on-click logic in the code that marks non-original posts, this simply marks non-original posts everywhere and toggles whether the extension is active by changing a data attribute on the timeline element and using a CSS descendent selector. (This could, mostly equivalently, have been a `~` selector and a data attribute on the controls container.)

(This has the nice side effect that you can really customize the behavior with user styles; I'm using one that slightly fades reblogs in "off" mode and I love it. Anyway.)

The onNewPosts callback always re-checks every timeline on the page (i.e. there is no excludeClass) because of the behavior described in #467; without ~~a fix being implemented for that issue and/or without~~ onNewPosts always querying like this, clicking a post to open Peepr and then clicking "show all posts" will result in Show Originals not running.

This also updates the show own reblogs option code to use the more accurate "is this post mine" code from the hide my posts tweak, which I forgot to co-author.

#### Issues this closes
Original discussion: #461
Resolves #499

---

- ~~maybe run in more places like own blog with default off? careful making too many saved settings though; nonobvious~~
- [x] less classes
- [x] test on mobile to see if pull to refresh looks bad
- [x] probably use different colors in peepr

### DIFF
--- a/src/scripts/show_originals.css
+++ b/src/scripts/show_originals.css
@@ -24,7 +24,7 @@
 }
 [data-show-originals="on"] .xkit-show-originals-controls a[data-mode="on"],
 [data-show-originals="off"] .xkit-show-originals-controls a[data-mode="off"],
-[data-show-originals="disabled"] .xkit-show-originals-controls a[data-mode="disabled"] {
+.xkit-show-originals-controls a[data-mode="disabled"] {
   box-shadow: inset 0 -2px 0 RGB(var(--accent));
   color: RGB(var(--accent));
 }
--- a/src/scripts/show_originals.css
+++ b/src/scripts/show_originals.css
@@ -14,20 +14,20 @@
   line-height: 1.333;
   margin-bottom: 20px;
 }
-.xkit-show-originals-controls a {
+.xkit-show-originals-controls > a {
   padding: 14px 16px;
   text-decoration: none;
   text-transform: capitalize;
 }
-.xkit-show-originals-controls a:not([data-mode="disabled"]) {
+.xkit-show-originals-controls > a:not([data-mode="disabled"]) {
   cursor: pointer;
 }
-.xkit-show-originals-controls a:hover {
+.xkit-show-originals-controls > a:hover {
   background: var(--blog-background-hover-color, rgba(var(--white-on-dark), .13));
 }
-[data-show-originals="on"] .xkit-show-originals-controls a[data-mode="on"],
-[data-show-originals="off"] .xkit-show-originals-controls a[data-mode="off"],
-.xkit-show-originals-controls a[data-mode="disabled"] {
+[data-show-originals="on"] .xkit-show-originals-controls > a[data-mode="on"],
+[data-show-originals="off"] .xkit-show-originals-controls > a[data-mode="off"],
+.xkit-show-originals-controls > a[data-mode="disabled"] {
   box-shadow: inset 0 -2px 0 var(--blog-link-color, rgb(var(--accent)));
   color: var(--blog-link-color, rgb(var(--accent)));
 }
--- a/src/scripts/show_originals.css
+++ b/src/scripts/show_originals.css
@@ -1,4 +1,4 @@
-.xkit-show-originals-timeline .xkit-show-originals-hidden article {
+[data-show-originals="on"] .xkit-show-originals-hidden article {
   display: none;
 }
 
--- a/src/scripts/show_originals.css
+++ b/src/scripts/show_originals.css
@@ -7,7 +7,7 @@
 }
 
 .xkit-show-originals-controls {
-  color: RGB(var(--white-on-dark));
+  color: var(--blog-title-color, rgb(var(--white-on-dark)));
   display: flex;
   font-size: 1.125rem;
   font-weight: 700;
@@ -23,11 +23,11 @@
   cursor: pointer;
 }
 .xkit-show-originals-controls a:hover {
-  background: rgba(var(--white-on-dark),.13);
+  background: var(--blog-background-hover-color, rgba(var(--white-on-dark), .13));
 }
 [data-show-originals="on"] .xkit-show-originals-controls a[data-mode="on"],
 [data-show-originals="off"] .xkit-show-originals-controls a[data-mode="off"],
 .xkit-show-originals-controls a[data-mode="disabled"] {
-  box-shadow: inset 0 -2px 0 RGB(var(--accent));
-  color: RGB(var(--accent));
+  box-shadow: inset 0 -2px 0 var(--blog-link-color, rgb(var(--accent)));
+  color: var(--blog-link-color, rgb(var(--accent)));
 }
--- a/src/scripts/show_originals.css
+++ b/src/scripts/show_originals.css
@@ -9,25 +9,30 @@
 .xkit-show-originals-controls {
   color: var(--blog-title-color, rgb(var(--white-on-dark)));
   display: flex;
-  font-size: 1.125rem;
   font-weight: 700;
-  line-height: 1.333;
   margin-bottom: 20px;
 }
+
 .xkit-show-originals-controls > a {
+  flex: 1;
   padding: 14px 16px;
+  text-align: center;
   text-decoration: none;
   text-transform: capitalize;
 }
-.xkit-show-originals-controls > a:not([data-mode="disabled"]) {
+
+.xkit-show-originals-controls > a {
   cursor: pointer;
 }
-.xkit-show-originals-controls > a:hover {
-  background: var(--blog-background-hover-color, rgba(var(--white-on-dark), .13));
+
+.xkit-show-originals-controls > a[data-mode="disabled"] {
+  cursor: not-allowed;
+  opacity: 0.65;
 }
+
 [data-show-originals="on"] .xkit-show-originals-controls > a[data-mode="on"],
 [data-show-originals="off"] .xkit-show-originals-controls > a[data-mode="off"],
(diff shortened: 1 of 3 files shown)


src/scripts/show_originals.js
```diff
@@ -94,6 +94,9 @@ const addControls = async (timeline, location) => {
     browser.storage.local.set({ [storageKey]: savedActive });
   };
 
+  let allPostsLabel = await translate('All posts');
+  if (allPostsLabel === 'All posts') allPostsLabel = 'All Posts';
```
AprilSylph, line 98: you could skip this line and use `text-transform: capitalize;` in the style(/sheet)

marcustyphoon: Ah, perfect.

marcustyphoon:
(maybe blogs on which the script has no effect should have a disabled controls selector rather than none, probably less confusing)

src/scripts/show_originals.js
```diff
@@ -70,25 +106,30 @@ const processPosts = async function (postElements) {
...
     showReblogsWithContributedContent,
-    whitelistedUsernames,
-    runOnDashboard,
-    runOnPeepr,
-    runOnBlogSubscriptions
+    whitelistedUsernames
   } = await getPreferences('show_originals'));
 
+  whitelist = whitelistedUsernames.split(',').map(username => username.trim());
+  const nonGroupUserBlogs = (await getUserBlogs().catch(() => []))
```
AprilSylph, line 117: When could `getUserBlogs()` reject? When you're logged out?

I'd rather build in error handling to the user blogs module itself if that's the case.

marcustyphoon:
I've looked over this a couple of times trying to reduce the number of lines of code. It's certainly a lot. The biggest thing that really isn't super necessary is the unique behavior that handles viewing a blog in Peepr whose posts would all be shown regardless of if the extension was active or not.

src/scripts/show_originals.js
marcustyphoon: When you're logged out, yeah. This is rather silly, but I just sort of left it as is because all of the other scripts that use it shouldn't do anything when you're logged out anyways, so user_blogs crashing them on init kind of worked out, at the moment. But yes, it should certainly be handled... not like that, ideally.

src/scripts/show_originals.js
```diff
@@ -3,19 +3,21 @@ import { timelineObject, exposeTimelines } from '../util/react_props.js';
 import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
 import { keyToCss } from '../util/css_map.js';
+import { translate } from '../util/language_data.js';
+import { getPrimaryBlogName, getUserBlogs } from '../util/user_blogs.js';
```
AprilSylph, line 7:
```suggestion
import { translate } from '../util/language_data.js';
import { getPrimaryBlogName, getUserBlogs } from '../util/user_blogs.js';
import { postSelector } from '../util/interface.js';
```

src/scripts/show_originals.js
```diff
@@ -25,43 +27,77 @@ const lengthenTimeline = async (timeline) => {
...
+  const controls = Object.assign(document.createElement('div'), { className: controlsClass });
+
+  const onButton = createButton(await translate('Original Posts'), handleClick, 'on');
+  const offButton = createButton(await translate('All posts'), handleClick, 'off');
+  const disabledButton = createButton(await translate('All posts'), null, 'disabled');
+
+  controls.append(...disabled ? [disabledButton] : [onButton, offButton]);
+
+  if (location === 'blogSubscriptions') {
+    timelineElement.querySelector('[data-id]')?.before(controls);
```
AprilSylph, line 54:
```suggestion
    timelineElement.querySelector(postSelector)?.before(controls);
```

src/scripts/show_originals.js
```diff
@@ -25,43 +27,77 @@ const lengthenTimeline = async (timeline) => {
...
+  } else {
+    timelineElement.querySelector('[data-id]')?.parentElement?.prepend(controls);
```
AprilSylph, line 56:
```suggestion
    timelineElement.querySelector(postSelector)?.parentElement?.prepend(controls);
```

AprilSylph (changes requested):

AprilSylph (commented):
The feature description is definitely going to need to be updated. The "on the dashboard" part is safe to just toss out, but I'm not sure what the most succinct way to communicate the toggle behaviour is.

Maybe just "Hide reblogged posts by default"? I was thinking along the lines of "Make reblogs opt-in"/"View reblogs on an opt-in basis", but that only communicates the main selling point of Show Originals through inference.

src/scripts/show_originals.js
marcustyphoon: 👍

marcustyphoon:
Hmm... yeah, that is tough. My brain goes to "filter" instead of "hide" for the verb, but I can't think of a logical justification for why that communicates optionality to me. Will have to think about it more.

AprilSylph (commented):
Testing this out on a fresh dev install, one thing that strikes me as very odd is how it's very possible to have a big enabled "Original Posts" button with a reblogged post right underneath it, due to `showOwnReblogs` and `showReblogsWithContributedContent` being enabled by default. With this change, I think it would be best to change them both to disabled by default.

Their descriptions could also use updating; I think "Always show" would make more sense than just "Show" given the new behaviour.

src/scripts/show_originals.css
```diff
@@ -1,7 +1,33 @@
...
+  padding: 14px 16px;
+  text-decoration: none;
+  text-transform: capitalize;
+}
+.xkit-show-originals-controls > a:not([data-mode="disabled"]) {
+  cursor: pointer;
+}
+.xkit-show-originals-controls > a:hover {
+  background: var(--blog-background-hover-color, rgba(var(--white-on-dark), .13));
+}
```
AprilSylph, line 27: Removes the hover background. I know this is copied from the `/tagged/` page's Recent/Top buttons, but the way there's no bottom border on the controls container means the visual effect is not quite the same. For me, I have to do a double-take on whether the enabled `box-shadow` is actually inset, since there's nothing for it or the hover background to both be flush against.

```suggestion
```

src/scripts/show_originals.css
```diff
@@ -1,7 +1,33 @@
...
+[data-show-originals="on"] .xkit-show-originals-controls > a[data-mode="on"],
+[data-show-originals="off"] .xkit-show-originals-controls > a[data-mode="off"],
+.xkit-show-originals-controls > a[data-mode="disabled"] {
+  box-shadow: inset 0 -2px 0 var(--blog-link-color, rgb(var(--accent)));
```
AprilSylph, line 31: Makes the enabled shadow thicker. No solid justification for this one either, it just looks better to me.

```suggestion
  box-shadow: inset 0 -3px 0 var(--blog-link-color, rgb(var(--accent)));
```

AprilSylph (changes requested):
Some style suggestions you'll hopefully agree with. Here's a screenshot of how it looks with all these changes applied:

Dashboard | Blog View
---|---
image | image

marcustyphoon:
Style changes sound good, and I trust your eye - going to accept changes and try them then comment if I think any deserve more discussion.

marcustyphoon:
> Testing this out on a fresh dev install, one thing that strikes me as very odd is how it's very possible to have a big enabled "Original Posts" button with a reblogged post right underneath it, due to `showOwnReblogs` and `showReblogsWithContributedContent` being enabled by default. With this change, I think it would be best to change them both to disabled by default.

Hm, user confusion vs users not realizing options exist that I think most should be aware of. I think that's fair, especially the case of the latter. The former feels like it has the potential to cause some confusion in its own right, even though the functionality of "hide reblogged posts" pretty clearly doesn't have an inherent exception for your own posts, so I'm ambivalent about that one. Again, I trust your judgement.

> Their descriptions could also use updating; I think "Always show" would make more sense than just "Show" given the new behaviour.

+1, certainly on the first two. Opinions on "Blogs to show reblogs from" vs "Blogs to always show reblogs from" vs "Always show reblogs from these blogs?"

AprilSylph:
"Always show reblogs from these blogs" sounds better to me!

AprilSylph (approved):
Great work!

I'd like to add this update to my next Week in Review given its significance. Would you like to be credited in it via your GitHub username or a Tumblr mention?

marcustyphoon:
Oh, thanks! Hmm... GitHub username works I suppose :)

PR #464, opened 2022-02-09
Experiment: Coalesce timelineObject calls

marcustyphoon:
#### User-facing changes
- bit faster if you have certain scripts enabled

#### Technical explanation
This fixes the inefficiency in which a) multiple extensions calling the unmemoized version of timelineObject in the same callback all do separate injections, and b) even if you only have one, it'll do an unnecessary extra call if it's after a timelineObjectMemoized, which is dependent upon what order you enable the scripts in.

This is done by adding a second cache, which in some sense "immediately" invalidates itself. Thus, unmemoized timelineObject requests are still guaranteed to be up to date; if anything, they can resolve a bit faster than before, since all concurrent invocations resolve on what would have been the first call.

<details>

<summary>This is a bit tunable:</summary>

```js
resultPromise.then(() => { delete immediateCache[postID]; }); 
```
^ This combines all timelineObject calls that would have been fired while one is still pending. This is the most robust way to collect unnecessary function calls, but is vulnerable to a race condition in which a mutation that changes the intended result of timelineObject happens after the first call, before the second call, and before we receive the message from the first call. Maybe if you spam clicked the like button just as a post loaded you could desync the Painter UI?

```js
queueMicrotask(() => { delete immediateCache[postID]; }); 
```
^ This only combines timelineObject calls made within the current synchronous event loop (i.e. if you call timelineObject immediately in your onNewPosts callback, you're fine; if you await something first, you may not be.) It 100% guarantees no race condition, though. 

```js
Promise.race([
  resultPromise,
  new Promise(setTimeout(resolve, 0)),
  new Promise(resolve => requestAnimationFrame(resolve))
]).then(() => { delete immediateCache[postID]; }); 
```
```js
queueMicrotask(queueMicrotask(queueMicrotask(queueMicrotask(queueMicrotask(() => { delete immediateCache[postID]; })))));
```
If you really wanted the best of both worlds, you could reduce the possibility of a race condition while ensuring calls are coalesced or reduce the possibility of calls not being coalesced while ensuring there is no race condition, but this gets fairly silly.

</details>

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/no_recommended/hide_recommended_posts.js
+++ b/src/scripts/no_recommended/hide_recommended_posts.js
@@ -1,6 +1,6 @@
 import { buildStyle, filterPostElements } from '../../util/interface.js';
 import { onNewPosts } from '../../util/mutations.js';
-import { exposeTimelines, timelineObject } from '../../util/react_props.js';
+import { exposeTimelines, timelineObjectImmediate } from '../../util/react_props.js';
 
 const excludeClass = 'xkit-no-recommended-posts-done';
 const hiddenClass = 'xkit-no-recommended-posts-hidden';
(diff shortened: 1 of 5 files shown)


marcustyphoon:
Closing this in favor of a somewhat cleaner method.

Issue #467, opened 2022-02-10
exposeTimelines utility can give stale results in peepr


marcustyphoon:
### Platform

MacOS 10.14.6

### Browser

Edge 98.0.1108.43

### Addon version

Master 25319e1

### Details

When Peepr is opened by clicking on a blog's username in certain cases, it opens the blog in question with the single post you clicked on and a "show all posts" button that navigates to the typical view of that user's blog.

This "show all posts" button's functionality is not implemented in Tumblr's code by unmounting the timeline element, evidently, because exposeTimelines doesn't detect any attribute-less elements after it's clicked, and the data attributes remain as what they were when Peepr was initially opened, i.e.`data-timeline="/v2/blog/nostalgebraist-autoresponder/posts/675762381622804480/permalink"` rather than `data-timeline="/v2/blog/nostalgebraist-autoresponder/posts"`.

This could be solved by unconditionally processing every timeline each time exposeTimelines is run, which feels a bit wasteful. Thinking about a better way to do this.

marcustyphoon:
I did just realize that, assuming the "show all posts" button in Peepr (both old and new beta versions) is the only place where this happens, this could be ignored and the state of the timeline could be inferred from `location.pathname` in scripts that care about this (currently only Show Originals) instead.

However, I still think it makes sense to fix this (instead of, I dunno, putting a warning in the JSDoc or something).

AprilSylph:
Just ran into the same problem with Tag Tracking+. Tried switching to a timeline RegExp approach instead of just peepr-excluding, but changing between followed tags doesn't unmount the timeline itself.

marcustyphoon:
...Wow. Huh. Okay. I guess the React developer side of me is like, "oh neat, you avoided unnecessary DOM manipulation!" and the extension developer side is like, really?

AprilSylph:
Hopefully this can someday be solved perfectly by Tumblr including the `data-timeline` attribute itself like I want it to...

PR #471, opened 2022-02-10
Expose Timelines Utility: Fix "show all posts" peepr button

marcustyphoon:
#### User-facing changes
- None directly; needed to fix Show Originals

#### Technical explanation

This unconditionally processes every timeline each time exposeTimelines is run, rather than checking an exclude class to only process unprocessed timelines.

This solves the problem described in #467 in which Tumblr changes a timeline's givenPath attribute without rerendering it.

Because the querySelectorAll used to check if anything needed to be processed was just as expensive as the querySelectorAll in the processing itself, this in itself is not a significant performance degradation. This is only true if the inject call resolves within the same event loop (i.e. within the animation frame before a repaint), however, which ~~is why #454 is valuable for this PR and why it uses the returnsVoid flag added there.~~ is now true because of 454b268bba7ed92fe008c8e83a4c5ffda6bb9529.

This also implements a small throttle effect as a performance optimization, in which multiple calls to exposeTimelines at the "same time" will only perform the injection once, and all calls will return the same Promise. The "same time" is fairly strict (the calls must be in the same microtask loop, so placing an await statement before one call but not another will break this, but there's no downside in that case, it just means you saved a few less CPU cycles than you could have.)

#### Issues this closes
Resoves #467.

### DIFF
--- a/src/scripts/seen_posts.js
+++ b/src/scripts/seen_posts.js
@@ -11,10 +11,10 @@ const dimClass = 'xkit-seen-posts-seen';
 const onlyDimAvatarsClass = 'xkit-seen-posts-only-dim-avatar';
 
 const dimPosts = async function (postElements) {
+  await exposeTimelines();
+
   const storageKey = 'seen_posts.seenPosts';
   const { [storageKey]: seenPosts = [] } = await browser.storage.local.get(storageKey);
-
-  await exposeTimelines();
 
   for (const postElement of filterPostElements(postElements, { excludeClass, timeline, includeFiltered })) {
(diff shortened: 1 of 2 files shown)


AprilSylph:
Ouch @ that timing.

If you want to open a new PR just for throttling `exposeTimelines`, please do. Given your commits, might not be necessary? Let me know!

marcustyphoon:
Hah, yeah, that was pretty funny!

Will do. It's certainly not particularly important, but if we do wind up removing the querySelector call from timelineObject, it'll be pretty much the only thing with a performance impact left!

PR #472, opened 2022-02-11
Implement dev-only global error notifications

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
This implements a background script that checks whether the extension has been loaded in development mode, and if so, it adds `error` and `unhandledrejection` listeners to the `window` object in the extension context which will use the XKit notification system to notify the developer if they have an unhandled error or rejection. (This, of course, only catches errors in injected functions if the inject error handler works correctly, and won't work if the main script or notify utility contain errors, but it's still pretty convenient sometimes.)

Mostly a proof of concept to see if I could get dev mode detection to work. I don't actually know if it does; of course I tested that both Chromium and Firefox dev modes are detected as dev, but I can't say for sure if it false positives in non-dev mode.

#### Issues this closes
n/a

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -2,7 +2,6 @@
 
 {
   const { getURL, sendMessage } = browser.runtime;
-  const extensionInfoPromise = sendMessage('browser.management.getSelf');
 
   const redpop = [...document.scripts].some(({ src }) => src.includes('/pop/'));
   const isReactLoaded = () => document.querySelector('[data-rh]') === null;
@@ -75,7 +74,7 @@
   };
 
   const runDevOnly = async () => {
-    const extensionInfo = await extensionInfoPromise;
+    const extensionInfo = await sendMessage('browser.management.getSelf');
     if (extensionInfo.installType === 'development') {
       console.log('XKit extension info:', extensionInfo);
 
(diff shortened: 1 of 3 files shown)


src/content_scripts/main.js
```diff
@@ -72,7 +73,21 @@
...
 
+  const runDevOnly = async () => {
+    const extensionInfo = await sendMessage('browser.management.getSelf');
+    if (extensionInfo.installType === 'development') {
+      console.log('XKit Rewritten is loaded in development mode. Extension info:', extensionInfo);
+
+      const notificationsPath = getURL('/util/notifications.js');
+      const { notify } = await import(notificationsPath);
+      window.addEventListener('error', (event) => { notify(`XKit Rewritten error: ${event.message}`); });
+      window.addEventListener('unhandledrejection', (event) => { notify(`XKit Rewritten error: ${event.reason}`); });
```
AprilSylph, line 84: Do both of these usually end up in the console?

marcustyphoon: Not having tested just now to be sure (since I'm going to bed): fairly sure the answer is an unconditional yes. There is also a `rejectionhandled` event on `window`, which would fire on caught exceptions, and those would not.

AprilSylph: I'm going to close the PR based on this answer. Dev-only behaviours are certainly a very interesting idea, but the benefits are outweighed by the complexity in this case. The only thing I really wanted it for is automatic CSS map exposure, which could be built in in simpler (not to mention more compatible) ways.

PR #475, opened 2022-02-11
Post footer redesign tweak: Improve overflow behavior

### LINKED ISSUE #435: Attempt to revert post footer design shifts to left side with Timestamps enabled
### Platform

Windows 10

### Browser

Firefox 96.0.3 (64-bit)

### Addon version

0.14.4

### Details

Reverted footer design gets shifted below notes when the buttons run out of room. This was made noticeable with the added note type icons next to the notes number. Easily reproduced with Timestamps enabled, & more noticeable on self reblogged/posted posts.

Footer without Timestamps:
image
image

Footer with Timestamps:
image
image
image
image

marcustyphoon:
#### User-facing changes

Improves the behavior of the post footer redesign tweak if there are a large number of icons.

| Before |
| --- |
|  |

| After |
| --- |
|  |

#### Technical explanation

I was hoping there would be some magic keyword like `justify-self` for this, but I guess it doesn't work in flexbox (kinda reasonable).

#### Issues this closes
Resolves #435.

### DIFF
--- a/src/scripts/tweaks/footer_unredesign.js
+++ b/src/scripts/tweaks/footer_unredesign.js
@@ -20,6 +20,10 @@ ${keyToCss('noteCount')} {
   gap: var(--post-padding);
 }
 
+${keyToCss('controls')} {
+  margin-left: auto;
+}
+
 .xkit-control-button-container {
   margin-left: 20px;
 }
--- a/src/scripts/tweaks/footer_unredesign.js
+++ b/src/scripts/tweaks/footer_unredesign.js
@@ -15,12 +15,12 @@ resolveExpressions`
   padding-bottom: 0;
 }
 
-${keyToCss('noteCount')} {
+article footer > ${keyToCss('noteCount')} {
   align-items: center;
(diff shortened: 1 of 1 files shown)


src/scripts/tweaks/footer_unredesign.js
```diff
@@ -20,6 +20,10 @@ ${keyToCss('noteCount')} {
   gap: var(--post-padding);
 }
 
+${keyToCss('controls')} {
```
AprilSylph, line 23: `controls` maps to seven classes, so let's increase the specificity here

```suggestion
article footer > ${keyToCss('controls')} {
```

(should probably apply this to the noteCount rule above, too)

AprilSylph (changes requested):

src/scripts/tweaks/footer_unredesign.js
marcustyphoon: Oh yikes, whoops. I was definitely thinking about that earlier and then completely forgot!

PR #478, opened 2022-02-12
Visible Alt Text Tweak: Improve wrapping on tiny images

marcustyphoon:
#### User-facing changes

Improves alt text captions wrapping on very small images.

Before:

After:

#### Technical explanation

`min-content` sets the minimum width to the width of the longest word. This is not perfect:

If you know of a way to improve this, that would be awesome. Perhaps just setting the minimum width to a reasonable quantity is better? But that doesn't adapt nicely to single word captions, so I dunno.

Edit: Oh, yeah, and I have no idea what the idiomatic css rule order is supposed to be.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -51,8 +51,9 @@ figure.accesskit-visible-alt-text {
 }
 
 figure.accesskit-visible-alt-text figcaption {
+  box-sizing: border-box;
   height: 100%;
-  min-width: min-content;
+  min-width: calc((min(540px, 100vw) / 3) - (4px * 2));
   padding: var(--post-header-vertical-padding) var(--post-padding);
   overflow: hidden;
 
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -52,6 +52,7 @@ figure.accesskit-visible-alt-text {
 
 figure.accesskit-visible-alt-text figcaption {
   height: 100%;
+  min-width: min-content;
   padding: var(--post-header-vertical-padding) var(--post-padding);
   overflow: hidden;
 


AprilSylph:
image

I don't think a perfect solution will be quite so simple

marcustyphoon:
Indeed. Simply setting a fixed minimum width the size of a reasonably short word is definitely less aesthetically pleasing on the single word example, but probably looks better with a few very short words and isn't prone to looking broken with, you know, endless screaming or whatever. (Not that one should put that in a caption designed for screen reader accessibility, I would imagine, but I digress :P)

[I may add screenshots to this comment later, but I may not.]

AprilSylph:
maybe a sensible minimum width would be the width of a 1/3 photoset photo? not quite 540/3 because of the padding, but close.

marcustyphoon:
126px min width, plus 2x20 margins = 166px container, which if multiplied by three yields 498, which is the width of the content in a textblock? Oh - you edited while I was typing this haha. So, not quite this. I think I was imagining a smaller value, but I think this works too.



(Also, it just dawned on me that the horizontal padding could be `min(var(--post-padding), /* some percentage */)` (or even  `max(var(--post-header-vertical-padding), calc(100% * 20/540))`) such that it wouldn't be as high when the caption block is small. The full 20px looks good in full width but pretty weird on the tiny stuff, IMO.)

src/scripts/accesskit.css
```diff
@@ -52,6 +52,7 @@ figure.accesskit-visible-alt-text {
 
 figure.accesskit-visible-alt-text figcaption {
   height: 100%;
+  min-width: min-content;
```
AprilSylph, line 55: Apparently the true number of pixels for the width of a single image in a three-image row is `177.33334350585938px`. Also, setting that without accounting for the padding breaks three-image photoset rows!

Beyond that, I realised that using a pixel value (or a `calc()` assuming the post width to be `540px` ) would break on viewport widths under 540px wide. I thought we might need to use a media query, but for once, CSS `min()` looks like it actually does its job!

```suggestion
  box-sizing: border-box;
  height: 100%;
  min-width: calc((min(540px, 100vw) / 3) - (4px * 2));
```

AprilSylph (changes requested):

src/scripts/accesskit.css
marcustyphoon: I... cannot believe I didn't know about `box-sizing: border-box;` :D

marcustyphoon: Ah, I see what you're doing here: setting a minimum size equal to the smallest an image could be in a photoset.

AprilSylph (approved):

marcustyphoon:
While I can't remember my exact thinking at the time, I just stumbled upon [the MDN article for overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) while investigating something unrelated and thought, "hey, wouldn't this have solved a problem I encountered previously?" Not sure if that is actually true, but I am putting a note here just in case.

AprilSylph:
The ruleset you modified in this PR already has `overflow-wrap: break-word`; specifying `anywhere` instead would reduce the intrinsic width down to what is necessary to fit one letter per line.

marcustyphoon:
Oh, I just looked at the CSS for the demo on that page and it's not doing what I thought it was - it's just a static max width ;( Never mind!

PR #501, opened 2022-02-28
Mutual Checker: Minor refactor of reprocessing

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
This is a minor refactor of the reprocessing logic in mutual checker which ensures that when you click to unfilter a post, it is processed again. Rather than querying the entire page for these posts, this just checks the new posts.

I initially kept the "remove excludeClass if necessary" logic:

```js
filterPostElements(postElements, { includeFiltered: true }).forEach(postElement => {
  if (postElement.classList.contains(mutualsClass) && postElement.querySelector(`.${mutualIconClass}`) === null) {
    postElement.classList.remove(excludeClass);
  }
});
```

which works just as well as the committed code, which replaces the built-in exposeTimelines logic with one with an exception.

This PR is entirely a matter of which you think is more clear. The performance benefit is not significant and the current code is fine.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -25,8 +25,7 @@ const needsReprocessing = postElement =>
 
 const addIcons = function (postElements) {
   filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
-    if (postElement.classList.contains(excludeClass) && !needsReprocessing(postElement)) return;
-    postElement.classList.add(excludeClass);
+    if (needsReprocessing(postElement) === false) return;
 
     const postAttribution = postElement.querySelector(postAttributionSelector);
     if (postAttribution === null) { return; }
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -19,13 +19,13 @@ let primaryBlogName;
 let postAttributionSelector;
 let icon;
 
-const needsReprocessing = postElement =>
+const alreadyProcessed = postElement =>
   postElement.classList.contains(mutualsClass) &&
-  postElement.querySelector(`.${mutualIconClass}`) === null;
+  postElement.querySelector(`.${mutualIconClass}`);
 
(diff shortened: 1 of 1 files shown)


src/scripts/mutual_checker.js
```diff
@@ -19,12 +19,15 @@ let primaryBlogName;
...
+  postElement.classList.contains(mutualsClass) &&
+  postElement.querySelector(`.${mutualIconClass}`) === null;
+
 const addIcons = function (postElements) {
-  [...document.querySelectorAll(`${postSelector}.${mutualsClass}`)]
-    .filter(postElement => postElement.querySelector(`.${mutualIconClass}`) === null)
-    .forEach(postElement => postElement.classList.remove(excludeClass));
+  filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
+    if (postElement.classList.contains(excludeClass) && !needsReprocessing(postElement)) return;
+    postElement.classList.add(excludeClass);
```
AprilSylph, line 29: I feel like we could remove `excludeClass` altogether here and simplify the condition down to

```suggestion
    if (needsReprocessing(postElement) === false) return;
```

since this by itself should prevent double mutual icons from appearing.

AprilSylph (changes requested):

src/scripts/mutual_checker.js
marcustyphoon: Er... oh. Oh, because this only fires on new postElements. Right. Because we implemented a mutation observer almost a month ago.

That took me way, way too long to figure out :P

marcustyphoon: Er, wait, no, this only processes posts that need to be reprocessed, right?

marcustyphoon:
Ah, yeah, logic error. The behavior we want is "process the post, unless it is one that has mutualsClass and an icon." Which... I may actually be impossible? But I think I would feel better leaving the check in anyways.

PR #504, opened 2022-03-01
TimelineObject utility: Process elements directly

marcustyphoon:
#### User-facing changes
- Slightly increased performance when loading new posts

#### Technical explanation
As we discussed in #416, it should be possible to skip using querySelector in some (all, I think, but at the expense of too much complexity in some cases) injected functions by injecting script elements into relevant DOM containers. I got worried it wouldn't work, so I tried it, and it seems to! 

This implements a parameter in `inject()` specifying which DOM element the script will be injected into. Because the script can find its location via `document.currentScript`, this removes the need for injected functions to communicate which element they are supposed to query or affect via serializable data and thus the need to use querySelector/querySelectorAll in the page context to find an element when the extension context already knows which one it is, which is the most expensive part of this whole process by far.

This takes advantage of this technique in `timelineObject`, converting its data id parameter into an element parameter and using a WeakMap to store its cache with elements as keys (yay, finally a use for WeakMap!). This means the cache gets invalidated more often than before, but it's really fast now, so who cares, really.

Finally, it removes the distinction between memoized and unmemoized `timelineObject`, relying on the fact that the resulting invalidation behavior should be up to date enough for any current use.

#### Issues this closes
n/a

---

~~Committing in case it saves time/sparks useful discussion. Feel free to ignore, peruse, cherry pick, etc, or I could convert this into a real PR with a real issue template description and a rebase off of master if you'd like. Sometimes it's fun to see the different ways we think to implement concepts when we try them independently, though - I'm curious if you were imagining something somewhat different than I was!~~

### DIFF
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -3,7 +3,7 @@ const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
 
 /**
  * @param {Function} func - Function to run in the page context (can be async)
- * @param {Array} args - Array of arguments to pass to the function via spread
+ * @param {Array} [args] - Array of arguments to pass to the function via spread
  * @param {Element} [target] - Element to append script to; will be accessible as
  *                             document.currentScript.parentElement in the injected function.
  * @returns {Promise<any>} The return value of the function, or the caught exception
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -4,7 +4,8 @@ const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
 /**
  * @param {Function} func - Function to run in the page context (can be async)
  * @param {Array} args - Array of arguments to pass to the function via spread
- * @param {Element} target
+ * @param {Element} [target] - Element to append script to; will be accessible as
+ *                             document.currentScript.parentElement in the injected function.
  * @returns {Promise<any>} The return value of the function, or the caught exception
  */
 export const inject = async (func, args = [], target = document.documentElement) => {
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -4,9 +4,10 @@ const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
 /**
  * @param {Function} func - Function to run in the page context (can be async)
  * @param {Array} args - Array of arguments to pass to the function via spread
+ * @param {Element} target
  * @returns {Promise<any>} The return value of the function, or the caught exception
  */
-export const inject = async (func, args = []) => {
+export const inject = async (func, args = [], target = document.documentElement) => {
   const script = document.createElement('script');
   const name = `xkit$${func.name || 'injected'}`;
   const async = func instanceof AsyncFunction;
@@ -46,11 +47,11 @@ export const inject = async (func, args = []) => {
         attributes: true,
         attributeFilter: ['data-result', 'data-exception']
       });
-      document.documentElement.append(script);
+      target.append(script);
       script.remove();
     });
   } else {
-    document.documentElement.appendChild(script);
+    target.append(script);
     script.remove();
     return JSON.parse(script.dataset.result);
   }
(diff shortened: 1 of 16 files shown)


src/util/inject.js
```diff
@@ -4,9 +4,10 @@ const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
 /**
  * @param {Function} func - Function to run in the page context (can be async)
  * @param {Array} args - Array of arguments to pass to the function via spread
+ * @param {Element} target
```
AprilSylph, line 7:
```suggestion
 * @param {Element} [target] - Element to append script to
```

marcustyphoon:
```diff
  /**
   * @param {Function} func - Function to run in the page context (can be async)
   * @param {Array} args - Array of arguments to pass to the function via spread
+  * @param {Element} [target] - Element to append script to; will be accessible as
+  *                             document.currentScript.parentElement in the injected function.
   * @returns {Promise<any>} The return value of the function, or the caught exception
   */
```

Perhaps?

(Or I guess we could get really cute and prepend `const target = document.currentScript.parentElement` into the injection body and set it as a global in eslint, or append it to the arguments... yeah, okay, no, I can't think of a clean way to do it, darn :D)

AprilSylph: Looks fine to me!

AprilSylph:
Now, I'll admit that I should have reviewed #463 already, but it is a fairly complex change, whereas this isn't really on its own. If you could drop 166cea1, implement the bare changes to Show Originals needed for it to continue to work with the new `timelineObject`, then force-push this branch, I would be happy to review this PR properly.

marcustyphoon:
Oh yeah, no worries! I just included the merged commit here since that was how I was testing it and I didn't feel the need to fix it unless you wanted to use this :) Will do!

src/util/inject.js
```diff
@@ -4,9 +4,11 @@ const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
 /**
  * @param {Function} func - Function to run in the page context (can be async)
  * @param {Array} args - Array of arguments to pass to the function via spread
```
AprilSylph, line 6: This isn't exactly to do with the changes in this PR, but reviewing it made me realise that `args` is optional too:

```suggestion
 * @param {Array} [args] - Array of arguments to pass to the function via spread
```

I'm not suggesting to pass `undefined` instead of `[]` anywhere, but it would (accurately) denote that an `inject()` call only needs `func` in order to work... though I can't think of any use-cases for that myself.

marcustyphoon: Indeed, we do use that: for getting the css map and language data!

AprilSylph: Oops, yeah.

I am not awake enough.

src/util/react_props.js
```diff
@@ -1,10 +1,10 @@
 import { inject } from './inject.js';
 import { keyToCss } from './css_map.js';
 
-const cache = {};
+const cache = new WeakMap();
```
AprilSylph, line 4:
```suggestion
const timelineObjectCache = new WeakMap();
```

src/util/react_props.js
```diff
@@ -19,19 +19,14 @@ const unburyTimelineObject = async id => {
...
- * @param {string} postID - The post ID of an on-screen post
+ * @param {Element} postElement - An on-screen post
  * @returns {Promise<object>} - The post's buried timelineObject property
  */
-export const timelineObject = async function (postID) {
-  cache[postID] = inject(unburyTimelineObject, [postID]);
-  return cache[postID];
+export const timelineObject = async function (postElement) {
+  if (!cache.has(postElement)) {
+    cache.set(postElement, inject(unburyTimelineObject, [], postElement));
```
AprilSylph, line 27:
```suggestion
  if (!timelineObjectCache.has(postElement)) {
    timelineObjectCache.set(postElement, inject(unburyTimelineObject, [], postElement));
```

AprilSylph (changes requested):
Just `cache` is too generic IMO

AprilSylph:
oh WHAT is up with my awful incomplete suggestions lately

marcustyphoon:
No worries ❤️

AprilSylph (approved):

PR #505, opened 2022-03-02
Notifications utility: Add persistent option

marcustyphoon:
Here's just an idea I had to allow notifications to optionally be persistent until the user manually closes them, in cases where the notification question is less an immediate response to a user action and more of an indicator of a background event that may have happened when the user was not looking at the tab. (I have wished XKit 7's update notifications were like this for years.) As I'm sure you could guess, I was primarily thinking about #451, but there may be other cases where this makes sense.

Annoyingly, this requires some kind of special handling for when the toast container target moves or a reversion of #422, which is not yet implemented here.

Anyway, before I think about how that would best be solved... any thoughts on this feature?

### DIFF
--- a/src/content_scripts/notifications.css
+++ b/src/content_scripts/notifications.css
@@ -10,10 +10,11 @@
   display: none;
 }
 
-#xkit-toasts div {
+#xkit-toasts > div {
   padding: 10px;
   border-top: 1px solid rgba(var(--white-on-dark), .13);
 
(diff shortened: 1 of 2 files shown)


AprilSylph:
I think displaying a modal for stuff the user _needs_ to see is still the best option. And I'm still convinced that #451 is not the best way to solve the problem that it tackles.

PR #506, opened 2022-03-02
Popup: Add additional searchable text to scripts

marcustyphoon:
#### User-facing changes
- Searching the popup for terms related to a script like the name of its XKit 7 equivalent will still find the script.

Probably should have proposed this a year ago when more people might have been switching from XKit 7. Then again, someday it will stop working, so maybe there will be another wave then.

#### Technical explanation
Uses a data attribute on the script element.

Would require an update to the manifest documentation.

#### Issues this closes
Discussion #113

### DIFF
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -132,7 +132,7 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, } = await file.json();
+    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {} } = await file.json();
 
     const scriptTemplateClone = document.getElementById('script').content.cloneNode(true);
 
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -132,12 +132,15 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', preferences = {} } = await file.json();
+    const { title = scriptName, description = '', icon = {}, help = '', preferences = {}, relatedTerms = '' } = await file.json();
 
     const scriptTemplateClone = document.getElementById('script').content.cloneNode(true);
 
+    const detailsElement = scriptTemplateClone.querySelector('details.script');
+
+    detailsElement.dataset.relatedTerms = relatedTerms;
+
     if (enabledScripts.includes(scriptName) === false) {
-      const detailsElement = scriptTemplateClone.querySelector('details.script');
       detailsElement.classList.add('disabled');
     }
 
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -132,12 +132,11 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', preferences = {}, relatedTerms = '' } = await file.json();
+    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, } = await file.json();
 
     const scriptTemplateClone = document.getElementById('script').content.cloneNode(true);
 
     const detailsElement = scriptTemplateClone.querySelector('details.script');
-
     detailsElement.dataset.relatedTerms = relatedTerms;
 
     if (enabledScripts.includes(scriptName) === false) {
--- a/src/scripts/quick_reblog.json
+++ b/src/scripts/quick_reblog.json
@@ -6,7 +6,7 @@
     "background_color": "#acd53a"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#quick-reblog",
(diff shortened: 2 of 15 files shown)


AprilSylph:
I'm allergic to strings with commas in them (when it's not strictly needed). Could `relatedTerms` be an array instead? Assigning an array to a `dataset` attribute will automatically add commas anyway.

marcustyphoon:
Yep! Right now it's just a straight text search, so it could be a space separated string, an array, whatever.

marcustyphoon:
Q: Do you like or dislike inline arrays in JSON?

AprilSylph:
Single-line arrays in JSON are fine, so long as there's a space after each comma. I'm also padding the brackets in `manifest.json` but I have a feeling I was just copying the style of some developer documentation there, it's not a hard preference.

https://github.com/AprilSylph/XKit-Rewritten/blob/7c19162b61b09e96dd37c3470c040bd07f75216c/src/manifest.json#L38-L39

src/browser_action/render_scripts.js
```diff
@@ -132,12 +132,15 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', preferences = {} } = await file.json();
+    const { title = scriptName, description = '', icon = {}, help = '', preferences = {}, relatedTerms = '' } = await file.json();
```
AprilSylph, line 135:
```suggestion
    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, } = await file.json();
```

src/browser_action/render_scripts.js
```diff
@@ -132,12 +132,15 @@ const renderScripts = async function () {
...
 
     const scriptTemplateClone = document.getElementById('script').content.cloneNode(true);
 
+    const detailsElement = scriptTemplateClone.querySelector('details.script');
+
+    detailsElement.dataset.relatedTerms = relatedTerms;
```
AprilSylph, line 141:
```suggestion
    const detailsElement = scriptTemplateClone.querySelector('details.script');
    detailsElement.dataset.relatedTerms = relatedTerms;
```

AprilSylph (changes requested):
I feel like `"relatedTerms"` fits best after the `"help"` feature manifest key; it seems that the existing order is descending priority for usability (`title`/`description`/`icon` are basically required, `help` is nice to have but technically redundant if the `description` is succinct enough, with `preferences` being the only key that is entirely missing in many features).

Unfortunately I can't suggest this change across the JSON files, so I'll leave that part to you.

src/scripts/seen_posts.json
```diff
@@ -1,13 +1,13 @@
...
   "title": "Seen Posts",
   "description": "Dim the posts you've seen already",
-  "relatedTerms": [ "read posts" ],
   "icon": {
     "class_name": "ri-eye-line",
     "color": "white",
     "background_color": "#849baf"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#seen-posts",
+  "relatedTerms": [ "read posts" ],
```
AprilSylph, line 10:
```suggestion
  "relatedTerms": [ "Read Posts" ],
```

AprilSylph (changes requested):
pedantry time 😏

I know these strings get `.toLowerCase()`d anyway, but the titles of XKit 7 extensions need to be capitalised correctly or it just doesn't zap my brain meat good enough.

src/scripts/quick_reblog.json
```diff
@@ -6,6 +6,7 @@
     "background_color": "#acd53a"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#quick-reblog",
+  "relatedTerms": [ "One-Click Postage", "One Click Postage", "Auto Tagger", "Autotagger" ],
```
AprilSylph, line 9:
```suggestion
  "relatedTerms": [ "One-Click Postage", "One Click Postage", "Auto Tagger", "AutoTagger" ],
```

AprilSylph (changes requested):
I think `"Autotagger"` is incorrect, actually. I would say it should be `"autotagger"` to match the extension filename, but nobody is going to be looking for it based on that - they're just going to be forgetting the space in "Auto Tagger".

Anyway, Quick Tags should also have
```json
"relatedTerms": [ "Auto Tagger", "AutoTagger", "XInbox" ],
```

src/scripts/tweaks.json
```diff
@@ -7,6 +7,7 @@
     "background_color": "#ecc84b"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#tweaks",
+  "relatedTerms": [ "Blacklist" ],
```
AprilSylph, line 10: "separat" already matches a preference, but since we're here, we might as well...

```suggestion
  "relatedTerms": [ "Blacklist", "Separator" ],
```

src/scripts/trim_reblogs.json
```diff
@@ -6,5 +6,6 @@
     "color": "white",
     "background_color": "#56bc8a"
   },
-  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#trim-reblogs"
+  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#trim-reblogs",
+  "relatedTerms": [ "Editable Reblogs" ]
```
AprilSylph, line 10:
```suggestion
  "relatedTerms": [ "Condensed Reblogs", "Editable Reblogs" ]
```

AprilSylph (changes requested):
Went through and checked the lists of excluded and removed features, found a couple which do fit despite their limited usefulness. Otherwise LGTM.

AprilSylph (approved):
I'll try to remember to add this to the documentation later, unless you beat me to it with a PR.

src/scripts/tweaks.json
marcustyphoon: Hah, I had to go git blame the XK7 spring cleaning list to to figure out what this was referring to.

PR #510, opened 2022-03-02
Popup: Fix iframe options sometimes capturing scroll

### LINKED ISSUE #509: Postblock: Options can capture scroll in popup
### Platform

MacOS 10.14.6 (Intel MacBook Pro)

### Browser

Edge 98.0.1108.62 

### Addon version

Master cd61ae3

### Details

With my 14 blocked posts in PostBlock, placing the mouse within the option menu PostBlock option menu and attempting to two-finger scroll causes the PostBlock options container to scroll slightly instead of scrolling the entire popup. Did not occur in Firefox, so perhaps this is Chromium-specific.

marcustyphoon:
#### User-facing changes
- iframe preferences no longer sometimes capture user scroll

#### Technical explanation
~~I thiiiiink this is the correct fix. Probably. Maybe? Well, it seems to work.~~

~~This adds one to the height set on iframe preference elements by the resize frames script, ensuring that even if the actual height of the document is fractional and `document.documentElement.scrollHeight` rounds down, all of the content will still fit in the iframe without a scroll bar appearing.~~

~~Not sure if this is better than changing the overflow css property, assuming that that also works.~~

This changes the overflow CSS property, which does work! It must be changed on the iframe contents, though, not the iframe itself.

#### Issues this closes
Resolves #509

### DIFF
--- a/src/browser_action/resize_frames.js
+++ b/src/browser_action/resize_frames.js
@@ -1,4 +1,4 @@
-const callback = () => { window.frameElement.height = document.documentElement.scrollHeight + 1; };
+const callback = () => { window.frameElement.height = document.documentElement.scrollHeight; };
 const observer = new ResizeObserver(callback);
 
 callback();
--- a/src/browser_action/resize_frames.js
+++ b/src/browser_action/resize_frames.js
@@ -1,4 +1,4 @@
-const callback = () => { window.frameElement.height = document.documentElement.scrollHeight; };
+const callback = () => { window.frameElement.height = document.documentElement.scrollHeight + 1; };
 const observer = new ResizeObserver(callback);
 
 callback();
--- a/src/scripts/postblock/options/index.css
+++ b/src/scripts/postblock/options/index.css
@@ -18,6 +18,7 @@ html {
   font-size: 14px;
   scrollbar-color: rgb(var(--grey)) transparent;
   scrollbar-width: thin;
+  overflow-y: hidden;
 }
 
 body {
--- a/src/scripts/quick_tags/options/index.css
+++ b/src/scripts/quick_tags/options/index.css
@@ -18,6 +18,7 @@ html {
   font-size: 14px;
   scrollbar-color: rgb(var(--dark-grey)) transparent;
   scrollbar-width: thin;
+  overflow-y: hidden;
 }
 
 body {


AprilSylph:
Could you test if setting `.script iframe { overflow-y: hidden; }` also fixes the issue? It seems like the more correct fix to me, but I can't reproduce the bug.

https://github.com/AprilSylph/XKit-Rewritten/blob/7c19162b61b09e96dd37c3470c040bd07f75216c/src/browser_action/configuration.css#L170-L173

marcustyphoon:
Yeah, I figure this might be retina display specific.

Setting `overflow-y: hidden` on the iframe does not, but as per https://stackoverflow.com/questions/15494568/html-iframe-disable-scroll/37345740#37345740, setting it on the html or body element of the iframe contents (i.e. in `src/scripts/postblock/options/index.css` and `src/scripts/quick_tags/options/index.css`) does! (That would cascade, right?)

AprilSylph:
`overflow` isn't inherited by default, but its behaviour means it doesn't need to be. If you have a parent element with `overflow-y: hidden` and two child elements, one of which overflows vertically outside the bounds of its parent, its height won't change based on this (unless it has `max-height: 100%` or the parent element has `display: flex`), and it won't scroll.

AprilSylph:
I'm happy for every `options/index.css` to specify `:root { overflow-y: hidden; }` - there are already a few styles that need to be applied to each one. (Maybe they should be moved to one stylesheet for each options page to use in a `<link>`?)

marcustyphoon:
Non-inherited properties! Interesting. No idea why I didn't put two and two together there; obviously if you set the `border` on an element, its children don't all have borders unless otherwise specified, but I didn't think of overflow as potentially working the same way. Makes sense, though!

Sounds good to me! `:root` and `html` are the same in this case, right?

AprilSylph:
yep, `:root` is `html` but ensuring it's the root one. Dunno what actually happens if you, for example, put a second `<html>` tag inside the document body, but using `:root` more clearly refers to the root `<html>` tag exclusively.

AprilSylph (approved):

marcustyphoon:
Oh, huh. I had assumed that besides that minor difference it would have a different meaning in shadow DOM trees, but apparently not.

PR #515, opened 2022-03-08
Scroll to Bottom: remove closure from clean()

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
This uses a CSS selector in Scroll to Bottom's clean function instead of a variable reference. This ensures that clean removes buttons added by a stale version of XKit that has been invalidated by an update or developer refresh in Firefox.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -83,6 +83,6 @@ export const clean = async function () {
   pageModifications.unregister(addButtonToPage);
   pageModifications.unregister(checkForButtonRemoved);
   stopScrolling();
-  scrollToBottomButton?.remove();
+  $(`#${scrollToBottomButtonId}`).remove();
   styleElement.remove();
 };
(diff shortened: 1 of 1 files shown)


src/scripts/scroll_to_bottom.js
```diff
@@ -83,6 +83,6 @@ export const clean = async function () {
   pageModifications.unregister(addButtonToPage);
   pageModifications.unregister(checkForButtonRemoved);
   stopScrolling();
-  scrollToBottomButton?.remove();
+  $(`#${scrollToBottomButtonId}`).remove();
```
AprilSylph, line 86: This only works if there is one stale button, which is a fair assumption most of the time, but not always. We can remove all stale buttons by using an attribute selector instead of an ID selector.

```suggestion
  $(`[id="${scrollToBottomButtonId}"]`).remove();
```

AprilSylph (changes requested):

src/scripts/scroll_to_bottom.js
marcustyphoon: Oh interesting! I assumed JQuery would passthrough to QSA, but clearly it doesn't (and in retrospect I was just reading the JQuery source and it's certainly not the kind of thing they would do :D) I accidentally did a sequence of events that made me think I had demonstrated that, too, but I think I had to throw a refresh in there.

AprilSylph (approved):

PR #516, opened 2022-03-08
Classic Search: remove closure from clean()

marcustyphoon:
~~I think I have to sleep on this one; there has to be a better way to do this.~~ Okay, I didn't spend any sleep time on it... but there was indeed a better way to do it!

#### User-facing changes
- Progress towards Classic Search not breaking if XKit Rewritten is initialized twice (by an update or developer refresh in Firefox.)

#### Technical explanation
This removes the reference to `searchInputElement` in Classic Search's `clean` function so it can be run without the extension being initialized, as discussed in #451.

In order to create the desired behavior ("Classic Search displays a copy of the form element containing the real input element, and clean() puts back the real form element containing the real input element"), the real and cloned input elements are left in the DOM along with the real and cloned form elements; they're just swapped. This way, the transformation can be reversed using the same swapping function based only on the DOM positions.

(This doesn't actually fix the double initialization by itself, of course)

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/classic_search.js
+++ b/src/scripts/classic_search.js
@@ -3,14 +3,22 @@ import { pageModifications } from '../util/mutations.js';
 
 const excludeClass = 'xkit-classic-search-done';
 const cloneClass = 'classic-search';
-const placeholderInputId = 'classic-search-placeholder';
 
 let newTab;
 
-const replaceSearchForm = function ([searchFormElement]) {
-  searchFormElement.classList.add(`${excludeClass}`);
+const swapNodes = (first, second) => {
+  const temporary = document.createElement('div');
+  first.replaceWith(temporary);
+  second.replaceWith(first);
+  temporary.replaceWith(second);
+};
 
+const replaceSearchForm = function ([searchFormElement]) {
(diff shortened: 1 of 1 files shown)


AprilSylph:
If it's not obvious by now, I'm likely not going to review this; I feel like feature scripts should instead clean up previous instances of themselves as part of module parsing (like util scripts do), which would allow us to keep closureful[^1] `clean()` functions.

[^1]: This probably isn't a real word, but hopefully you understand what I mean by it.

marcustyphoon:
Hm, I don't think I fully understand the scenario here! If I recall correctly, the changes in this PR were what I would do if I were tasked with implementing the change you just stated. I haven't looked at this in a while, though, and if there's a better way of doing this, awesome. Skimming the previous code, I think I could sort of see a simpler path, but it would still involve not having a "closureful" function, so I may not be seeing the whole picture here.

AprilSylph:
Declining this one too as my vision for dealing with extension reloads has changed - if a feature needs to clean a previous version of itself up, it should be done as a module side-effect.

PR #522, opened 2022-03-14
AccessKit: Make alt text user-selectable in the radar

marcustyphoon:
#### User-facing changes
- AccessKit's alt text captions can now be selected in the radar in the same way that they can be in posts in other locations.

#### Technical explanation
I had actually originally assumed `preventDefault()` would be necessary in #520, but it worked without it, so I committed it that way, thinking, "wow, button elements are weird." Well... turns out that it is necessary!

#### Issues this closes
n/a


AprilSylph (approved):
I had also tested this and it didn't work, but it looks like that's Firefox-specific.

PR #523, opened 2022-03-14
Seen Posts: Cache data to remove flicker

marcustyphoon:
~~Ugh this diff. Excessive bikeshedding is unproductive etc etc but there must be a way to represent this in code that sparks joy. (Maybe there isn't.)~~ I did my best!

#### User-facing changes
- Posts instantly load as seen when running seen posts instead of loading as unseen and becoming seen momentarily thereafter.

(Yes, I tried to compete with bec57e1.)

#### Technical explanation
~~Seen posts processes posts twice: once synchronously using a cached copy of the seen posts data (which is most likely correct), and then again with the updated data from local storage once the asynchronous `get` finishes (in case more posts have been seen in another of the user's tabs since the last time this tab had new posts).~~

~~Of course, unless we find a way to make MV3 have synchronous injection, this will become irrelevant in Chrome eventually, but hey, Firefox users can still have nice things right?~~

Another line and a half of code could also skip the `browser.storage.local.set` if there have been no mutations to the seen posts list, if you wanna save ~8ms every time you load posts on a non-dashboard page. Not really relevant but hey.

#### Issues this closes
n/a


marcustyphoon:
(This will merge conflict with #567; I can fix)

AprilSylph (changes requested):
Couldn't we use a more PostBlock-like approach to the cache - always assume it's correct, then process changes to storage when they happen?

Just updating the cache from storage doesn't inherently reprocess newly-seen posts; it would just ensure they're dimmed the *next* time they're seen in that session, which is precisely the intended behaviour of the feature.

AprilSylph (approved):

PR #529, opened 2022-03-16
Add XKit header to all API requests

cyle:
#### User-facing changes

- None!

#### Technical explanation

Like the previous version of XKit, it'd be helpful for us on the Tumblr side to know when requests are coming from XKit. I think this should do the trick?

#### Issues this closes

- None!

@AprilSylph hello

### DIFF
--- a/src/util/tumblr_helpers.js
+++ b/src/util/tumblr_helpers.js
@@ -20,7 +20,7 @@ export const apiFetch = async function (...args) {
 
       // make sure every API request we make is telling Tumblr that we're XKit
       init.headers['X-XKit'] = '1';
-      window.tumblr.apiFetch(resource, init);
+      return window.tumblr.apiFetch(resource, init);
     },
     args
   );
--- a/src/util/tumblr_helpers.js
+++ b/src/util/tumblr_helpers.js
@@ -7,16 +7,8 @@ import { inject } from './inject.js';
  */
 export const apiFetch = async function (...args) {
   return inject(
-    async (resource, init) => {
-      // make sure init is an object we can modify
-      if (typeof init !== 'object' || init === null) {
-        init = {};
-      }
-
-      // make sure init.headers is an object we can modify
-      if (typeof init.headers !== 'object' || init.headers === null) {
-        init.headers = {};
-      }
+    async (resource, init = {}) => {
+      init.headers ??= {}; 
 
       // make sure every API request we make is telling Tumblr that we're XKit
       init.headers['X-XKit'] = '1';
--- a/src/util/tumblr_helpers.js
+++ b/src/util/tumblr_helpers.js
@@ -7,7 +7,21 @@ import { inject } from './inject.js';
  */
 export const apiFetch = async function (...args) {
   return inject(
-    async (resource, init) => window.tumblr.apiFetch(resource, init),
+    async (resource, init) => {
+      // make sure init is an object we can modify
+      if (typeof init !== 'object' || init === null) {
+        init = {};
+      }
+
+      // make sure init.headers is an object we can modify
+      if (typeof init.headers !== 'object' || init.headers === null) {
+        init.headers = {};
+      }
+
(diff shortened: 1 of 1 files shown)


AprilSylph:
> Also what's the best way to test this locally? 🤔

https://github.com/AprilSylph/XKit-Rewritten/blob/master/docs/Chapter%202%20-%20Getting%20started.md!

P.S. hi

src/util/tumblr_helpers.js
```diff
@@ -7,7 +7,21 @@ import { inject } from './inject.js';
...
+      }
+
+      // make sure init.headers is an object we can modify
+      if (typeof init.headers !== 'object' || init.headers === null) {
+        init.headers = {};
+      }
+
+      // make sure every API request we make is telling Tumblr that we're XKit
+      init.headers['X-XKit'] = '1';
+      window.tumblr.apiFetch(resource, init);
```
AprilSylph, line 23: This needs to be explicitly returned now that it's no longer the sole statement; arrow functions only return implicitly in their braceless form.

```suggestion
      return window.tumblr.apiFetch(resource, init);
```

AprilSylph (changes requested):

src/util/tumblr_helpers.js
```diff
@@ -7,7 +7,21 @@ import { inject } from './inject.js';
...
+    async (resource, init) => {
+      // make sure init is an object we can modify
+      if (typeof init !== 'object' || init === null) {
+        init = {};
+      }
+
+      // make sure init.headers is an object we can modify
+      if (typeof init.headers !== 'object' || init.headers === null) {
+        init.headers = {};
+      }
```
AprilSylph, line 19: i'd say it's acceptable given the lacking jsdoc. (hopefully the eslint plugin doesn't complain about me documenting specific arguments within rest parameters...)

that said, it would definitely be possible to make use of some newer and cuter language features here, since XKit Rewritten has very strict browser requirements:
1. it's safe to assume `init` is either going to be an object or omitted entirely, so ensuring it exists is as simple as setting it as a [default parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)
2. it's also fairly safe to assume `init.headers` is either going to be an object or omitted (I'm sure setting it to another type would likely have some adverse effects), so ensuring it exists is also straightforward using [logical nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment)

```suggestion
    async (resource, init = {}) => {
      init.headers ??= {}; 
```

marcustyphoon: > [logical nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment)

Ooh, I was avoiding using that one because I wasn't sure if people would be familiar with it! Will make use in future :)

cyle: awesome, thank you for explaining that so thoroughly 👌

cyle:
awesome, i can confirm the header is being sent via Quote Replies, for example 👌

marcustyphoon:
I don't think it makes any difference (I guess if there's an exception in the code added here it would produce a more informative stack trace, but whatever), but just a note that this manipulation could also be done in the function body instead of in the inject body.

Anyway, certainly not the same scenario, but would it also be useful to add a similar header to the mega-edit utility?

AprilSylph:
Moving the manipulation to the outer function body would include switching away from using rest parameters, so I'd say that's out of scope for this PR.

It's pretty late for me, but I'll test this locally and merge tomorrow morning, @cyle. And thanks for putting the work in!

AprilSylph (approved):
LGTM! I will note though that the way XKit Rewritten bundles all its code, this will have to be included in a new release (v0.17.1) and won't reach previous versions - so any stats you gather won't be accurate for Firefox users, who can prevent addons from auto-updating and can choose to install any previous version of the addon.

PR #537, opened 2022-04-01
Fix z-index thrashing with native shorten posts

### LINKED ISSUE #536: Accesskit's pause gifs is incompatible with tumblr's shorten post setting
### Platform

Windows 10

### Browser

Chrome 99.0.4844.84 (Official Build) (64-bit)

### Addon version

Beta 0.19.0

### Details

When using tumblr's new 'Shorten Posts' setting along with Accesskit's 'Pause gifs until hovered over' setting it creates an issue in any shortened post where the 'Expand' bar would lay on top of a gif. When going to expand the post you can't because the gif tries to play at the same time the expand banner is trying to pop up, so you just get a ton of flickering. This happens both with xKit's Shorten Post turned on and when it's turned off. If I turn off the pause gifs setting then tumblr's expand bar works as intended. 
flicker2

marcustyphoon:
...by not using z-indexes at all. Certainly not the only way to solve this.

#### User-facing changes
- AccessKit paused gifs and CleanFeed hidden images no longer appear over Tumblr's native shorten post control element and flicker when hovered.

#### Technical explanation
Rather than adding a z-index to the Tumblr native shorten post control element, this removes the z-indexes from the Disable Gifs elements and the CleanFeed element and uses DOM ordering to ensure that they appear on top of the images they are supposed to replace.

Cleanfeed's element is an `::after` element, so I don't think it's possible for it to be out of order(?) and its z-index seems unnecessary.

Disable Gifs, as you pointed out in the relevant GitHub issue, needs to handle gifs being removed from the DOM and reattached as the last child of their containers, since React does not know about our placeholders. Rather than just exiting when this occurs, this moves the existing placeholders after the image before exiting.

I think `pausedGifElements` is an unclear variable name, but I didn't have any better ideas.

#### Issues this closes
#536

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -2,7 +2,6 @@
   position: absolute;
   top: 1ch;
   right: 1ch;
-  z-index: 1;
 
   height: 1em;
   padding: 5px;
@@ -20,7 +19,6 @@
 
 .xkit-paused-gif {
   position: absolute;
-  z-index: 1;
 }
 
 figure:hover .xkit-paused-gif,
--- a/src/scripts/accesskit/disable_gifs.js
+++ b/src/scripts/accesskit/disable_gifs.js
@@ -17,13 +17,18 @@ const pauseGif = function (gifElement) {
     const gifLabel = document.createElement('p');
     gifLabel.className = 'xkit-gif-label';
 
-    gifElement.parentNode.prepend(canvas, gifLabel);
+    gifElement.parentNode.append(canvas, gifLabel);
   };
 };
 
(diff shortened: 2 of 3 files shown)


AprilSylph (approved):
Looks good!

CleanFeed only had a `z-index` set on its placeholder elements so that it would override paused GIFs (a8418fc09257fd737b0d63cbd05a4aa434dd8f1c). With this approach to pausing GIFs, it is indeed unnecessary.

I still don't have Tumblr's shorten posts feature on either my main or my XKit alt, so I'm not able to test the fix per se, but I'm satisfied that this is an improvement regardless.

PR #548, opened 2022-04-09
Meatballs util: Fix functionality with React portals 

marcustyphoon:
To do: Write documentation.

```
#### User-facing changes
- 

#### Technical explanation


#### Issues this closes
```

### DIFF
--- a/src/util/meatballs.js
+++ b/src/util/meatballs.js
@@ -5,7 +5,7 @@ import { pageModifications } from './mutations.js';
 import { inject } from './inject.js';
 import { timelineObject } from './react_props.js';
 
-const postHeaderSelector = `${postSelector} article header`;
+const postHeaderSelector = `${postSelector} article > header`;
 
 const testHeaderElement = (selector) => {
   const postElement = document.currentScript.parentElement;
--- a/src/util/meatballs.js
+++ b/src/util/meatballs.js
@@ -46,7 +46,7 @@ const addMeatballItems = meatballMenus => meatballMenus.forEach(async meatballMe
   const inPostHeader = await inject(testHeaderElement, [postHeaderSelector], meatballMenu);
   if (!inPostHeader) return;
 
-  const timelineObjectData = await timelineObject(meatballMenu);
+  const __timelineObjectData = await timelineObject(meatballMenu);
 
   $(meatballMenu).children('[data-xkit-meatball-button]').remove();
 
@@ -61,12 +61,12 @@ const addMeatballItems = meatballMenus => meatballMenus.forEach(async meatballMe
     }, [
       label
     ]);
-    meatballItemButton.timelineObjectData = timelineObjectData;
+    meatballItemButton.__timelineObjectData = __timelineObjectData;
 
     meatballMenu.append(meatballItemButton);
 
     if (postFilter instanceof Function) {
-      const shouldShowItem = postFilter(timelineObjectData);
+      const shouldShowItem = postFilter(__timelineObjectData);
       meatballItemButton.hidden = shouldShowItem !== true;
 
       if (shouldShowItem instanceof Promise) {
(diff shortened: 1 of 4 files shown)


src/util/meatballs.js
```diff
@@ -2,6 +2,24 @@ import { keyToCss } from './css_map.js';
...
+
+const postHeaderSelector = `${postSelector} article header`;
+
+const testHeaderElement = (selector) => {
+  const postElement = document.currentScript.parentElement;
+  const reactKey = Object.keys(postElement).find(key => key.startsWith('__reactFiber'));
+  let fiber = postElement[reactKey];
+
+  while (fiber !== null) {
+    if (fiber.elementType === 'header') {
```
AprilSylph, line 16: I assume this isn't just testing for the tag name of the element, correct? This ensures the `stateNode` is whatever the portal is attached to?

AprilSylph (approved):
I can read this code and it looks like it works fine. No issues with meatball options appearing in menus where they shouldn't. 👍

src/util/meatballs.js
marcustyphoon: It is testing for the tag name only. I'm relying on the fact that "the first header element you encounter when you go up the react tree from the meatball menu" is a) through the portal and b) usable to identify that this is indeed a post meatball menu.

...Actually, now that I think about it, `${postSelector} article header` is maybe not the best test selector. The header selector is just a formality when the element is by definition a header, so effectively I guess this is just saying "the header element that this meatball menu is supposed to be attached to is somewhere within the body of a post." This code does successfully exclude blog meatball menus from blog cards spawned from within post bodies, but only because the blog cards are also in portals.

`${postSelector} article > header` (or, equivalently, `${postSelector} article > *`) should make this completely unambiguous though. "My first parent header is the one at the top of a post" seems correct.

marcustyphoon:
I think the one bikeshedding-style question I have is: is it weird to add the timelineObject data directly as a property of the button elements, or would it make more sense to add it to the meatball menu element (makes imo more semantic sense, but adds more lines of code to the handlers for no reason) or to store them in a WeakMap (as putting data as a DOM element property is... I mean, I'm sure it's normal, React is doing it after all, but I had to read some MDN docs pretty carefully trying to figure out if it's even a thing you can do?)

AprilSylph:
Storing data in the DOM properties is fine as long as we're doing it to our own elements. The big consideration is not to overwrite data that other things rely on, but if they're our own elements, it's not a concern at all.

I was going to comment about maybe adding `__` to the beginning of the property name to make absolutely sure it doesn't clash with any native DOM properties but I don't feel super strongly about that.

AprilSylph (dismissed):

marcustyphoon:
Yeah, I think we only have to worry about native DOM stuff; any properties we attach are gonna be in the extension sandbox, so they're ours to do what we please with from the Javascript side (unless we find a new and exciting way to implement `timelineObject()` anyway :D) No strong feelings either way about the prefix either - sure, why not.

(I'll write a retroactive description for this PR I guess :D)

AprilSylph (approved):

marcustyphoon:
Putting some of my test code here in case I someday need it for whatever reason:
```diff
const unburyTimelineObject = () => {
  const postElement = document.currentScript.parentElement;
  const reactKey = Object.keys(postElement).find(key => key.startsWith('__reactFiber'));
  let fiber = postElement[reactKey];
+ console.warn('begin');
  while (fiber !== null) {
+   // console.log('props: ', fiber.memoizedProps);
+   // console.log('fiber: ', fiber);
+   const entries = Object.entries(fiber.memoizedProps).filter(([k, v]) => typeof v !== 'object');
+   console.log(JSON.stringify(Object.fromEntries(entries), null, 2));
+   console.log('stateNode: ', fiber.stateNode);
    const { timelineObject } = fiber.memoizedProps || {};
    if (timelineObject !== undefined) {
+     console.warn('end');
      return timelineObject;
    } else {
      fiber = fiber.return;
    }
  }
};
```

Issue #413, opened 2022-01-25
"Compact buttons" CSS to put reblog/like/etc buttons back in the bottom right corner of posts


e-t-l:
### Concept

This CSS should put the buttons back into the bottom right corner and make the whole post footer more compact. I haven't tested it in multiple screen resolutions, so let me know if there's any problems. I can open a PR with this ass a css src script if you like, but I don't know the details of the other scripts required to add a button to the popup, trigger the script, etc. 
```
.tOKgq {
  margin-top: 15px !important;
  padding-bottom: 10px !important;
}
.gstmW {
  height: 0px !important;
}
.G6mnk .MCavR {
  border-top: none !important;
  justify-content: flex-end !important;
  margin: 0px var(--post-padding) 0 var(--post-padding) !important;
  margin-top: -5px !important;
  padding: 0px !important;
  position: relative !important;
  bottom: 20px !important;
}
.sfGru {
  padding-left: 20px !important;
  top: 15px !important;
}
```

And this should be added if timestamps are enabled
```
.tOKgq {
  margin-top: 0px !important;
}
.sfGru {
  top: 5px !important;
}
:not(:empty) + .xkit-timestamp {
  position: relative !important;
  top: 25px !important;
}
```

AprilSylph:
You're looking for the Tweaks option "Attempt to revert the post footer redesign".

e-t-l:
Ah. Well now I just feel silly.

AprilSylph:
No worries! I will probably need a CSS approach in future, so I might come back and reference this if I don't manage to grab the raw styles in time. Thanks for showing such a vested interest in the project.

If you'd like to learn more about the framework, check out the `doc/` folder at the repository root. If anything doesn't make sense, open a documentation issue and I'll see how I can improve it!

AprilSylph:
Here's my CSS solution (using the util functions for taking advantage of Tumblr's CSS map):

```js
Promise.all([
  descendantSelector('post', 'footer'),
  descendantSelector('post', 'footer', 'controls'),
  descendantSelector('post', 'footer', 'controlIcon'),
  descendantSelector('post', 'footer', 'postActivityWrapper')
]).then(([postFooterSelector, controlsSelector, controlIconSelector, postActivityWrapperSelector]) => {
  styleElement.textContent += `
    ${postFooterSelector} {
      display: flex;
      justify-content: space-between;
      align-items: center;
      padding-bottom: 12px;
    }
[...]

PR #554, opened 2022-04-14
Footer redesign revert tweak: New implementation

### LINKED ISSUE #553: Old Footer Change not Working
### Platform

macOS 11.6.5

### Browser

Chrome Version 100.0.4896.88

### Addon version

v0.19.2

### Details

The footer change to the original one is back to the current update. This only happened a couple of minutes ago, and worked fine during the rest of the day. Restarting Tumblr did not help. I have the "tweaks" part on, along with "attempt to revert the post footer design" checked off.

marcustyphoon:
I haven't done what I think of as a  "self-review" of this yet, per se. Seems to work though.

#### User-facing changes
- Fixes the footer redesign revert tweak not working after Staff removed the old footer CSS.

#### Technical explanation
Primarily based on the comment in #413. This removes the manipulation of the `footerRedesign` class, which no longer exists, implementing the old footer css explicitly.

Staff added a `footerRow` element in between the footer element and its children, necessitating a few tweaks.

The functionality on #459 is done how I personally think is the most clear: always putting a bottom ~~margin~~ *padding* on the footer; never putting a top margin on the `postActivityWrapper`.

#### Issues this closes
#553

### DIFF
--- a/src/scripts/tweaks/footer_unredesign.js
+++ b/src/scripts/tweaks/footer_unredesign.js
@@ -1,12 +1,7 @@
-import { keyToClasses, keyToCss, resolveExpressions } from '../../util/css_map.js';
+import { keyToCss, resolveExpressions } from '../../util/css_map.js';
 import { buildStyle } from '../../util/interface.js';
-import { pageModifications } from '../../util/mutations.js';
 
 const removePaddingClass = 'xkit-footer-padding-fix';
-
-let footerRedesignClasses;
-let footerRedesignSelector;
-let postActivityWrapperSelector;
 
 const styleElement = buildStyle();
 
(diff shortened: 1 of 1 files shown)


src/scripts/tweaks/footer_unredesign.js
```diff
@@ -1,32 +1,29 @@
...
-}
-
-article footer > ${keyToCss('noteCount')} {
+article footer ${keyToCss('footerRow')} {
+  display: flex;
+  justify-content: space-between;
   align-items: center;
-  gap: var(--post-padding);
+  flex-wrap: wrap;
+  margin-bottom: 12px;
```
AprilSylph, line 12:
```suggestion
  padding-bottom: 12px;
```

AprilSylph (changes requested):
Before | After
-------|------
image | image

src/scripts/tweaks/footer_unredesign.js
marcustyphoon: Whoops! 👍

AprilSylph (approved):

PR #555, opened 2022-04-15
Tweaks: Add blaze button hiding 

marcustyphoon:
I don't know if this button will be togglable in Tumblr's settings, but in case it never is, here's a way to hide it.

The one thing I don't know is whether the button appears on other people's blazed posts as an indicator that that is the reason for them appearing on your dash. If that is the case, I guess this should be changed to exclude selected icons or to an onNewPosts-based script filtering for... oh, filterPostElements doesn't have an option for editable, does it. Interesting. Anyway, I need to get a blazed post to check on that. **Edit:** Looks like they do not have blaze buttons.

#### User-facing changes
- Adds a Tweaks option to hide the blaze button.

#### Technical explanation
`${keyToCss('controlIcon')} ${keyToCss('igniteButton')}` works too, though I assume it's less robust; the other control buttons don't have their own CSS classes.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -50,6 +50,11 @@
       "label": "Hide the note type badges in post footers",
       "default": false
     },
+    "hide_blaze": {
+      "type": "checkbox",
+      "label": "Hide the blaze button in post footers",
+      "default": false
+    },
     "hide_filtered_posts": {
       "type": "checkbox",
       "label": "Hide filtered posts entirely",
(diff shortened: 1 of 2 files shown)


AprilSylph:
I think I'm gonna have to reject this one. COI aside, I think the Blaze button is just not intrusive enough to warrant hiding; it's basically impossible to Blaze a post by complete accident, and I've actually seen... zero demand to get rid of the button.

Good code, though!

marcustyphoon:
Yep, makes sense to me! I was expecting to see at least some "hey, I'm never gonna use this button" but apparently no one minds! (Out of curiosity, what's COI?)

AprilSylph:
COI = Conflict Of Interest

PR #558, opened 2022-04-20
AccessKit: : Disable visible alt text in Gutenberg editor

### LINKED ISSUE #557: Visible Alt Text: Causes visual artifact in Gutenberg editor
### Platform

MacOS 10.14.6 x86

### Browser

Edge 100.0.1185.44

### Addon version

0.19.3 (actually #552)

### Details

These boxes appear and disappear for reasons I'm not entirely clear on - looks like it might only be during certain loading states of  the Gutenberg editor photo element.

image

marcustyphoon:
This is one way of solving #557. Perhaps not the best way; it would be convenient to be able to preview the alt text that you have written for an image while composing it. Getting the styling correct seems nontrivial without a better understanding of the Gutenberg code than I currently have, however.

#### User-facing changes
- There is no visible alt text in the post compose form, but at least it isn't buggy and doesn't change formatting suddenly without user interaction.

#### Technical explanation
n/a

#### Issues this closes
#557

### DIFF
--- a/src/scripts/accesskit/visible_alt_text.js
+++ b/src/scripts/accesskit/visible_alt_text.js
@@ -17,7 +17,7 @@ const processImages = function (imageElements) {
   imageElements.forEach(imageElement => {
     const { alt } = imageElement;
     const imageBlock = imageElement.closest(imageBlockSelector);
-    if (imageBlock !== null) imageBlocks.set(imageBlock, alt);
+    imageBlocks.set(imageBlock, alt);
   });
 
   for (const [imageBlock, alt] of imageBlocks) {
@@ -60,7 +60,7 @@ export const main = async function () {
   styleElement.textContent = `${imageBlockLinkSelector}, ${imageBlockButtonInnerSelector} { height: 100%; }`;
   document.head.append(styleElement);
 
-  pageModifications.register('img[alt]', processImages);
+  pageModifications.register(`article ${imageBlockSelector} img[alt]`, processImages);
 
   browser.storage.onChanged.addListener(onStorageChanged);
 };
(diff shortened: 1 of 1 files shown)


AprilSylph (changes requested):
I think we can agree that this feature was only designed to work within posts, where we can guarantee there is room for a caption.

We could reflect this in the code and solve the issue more simply by changing the selector used to register `processImages()` to `pageModifications` to something more specific:

```diff
-  pageModifications.register('img[alt]', processImages);
+  pageModifications.register(`${postSelector} ${imageBlockSelector} img[alt]`, processImages);
```

---

As a bonus, this would allow us to skip the check for `imageBlock` in the function itself, too, since we'll only be processing descendants of `imageBlock`s in the first place:
```diff
-    const imageBlock = imageElement.closest(imageBlockSelector);
-    if (imageBlock !== null) imageBlocks.set(imageBlock, alt);
+    imageBlocks.set(imageElement.closest(imageBlockSelector), alt);
```

marcustyphoon:
Right! Thought about that - it would require a custom `postSelector` in order to still work on the radar post, right?

Actually, it looks like simply adding `${imageBlockSelector}` would work, assuming that that css map key is never used inside Gutenberg. Do you think that's better, or do you think explicitly specifying e.g. `is(${postSelector}, ${radarSelector})` is more clear about exactly where the feature is supposed to function?

(I assume there are no other locations where this is relevant besides timeline posts and radar? Trying to remember.)

AprilSylph:
Ah, right, the Radar. Well, I was originally going to recommend more simply using `article ${imageBlockSelector} img[alt]`, which I guess would cover it.

As for omitting a post selector entirely, we're... already checking for `imageElement.closest(imageBlockSelector)` before sending the image blocks to be processed. Checking for it twice won't change anything.

marcustyphoon:
Right, it would only matter if there are or might ever be `imageBlock`s that are not within `article`s somewhere.

AprilSylph (approved):

PR #560, opened 2022-04-20
Seen Posts: Fix dimming blazed posts

### LINKED ISSUE #559: Seen Posts: Visual inconsistency on Blazed posts
### Platform

MacOS 10.14.6 x86

### Browser

Edge 100.0.1185.44

### Addon version

0.19.3

### Details

Seen posts dims posts, including their white backgrounds, by setting the opacity on the post's `article` element. Blazed posts contain an `instreamAd ` wrapper around their `article` element with a white background, so this doesn't work correctly in that case.

marcustyphoon:
~~I think this works. Ran out of posts to test on at the moment.~~

#### User-facing changes
- Seen posts consistently dims blazed posts.

#### Technical explanation
Sets the opacity on the entire listTimelineObject instead of just the article element. I can't think of a place where this would be a problem, but I could be wrong?

#### Issues this closes
Resolves #559.


AprilSylph:
I have no way of testing this as a fix since Blazed posts are only visible to US users, but I can test for regressions!

marcustyphoon:
Ah, there we go, got another one. Seems to work!

AprilSylph (approved):

PR #562, opened 2022-04-21
AccessKit: Unpause all gifs on imageblock hover

marcustyphoon:
I am kind of surprised React is okay with this. Anyway, I wanted to see if it worked and if the resulting functionality felt enjoyable and natural.

#### User-facing changes
- Hovering over any paused gif in an imageblock unpauses the entire image block

#### Technical explanation
~~Branch based off of #552; presumably I'll force push rebase this when it gets merged and if this is something we want to pursue.~~

This adds a wrapper div element around entire photosets and hides the paused gif overlay if the wrapper is hovered.

Not yet implemented: only performing this manipulation on gifs (cannot be synchronous; would have to be in processGifs checking for `.closest('.xkit-paused-gif-container')`), ~~clean function (`wrapper.replaceWith(...wrapper.children)`)~~

Edit: That I can think of, it's somewhat convoluted to only do this on photosets with a gif in them. Also, there's no way to make this behavior optional with AccessKit's preferences, and I don't know if most users would think this way makes more sense, though I do personally.

#### Issues this closes
Discussion #561

### DIFF
--- a/src/scripts/accesskit/disable_gifs.js
+++ b/src/scripts/accesskit/disable_gifs.js
@@ -1,6 +1,7 @@
 import { pageModifications } from '../../util/mutations.js';
 import { keyToCss } from '../../util/css_map.js';
 import { dom } from '../../util/dom.js';
+import { postSelector } from '../../util/interface.js';
 
 const className = 'accesskit-disable-gifs';
 
@@ -85,7 +86,7 @@ export const main = async function () {
   `;
   pageModifications.register(gifBackgroundImage, processBackgroundGifs);
 
-  pageModifications.register(await keyToCss('rows'), processRows);
+  pageModifications.register(`${postSelector} ${keyToCss('rows')}`, processRows);
 };
 
 export const clean = async function () {
--- a/src/scripts/accesskit/disable_gifs.js
+++ b/src/scripts/accesskit/disable_gifs.js
@@ -1,5 +1,6 @@
 import { pageModifications } from '../../util/mutations.js';
 import { keyToCss, resolveExpressions } from '../../util/css_map.js';
+import { dom } from '../../util/dom.js';
 
 const className = 'accesskit-disable-gifs';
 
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Undrafting this because as per discussion elsewhere about how NPF photosets are just rows of images that happen to be next to one another, I don't see another way to do this.

(Well, making it such that hovering over a post unpauses any gifs in the post, would be the other way.)

I greatly prefer this functionality, but I don't know if it feels natural or not.

src/scripts/accesskit/disable_gifs.js
```diff
@@ -1,5 +1,6 @@
 import { pageModifications } from '../../util/mutations.js';
 import { keyToCss } from '../../util/css_map.js';
+import { dom } from '../../util/dom.js';
```
AprilSylph, line 3:
```suggestion
import { dom } from '../../util/dom.js';
import { postSelector } from '../../util/interface.js';
```

src/scripts/accesskit/disable_gifs.js
```diff
@@ -67,13 +84,20 @@ export const main = async function () {
     ${keyToCss('communityHeaderImage', 'bannerImage')}[style*=".gif"]
   `;
   pageModifications.register(gifBackgroundImage, processBackgroundGifs);
+
+  pageModifications.register(await keyToCss('rows'), processRows);
```
AprilSylph, line 88: `keyToCss()` is synchronous now so we can also drop the await

```suggestion
  pageModifications.register(`${postSelector} ${keyToCss('rows')}`, processRows);
```

AprilSylph (changes requested):
Wow, React really is okay with this! I thought it might be an issue while editing posts with images, but I couldn't cause a crash like that.

I think this feels good, but I'm also not a daily user of this feature. (Maybe I should put the work in to have AccessKit/No Recommended/Tweaks support `select`-type preferences properly...)

My only concern is the targeting of `rows` globally. Doesn't feel too safe and doesn't read too clearly.

marcustyphoon:
Yeah, it... seems to be? I don't know enough about reconciliation to be sure exactly, but I feel like there's probably a way to get an undesirable result from "render a component tree, wrap a div around part of it, and render some subset again," in which case implementing this would be relying somewhat on redpop never doing that exact thing. Doesn't make me feel amazing.

AprilSylph (approved):

PR #568, opened 2022-04-22
new script: Themed posts

marcustyphoon:
This is definitely <https://www.tumblr.com/settings/labs>, I don't know what you're talking about.

#### User-facing changes

Experimental script to color posts using the theme colors that are used in the blog cards.

||||
| - | - | - |
| image | image | image |
 | image | image | image | 
|image | image | image |

#### Technical explanation
Blogs' background and title colors are applied to their posts as the `white` and `black` theme colors, respectively. I implemented a small override in which the default uncustomized background color, which is almost-but-not-quite-white, is converted to a pure white background color (as would be used in the blog view modal and the default true blue theme) since it seemed kind of weird to have most blogs' backgrounds slightly grey.

My hex to RGB conversion code is, um, not robust.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -21,20 +21,18 @@ const processPosts = async function (postElements) {
   filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
     if (postElement.matches(blogViewSelector)) return;
 
-    const { blog: { name, theme: { backgroundColor, titleColor } } } =
-      await timelineObject(postElement);
+    const { blog: { name, theme } } = await timelineObject(postElement);
 
     if (blacklist.includes(name)) return;
 
-    let backgroundColorRGB = hexToRGB(backgroundColor);
-    const titleColorRGB = hexToRGB(titleColor);
-
-    if (backgroundColor === defaultBackgroundColor) {
-      backgroundColorRGB = '255, 255, 255';
-    }
-
     if (!blogs.has(name)) {
       blogs.add(name);
+
+      const { backgroundColor, titleColor } = theme;
+
+      const backgroundColorRGB = hexToRGB(backgroundColor);
+      const titleColorRGB = hexToRGB(titleColor);
+
       styleElement.textContent += `
         [data-xkit-themed="${name}"] {
           --white: ${backgroundColorRGB};
@@ -42,6 +40,8 @@ const processPosts = async function (postElements) {
         }
       `;
     }
+
+    postElement.dataset.xkitThemed = name;
     postElement.dataset.xkitThemed = name;
   });
 };
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -28,19 +28,19 @@ const processPosts = async function (postElements) {
 
     if (blacklist.includes(name)) return;
 
-    let backgroundColorHex = hexToRGB(backgroundColor);
-    const titleColorHex = hexToRGB(titleColor);
+    let backgroundColorRGB = hexToRGB(backgroundColor);
+    const titleColorRGB = hexToRGB(titleColor);
 
     if (backgroundColor === defaultBackgroundColor) {
-      backgroundColorHex = '255, 255, 255';
+      backgroundColorRGB = '255, 255, 255';
     }
 
     if (!blogs.has(name)) {
       blogs.add(name);
       styleElement.textContent += `
         [data-xkit-themed="${name}"] {
-          --white: ${backgroundColorHex};
-          --black: ${titleColorHex};
+          --white: ${backgroundColorRGB};
+          --black: ${titleColorRGB};
         }
       `;
     }
@@ -56,9 +56,12 @@ export const main = async function () {
 };
 
 export const clean = async function () {
-  $('[data-xkit-themed]').removeAttr('data-xkit-themed');
-  styleElement.remove();
-  blogs.clear();
-  styleElement.textContent = '';
   onNewPosts.removeListener(processPosts);
+
+  styleElement.remove();
+  styleElement.textContent = '';
+
+  $('[data-xkit-themed]').removeAttr('data-xkit-themed');
+
+  blogs.clear();
 };
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -29,13 +29,13 @@ const processPosts = async function (postElements) {
     if (!blogs.has(name)) {
       blogs.add(name);
       styleElement.textContent += `
-        .xkit-themed-${name} {
+        [data-xkit-themed="${name}"] {
           --white: ${hexToRGB(backgroundColor)};
           --black: ${hexToRGB(titleColor)};
         }
       `;
     }
-    postElement.classList.add(`xkit-themed-${name}`);
+    postElement.dataset.xkitThemed = name;
   });
 };
 
@@ -47,9 +47,9 @@ export const main = async function () {
 };
 
 export const clean = async function () {
-  blogs.forEach(name => $(`.xkit-themed-${name}`).removeAttr(`xkit-themed-${name}`));
-  blogs.clear();
+  $('[data-xkit-themed]').removeAttr('data-xkit-themed');
   styleElement.remove();
+  blogs.clear();
   styleElement.textContent = '';
   onNewPosts.removeListener(processPosts);
 };
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -42,7 +42,6 @@ const processPosts = async function (postElements) {
     }
 
     postElement.dataset.xkitThemed = name;
-    postElement.dataset.xkitThemed = name;
   });
 };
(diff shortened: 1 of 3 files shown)


src/scripts/themed_posts.js
```diff
@@ -0,0 +1,64 @@
...
+  document.head.append(styleElement);
+  onNewPosts.addListener(processPosts);
+};
+
+export const clean = async function () {
+  $('[data-xkit-themed]').removeAttr('data-xkit-themed');
+  styleElement.remove();
+  blogs.clear();
+  styleElement.textContent = '';
+  onNewPosts.removeListener(processPosts);
```
AprilSylph, line 63: order order order order

ideal order of cleanup:
1. remove listeners (makes sure none of the rest of the script runs while the rest of cleanup happens
2. remove visual changes (remove elements, remove styles)
3. remove dom changes
4. reset data

technically i'm breaking this a little bit with my suggestion but if you're manipulating the same thing multiple times i think it's fine to group that together too. clarity!

```suggestion
  onNewPosts.removeListener(processPosts);

  styleElement.remove();
  styleElement.textContent = '';

  $('[data-xkit-themed]').removeAttr('data-xkit-themed');

  blogs.clear();
```

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,64 @@
...
+  filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
+    if (postElement.matches(blogViewSelector)) return;
+
+    const { blog: { name, theme: { backgroundColor, titleColor } } } =
+      await timelineObject(postElement);
+
+    if (blacklist.includes(name)) return;
+
+    let backgroundColorHex = hexToRGB(backgroundColor);
+    const titleColorHex = hexToRGB(titleColor);
```
AprilSylph, line 32: \>output of `hexToRGB`
\>name implies they're still hex
:thinking:

```suggestion
    let backgroundColorRGB = hexToRGB(backgroundColor);
    const titleColorRGB = hexToRGB(titleColor);
```

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,64 @@
...
+
+    if (backgroundColor === defaultBackgroundColor) {
+      backgroundColorHex = '255, 255, 255';
```
AprilSylph, line 35:
```suggestion
      backgroundColorRGB = '255, 255, 255';
```

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    if (!blogs.has(name)) {
+      blogs.add(name);
+      styleElement.textContent += `
+        [data-xkit-themed="${name}"] {
+          --white: ${backgroundColorHex};
+          --black: ${titleColorHex};
```
AprilSylph, line 43:
```suggestion
          --white: ${backgroundColorRGB};
          --black: ${titleColorRGB};
```

AprilSylph (changes requested):
i'm going to be a bit more casual reviewing your PRs from now on if that's okay, makes it more fun

REVIVAL

src/scripts/themed_posts.js
marcustyphoon: ...huh, I have no idea what I was thinking there :D

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,67 @@
...
+
+const hexToRGB = (hex) => {
+  const chars = hex.replace('#', '').split('');
+  const splitnum = chars.length > 3 ? 2 : 1;
+  const arr = [];
+  while (chars.length) {
+    arr.push(parseInt(chars.splice(0, splitnum).join(''), 16));
+  }
+  return arr.join(', ');
+};
```
AprilSylph, line 20: i see the spirit of what this is trying to do, but i think a regex approach could be a lot more fun...

patch notes:
- made the function shorter in terms of sloc
- probably made the function longer in terms of total characters
- now supports `#RGB` syntax properly
- now supports alpha values even less! (`#RGBA` will be interpreted as `#RRGB`, `#RRGGBBAA` will throw an error)

```suggestion
const hexToRGB = (hex) => {
  const { red, green, blue } = hex.match(/^#(?<red>[A-Fa-f0-9]{1,2})(?<green>[A-Fa-f0-9]{1,2})(?<blue>[A-Fa-f0-9]{1,2})$/).groups;
  return [red, green, blue]
    .map(color => color.padEnd(2, color))
    .map(color => parseInt(color, 16))
    .join(', ');
};
```

we could support alpha values properly by having four perfect regexes (one per valid hex syntax) but i think it's safe enough to operate on the assumption that blog theme colours will never support alpha

marcustyphoon: Wait, how does this - oh, greedy quantifiers give back. Wild.

(Also, whoops, I didn't realize I wasn't correctly handling `#RGB` colors. That is a weird but convenient shorthand!)

Adding a `{0,2}` alpha term would work for `#RRGGBBAA` but still parse `#RGBA` as `#RRGB`, right? Darn.

marcustyphoon: I guess one could do `hex.replace(`#`, ``).slice(0, 6).match(` to at least not crash on... ah, still doesn't work for #RGBA, never mind.

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (!blogs.has(name)) {
+      blogs.add(name);
+      styleElement.textContent += `
+        [data-xkit-themed="${name}"] {
+          --white: ${backgroundColorRGB};
+          --black: ${titleColorRGB};
+        }
+      `;
+    }
```
AprilSylph, line 44: it looks like this is calculating RGB values and then discarding them if the blog is already registered

```suggestion
    const { blog: { name, theme } } = await timelineObject(postElement);

    if (blacklist.includes(name)) return;

    if (!blogs.has(name)) {
      blogs.add(name);

      const { backgroundColor, titleColor } = theme;

      const backgroundColorRGB = hexToRGB(backgroundColor);
      const titleColorRGB = hexToRGB(titleColor);

      styleElement.textContent += `
        [data-xkit-themed="${name}"] {
          --white: ${backgroundColorRGB};
          --black: ${titleColorRGB};
        }
      `;
    }

    postElement.dataset.xkitThemed = name;
```

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,65 @@
...
+import { getPreferences } from '../util/preferences.js';
+import { onNewPosts } from '../util/mutations.js';
+import { timelineObject } from '../util/react_props.js';
+
+const styleElement = buildStyle();
+const blogs = new Set();
+let blacklist;
+
+const defaultBackgroundColor = '#FAFAFA';
+
```
AprilSylph, line 11: i already removed the logic in my first suggestion, but while i was working on that, i realised how it just doesn't fit the spirit of the feature to have this override, IMO. if someone has an off-white background i want to see an off-white background.

AprilSylph (changes requested):
code otherwise LGTM!

marcustyphoon:
Not sure what one would call the variable, but I think I would prefer if the regex were extracted as a variable for line length reasons. (I'm sure some people use wrapping, but I don't.)

src/scripts/themed_posts.js
marcustyphoon: Yeah, I thought about it. Realistically, if this were first-party, I would want a valid (default) theme color of undefined and a way to revert to that, which would result in this script falling back to the current theme's coloration as if the script were not enabled. Seems too cute to just skip theming posts from blogs with `#FAFAFA` set, though, I think; it would feel more natural at first IMO, but the idea that you must reset your theme color to a magic value if you want to make your posts not colored anymore is definitely not intuitive.

src/scripts/themed_posts.js
```diff
@@ -0,0 +1,62 @@
...
+import { getPreferences } from '../util/preferences.js';
+import { onNewPosts } from '../util/mutations.js';
+import { timelineObject } from '../util/react_props.js';
+
+const styleElement = buildStyle();
+const blogs = new Set();
+let blacklist;
+
+const hexToRGB = (hex) => {
+  const { red, green, blue } = hex.match(/^#(?<red>[A-Fa-f0-9]{1,2})(?<green>[A-Fa-f0-9]{1,2})(?<blue>[A-Fa-f0-9]{1,2})$/).groups;
```
AprilSylph, line 11: > Not sure what one would call the variable, but I think I would prefer if the regex were extracted as a variable for line length reasons.

```suggestion
const groupsFromHex = /^#(?<red>[A-Fa-f0-9]{1,2})(?<green>[A-Fa-f0-9]{1,2})(?<blue>[A-Fa-f0-9]{1,2})$/;

const hexToRGB = (hex) => {
  const { red, green, blue } = hex.match(groupsFromHex).groups;
```

AprilSylph (approved):
Beautiful!

PR #570, opened 2022-04-22
Update "blog view" exclusion filtering

### LINKED ISSUE #569: Inconsistency in noPeepr utility option
### Platform

MacOS 10.14.6 x86

### Browser

Edge 100.0.1185.44

### Addon version

0.19.3

### Details

Shorten posts is active in the blog view modal if it is opened in a glass container by clicking on a blog, but not if you navigate directly to /blog/view. This is because the `noPeepr` option in `filterPosts` checks for the modal popup.

This can be fixed by using a different method to detect Peepr, like checking for the blog-theme-specific styles. (Show Originals uses `getComputedStyle`, but I think it's faster to do `postElement.matches('[style*="--blog-title-color"] *')`.

marcustyphoon:
Question: Do we want to rename this option flag to `noBlogView`? Edit: Done!

#### User-facing changes
- Shorten posts is consistently inactive in the blog view.

#### Technical explanation
The noPeepr flag now checks for the blog-theme-specific styles that are applied to the view formerly know as Peepr in both glass and non-glass modes.

Checking for a parent with a style attribute is a bit faster than calling `getComputedStyle`.

#### Issues this closes
Resolves #569

### DIFF
--- a/src/scripts/shorten_posts.js
+++ b/src/scripts/shorten_posts.js
@@ -10,7 +10,7 @@ let maxHeight;
 let tagsSelector;
 
 const excludeClass = 'xkit-shorten-posts-done';
-const noPeepr = true;
+const noBlogView = true;
 
 const shortenClass = 'xkit-shorten-posts-shortened';
 const tagsClass = 'xkit-shorten-posts-tags';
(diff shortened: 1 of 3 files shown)


AprilSylph:
`noBlogView` looks good to me!

AprilSylph (approved):

PR #571, opened 2022-04-22
Add show all tags tweak

marcustyphoon:
Feel free to close this; it was already kind of rejected with a batch of XK7 tweaks, IIRC, but I don't think I actually put it in the list for some reason.

#### User-facing changes
- Adds a tweaks option to show all post tags instead of requiring you to click a button.

#### Technical explanation
n/a

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/tweaks/show_all_tags.js
+++ b/src/scripts/tweaks/show_all_tags.js
@@ -1,0 +1,20 @@
+import { keyToCss, resolveExpressions } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const styleElement = buildStyle();
+resolveExpressions`
+${keyToCss('tags')}${keyToCss('collapsed')} {
+  max-height: inherit;
+}
+${keyToCss('seeAll')} {
+  display: none;
+}
+`.then(css => { styleElement.textContent = css; });
+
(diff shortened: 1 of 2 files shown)


src/scripts/tweaks/show_all_tags.js
```diff
@@ -0,0 +1,20 @@
+import { keyToCss, resolveExpressions } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const styleElement = buildStyle();
+resolveExpressions`
+${keyToCss('tags')}${keyToCss('collapsed')} {
+  max-height: inherit;
```
AprilSylph, line 7: According to Firefox devtools, the default value for `max-height` is `none`, which I think is more intuitive to communicate what this CSS is doing anyhow:

```suggestion
  max-height: none;
```

AprilSylph (changes requested):
Otherwise LGTM!

src/scripts/tweaks/show_all_tags.js
marcustyphoon: Oh yeah, makes sense!

AprilSylph (approved):

Issue #588, opened 2022-05-10
[Feature Request] Add option to hide non-mutuals in Mutual Checker


salazarbarrera:
### Concept

What the title says, allow to hide non-mutuals in dashboard, similar to how Show Originals allows to hide posts without original content.

marcustyphoon:
Suggest this to Tumblr Staff as well! They've been working on new filtered dashboards like https://www.tumblr.com/timeline/crushes and https://www.tumblr.com/timeline/blog_subscriptions; I bet a "mutuals" dashboard would make a lot of sense along with those.

salazarbarrera:
Great idea! Also, I didn't know it was possible to access to those on web. I saw the Labs experiments in the app, but nothing on browser.

marcustyphoon:
Yeah, they certainly feel a bit unfinished :D They've been hanging out semi-secretly for a while, but I think they might get an actual release with a proper UI soon!

salazarbarrera:
I just realized one single bug after two days using my patch: the browser scrolls to the top when 'k' is pressed. I know `show_originals` already solved this, but since this is the first time code a web extension, I'm not sure how. Does it have anything to do with that `getLocation()` function? I've been reading the [commit history](https://github.com/AprilSylph/XKit-Rewritten/commits/master/src/scripts/show_originals.js) of that script, but I can't find anything about that specific problem.

marcustyphoon:
This happens when you set `display:none` on a post directly, and it's the one Tumblr's code tries to scroll to when you press K. We hide the `article` element to avoid this!

(I don't know if April thinks this feature is a good idea for the script or not and was waiting for her comment, but I could review your code's implementation details!)

salazarbarrera:
Thanks! It was way easier than I thought and was overly complicating things. And in any case, even if it isn't accepted, it has been an entertaining thing to do. It has been a long time since I coded anything just for fun.

PR #589, opened 2022-05-10
Implementing option to only show posts from mutuals requested in #588

salazarbarrera:
#### User-facing changes
- Added option to Mutual Checker script to only show posts by mutuals on dashboard.

#### Technical explanation
- Hides via CSS all the posts made by non-mutual blogs.

#### Issues this closes
Issue #588

### DIFF
--- a/src/scripts/mutual_checker.css
+++ b/src/scripts/mutual_checker.css
@@ -7,6 +7,6 @@ svg.xkit-mutual-icon {
 	margin-right: 0.5ch;
 }
 
-.no-mutual {
+.xkit-mutual-checker-hidden article {
   display: none;
 }
--- a/src/scripts/mutual_checker.css
+++ b/src/scripts/mutual_checker.css
@@ -7,6 +7,6 @@ svg.xkit-mutual-icon {
 	margin-right: 0.5ch;
 }
 
-.xkit-mutual-checker-hidden article {
+[data-timeline="/v2/timeline/dashboard"] .xkit-mutual-checker-hidden article {
   display: none;
 }
--- a/src/scripts/mutual_checker.css
+++ b/src/scripts/mutual_checker.css
@@ -6,3 +6,7 @@ svg.xkit-mutual-icon {
   margin-left: 0;
 	margin-right: 0.5ch;
 }
+
+.no-mutual {
+  display: none;
+}
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -60,7 +60,7 @@ const addIcons = function (postElements) {
     if (isMutual) {
       postElement.classList.add(mutualsClass);
       postAttribution.prepend(icon.cloneNode(true));
-    } else if(showOnlyMutuals) {
+    } else if (showOnlyMutuals) {
       postElement.classList.add('no-mutual');
     } else if (!showOnlyMutuals) {
       postElement.classList.remove('no-mutual');
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -60,9 +60,9 @@ const addIcons = function (postElements) {
     if (isMutual) {
       postElement.classList.add(mutualsClass);
       postAttribution.prepend(icon.cloneNode(true));
-    } else  if(showOnlyMutuals){
+    } else if(showOnlyMutuals) {
       postElement.classList.add('no-mutual');
-    } else if(!showOnlyMutuals){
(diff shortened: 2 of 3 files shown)


salazarbarrera (commented):
Changes made to pass test.

salazarbarrera (commented):
More style changes.

src/scripts/mutual_checker.css
```diff
@@ -6,3 +6,7 @@ svg.xkit-mutual-icon {
   margin-left: 0;
 	margin-right: 0.5ch;
 }
+
+.xkit-mutual-checker-hidden article {
```
AprilSylph, line 10: If the option is only supposed to hide non-mutual posts on the dashboard, we need to make this selector more specific. Tumblr has `data-timeline` attributes on each timeline, which makes this very easy!

```suggestion
[data-timeline="/v2/timeline/dashboard"] .xkit-mutual-checker-hidden article {
```

src/scripts/mutual_checker.js
```diff
@@ -58,11 +61,16 @@ const addIcons = function (postElements) {
     if (isMutual) {
       postElement.classList.add(mutualsClass);
       postAttribution.prepend(icon.cloneNode(true));
+    } else if (showOnlyMutuals) {
+      postElement.classList.add(hiddenClass);
+    } else if (!showOnlyMutuals) {
+      postElement.classList.remove(hiddenClass);
```
AprilSylph, line 67: It looks like this condition only exists for cleanup when the preference is toggled... is that right? If so, it's sadly insufficient; while it does work for disabling the preference, it won't get called when the feature itself is disabled. Features should always clean themselves up in the `clean()` function.

```suggestion
```
---
Unfortunately I can't code-suggest this properly, and you won't be able to copy-paste this entire block, but this is what I'd do further down this file:
```diff
 export const clean = async function () {
   onNewPosts.removeListener(addIcons);

   $(`.${mutualsClass}`).removeClass(mutualsClass);
+  $(`.${hiddenClass}`).removeClass(hiddenClass);
   $(`.${mutualIconClass}`).remove();
 };
```

src/scripts/mutual_checker.json
```diff
@@ -6,5 +6,12 @@
     "color": "white",
     "background_color": "#777777"
   },
-  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#mutual-checker"
+  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#mutual-checker",
+  "preferences": {
+    "showOnlyMutuals": {
+      "type": "checkbox",
+      "label": "Only show mutuals on dashboard",
```
AprilSylph, line 13: This is a nitpick, so feel free to resolve without committing this if you disagree, but I feel like the description here could stand to be a little more verbose...

```suggestion
      "label": "Only show posts from mutuals on dashboard",
```

AprilSylph (changes requested):
Very nice for a first contribution! It's always wonderful to see someone new joining in with the project. I hope that between the documentation and the framework itself, it wasn't too hard to get your bearings in the codebase.

Other than these few little details, this looks good to go!

src/scripts/mutual_checker.js
salazarbarrera: Thanks! Effectively, that else was to undo the changes when someone wanted to stop using that feature. I'll add that cleanup and test it later.

salazarbarrera: Worked wonderfully.

src/scripts/mutual_checker.json
```diff
@@ -6,5 +6,12 @@
     "color": "white",
     "background_color": "#777777"
   },
-  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#mutual-checker"
+  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#mutual-checker",
+  "preferences": {
+    "showOnlyMutuals": {
+      "type": "checkbox",
+      "label": "Only show posts from mutuals on dashboard",
```
marcustyphoon, line 13:
```suggestion
      "label": "Only show posts from mutuals on the dashboard",
```

Minor phrasing suggestion!

AprilSylph: gonna commit this one myself, i should have looked more carefully at my original suggestion

salazarbarrera: Thanks! I too should have checked it better. Since English is my second language, on first inspection it sounded fine.

PR #591, opened 2022-05-10
Quick Reblog: Add a CSS class to already reblogged posts

marcustyphoon:
This only makes sense if we in fact do not want to implement the "hide alreadyreblogged posts" feature that has been requested.

#### User-facing changes
- n/a unless the user wants to create a user style

#### Technical explanation
Adds an `alreadyReblogged` class to posts which have already been reblogged with Quick Reblog, so that Stylus userstyles can target them.

```css
.alreadyReblogged article {
  outline: 3px solid purple;
}
```

#### Issues this closes
n/a


AprilSylph:
I think we should implement the option for real!

Issue #592, opened 2022-05-12
Seen Posts: Only register posts that appear on screen


marcustyphoon:
### Concept

Seen Posts currently remembers any post that has been loaded while the script is running, even if you never scrolled down to the post in question. As per the script title, it makes sense to only mark posts the user has, well, seen.

This is pretty easy using an intersection observer, but there are some notes:
- Should scrolling past posts rapidly mark them all as seen, or should the post need to be visible for more than a few milliseconds? (Or should this be a preference?) I can see the logic of both; the latter implements the script name more literally. Similarly, should any pixels of a post becoming visible count as seeing it, or should the center of the post be visible? I am a user of this script but not a passionate one (partially for reasons this issue addresses) and I would like to know what passionate users would prefer.
- Should posts hidden by extensions be marked as seen? This seems like a clear "no" - you haven't seen them! This basically happens anyway, in the simplest implementation (display:none means an element does not intersect the viewport), though ensuring it is reliable would require some complexity if MV3 eventually forces us to process mutations asynchronously.
- Do we use Tumblr's `hiddenObserverMiddlePoint` element, or is that getting too cute?

AprilSylph:
I believe the `hiddenObserverMiddlePoint` is how the dashboard tracks where to put you with the "where were we?" button, and consistency would be nice.

marcustyphoon:
Interesting side effect of this: posts taller than 2x viewport height are not marked as seen if you j/k scroll through them and are not intrigued enough to scroll down.

PR #593, opened 2022-05-12
Hide post element contents instead of articles

marcustyphoon:
Presumably it would take knowledge of the Tumblr code that I do not have to know if this is a good or bad idea, but I will PR it for discussion.

#### User-facing changes
- n/a

#### Technical explanation
This changes the CSS in all five scripts/tweaks that hide posts to setting `display:none` on the sibling `hiddenObserver` and `hiddenObserverMiddlePoint` elements as well as the `article` elements in hidden posts.

For the most part, this does nothing, but I assume that it affects the data Tumblr collects, which could indirectly affect the user experience as mentioned in #592.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/no_recommended/hide_recommended_posts.js
+++ b/src/scripts/no_recommended/hide_recommended_posts.js
@@ -7,7 +7,7 @@ const hiddenClass = 'xkit-no-recommended-posts-hidden';
 const timeline = /\/v2\/timeline\/dashboard/;
 const includeFiltered = true;
 
-const styleElement = buildStyle(`.${hiddenClass} article { display: none; }`);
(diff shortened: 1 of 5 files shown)


marcustyphoon:
It doesn't seem to. My understanding for why the bug happened was because hiding the listTimelineObject caused it to have no bounding box and thus supply no location to scroll to, while hiding its contents gives it 0 height, but still results in a valid location!

AprilSylph:
Good to know!

...It still seems like we have no way of knowing the extent of what the hidden observers really do, though. Unless we can determine exactly what they're for, it's hard to justify using a more vague selector (which is both a minor penalty for both performance and readability).

marcustyphoon:
Makes sense! I have no intention of trying to read more minified redpop to find references to them.

PR #596, opened 2022-05-14
new tweak: Hide liked posts

### LINKED ISSUE #594: feature request (from new xkit) - hide liked posts
### Concept

I saw it's not in the xkit rewritten removed/wont implement list and it's broken in new xkit

the extension would hide posts that are already "liked"

marcustyphoon:
#### User-facing changes
Adds a tweak to hide liked posts on the dashboard.

Two functionality considerations:
- As per my comment in the linked issue, should this only check posts on load or should it immediately hide a post when you press the like button? (Currently implemented the former.)
- Should this exclude your own posts? Tweaks can't really have sub-preferences. I personally assume that, by default, no script hides my own posts besides the one whose function is specifically "hide my own posts," but I can definitely imagine disagreement on this.

#### Technical explanation
n/a

#### Issues this closes
resolves #594

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -65,6 +65,11 @@
       "label": "Hide my own posts on the dashboard",
       "default": false
     },
+    "hide_liked_posts": {
+      "type": "checkbox",
+      "label": "Hide posts that I've liked on the dashboard",
+      "default": false
+    },
     "hide_follower_counts": {
       "type": "checkbox",
       "label": "Hide my follower count where possible",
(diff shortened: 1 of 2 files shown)


AprilSylph (changes requested):
i love that what i'm about to type is true but i'm sorry that it makes work for you:

we don't need `excludeClass` here! repetition prevention is built into the modification engine, and processing posts twice won't have an adverse effect.

marcustyphoon:
No worries, it's certainly not much work :D Good point; it did cross my mind but I copied it from the hide my posts tweak without really thinking about it. I suppose that script doesn't need one either.

I was thinking a bit about unfiltering a filtered post, but that would only have an effect if you liked a post in between loading it and unfiltering it, which presumably you can't do (no button). (Now I wonder if the l-to-like shortcut works on filtered posts, which I would view as a bug.)

AprilSylph (approved):

PR #597, opened 2022-05-14
new tweak: Hide footer tooltips

### LINKED ISSUE #595: hide Tumblr's pop-in labels for control buttons
### Discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/575

<sup>Originally posted by **faefangirl** April 25, 2022</sup>
I would like to see a feature that turns off the hover animation/text when you hover over the Like button. The hover delay often causes my likes to not register and I have to like again or wait for the animation. </div>

marcustyphoon:
No idea if this works with the newest footer redesign, as I don't have it.

#### User-facing changes
- Adds a tweak to hide the tooltips for the post footer control buttons.

#### Technical explanation
This could also target `article footer ${keyToCss('controlIcon')} ${keyToCss('popoverHolder')}`; not sure which is more robust against changes. `popoverHolder` is used for footer popovers that we don't want to target (the "oldest first" / "newest first" selector), so care must be taken here.

#### Issues this closes
resolves #595, #575

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -55,6 +55,11 @@
       "label": "Hide the note type badges in post footers",
       "default": false
     },
+    "hide_footer_tooltips": {
+      "type": "checkbox",
+      "label": "Hide the tooltips for the post footer control buttons",
+      "default": false
+    },
     "hide_filtered_posts": {
       "type": "checkbox",
       "label": "Hide filtered posts entirely",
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
the current approach is a little less elegant in code but definitely more robust. it's not unthinkable that another type of popover could be attached to a control icon - i mean, we do it ourselves with quick reblog!

PR #601, opened 2022-05-17
Create accurate isMyPost utility

### LINKED ISSUE #627: NotificationBlock can't be used on Blazed posts
### Platform

macOS Big Sur 11.6.5

### Browser

Firefox

### Addon version

v0.19.6

### Details

I can't block notifications. When I press the meatball thing, I only see [this](https://64.media.tumblr.com/2898e3f0b959eed44c63f86b31d2a0d4/690cc01f5c73a6db-75/s540x810/903236bca8b0a7b2bcbf917220de3ac9d3bb2f33.png). I've for sure got it turned on as you can see [here](https://64.media.tumblr.com/a64d0fb739951858978092c62af0f802/690cc01f5c73a6db-db/s1280x1920/714095fe5509fad2a6a57f1a3b4f0dc192ec64e6.png) but it's still giving me notifications for this post in my activity tabs. Thank you please help.

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
Extracts the reused "is this my own post" logic into a utility function.

Improves own-post finding logic to not use editability status (is that a word?), ensuring that blazing a post does not cause it to be incorrectly detected as not yours.

#### Issues this closes
resolves #627


src/util/react_props.js
```diff
@@ -28,6 +29,12 @@ export const timelineObject = async function (postElement) {
   return timelineObjectCache.get(postElement);
 };
 
+const primaryBlogName = await getPrimaryBlogName().catch(() => undefined);
```
AprilSylph, line 32: we wouldn't need to `await` this if `user.js` exported synchronous functions, which would be completely possible if it top-level awaited the `apiFetch` call

(not a blocker, i don't expect you to go overhaul the whole utility; you are of course free to, though)

marcustyphoon: Oh yeah!

I can do most of the top level `await` chores if you'd like.

AprilSylph: fine by me, just make sure to split it up per util file. this is where i'd probably get tempted to open one giant PR

AprilSylph:
Do you still want this reviewed as-is or should we convert the user util functions to be synchronous first?

marcustyphoon:
No preference here. Would you like that conversion after merging this, in this PR, or as another PR to review first?

AprilSylph:
I think I'd prefer a separate PR first!

AprilSylph:
As per https://github.com/AprilSylph/XKit-Rewritten/issues/627#issuecomment-1146687116, relying on `canEdit` may disqualify the user's own Blazed posts from being counted as own posts by this logic. It would be nice to fix that!

An idea I have is to export two new constants in `user.js`: one of blogs where the user is an admin, and another of just those blogs' names. We could use the latter as an alternative to `canEdit`, something like...

```js
const isAdmin = adminBlogNames.includes(blog.name);

return (canEdit || isAdmin) && <group blog is-actually-mine assurance>;
```

There might be cases I'm not thinking of that this wouldn't cover, but intuitively it seems like this would be a solid fallback. If you can't edit it because you're not a blog admin and it's not your post, it's not your post. If you can't edit it but you are a blog admin, you'll probably be able to edit it later anyway, so continue with the checks.

As an aside, this actually has me happy that Quick Tags & Trim Reblogs literally only check for the edit button. I didn't know it was going to be possible for a post to be yours but not editable, but using dead simple assumptions has covered that case for those features already!

marcustyphoon:
Wait, if you're part of a group blog where you're not an admin... currently, if you make a post... then you can edit the post, `postAuthor === primaryBlogName` is true, and this returns true, right? Or do those not count as your own posts - I forget, now.

AprilSylph:
A post is "yours" if you made it or (regarding submissions) it can't be proven that you didn't publish it. Posts made on group blogs are editable by the post author (regardless of admin status) and all group admins.

marcustyphoon:
Right, so this would still lead to inconsistent results on posts you make on group blogs. `(canEdit || isAdmin)` would still depend on the blaze status of the post.

src/util/react_props.js
```diff
@@ -1,4 +1,5 @@
 import { inject } from './inject.js';
+import { primaryBlogName, adminBlogNames } from './user.js';
```
AprilSylph, line 2:
```suggestion
import { primaryBlogName, userBlogNames, adminBlogNames } from './user.js';
```

src/util/react_props.js
```diff
@@ -28,6 +29,13 @@ export const timelineObject = async function (postElement) {
   return timelineObjectCache.get(postElement);
 };
 
+export const isMyPost = async (postElement) => {
+  const { blog, canEdit, isSubmission, postAuthor } = await timelineObject(postElement);
+  const isAdmin = adminBlogNames.includes(blog.name);
```
AprilSylph, line 34:
```suggestion
  const userIsMember = userBlogNames.includes(blog.name);
  const userIsAdmin = adminBlogNames.includes(blog.name);
```

src/util/react_props.js
```diff
@@ -28,6 +29,13 @@ export const timelineObject = async function (postElement) {
...
+
+  return (canEdit || isAdmin) && (isSubmission || postAuthor === primaryBlogName || postAuthor === undefined);
```
AprilSylph, line 36:
```suggestion
  // Post belongs to the user's primary blog
  if (blog.name === primaryBlogName) return true;

  // Post belongs to the user's single-member sideblog
  if (postAuthor === undefined && userIsMember) return true;

  // Post was created by the user on a group blog
  if (postAuthor === primaryBlogName && !isSubmission) return true;

  // Submission belongs to group blog which the user is admin of
  if (isSubmission && userIsAdmin) return true;

  return false;
```

src/util/react_props.js
```diff
@@ -28,6 +29,13 @@ export const timelineObject = async function (postElement) {
   return timelineObjectCache.get(postElement);
 };
 
+export const isMyPost = async (postElement) => {
+  const { blog, canEdit, isSubmission, postAuthor } = await timelineObject(postElement);
```
AprilSylph, line 33:
```suggestion
  const { blog, isSubmission, postAuthor } = await timelineObject(postElement);
```

AprilSylph (changes requested):
stupid problems require foolproof solutions

(this comment will be funnier if i actually missed a case here.)

src/util/user.js
```diff
@@ -26,3 +26,13 @@ export const primaryBlog = userBlogs.find(blog => blog.primary === true);
  * {string?} primaryBlogName - The name of the user's primary blog
  */
 export const primaryBlogName = primaryBlog?.name;
+
+/**
+ * {object[]} userBlogs - An array of blog objects the current user has post access to
```
AprilSylph, line 31:
```suggestion
 * {object[]} adminBlogs - An array of blog objects the current user is admin of
```

src/util/user.js
```diff
@@ -26,3 +26,13 @@ export const primaryBlog = userBlogs.find(blog => blog.primary === true);
...
+ */
+export const adminBlogs = userInfo?.blogs?.filter(blog => blog.admin) ?? [];
+
+/**
+ * {string[]} userBlogNames - An array of blog names the current user has post access to
```
AprilSylph, line 36:
```suggestion
 * {string[]} adminBlogNames - An array of blog names the current user is admin of
```

AprilSylph (changes requested):
also jsdoc on the new exports

marcustyphoon:
Perfect, thanks! Had a hell of a time trying to figure this one out. All of this looks correct to me!

Make sure you don't lose coauthor on this one :D

AprilSylph (approved):

Issue #602, opened 2022-05-18
Anti-Capitalism hides asks in the inbox


AprilSylph:
Looks like `videoCTA` is no longer in Tumblr's CSS map, somehow causing Anti-Capitalism to go a little haywire here:

https://github.com/AprilSylph/XKit-Rewritten/blob/1695fbe056dee86a92f1c6397e5c3135e7478a5e/src/scripts/anti_capitalism.js#L63

I wonder what `keyToCss('videoCTA')` is being evaluated to. We really should have better protections in place for CSS map key removal.

marcustyphoon:
Oh my god. Asks' postAttribution elements actually have the css class `undefined`. Legendary.

marcustyphoon:
I'll run anti-capitalism with show all ads on to see if a video ad pops up. I guess there's no need to specifically remove this code once the fix is implemented.

AprilSylph:
I don't think Tumblr is running any takeovers right now, so... the code might work again when they run their next one? Or it won't, but might be similar enough to be easily adapted. I guess we'll have to see.

PR #603, opened 2022-05-18
Fix CSS map removed key behavior

marcustyphoon:
#### User-facing changes
- Fixes script misbehavior when Tumblr removes CSS keys

#### Technical explanation
If `keyToCss` is called with an invalid/no-longer-valid key name, it now outputs `:is()`, which doesn't appear to match anything. Previously, it cast `undefined` to `'undefined'`, and thus output `:is(.undefined)`, which actually did match a `postAttribution` element briefly rendered inside asks as they load that had the css class `undefined`.

#### Issues this closes
Addresses the core issue in #602

### DIFF
--- a/src/util/css_map.js
+++ b/src/util/css_map.js
@@ -6,7 +6,7 @@ import { getCssMap } from './tumblr_helpers.js';
  */
 export const keyToClasses = async function (...keys) {
   const cssMap = await getCssMap;
-  return keys.flatMap(key => cssMap[key]);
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
merging without testing due to simplicity of change, plus i'll probably test it before release anyway

...it'll be fun to make the css map functions synchronous via top-level await and do away with resolveExpressions entirely, won't it!

PR #610, opened 2022-05-20
Convert translate utility to synchronous

marcustyphoon:
#### User-facing changes
- n/a

#### Technical explanation
Converts the `translate` utility to be synchronous using top-level `await`.

#### Issues this closes
n/a

### DIFF
--- a/src/util/language_data.js
+++ b/src/util/language_data.js
@@ -1,10 +1,9 @@
-import { getLanguageData } from './tumblr_helpers.js';
+import { inject } from './inject.js';
+
+export const languageData = await inject(() => window.tumblr.languageData);
 
 /**
  * @param {string} rootString - The English string to translate
- * @returns {Promise<string>} - The translated string in the current Tumblr locale
+ * @returns {string} - The translated string in the current Tumblr locale
  */
-export const translate = async function (rootString) {
-  const languageData = await getLanguageData;
-  return languageData.translations[rootString] || rootString;
-};
+export const translate = rootString => languageData.translations[rootString] || rootString;
--- a/src/scripts/accesskit/visible_alt_text.js
+++ b/src/scripts/accesskit/visible_alt_text.js
@@ -7,7 +7,6 @@ import { getPreferences } from '../../util/preferences.js';
 let mode;
 
 let imageBlockSelector;
-let imageString;
 
 const styleElement = buildStyle();
 const processedClass = 'accesskit-visible-alt-text';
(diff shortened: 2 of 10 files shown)


src/util/language_data.js
```diff
@@ -1,10 +1,9 @@
-import { getLanguageData } from './tumblr_helpers.js';
+import { inject } from './inject.js';
+
+export const languageData = await inject(() => window.tumblr.languageData);
```
AprilSylph, line 3: What's the rationale behind using `inject` instead of just `await getLanguageData`? Injecting a synchronous function?

(If so, feel free to modify tumblr_helpers just for this; it's not like `getLanguageData` is used anywhere else)

marcustyphoon: Oh, I had moved `getLanguageData` into `language_data.js` and then converted it this way - whoops, I didn't remove it from `tumblr_helpers.js`. Do you think it makes sense for it to be there? I'd buy that.

marcustyphoon: I suppose it doesn't matter which file actually has the side effect, right?
```js
// tumblr_helpers.js
export const getLanguageData = () => inject(async () => window.tumblr.languageData);

// language_data.js
import { getLanguageData } from './tumblr_helpers.js';
const languageData = await getLanguageData();
```

```js
// tumblr_helpers.js
export const languageData = await inject(async () => window.tumblr.languageData);

// language_data.js
import { languageData } from './tumblr_helpers.js';
```

AprilSylph: last option looks best to me, but you're right that these are all basically the same in terms of behaviour.

AprilSylph: there's definitely a point to be made about making util files non-dependent where possible... not sure what's best for performance

marcustyphoon: Oops, I forgot the `inject` import in the last one; I guess it's similar in line count. The real difference is which file one would import `languageData` from if for some reason a script wanted access to it. Well, okay, that's not a "real" difference either, but as far as semantic correctness goes.

Anyway, I'm not entirely sure off the top of my head exactly how the not-a-waterfall works, but it seemed from the descriptions of top level await that it was designed to minimize the performance impact. Certainly if top level code has bugs that throw or never resolve, there would be some effects from that, but I think that goes without saying.

AprilSylph: oh, yeah, it would make way more sense to import `languageData` from `language_data.js`. i wasn't thinking of it that way round!

👍

AprilSylph (approved):
a beautiful start to an exciting journey

PR #614, opened 2022-05-25
Convert user utilities to synchronous

marcustyphoon:
Okay, question: do you think it makes more sense to make the `user.js` utility exports constants, or functions?

The major difference I can see is in what happens when the API query fails (generally when the user is logged out). I assume - I just realized while writing this that I didn't test it - that if there were no error handling in a top level await, the dynamic `import()` in `main.js` would crash, and while this is handled, it seems to me like that's not ideal. Thus it makes sense to handle errors gracefully. If `user.js` exports a constant `primaryBlogName`, it has to be set to something if the API fetch fails (in this case, `undefined`), while keeping the `getPrimaryBlogName` function export could have that function throw. Before this PR, generally speaking, `user.js` throwing would crash the main function on any script that required it, nicely ensuring that scripts which expect a user do not run when there is none.

Of course, scripts should be responsible for not having buggy behavior unless `main.js` checked if you were logged in and just stopped the extension entirely if not, though some scripts do work when logged out. Even in that case, though, API queries aren't guaranteed to work, and the `.undefined` css map situation was a good reminder that graceful failure modes are nice.

---

#### User-facing changes
- n/a

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/anti_capitalism.js
+++ b/src/scripts/anti_capitalism.js
@@ -25,7 +25,8 @@ const showPremiumPlea = async () => {
   const { [storageKey]: hasShownPlea } = await browser.storage.local.get(storageKey);
   if (hasShownPlea) return;
 
-  const { response: { user: { hasTumblrPremium, usedToHaveTumblrPremium } } } = await userInfo;
+  if (userInfo === undefined) return;
+  const { hasTumblrPremium, usedToHaveTumblrPremium } = userInfo;
   if (hasTumblrPremium || usedToHaveTumblrPremium) {
     browser.storage.local.set({ [storageKey]: true });
     return;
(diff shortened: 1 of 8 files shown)


AprilSylph:
This is something I was thinking about too! It definitely makes sense to me to just export constants, with non-logged-in values of empty array for blogs/blognames, `undefined` for primary blog name, etc.

This would require a lot more adaptation to feature scripts than just removing `await` statements though; if you want to split it into a separate PR, that's totally fine, but yes please definitely exported constants when you feel like it.

AprilSylph (approved):
Very nice work! I couldn't see any issues in the affected features while logged out.

TIL reblog buttons while logged out actually have no `href`, so Quick Reblog can't be shown, but even if it is, the blog selector being empty actually isn't an issue (though reblogging will obviously fail, with a message from Tumblr saying you need to be logged in).

PR #615, opened 2022-05-25
Fix disable gifs "poster" incompatibility

marcustyphoon:
Well, this is kind of a mess (a least temporarily?)

It appears that Tumblr now places a "poster" static image element over GIFs, with CSS that defaults both the "poster" and actual animated image element to `visibility: hidden`, and moves a class that sets `visibility: visible` from the poster to the animated element once the gif loads.

If this is fully rolled out, usable everywhere, and we want to rely on it, this makes the feature extremely easy to implement, as we can just reuse the "poster" element instead of using canvases. I lack the forbidden knowledge to know if this is currently true.

In the meantime (?), as a quick and dirty fix, this PR stops a second canvas element from being created by the "poster" element and ensures that whether or not the `visible` class gets copied, the canvas element is set to `visibility: visible`.

Not sure what happens to offscreen gifs that get scrolled back onscreen at the moment.

#### User-facing changes
- Fixes disable gifs with new Tumblr gif code

#### Technical explanation
see above

#### Issues this closes
Discussion #612

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -19,6 +19,7 @@
 
 .xkit-paused-gif {
   position: absolute;
+  visibility: visible;
 }
 
 figure:hover .xkit-paused-gif,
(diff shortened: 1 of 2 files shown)


AprilSylph:
This isn't fully rolled out (I don't have it on my main) so quick and dirty it stays

AprilSylph (approved):
Code looks good and there's no regressions with non-postered GIFs; LGTM!

PR #629, opened 2022-06-05
Themed Posts: Implement per-trail-item theming

marcustyphoon:
Okay, now I see what the "oh no" was all about. Ow, my eyes.

I have no idea how to fix the margin issues with `reblog` elements here. When a reblog ends in a paragraph, the paragraph has a 15px margin that collapses with the `reblog` element margin, but that makes a gap in the colored sections. Adding a 15px bottom padding makes the gap too large in this case but correct when the `reblog` ends in a media element; adding a tiny bottom padding just to remove the margin collapse makes the gap too small when the `reblog` ends in a media element; you can't color in margins.

Unless someone wants to solve this or tell me the forbidden knowledge... yeah, I've got nothing.

#### User-facing changes
- Adds an option to individually theme every reblog trail item, as requested by Cyle. Blinds the user.

#### Technical explanation
The box model hates me.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -81,8 +81,11 @@ export const main = async function () {
 
   styleElement.textContent += `
     article ${reblogSelector} {
-      margin-top: 0;
-      padding-bottom: 1px;
+      box-shadow: 0px 15px 0px 0px RGB(var(--white));
+    }
+    article ${await keyToCss('holder')} {
+      padding-bottom: 15px;
+      margin-bottom: -15px;
     }
   `;
 
--- a/src/scripts/themed_posts.js
+++ b/src/scripts/themed_posts.js
@@ -2,15 +2,20 @@ import { buildStyle, filterPostElements, blogViewSelector } from '../util/interf
 import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
 import { timelineObject } from '../util/react_props.js';
+import { keyToCss } from '../util/css_map.js';
 
 const styleElement = buildStyle();
 const blogs = new Set();
 const groupsFromHex = /^#(?<red>[A-Fa-f0-9]{1,2})(?<green>[A-Fa-f0-9]{1,2})(?<blue>[A-Fa-f0-9]{1,2})$/;
 
+let reblogSelector;
+
 let enableOnPeepr;
 let blacklistedUsernames;
 
 let blacklist;
+
+let reblogTrailTheming = true;
 
 const hexToRGB = (hex) => {
   const { red, green, blue } = hex.match(groupsFromHex).groups;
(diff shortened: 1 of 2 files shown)


marcustyphoon:
No idea how to make any progress on the CSS here. (And anyway, does it count as a joke extension?) Opinion on close/leave/undraft/etc?

src/scripts/themed_posts.js
```diff
@@ -24,41 +29,63 @@ const processPosts = async function (postElements) {
...
+  ({ reblogTrailTheming, enableOnPeepr, blacklistedUsernames } = await getPreferences('themed_posts'));
   blacklist = blacklistedUsernames.split(',').map(username => username.trim());
 
+  reblogSelector = await keyToCss('reblog');
+
+  styleElement.textContent += `
+    article ${reblogSelector} {
+      margin-top: 0;
+      padding-bottom: 1px;
+    }
```
AprilSylph, line 86: Here's that forbidden knowledge you ordered:

```suggestion
    article ${reblogSelector} {
      overflow-y: hidden;
      margin-top: 0;
    }

    article ${reblogSelector}:not(:last-child) > :last-child {
      margin-bottom: 15px;
    }
```

Setting `overflow: hidden` makes descendant margins count towards the content box size, instead of collapsing through it. Descendant margins can still collapse into _each other_ though, allowing us to set a minimum amount of bottom spacing without padding at all.

marcustyphoon: I think I have read this paragraph like 10 times :D I think I get most of it. Out of curiosity, Is this related to this paragraph?

> Specifying a value other than visible (the default) or clip creates a new [block formatting context](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context). This is necessary for technical reasons — if a float intersected with the scrolling element it would forcibly rewrap the content after each scroll step, leading to a slow scrolling experience.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#description

src/scripts/themed_posts.js
```diff
@@ -24,41 +29,67 @@ const processPosts = async function (postElements) {
...
+  ({ reblogTrailTheming, enableOnPeepr, blacklistedUsernames } = await getPreferences('themed_posts'));
   blacklist = blacklistedUsernames.split(',').map(username => username.trim());
 
+  reblogSelector = await keyToCss('reblog');
+
+  styleElement.textContent += `
+    article ${reblogSelector} {
+      overflow-y: hidden;
+      margin-top: 0;
+    }
```
marcustyphoon, line 86:
```suggestion
  styleElement.textContent += `
    article ${reblogSelector} {
      display: flow-root;
      margin-top: 0;
    }
```

Does this serve the same purpose as setting the overflow? I saw it in the infobox in https://www.joshwcomeau.com/css/rules-of-margin-collapse/#margins-can-collapse-in-the-same-direction - never heard of it before.

AprilSylph: Ah, once again my instinct about more simple things existing turns out to be right. I wouldn't have thought it was a `display` value...

Tested and works.

marcustyphoon: Me neither! I feel like it should have its own MDN. The caniuse looks fine, though.

src/scripts/themed_posts.js
```diff
@@ -24,41 +28,65 @@ const processPosts = async function (postElements) {
...
+  styleElement.textContent += `
+    article ${reblogSelector} {
+      display: flow-root;
+      margin-top: 0;
+    }
+
+    article ${reblogSelector}:not(:last-child) > :last-child {
+      margin-bottom: 15px;
+    }
+  `;
```
AprilSylph, line 88: Two things about this:
1. I'd like to see it wrapped in an `if (reblogTrailTheming)` block, so that it's not needlessly applied when it would have no effect
2. Instead of `article` I'd prefer to see `${postSelector}`

AprilSylph (changes requested):
Almost there!

marcustyphoon:
Interesting that the the 12px margin on top of the footer doesn't collapse into... something... with this active, on one of the posts I was looking at, but not on another one. Didn't fully investigate why. I don't think it looks broken or anything, but it is a visible difference in positioning.

AprilSylph (approved):
TASTE THE RAINBOW

PR #630, opened 2022-06-06
NotificationBlock: Fix menu item missing on non-editable posts

### LINKED ISSUE #627: NotificationBlock can't be used on Blazed posts

marcustyphoon:
#### User-facing changes
- NotificationBlock menu items appear on non-editable posts

#### Technical explanation
The notificationblock menu item appears on any post from any of the user's blogs, regardless of whether the user has edit access to that post.

This makes sense for blazed posts the user has made, as well as group blog posts that the user didn't make and can't edit, as the user can still see notifications on those.

#### Issues this closes
resolves #627
(for real this time)

### DIFF
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -80,12 +80,9 @@ const onButtonClicked = async function ({ currentTarget }) {
   });
 };
 
-const blockPostFilter = async ({ blog, rebloggedFromName, id, rebloggedRootId }) => {
+const blockPostFilter = async ({ blog, id, rebloggedRootId }) => {
   const rootId = rebloggedRootId || id;
-  return (
-    (userBlogNames.includes(blog.name) || userBlogNames.includes(rebloggedFromName)) &&
-    blockedPostTargetIDs.includes(rootId) === false
-  );
+  return userBlogNames.includes(blog.name) && blockedPostTargetIDs.includes(rootId) === false;
 };
 
 const unblockPostFilter = async ({ id, rebloggedRootId }) => {
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -80,9 +80,12 @@ const onButtonClicked = async function ({ currentTarget }) {
   });
 };
 
-const blockPostFilter = async ({ blog, id, rebloggedRootId }) => {
+const blockPostFilter = async ({ blogName, rebloggedRootName, rebloggedFromName, id, rebloggedRootId }) => {
   const rootId = rebloggedRootId || id;
-  return userBlogNames.includes(blog.name) && blockedPostTargetIDs.includes(rootId) === false;
+  const yourOriginalPost = userBlogNames.includes(rebloggedRootName);
+  const yourReblog = userBlogNames.includes(blogName) || userBlogNames.includes(rebloggedFromName);
+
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Hm, you definitely are right that adding the UI to the notifications themselves solves a lot of weirdness.

I just noticed that it would - in theory - make sense to also have these buttons on any post containing one of your blogs in the reblog chain, in a way. After all, if I reblog user A's post, and I receive a notification that user B has reblogged my reblog, and I click that notification... it opens user B's blog. In that case, I have open a post which I might want to block future similar notifications from, in which I am neither the current post author nor the root post author.

marcustyphoon:
`blog.name` is equivalent to `blogName`, right?

AprilSylph:
I think `blogName` and `blog.name` should be the same, yeah.

src/scripts/notificationblock.js
```diff
@@ -79,9 +80,12 @@ const onButtonClicked = async function ({ currentTarget }) {
...
 };
 
-const blockPostFilter = async ({ id, canEdit, rebloggedRootId }) => {
+const blockPostFilter = async ({ blogName, rebloggedRootName, rebloggedFromName, id, rebloggedRootId }) => {
   const rootId = rebloggedRootId || id;
-  return canEdit && blockedPostTargetIDs.includes(rootId) === false;
+  const yourOriginalPost = userBlogNames.includes(rebloggedRootName);
+  const yourReblog = userBlogNames.includes(blogName) || userBlogNames.includes(rebloggedFromName);
+
+  return (yourOriginalPost || yourReblog) && blockedPostTargetIDs.includes(rootId) === false;
```
AprilSylph, line 88: These names aren't strictly accurate... `rebloggedRootName` will be undefined when viewing _any_ original post, which means all original posts when viewing your own blog would be counted as `yourReblog` rather than `yourOriginalPost`.

What's being tested for *is* correct though, so let's just combine the two constants.

```suggestion
  const canReceiveActivity = userBlogNames.includes(blogName) ||
    userBlogNames.includes(rebloggedFromName) ||
    userBlogNames.includes(rebloggedRootName);

  return canReceiveActivity && blockedPostTargetIDs.includes(rootId) === false;
```

AprilSylph (changes requested):

src/scripts/notificationblock.js
marcustyphoon: Oh, right! Whoops. Sometimes I think I need to just write documentation for the timelineObject fields for myself as if they were official API endpoints :D

AprilSylph (approved):

PR #638, opened 2022-06-18
No Recommended: hide "more like this" after liking posts

### LINKED ISSUE #637: Block the "more like this" section via No Recommended
### Concept

I've encountered a new (or at least new-to-me) way for Tumblr to shove recommendations in my face. Below is a screenshot of a "more like this" section which appeared when I liked a post.

image

This could be blocked by No Recommended, I assume, and it would be nice to have it do so!

marcustyphoon:
my brain is fried, I don't think this is good UI copy (should it use "more like this" maybe?), and I can't remember if the "more like this" string is already used elsewhere

#### User-facing changes
- Adds an option to No Recommended to hide the "more like this" recommended posts that appear after you like a post

#### Technical explanation
In a departure from the other carousel hiding scripts, this targets the header text instead of the DOM structure of the recommendation bit; I don't know if `chicletRow` is already used elsewhere/will get used elsewhere.

If Staff adds RNG to the title, then idk, I give up :P

#### Issues this closes
resolves #637

### DIFF
--- a/src/scripts/no_recommended.json
+++ b/src/scripts/no_recommended.json
@@ -33,6 +33,11 @@
       "label": "Hide related posts in the photo lightbox",
       "default": false
     },
+    "hide_liked_related_posts": {
+      "type": "checkbox",
+      "label": "Hide related posts after you like a post",
+      "default": false
+    },
     "hide_recommended_blogs": {
       "type": "checkbox",
       "label": "Hide recommended blogs in the sidebar",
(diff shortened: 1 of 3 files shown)


AprilSylph:
There's a "More like this" title for recommendations in the photo lightbox. Do the chiclet posts themselves have no useful react data?

marcustyphoon:
Hm. I assume not a lot of unrelated components would subscribe to `relatedPostsContext`. Seems reasonable.

marcustyphoon:
I assume(?) that the test feature this PR is for is not currently active; feel free to close if so, and this can be reopened if it resurfaces.

AprilSylph:
It's 5 hours too late for me to check, so I'll have to try to remember to look on Monday.

AprilSylph:
It's still active. My main account has it.

marcustyphoon:
Gotcha. Darn, my test account appears to have lost it.

marcustyphoon:
Closing this because whether or not the feature it's supposed to hide still exits, iirc the code doesn't work.

PR #639, opened 2022-06-22
Add tweak 'Hide activity notification badge'

krc:
'Hide activity notification badge' based on the version in Xkit 7

#### User-facing changes
Option to hide the notification badge on the activity button on desktop

#### Technical explanation
I got the names from the tweak in Xkit 7

#### Issues this closes
no open issue, I went right to (draft) PR

I am not a "real" developer, and this is my first ever PR. Hope I did things somewhat right.

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -80,6 +80,11 @@
       "label": "Hide my follower count where possible",
       "default": false
     },
+    "hide_activity_notification_badge": {
+      "type": "checkbox",
+      "label": "Hide the activity notification badge",
+      "default": false
+    },
     "no_where_were_we": {
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
--- a/src/scripts/tweaks/hide_activity_notification_badge.js
+++ b/src/scripts/tweaks/hide_activity_notification_badge.js
@@ -1,9 +1,12 @@
-import { keyToClasses } from '../../util/css_map.js';
+import { keyToCss } from '../../util/css_map.js';
 import { buildStyle } from '../../util/interface.js';
 import { translate } from '../../util/language_data.js';
 
-const badgeSelector = keyToClasses('notificationBadge').map(cssClass => `button[aria-label="${translate('Activity')}"] .${cssClass}`).join(',');
-const styleElement = buildStyle(`${badgeSelector} { display: none; }`);
+const styleElement = buildStyle(`
+button[aria-label="${translate('Activity')}"] ${keyToCss('notificationBadge')} {
+  display: none;
+}
+`);
 
 export const main = async () => document.head.append(styleElement);
 export const clean = async () => styleElement.remove();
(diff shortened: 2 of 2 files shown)


krc:
Ok that was an embarrassing number of tries but it passes now at least. Let me know if there is anything else I could fix/do better.

AprilSylph:
This is really good for a first contribution! Don't worry about the number of tries, without a local dev setup there's not much else to do but push and let the linter run here.

I believe what you have here should work, but there's actually an even easier way to accomplish what this new tweak does. I'll be very happy to show you as part of a proper review.

Also, thanks for doing this. If the activity badge worked for my main account, I'd probably want this option too.

Now that the tests are passing, do you want to mark this as ready for review?

krc:
> I'll be very happy to show you as part of a proper review.

That would be great thank you! :-)

AprilSylph (approved):
Alright, so: first of all, congratulations! This works exactly as intended.

But as I mentioned, there's actually a way to simplify the code. I can't blame you at all for just copying from XKit 7, but XKit Rewritten's CSS map utilities are a bit more powerful.

---

In XKit 7, `keyToCss` just spits out a list of selectors which match the key. For example, `keyToCss('timeline')` will give us `.j8ha0, .AiNLD, .wDihp, .dOTgj`. This is useful sometimes, but if you want to add stuff to the beginning or end, you have to use `keyToClasses().map().join()` instead to construct multiple selectors with each classname, as this code does.

Say you want to target any Tumblr `'timeline'` in any `<main>` element. If you just do this...
```js
css = `main ${keyToCss('timeline')`;
```
You'll end up with this:
```css
main .j8ha0, .AiNLD, .wDihp, .dOTgj
```
Which doesn't do what you want, because commas mark the beginning of a new selector entirely. This will target any elements matching `.AiNLD`, `.wDihp`, `.dOTgj` or `main .j8ha0`.

---

In XKit Rewritten, `keyToCss()` instead outputs a much more useful [`:is()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) selector. This allows us to group selectors and let the browser figure out what we mean. Where we'd need to use `keyToClasses().map().join()` previously to produce a selector like this:
```css
main .j8ha0,
main .AiNLD,
main .wDihp,
main .dOTgj
```
We actually _can_ use XKit Rewritten's `keyToCss()` as if it only produced one selector! For example:
```js
const styleElement = buildStyle(`main ${keyToCss('timeline') { /* some styles */ }`);
```
Results in:
```html
<style class="xkit">
main :is(.j8ha0, .AiNLD, .wDihp, .dOTgj) { /* some styles */ }
</style>
```
Which targets the exact same elements but without repeating the `main` part at all.

This is not only much easier for us to read - it's easier for the browser, too. We saw a massive performance boost when we crunched rules which used a ton of separate selectors down to just one selector with multiple usages of `:is()`.

Ironically, this actually isn't too relevant to this new tweak, as `'badgeSelector'` only has a single match in Tumblr's CSS map. But if you plan to contribute more in future, it's a very handy tool!

---

I've added some code suggestions below which implements the simplified syntax. Just be sure to commit both in a batch, or the tests will fail again until you commit the other!

AprilSylph (approved):
Committed the suggestions for you.

This will go out with the next release; thanks again for contributing!

PR #667, opened 2022-07-19
AccessKit Disable GIFs: Use native poster elements 

### LINKED ISSUE #664: AccessKit: pause GIFs using Tumblr's poster elements
GIF posters are out to everyone now, so if GIFs as MP4 have the same poster structure, the whole option could be changed to take advantage of them and implement this in the process.

Otherwise, they will need to be specifically dealt with.

marcustyphoon:
"special cases: the ~~movie~~ xkit extension"

#### User-facing changes
- Fixes disable gifs not working when gifs are served as mp4.

#### Technical explanation
Takes advantage of Tumblr's non-animated "poster" elements that are used to hide gifs while they load... but only in gifs that are part of a post body (currently checking for `figure` elements); other locations where gifs are rendered do not use posters, so the old canvas-based code is still used for those.

A huge mess, basically.

#### Issues this closes
resolves #664

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -20,16 +20,6 @@
 
 .xkit-paused-gif {
   position: absolute;
-  visibility: visible;
-
-  background-color: rgb(var(--white));
-}
-
-*:hover > .xkit-paused-gif,
-*:hover > .xkit-paused-gif-label,
-.xkit-paused-gif-container:hover .xkit-paused-gif,
-.xkit-paused-gif-container:hover .xkit-paused-gif-label {
-  display: none;
 }
 
 .xkit-paused-background-gif:not(:hover) {
--- a/src/scripts/accesskit/disable_gifs.js
+++ b/src/scripts/accesskit/disable_gifs.js
@@ -5,14 +5,16 @@ import { buildStyle, postSelector } from '../../util/interface.js';
 
 const className = 'accesskit-disable-gifs';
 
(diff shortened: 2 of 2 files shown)


marcustyphoon:
This appears to do fairly bad things to the foreground/background/postGlass setup in the cardPost element used in the recommended blog carousel, but I'm not sure why at the moment.

marcustyphoon:
Note that `tagImage` elements* and the post previews in the blog cards** do not use gif posters. I don't know about animated `takeoverBanner`s*.

*as of this commit, implemented using the previous canvas-based code. I do not know if these elements can contain mp4 gifs, which would necessitate modification to that code

**just noticed this; not paused as of this commit

marcustyphoon:
Hm. Setting the background color to white requires a really high specificity in the case of the recommended blog carousel, and also doesn't look right. Would probably be more correct to visibility:hidden the background image when not hovered, but to be honest the logic to make that correct sounds like a nightmare, so, eh.

AprilSylph:
> User-facing changes: Fixes disable gifs not working when gifs are served as mp4.

Serving GIFs as MP4 is a dead experiment, so I'm going to close this as unnecessary. Using native GIF posters to pause GIFs is a neat idea for simplifying this option's code, but not worth complicating it for.

marcustyphoon:
👍  If it's ever able to be used for every element, it would be great; given that that is not the case this is absolutely not worth it.

PR #672, opened 2022-07-22
No Recommended: Fix visible MP4 gifs from hidden blog recommendations

### LINKED ISSUE #666: No Recommended: broken carousel hiding with GIFs served as MP4
#616... 2!

image

marcustyphoon:
#### User-facing changes
- Fixes GIFs from carousels sometimes appearing in the background between posts. Again.

#### Technical explanation
To avoid specificity/`!important` issues between this and disable gifs when toggling visibility on and off, I just used display:none here. As long as the blog carousel elements still exist, this shouldn't trigger the infinite loading thing.

#### Issues this closes
resolves #666

### DIFF
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -7,7 +7,7 @@ const hiddenClass = 'xkit-no-recommended-blog-carousels-hidden';
 const styleElement = buildStyle(`
   .${hiddenClass} { position: relative; }
   .${hiddenClass} > div { visibility: hidden; position: absolute; max-width: 100%; }
-  .${hiddenClass} > div ${keyToCss('visible', 'poster')}, .${hiddenClass} > div canvas { display: none }
+  .${hiddenClass} > div :is(img, video, canvas) { display: none }
 `);
 
 const listTimelineObjectSelector = keyToCss('listTimelineObject');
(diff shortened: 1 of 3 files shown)


src/scripts/no_recommended/hide_blog_carousels.js
```diff
@@ -7,7 +7,7 @@ const hiddenClass = 'xkit-no-recommended-blog-carousels-hidden';
 const styleElement = buildStyle(`
   .${hiddenClass} { position: relative; }
   .${hiddenClass} > div { visibility: hidden; position: absolute; max-width: 100%; }
-  .${hiddenClass} > div img, .${hiddenClass} > div canvas { visibility: hidden; }
+  .${hiddenClass} > div ${keyToCss('visible', 'poster')}, .${hiddenClass} > div canvas { display: none }
```
AprilSylph, line 10: 1. Is there a purpose to using the CSS map here other than future-proofing? I can't imagine Tumblr would start using a third image-holding element... if there were any advantages to `<picture>` I'm sure it would be being used already
2. If we were to go back to just targeting element types, we could simplify this line down very nicely

```suggestion
  .${hiddenClass} > div :is(img, video, canvas) { display: none }
```

marcustyphoon: Nope, not particularly! I think my line of logic was just "target the classes that redpop would be setting `visibility` on and the elements xkit would be setting `visibility` on."

AprilSylph: Solid process, unfortunately "visible" by itself is not too descriptive. Ah the joys of dealing with the result of SCSS without seeing the source sheets... in the same context, anyway.

AprilSylph (approved):

Issue #673, opened 2022-07-25
Tag Replacer only replacing 5 posts at a time


the-graves-family:
### Platform

Windows 10

### Browser

Chrome 103.0.5060.134 and Firefox 102.0.1

### Addon version

v.0.20.4 (Beta)

### Details

Hello! 

When using the Tag Replacer, it says it finds all the posts with the tag I'm trying to replace (say about 78 posts), and then, it says it's replaced all of them with the new tag. But I went and did it again, just to check, and found out it only replaced five of the posts. Every time I did it again, it replaced five posts. Reloading, logging out, restarting did nothing, nor did changing into Firefox. This happened both with tags that had a lots of posts (over 1000) and tags that had like, 20 posts.

I could use some help, even if just to know if it's a 'me' thing, or something that's happened to several people.

(I'm very very new to both XKit and GitHub, so I'm sorry if there's anything wrong or missing from this.)

AprilSylph:
Strange, not sure what could be causing this. The Mass Post Editor endpoints should be adding and then removing tags in batches of 100. Maybe the post IDs are getting mangled somewhere...

Can you add or remove tags successfully on more than 5 posts at a time via the actual Mass Post Editor? This could be a Tumblr bug.

the-graves-family:
I just checked, and I can both add and remove tags via the Mass Post Editor on more than 5 posts. Still no luck on the Tag Replacer, though.

marcustyphoon:
Was it always the first (or last) five posts that were edited, or a random selection of five? Five seems like such a specific number!

the-graves-family:
Well, it's actually weirder now. I was away from my PC all day and tried replacing tags again to take screenshots to put here and, I don't even know what it's doing anymore. It doesn't find the same number of posts it replaces, and it keeps not replacing everything. I'm really at a loss. (I'm not sure if the screenshots will work, but they were taken in order they appeared.)

![replace1](https://user-images.githubusercontent.com/109947823/181083391-cb5f6905-1a5f-4014-995d-f3a80ec5a7ff.JPG)
![replace2](https://user-images.githubusercontent.com/109947823/181083385-850e86b6-568f-4cfa-b0b0-b193a17d769d.JPG)
![replace3](https://user-images.githubusercontent.com/109947823/181083379-9e4b6e8b-fdf6-4fa8-ab65-ce9a573190fc.JPG)
[...]

PR #675, opened 2022-07-26
Tag Replacer/Mass Unliker/Mass Deleter: Remove ads from api response

marcustyphoon:
#### User-facing changes
- Tag replacer's counts will no longer be incorrect because Tumblr serves ads in the API request now (???)
- Maybe this fixes tag replacer failures (#673), but I kind of doubt it. The number always being exactly five seems too specific. (I did not observe that problem in my brief test, so I naturally didn't observe this change fixing it, either.)

#### Technical explanation
My ads had `"objectType": "client_side_ad_waterfall",` and my posts had `"objectType": "post",`.

There are a whole bunch of other API calls in the codebase to check for ads and filter, I guess.

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/tag_replacer.js
+++ b/src/scripts/tag_replacer.js
@@ -137,7 +137,8 @@ const replaceTag = async ({ uuid, oldTag, newTag, appendOnly }) => {
   while (resource) {
     await Promise.all([
       apiFetch(resource).then(({ response }) => {
-        taggedPosts.push(...response.posts);
+        const posts = response.posts.filter(object => object.objectType === 'post');
+        taggedPosts.push(...posts);
         resource = response.links?.next?.href;
 
         gatherStatus.textContent = `Found ${taggedPosts.length} posts${resource ? '...' : '.'}`;
--- a/src/scripts/tag_replacer.js
+++ b/src/scripts/tag_replacer.js
@@ -137,7 +137,7 @@ const replaceTag = async ({ uuid, oldTag, newTag, appendOnly }) => {
   while (resource) {
     await Promise.all([
       apiFetch(resource).then(({ response }) => {
-        const posts = response.posts.filter(object => object.objectType === 'post');
+        const posts = response.posts.filter(({ canEdit }) => canEdit === true);
         taggedPosts.push(...posts);
         resource = response.links?.next?.href;
 
--- a/src/scripts/mass_deleter.js
+++ b/src/scripts/mass_deleter.js
@@ -76,7 +76,11 @@ const deleteDrafts = async function ({ blogName, before }) {
   while (resource) {
     await Promise.all([
       apiFetch(resource).then(({ response }) => {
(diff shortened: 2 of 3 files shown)


AprilSylph:
A clearer sanity check would probably be `.filter({ canEdit } => canEdit === true)`. Would cover other cases of uneditable posts, too (group blogs & Blaze).

src/scripts/tag_replacer.js
```diff
@@ -137,7 +137,8 @@ const replaceTag = async ({ uuid, oldTag, newTag, appendOnly }) => {
   while (resource) {
     await Promise.all([
       apiFetch(resource).then(({ response }) => {
-        taggedPosts.push(...response.posts);
+        const posts = response.posts.filter(object => object.objectType === 'post');
```
AprilSylph, line 140:
```suggestion
        const posts = response.posts.filter(({ canEdit }) => canEdit === true);
```

marcustyphoon: Oh yeah, I guess this makes sense for mass deleter/mass unliker as well; in all cases we are collecting posts to edit them, so non-editable entries of any sort would be bad.

Come to think of it... I feel like I would want a warning instead of a silent failure if I tried to tag replace a blazed post hmmmmm

AprilSylph: Mass Deleter might need it for group blog purposes, I actually can't remember if non-admin group members can access a group blog's drafts at all, or if draft editing is free for any group member, or not.

But liking a post doesn't create an editable copy... unliking does require more than a post ID though, so filtering to objects with the necessary data to send an unlike request would be good.

marcustyphoon: Oh, right. There is a `canLike` field, though I don't know what toggles it, and thus whether it helps with any edge cases (I guess it would be about the user-has-blocked-you case?)

I guess re: "warning instead of silent failure" maybe the best way to do that would be to remove ads, then filter a second time and display "found x posts (y uneditable)." Or do you think that's showing how the sausage is made too much, so to speak, and it should just be a silent failure?

AprilSylph: You can always unlike posts regardless of blocks. If a blog has you blocked, you can't like their posts but can unlike them.

I don't see that there's a clear benefit to displaying the count of uneditable posts.., it's already an edge-case that they will occur at all, much simpler to just filter them out silently. We can always follow up and implement this later if anyone actually notices.

marcustyphoon:
Undrafting, but note that as of this message I haven't tested these commits.

AprilSylph (approved):
looks good! no trace of ads in the mass post editor requests.

didn't get to test mass deleter properly since i can't test on firefox on this laptop (thanks snap store) but clearing likes and queue also work fine.

AprilSylph:
okay technically i can test on firefox but not from the command line so it's much more annoying. so annoying that i forget it's an option.

PR #689, opened 2022-08-01
Tweaks: Add full note count tweak

### LINKED ISSUE #576: Tweak to go back to original notes
### Concept

Tweak to change how the notes look now to how they looked originally, with the full number and notes next to it

marcustyphoon:
#### User-facing changes
- Adds a tweak to display full note counts (1,234) instead of compact note counts (1.2k)

#### Technical explanation
Very similar to timeformat.

Using a pseudo element makes the note count non-selectable, which I personally find sort of annoying, but it looks like the note count isn't currently selectable anyway. Opinions on changing this?

#### Issues this closes
resolves #576

### DIFF
--- a/src/scripts/tweaks/show_full_note_count.js
+++ b/src/scripts/tweaks/show_full_note_count.js
@@ -1,10 +1,16 @@
 import { pageModifications } from '../../util/mutations.js';
-import { translate } from '../../util/language_data.js';
+import { translate, languageData } from '../../util/language_data.js';
 import { buildStyle } from '../../util/interface.js';
 import { timelineObject } from '../../util/react_props.js';
 import { keyToCss } from '../../util/css_map.js';
 
-const numberFormat = new Intl.NumberFormat();
+const createNumberFormat = async () => {
+  const { code } = await languageData;
+  const locale = code.replaceAll('_', '-');
+  return new Intl.NumberFormat(locale);
+};
+
+const numberFormat = await createNumberFormat().catch(() => new Intl.NumberFormat());
 
 const notes = translate('notes');
 const note = translate('note');
--- a/src/scripts/tweaks/show_full_note_count.js
+++ b/src/scripts/tweaks/show_full_note_count.js
@@ -1,12 +1,11 @@
 import { pageModifications } from '../../util/mutations.js';
-import { translate, languageData } from '../../util/language_data.js';
+import { translate } from '../../util/language_data.js';
 import { buildStyle, postSelector } from '../../util/interface.js';
 import { timelineObject } from '../../util/react_props.js';
 import { keyToCss } from '../../util/css_map.js';
 
 const createNumberFormat = async () => {
-  const { code } = await languageData;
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Oh, interesting performance note: this won't use the `timelineObject` cache unless `timelineObject(noteElement)` is replaced by e.g. `timelineObject(noteElement.closest(postSelector))`, but that would require an exception to be made to work in the radar since postSelector doesn't select the radar post.

In MV2, this doesn't reeeeeeeally matter, but still. Will look for other edge cases with postSelector and the radar post.

src/scripts/tweaks/show_full_note_count.js
```diff
@@ -0,0 +1,49 @@
+import { pageModifications } from '../../util/mutations.js';
+import { translate, languageData } from '../../util/language_data.js';
+import { buildStyle, postSelector } from '../../util/interface.js';
+import { timelineObject } from '../../util/react_props.js';
+import { keyToCss } from '../../util/css_map.js';
+
+const createNumberFormat = async () => {
+  const { code } = await languageData;
+  const locale = code.replaceAll('_', '-');
```
AprilSylph, line 9:
```suggestion
  const locale = document.documentElement.lang;
```

marcustyphoon: Ah, excellent!

src/scripts/tweaks/show_full_note_count.js
```diff
@@ -0,0 +1,48 @@
...
+import { buildStyle, postSelector } from '../../util/interface.js';
+import { timelineObject } from '../../util/react_props.js';
+import { keyToCss } from '../../util/css_map.js';
+
+const createNumberFormat = async () => {
+  const locale = document.documentElement.lang;
+  return new Intl.NumberFormat(locale);
+};
+
+const numberFormat = await createNumberFormat().catch(() => new Intl.NumberFormat());
```
AprilSylph, line 12: I don't believe this can fail now, so
```suggestion
const { lang } = document.documentElement;
const numberFormat = new Intl.NumberFormat(lang);
```

src/scripts/tweaks/show_full_note_count.js
```diff
@@ -0,0 +1,44 @@
...
+
+export const main = async function () {
+  document.head.append(styleElement);
+  pageModifications.register(`article footer ${keyToCss('noteCountContainer')} > span`, formatNoteElements);
+};
+
+export const clean = async function () {
+  styleElement.remove();
+  pageModifications.unregister(formatNoteElements);
+  $('[data-full-notes]').removeAttr('data-full-notes');
```
AprilSylph, line 43: Listeners first!

```suggestion
  pageModifications.unregister(formatNoteElements);
  styleElement.remove();
  $('[data-full-notes]').removeAttr('data-full-notes');
```

AprilSylph (changes requested):

AprilSylph (approved):

PR #697, opened 2022-08-13
Replace jQuery cleanup functions with vanilla JS

marcustyphoon:
Here's a totally random idea I had: rather than using jQuery's ergonomic "do XYZ to every element matched by this selector" functions in `clean()`, we could import them from a utility file.

This is mostly because I occasionally mistype ```$(`.${className}`).removeClass(className)```, or have to pause to dash-case a data attribute, and because VS Code automatically imports XKit Rewritten's utility files when it autocompletes.

#### User-facing changes
- n/a

#### Technical explanation
n/a

#### Issues this closes
n/a

### DIFF
--- a/src/util/cleanup.js
+++ b/src/util/cleanup.js
@@ -2,37 +2,48 @@
  * Removes all elements on the current page with the given css class.
  * i.e. jQuery $(`.${className}`).remove();
  *
- * @param {string} className - class name to select elements to remove
+ * @param {...string} classNames - one or more class names to select elements to remove
  * @returns {void}
  */
-export const remove = className =>
-  [...document.querySelectorAll(`.${className}`)].forEach(element => element.remove());
+export const remove = (...classNames) =>
+  [...document.querySelectorAll(classNames.map(className => `.${className}`).join(','))]
+    .forEach(element => element.remove());
 
 /**
  * Removes the given css class from all elements on the current page that have it.
  * i.e. jQuery $(`.${className}`).removeClass(className);
  *
- * @param {string} className - class name to remove
+ * @param {...string} classNames - one or more class names to remove
  * @returns {void}
  */
-export const removeClass = className =>
-  [...document.querySelectorAll(`.${className}`)].forEach(element => element.classList.remove(className));
+export const removeClass = (...classNames) =>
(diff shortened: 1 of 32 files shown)


marcustyphoon:
(currently untested and unimplemented; will do properly if you are interested)

src/util/cleanup.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Removes all elements on the current page with the given css class.
+ * i.e. jQuery $(`.${className}`).remove();
+ *
+ * @param {...string} classNames - one or more class names to select elements to remove
+ * @returns {void}
+ */
+export const remove = (...classNames) =>
+  [...document.querySelectorAll(classNames.map(className => `.${className}`).join(','))]
+    .forEach(element => element.remove());
```
AprilSylph, line 10: :thinking: we could go even more vanilla

```suggestion
export const removeElementsByClassName = (...classNames) => classNames
    .flatMap(className => [...document.getElementsByClassName(className)])
    .forEach(element => element.remove());
```

AprilSylph: hmm, actually, `classNames.flatMap(className => [...document.getElementsByClassName(className)])` could be extracted out into its own function (`getElementsByClassNames`?) and then used in both this and the below function.

marcustyphoon: This is true, and indeed feels more elegant and composable! I used to have it as, iirc

```js
export const remove = (...classNames) =>
  classNames.forEach(className =>
    [...document.querySelectorAll(`.${className}`)].forEach(element =>
      element.remove()
    )
  );
```
but combined it into one QSA call for "performance." <- air quotes intended as absolutely giant

marcustyphoon: Well, it didn't clean things up as much as I thought it would, turns out :D

src/util/cleanup.js
AprilSylph, line 1: I even unknowingly used the same filename in my version... spooky!!

marcustyphoon:
I was writing full jsdoc on utils in 2022? Man, I would have bet against that.

Anyway, we could adopt the multiple-arguments-are-allowed thing in the new PR if desired.

PR #698, opened 2022-08-14
New Feature: Find Inactives

### LINKED ISSUE #989: Clear Innactives
### Concept

NewXKit had a feature where you could find inactive blogs that you followed and it would put them all in a list with their last known activity (i.e. JohnSmith ..... 2 years ago) I found this really useful to clear out old blogs that I not only didn't want to follow anymore, but that were just straight up dead.

marcustyphoon:
~~welcome to: fun with canvas elements~~

| | |
| - | - |
|  |  |

### Description

I'm committing the results of playing around with an interactive UI for a Find Inactives script, which provides a (kind of weird, kind of neat) visual for the distribution of last-updated times for all of your followed blogs along with the primary features of helping the user unfollow specific long-inactive blogs if they're near or at the following limit.

### Testing steps

Confirm that each UI element functions as expected; I don't have specific tests to list at the moment.


marcustyphoon:
to be honest, I dunno if I like this better aesthetically, but here's a histogram view that I assume is less confusing

| | |
| - | - |
|  |  |

src/features/find_inactives.js
```diff
@@ -0,0 +1,391 @@
...
+    showModal({
+      title: 'Select Inactive Blogs',
+      message: [
+        dom('small', null, null, [
+          'Use the slider to display blogs with no posts after the selected date.'
+        ]),
+        canvasElement,
+        slider,
+        selectionInfo,
+        dom('div', { style: 'height: 0.5em' }),
```
AprilSylph, line 298: CSS margin instead of this please, i'd recommend 1em bottom margin here and 1ch bottom margin on the proceeding div

src/features/find_inactives.js
```diff
@@ -0,0 +1,391 @@
...
+
+  const createButton = (text, onClick) =>
+    dom('button', { class: buttonClass }, { click: onClick }, [text]);
+
+  const selectNoneButton = createButton('none', () =>
+    visibleBlogs.forEach(({ checkbox }) => {
+      checkbox.checked = false;
+    })
+  );
+  const selectAllNoMutualsButton = createButton('all (no mutuals)', () =>
```
AprilSylph, line 264:
```suggestion
  const selectAllNoMutualsButton = createButton('non-mutuals', () =>
```

src/features/find_inactives.js
```diff
@@ -0,0 +1,391 @@
...
+    visibleBlogs.forEach(({ checkbox, isFollowingYou }) => {
+      checkbox.checked = !isFollowingYou;
+    })
+  );
+  const selectAllWithMutualsButton = createButton('all (including mutuals)', () =>
```
AprilSylph, line 269:
```suggestion
  const selectAllWithMutualsButton = createButton('all', () =>
```

AprilSylph (changes requested):
I wasn't expecting to like the canvas thingy but it's rather nice.
A couple of brush-up suggestions left below.
(Renaming the constant names to match the proposed copy changes would be welcome.)

src/features/find_inactives.js
marcustyphoon: Honestly, no idea why I even did that. Sort of surprised I didn't go for a vertical flex container. Anyway, could you restate your margin suggestion again? Not sure which elements you're referring to.

marcustyphoon:
> I wasn't expecting to like the canvas thingy but it's rather nice.

Yeah, it's definitely both one of the more interesting and one of the stranger things I've ever coded.

src/features/find_inactives.js
AprilSylph: - `selectionInfo`: 1em bottom margin
- result of `dom()` below: 1ch bottom margin

I originally thought 1em (or just a double `<br>`) for both would be best, but then I realised the "select: all / non-mutuals / none" control should be more visually attached to the selection table, since that's what the buttons affect.

marcustyphoon: Oh—the selection controls div—gotcha.

PR #703, opened 2022-08-16
Seen Posts: Only register posts that appear on screen

### LINKED ISSUE #592: Seen Posts: Only register posts that appear on screen

marcustyphoon:
#### User-facing changes
- Seen Posts does not treat posts rendered off-screen and never scrolled into view as "seen." It also does not treat posts hidden by scripts (e.g. Show Originals) as seen.

#### Technical explanation
As of the initial commit, this implements this exact behavior:

`Posts are marked as seen if any part of the article element is visible for a continuous, uninterrupted 300ms interval.`

This feels pretty good to me. (Initial commit contains testing-only visible behavior indicators; it is very hard to tell what's going on with them removed). It is somewhat overcomplicated, though, of course.

Removing the timers is an option to simplify this code significantly, though I would probably add a processing delay to attempt to run this code after post-hiding scripts complete. That is, `after posts are hidden by other scripts, posts are marked as seen if any part of the article element is visible for any amount of time.`

#### Issues this closes
resolves #592


marcustyphoon:
https://github.com/thebuilder/react-intersection-observer/blob/3041b8b5027025ca2b742ab13856625073d53a5d/src/observe.ts#L68-L69

Note to self: test in Firefox due to this comment

marcustyphoon:
I bothered to actually use this in practice, and... I still get some "hey, that post is greyed out but I definitely have never seen it before." But it's better!

AprilSylph (approved):

PR #706, opened 2022-08-16
Trim Reblogs: Fix silent failure if original root post is deleted

### LINKED ISSUE #563: Trim Reblogs: Silent failure if original root post is deleted
### Platform

MacOS 10.14.6 x86

### Browser

Edge 100.0.1185.44

### Addon version

0.19.3

### Details

Pressing the trim reblogs button on a reblog chain whose original root post has been deleted appears to do nothing; the apiFetch to check `shouldOpenInLegacy` 404s and the rejected promise kills the callback without anything else happening like notifying the user.

(Apologies if I have some details wrong here; reporting the bug before I forget. May test more carefully later.)

marcustyphoon:
I feel like our understanding of when trim reblogs does and doesn't work must be incomplete. Nevertheless, this seems like an improvement.

#### User-facing changes
- Trim Reblogs will warn the user but allow them to continue instead of silently failing if the root post cannot be retrieved to determine if the thread in question is legacy.

#### Technical explanation
Maybe `.catch(() => ({ response: {} }))` is more elegant? I find it a bit hard to proofread.

#### Issues this closes
resolves #563

### DIFF
--- a/src/scripts/trim_reblogs.js
+++ b/src/scripts/trim_reblogs.js
@@ -25,12 +25,17 @@ const onButtonClicked = async function ({ currentTarget }) {
     rebloggedRootId
   } = await timelineObject(postElement);
 
+  let unsureOfLegacyStatus = true;
+
   if (rebloggedRootUuid && rebloggedRootId) {
-    const { response: { shouldOpenInLegacy } } = await apiFetch(`/v2/blog/${rebloggedRootUuid}/posts/${rebloggedRootId}`);
-    if (shouldOpenInLegacy) {
-      notify('Legacy posts cannot be trimmed.');
-      return;
-    }
+    try {
+      const { response: { shouldOpenInLegacy } } = await apiFetch(`/v2/blog/${rebloggedRootUuid}/posts/${rebloggedRootId}`);
+      if (shouldOpenInLegacy) {
+        notify('Legacy posts cannot be trimmed.');
+        return;
+      }
+      unsureOfLegacyStatus = false;
+    } catch (e) {}
   }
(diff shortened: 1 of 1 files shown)


src/scripts/trim_reblogs.js
```diff
@@ -25,12 +25,17 @@ const onButtonClicked = async function ({ currentTarget }) {
     rebloggedRootId
   } = await timelineObject(postElement);
 
+  let unsureOfLegacyStatus = true;
```
AprilSylph, line 28:
```suggestion
  let unsureOfLegacyStatus;
```

src/scripts/trim_reblogs.js
```diff
@@ -25,12 +25,17 @@ const onButtonClicked = async function ({ currentTarget }) {
...
-      return;
-    }
+    try {
+      const { response: { shouldOpenInLegacy } } = await apiFetch(`/v2/blog/${rebloggedRootUuid}/posts/${rebloggedRootId}`);
+      if (shouldOpenInLegacy) {
+        notify('Legacy posts cannot be trimmed.');
+        return;
+      }
+      unsureOfLegacyStatus = false;
+    } catch (e) {}
```
AprilSylph, line 38:
```suggestion
    } catch (exception) {
      unsureOfLegacyStatus = true;
    }
```

AprilSylph (changes requested):
I think the most elegant thing is to make what we're doing, in the event of `apiFetch` rejecting, more explicit

AprilSylph (approved):

PR #711, opened 2022-08-19
New Script: Mass Privater

marcustyphoon:
Mass Deleter and Tag Replacer had a baby.

The intent of the filter-by-tag thing was "only posts containing any of these tags," right? I assume we don't need "only posts containing all of these tags" unless someone really, really wants it.

Should this sidebar element show up everywhere like tag replacer's does?

#### User-facing changes
- Adds a script that converts posts on a blog to private, with filtering by date-before and by tag.

#### Technical explanation
It's pretty much just mass deleter and tag replacer mashed together with a few extra error messages.

#### Issues this closes
n/a


AprilSylph:
> Should this sidebar element show up everywhere like tag replacer's does?

I think I intended to move the Tag Replacer controls to only `/blog/:blogName` pages (excluding queue/drafts/etc subpages) at one point but don't remember why I didn't.

src/scripts/mass_privater.json
```diff
@@ -0,0 +1,10 @@
+{
+  "title": "Mass Privater",
+  "description": "Make some/all of your posts private",
```
AprilSylph, line 3:
```suggestion
  "description": "Unpublish posts by time range or tag",
```

AprilSylph (changes requested):
i am so sorry for taking SIX MONTHS to review this

AprilSylph:
Oh no... this PR has been open for over a year now...

I need to get better at forming habits. I would love to actually keep on top of my review queue, and prevent tragedies like this from happening.

AprilSylph (approved):

Issue #619, opened 2022-05-26
Scroll to Bottom stays stuck "on" between pages


lihatt:
### Platform

Windows 10

### Browser

Firefox 100.0.2 (64 bit)

### Addon version

v0.19.4

### Details

When I hit the Scroll to Bottom arrow, it turns yellow and does its thing. Cool. But then if I change pages, say going from my queue to my drafts, the arrow is still yellow upon page load and immediately scrolls the page down to the bottom without me interacting with it. I can work around this by remembering to click the arrow again before changing pages, but I feel like this script didn't used to behave like a cross-page toggle, and would always begin in the "off" state when a page where it could apply was loaded. (I don't remember for sure when I updated from 0.19.3 to 0.19.4, although it was within the past 10 days most likely. Unfortunately I think I also applied a Firefox update within the same timeframe, so that possibly muddies the waters.)

I noticed also that at first (meaning right after one or the other update) the down arrow was duplicated on the pages where it appears. That was fixed by turning off the toggle for the script and then turning it on again (I haven't seen it since).

marcustyphoon:
Hmm, I think I remember the feature turning itself off successfully on navigation before, so I think you're right. Based on our current code (which turns off the feature if the button that we add is removed from the page), my guess is that a slight change to Tumblr's code results in the button no longer being removed, and so we need a different way to detect this.

lihatt:
Ah, if Tumblr changed something slightly then it might not be related to either extension or browser update. I just remember seeing it start to happen sometime within the past few days on two different computers (but both Win/Firefox of similar versions).

AprilSylph:
It looks like the main problem here is that scrolling doesn't stop when there are no more posts on the page. I just experienced this after loading my whole queue, then going to a different page. When all posts are loaded, scrolling should stop.

https://github.com/AprilSylph/XKit-Rewritten/blob/e764557adf7d1f8212ecec8141bb74129f97d8b9/src/scripts/scroll_to_bottom.js#L23-L25

PR #713, opened 2022-08-20
Scroll to Bottom: Stop at bottom + Fix button color

### LINKED ISSUE #712: Scroll to Bottom: Button does not change color
### Platform

MacOS 10.14.6 x86

### Browser

Edge 104.0.1293.54

### Addon version

v0.20.7

### Details

The Scroll to Bottom button does not turn yellow when active.

Almost certainly the same issue and fix as #644.

marcustyphoon:
#### User-facing changes
- Scroll to bottom's button correctly turns yellow when active.
- Scroll to bottom correctly stops when it gets to the "bottom" of the page.
- ~~Scroll to bottom stops if you open the blog view modal while scrolling (otherwise it would continue in the background; #488 was not implemented)~~
- Minor visual fix on non-glass /blog/view pages, which have a visible scroll to top button but no visible scroll to bottom button (as per reasoning in #488)

Note: the script does not work on https://www.tumblr.com/following before or after this PR due to Tumblr quirks (there isn't a persistent loading indicator on the page and it looks to me like the site doesn't change behavior when you get to the end of the list)

#### Technical explanation
n/a

#### Issues this closes
resolves #712
addresses only the comment in #619 ("scrolling doesn't stop when there are no more posts on the page"); do not close this issue

### DIFF
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -7,7 +7,10 @@ const scrollToBottomButtonId = 'xkit-scroll-to-bottom-button';
 $(`[id="${scrollToBottomButtonId}"]`).remove();
 const activeClass = 'xkit-scroll-to-bottom-active';
 
-const loaderSelector = `${keyToCss('timeline')} > ${keyToCss('loader')}`;
+const loaderSelector = `
+${keyToCss('timeline', 'blogRows')} > ${keyToCss('loader')},
+${keyToCss('notifications')} + ${keyToCss('loader')}
+`;
 
 let scrollToBottomButton;
 let active = false;
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -30,9 +30,7 @@ const scrollToBottom = () => {
   const loaders = [...document.querySelectorAll(loaderSelector)]
     .filter(element => element.matches(blogViewSelector) === false);
 
-  const glassContainerOpen = document.getElementById('glass-container')?.hasChildNodes();
-
-  if (loaders.length === 0 || glassContainerOpen) {
+  if (loaders.length === 0) {
     stopScrolling();
   }
 };
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -14,8 +14,7 @@ let active = false;
 
 const styleElement = buildStyle(`
 ${keyToCss('isPeeprShowing')} #${scrollToBottomButtonId} {
-  opacity: 0;
-  pointer-events: none;
+  display: none;
 }
 
 .${activeClass} svg use {
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -17,7 +17,8 @@ let active = false;
 
 const styleElement = buildStyle(`
 ${keyToCss('isPeeprShowing')} #${scrollToBottomButtonId} {
-  display: none;
+  opacity: 0;
+  pointer-events: none;
 }
 
 .${activeClass} svg use {
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -1,13 +1,13 @@
 import { keyToClasses, keyToCss } from '../util/css_map.js';
 import { translate } from '../util/language_data.js';
 import { pageModifications } from '../util/mutations.js';
-import { buildStyle } from '../util/interface.js';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
- [ ] call stopScrolling on navigation somehow

Man, I wish #453 was less... hacky.

src/scripts/scroll_to_bottom.js
```diff
@@ -30,7 +30,9 @@ const scrollToBottom = () => {
   const loaders = [...document.querySelectorAll(loaderSelector)]
     .filter(element => element.matches(blogViewSelector) === false);
 
-  if (loaders.length === 0) {
+  const glassContainerOpen = document.getElementById('glass-container')?.hasChildNodes();
```
AprilSylph, line 33: Is this reliable again? Last I knew, the glass container doesn't have an ID at all if the first route has glass (e.g, `/new/text`), hence the usage of `[role="dialog"]` glass checks everywhere.

marcustyphoon: a) That is a good question! I didn't see any instances where it didn't on a quick check, but I'm not sure exactly where to look. However:
b) I just noticed this makes scroll to bottom stop if you accidentally mouse over a blog avatar.

Uh. Oops.

src/scripts/scroll_to_bottom.js
```diff
@@ -1,27 +1,36 @@
...
-import { buildStyle } from '../util/interface.js';
+import { blogViewSelector, buildStyle } from '../util/interface.js';
 
 const scrollToBottomButtonId = 'xkit-scroll-to-bottom-button';
 $(`[id="${scrollToBottomButtonId}"]`).remove();
+const activeClass = 'xkit-scroll-to-bottom-active';
 
-const knightRiderLoaderSelector = `main ${keyToCss('loader')} ${keyToCss('knightRiderLoader')}`;
+const loaderSelector = `
+${keyToCss('timeline', 'blogRows')} > ${keyToCss('loader')},
```
AprilSylph, line 11: What are `blogRows`?

AprilSylph: Oh wait, I see, it's follower/following lists. I guess those are rows of blogs.

src/scripts/scroll_to_bottom.js
```diff
@@ -1,27 +1,36 @@
...
 
 let scrollToBottomButton;
-let scrollToBottomIcon;
 let active = false;
 
 const styleElement = buildStyle(`
 ${keyToCss('isPeeprShowing')} #${scrollToBottomButtonId} {
-  opacity: 0;
-  pointer-events: none;
+  display: none;
```
AprilSylph, line 20: The hiding CSS was the way it was for a reason; it allows the scroll to bottom button to fade out/in with the scroll to top button when peepr is opened/closed.
```suggestion
  opacity: 0;
  pointer-events: none;
```

AprilSylph (changes requested):

src/scripts/scroll_to_bottom.js
marcustyphoon: Hmmm. Makes sense; I guess non-modal blog view will need a different solution.

marcustyphoon: Almost works, but it glitches briefly when you scroll up in non modal blog view. Oh well, really not important.

```diff
  ${keyToCss('isPeeprShowing')} #${scrollToBottomButtonId} {
    opacity: 0;
    pointer-events: none;
  }

+ ${keyToCss('isPeeprShowing')} button:not(${keyToCss('hidden')}) + #${scrollToBottomButtonId} {
+   display: none;
+ }
```

AprilSylph (approved):
Shame to merge this without a fix for the non-modal blog view issue, but it's minor enough (doesn't impair use) and the other fixes are more important.

Also, I was going to say this fixes #619 for me, but I only tested it going from my drafts to the dashboard. If going between drafts and the queue still doesn't deactivate scroll to bottom by itself, I don't think it ever did, and I'd be tempted to classify that behaviour as working as intended.

marcustyphoon:
Interesting; I swore that going from drafts to dashboard wasn't stopping it for me, but it certainly is now. It's triggering the "no loader" code; it looks like `checkForButtonRemoved` isn't doing much of anything at the moment.

Anyway, yes, I think "I clicked a UI element while scroll to bottom was actively making my page jump around" having nonobvious behavior doesn't even really count as a bug almost regardless of what that behavior is, because... you know, why, and what were you expecting :D

I do like the currently unimplemented behavior of "clicking the Tumblr logo stops scroll to bottom on the dashboard" though.

PR #718, opened 2022-08-22
Quick Reblog: Display blog avatars in blog selector

### LINKED ISSUE #716: Quick Reblog: Blog icon next to blog selector
### Discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/481

<sup>Originally posted by **marcustyphoon** February 14, 2022</sup>
The CSS might be a bit of a nightmare, and honestly it might not actually look any good, but I had the idea to add the icon of the selected blog to the side of the blog selector in the quick reblog popup. This is nice if, like me, you have multiple blogs with visually similar names (a poor decision on my part, perhaps, but oh well.)

I don't plan to try implementing this at the moment, but I don't think it would be difficult. The hard part, I would imagine, would be making the layout look good; there's not a lot of space :D</div>

marcustyphoon:
#### User-facing changes
- Adds an option to display avatars beside the Quick Reblog blog selector.

#### Technical explanation
I threw CSS at the wall until the select element correctly resized. Can you do this without `display: flex` and `flex: none` (and a bunch of manual padding)? Probably, but apparently I can't figure out how to.

Quick Reblog could use a bit of a rewrite to use `dom()`.

#### Issues this closes
resolves #716

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -44,7 +44,6 @@ let accountKey;
 
 let popupPosition;
 let showBlogSelector;
-let showBlogAvatar;
 let rememberLastBlog;
 let showCommentInput;
 let quickTagsIntegration;
@@ -290,7 +289,6 @@ export const main = async function () {
   ({
     popupPosition,
     showBlogSelector,
-    showBlogAvatar,
     rememberLastBlog,
     showCommentInput,
     quickTagsIntegration,
@@ -329,7 +327,6 @@ export const main = async function () {
   renderBlogAvatar();
 
   blogSelectorContainer.hidden = !showBlogSelector;
-  blogAvatar.hidden = !showBlogAvatar;
   commentInput.hidden = !showCommentInput;
   quickTagsList.hidden = !quickTagsIntegration;
   tagsInput.hidden = !showTagsInput;
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -307,7 +307,6 @@ export const main = async function () {
   blogSelector.replaceChildren(
     ...userBlogs.map(({ name, uuid }) => Object.assign(document.createElement('option'), { value: uuid, textContent: name }))
   );
-  renderBlogAvatar();
 
   if (rememberLastBlog) {
     for (const { uuid } of userBlogs) {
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Oh, dammit. I love the `hidden` property, it's the best.

AprilSylph:
I think this makes more sense as just how Quick Reblog works rather than a preference. Also yes, please rewrite the entire structure with `dom`, it's been on my mental to-do list for a while.

src/scripts/quick_reblog.js
```diff
@@ -56,6 +59,14 @@ const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
 const blogHashes = {};
 
+const renderBlogAvatar = async () => {
+  const { value: selectedUuid } = blogSelector;
+  const { avatar } = userBlogs.find(({ uuid }) => uuid === selectedUuid);
+  const { url } = avatar[avatar.length - 1];
```
AprilSylph, line 65:
```suggestion
  const { url } = avatar.at(-1);
```

AprilSylph:
I have some CSS changes that I'm just going to commit directly, as the changes touch lines I can't code-suggest on.

src/scripts/quick_reblog.js
marcustyphoon: I got super excited about this when I discovered it recently before I realized that it would require a minimum version bump. Do we care about that? https://caniuse.com/mdn-javascript_builtins_array_at

AprilSylph: I misremembered our minimum versions! Probably not worth doing, then.

I guess a different way to shortcut it would be
```suggestion
  const [{ url }] = avatar.slice(-1);
```
but it's too late at night for me to tell if this is silly or not.

marcustyphoon: To be fair, it's Firefox 89 -> 90 and Chromium 89 -> 92, so a very small bump. Requiring 2022 Safari if and when Safari support becomes a thing at all is the more notable change browser-release-date-wise, but you know... shrug.

marcustyphoon:
Quick Firefox screenshot. Looks fine to me!

AprilSylph (approved):

PR #719, opened 2022-08-22
feat(Postblock): Make options UI items links if possible

marcustyphoon:
#### User-facing changes
- Postblock's options UI links to posts you have blocked, so you can tell what they actually are if you want to unblock them. Posts that have not been seen since this upgrade won't have links, since we didn't store the required information previously.
- When viewing a single blocked post in /blog/view, there is a visual indicator with an unhide and unblock button instead of, you know, nothing. 

#### Technical explanation

A map* of post ID -> blog UUID entries is stored as a separate storage entry, rather than migrating the blocked posts list to e.g. an array of `[id, ?uuid]`. You could definitely do the latter, but that kind of thing requires a bit of thinking to ensure you correctly migrate old entries and don't have race conditions.

*I unapologetically call POJO records "maps." Sue me.

#### Issues this closes
discussion #402

### Testing steps

- Load a saved set of preference data with blocked posts. Confirm that they function correctly but are not links in the options UI.
- Block a post. Confirm that it is a link in the options UI, and that the link opens the post in question when clicked.
- Confirm that blocked posts have a new UI when viewed alone, that the new UI looks correct both with and without the "use blog colors when viewing blogs" option, and that its buttons function.
- Navigate to a post that was blocked before switching to this code. Confirm that it becomes a link.


marcustyphoon:
Uhhhhhhh. Is this impossible now? Can you no longer go to a post on a blog given the blog UUID and post ID?

AprilSylph:
Shit, really? What happens if you do? (Mainly want to see if it ends up on a Redpop 404 page; if it does, my redirection code handling this is busted but can probably be fixed.)

marcustyphoon:
does work (post, name):
**Blog network:** https://april.tumblr.com/post/661324787743948800/
**Blog view (new):** https://www.tumblr.com/april/661324787743948800/, https://tumblr.com/april/661324787743948800/
**Blog view (new, @):** https://www.tumblr.com/@april/661324787743948800/, https://tumblr.com/@april/661324787743948800/
**Blog view (old):** https://www.tumblr.com/blog/view/april/661324787743948800/, https://tumblr.com/blog/view/april/661324787743948800/
**Branch.io:** https://at.tumblr.com/april/im-april-im-a-web-developer-who-makes-things/sso2gjczjg24

does work (blog, uuid):
**Blog view (new, @):** https://www.tumblr.com/@t:Fft_7iTGnBlHdl-UtlllwQ, https://tumblr.com/@t:Fft_7iTGnBlHdl-UtlllwQ
**Blog view (old):** https://www.tumblr.com/blog/view/t:Fft_7iTGnBlHdl-UtlllwQ, https://tumblr.com/blog/view/t:Fft_7iTGnBlHdl-UtlllwQ

doesn't work (blog, uuid):
**Blog view (new):** https://www.tumblr.com/t:Fft_7iTGnBlHdl-UtlllwQ

doesn't work (post, uuid):
**Blog view (new, @):** https://www.tumblr.com/@t:Fft_7iTGnBlHdl-UtlllwQ/661324787743948800, https://tumblr.com/@t:Fft_7iTGnBlHdl-UtlllwQ/661324787743948800
**Blog view (old):** https://www.tumblr.com/blog/view/t:Fft_7iTGnBlHdl-UtlllwQ/661324787743948800, https://tumblr.com/blog/view/t:Fft_7iTGnBlHdl-UtlllwQ/661324787743948800

doesn't work: blog/view/@username, anything with /post/

marcustyphoon:
Oh, I guess that's a silly reason to abandon this; it can just look up the blog name via an API request before navigation. Well, not "just," actually, since we can't access apiFetch from a settings panel. But it's solvable.

marcustyphoon:
To fix: #1544 breaks this, as this currently inserts an element before the `article` in a hidden post with the intention that the element is shown when the post is hidden.

marcustyphoon:
- [x] make link color reasonable in firefox dark mode

marcustyphoon:
- [x] check/sync #1755, #1445 #2115, #2117, other potential changes

AprilSylph:
Hold on, couldn't this be massively simplified by just storing the `shortUrl` from each blocked post? It wouldn't always link to the blog view, but the blog network always links to the blog view... it wouldn't be quite as good for UX, but it would be extremely reliable, no?

marcustyphoon:
Hm, probably, assuming that posts always have those, they don't change when a user changes blog names, and they're stored/hosted/whatever in a way we think will persist. I would trust an ex-tumblr-employee to have a better idea of that than I would! Do you think it's a good idea?

AprilSylph:
Reliability-wise they are exactly as good as a blog UUID and post ID, if not better. They continue to work through blog name changes and, although this has never happened, I _believe_ through a blog UUID change as well. They're a very _very_ old part of Tumblr. The only downside is that they will always go to the post on the blog's custom theme, if it has one. (I'm not even sure it goes to the blog view immediately if the blog doesn't? It might be a chain-redirect in that case.)

marcustyphoon:
> They're a very _very_ old part of Tumblr.

You know, I'm sure that's meant to sound reassuring, but... :D

marcustyphoon:
Ah: this will require some network fetching because we block posts by root id, and "the short url of the root post of this reblog" naturally isn't included in a post's data. Not the end of the world, but it does mean we have to think about no-longer-accessible posts and a bit more async stuff. `rebloggedRootUuid` is a pretty convenient field.

marcustyphoon:
#2141 / https://github.com/AprilSylph/XKit-Rewritten/pull/2141/changes/2ff88469ed834e37a57582aef308eff78affae49. Not too bad.

AprilSylph:
> > They're a very _very_ old part of Tumblr.
> 
> You know, I'm sure that's meant to sound reassuring, but... :D

Okay, fine: the system is very old _and_ is still driving enough traffic that it can't just be axed.

It's tried and true, here to stay, proven working, decades-scale reliable.

If it ever does get cut... the broken links in PostBlock will be a drop in the ocean.

marcustyphoon:
> I would love to do this, but those links would rot very easily. Tumblr needs a way for UUIDs to be translated into blog URLs without an authorised API call tbqh. I think I tested at some point that `shortUrl` links break across blog name changes, but maybe I'm wrong/misremembering/outdated. Something to test again for sure.
> 
> _Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/discussions/402#discussioncomment-2016210_

Oh hey. That's fairly funny. I guess we never did test again for sure, huh.

AprilSylph:
Well, I'm completely misremembering in one of these comments. Wouldn't it be really funny if it was the one I made after working at Tumblr, and not the one before? We really do need to test it to be sure.

marcustyphoon:
https://www.tumblr.com/url-change-test-54375/810890566411878400/sdlfjn

Ah, fuck.

marcustyphoon:
~~Wait. Why does the short link in the post I linked work. What.~~

edit: now it doesn't? did I mix up my tabs? what am I smoking

src/features/postblock/options/index.js
```diff
@@ -93,12 +95,20 @@ class PostBlockBlockedPostsElement extends CustomElement {
       unblockButton.dataset.postId = blockedPostID;
       unblockButton.addEventListener('click', this.unblockPost);
 
+      if (uuids[blockedPostID]) {
+        const a = document.createElement('a');
+        a.href = `https://www.tumblr.com/?xkit-postblock-open-post-id=${blockedPostID}`;
```
AprilSylph: I think we shouldn't potentially open the following feed during this process; it unexpectedly moves the "You're caught up" carousel (even if it's not called that anymore).

Could we change this to use the blog view as a base instead? That way we even load the right route chunks! e.g. `https://www.tumblr.com/@${uuids[blockedPostID]}?postblock-id=${blockedPostID}`

src/features/postblock/options/index.js
```diff
@@ -38,6 +38,7 @@ const adoptedStyleSheets = await fetchStyleSheets([
 ].map(import.meta.resolve));
 
 const storageKey = 'postblock.blockedPostRootIDs';
+const uuidsStorageKey = 'postblock.uuids';
```
AprilSylph: nit: prefer less-ambiguous naming

suggestion: `'postblock.blockedPostBlogUUIDs'`

AprilSylph (changes requested):

src/features/postblock/index.js
```diff
@@ -3,27 +3,81 @@ import { getTimelineItemWrapper, filterPostElements } from '../../utils/interfac
...
     } else {
       getTimelineItemWrapper(postElement).removeAttribute(hiddenAttribute);
     }
+
+    const saveUuidPair = (id, uuid) => {
+      if (blockedPostRootIDs.includes(id) && !uuids[id]) {
+        uuids[id] = uuid;
+        browser.storage.local.set({ [uuidsStorageKey]: uuids });
+      }
+    };
```
AprilSylph: This should be extracted out of `processPosts`, it doesn't need to be recreated every time here, right?

src/features/postblock/index.js
```diff
@@ -3,27 +3,81 @@ import { getTimelineItemWrapper, filterPostElements } from '../../utils/interfac
...
+
+    saveUuidPair(rebloggedRootId, rebloggedRootUuid);
+    saveUuidPair(id, uuid);
+    saveUuidPair(rebloggedFromId, rebloggedFromUuid);
```
AprilSylph: nit: this is just... unsatisfying. least-to-most immediate or most-to-least, I don't care, just pick one

```suggestion
    saveUuidPair(rebloggedRootId, rebloggedRootUuid);
    saveUuidPair(rebloggedFromId, rebloggedFromUuid);
    saveUuidPair(id, uuid);
```

```suggestion
    saveUuidPair(id, uuid);
    saveUuidPair(rebloggedFromId, rebloggedFromUuid);
    saveUuidPair(rebloggedRootId, rebloggedRootUuid);
```

src/features/postblock/options/index.js
marcustyphoon: I like it. Potentially. The redirect from `@${uuid}` appears to strip the query param, unfortunately; any idea how to get around that if I'm right?

AprilSylph: Ah, bum.

Well, hashes aren't stripped: https://www.tumblr.com/@t:Fft_7iTGnBlHdl-UtlllwQ#813969316148051968

src/features/postblock/index.js
marcustyphoon: Honestly, I might make this a forEach.

marcustyphoon:
Hm. Not sure if:

- this url hash should be prefixed (`post-id-${blockedPostID}` or whatever)
- this code should skip the extra api fetch since we have the blog name in location.pathname now

AprilSylph:
I would prefix it with `postBlock:` and drop the API fetch, I think.

* Test that the hash starts with `postBlock:`
* Check that the rest of the hash consists of only `0-9`
* Check that the rest of the hash is a valid length for a post ID (don't remember what the upper bound is, sorry)
* ```js
  tumblr.navigate(`./${postId}`);
  ```

marcustyphoon:
re: dropping the api fetch, do we know if it's possible for there to be a trailing slash on the url post redirect? I assume there's no reason that would ever happen?

(aside: I feel like js needs better primitives for url path segments; the moment I write .split('/').filter(Boolean) I feel very silly, and that's even avoiding the magic number stuff most people use)

marcustyphoon:
...oh, I missed the implications of the last line of your comment. never mind!

src/features/postblock/index.js
```diff
@@ -49,26 +102,49 @@ const blockPost = async rootID => {
...
-
   onNewPosts.addListener(processPosts);
+
+  const blockedPostID = location.hash.match(/(?<=^#postblock:)\d{1,20}$/)?.[0];
+
+  if (blockedPostID) {
+    // remove search param now, so it doesn't persist if after we successfully
+    // navigate, the user presses the blog view back button once
+    navigate(location.pathname);
+
```
AprilSylph: I think it's fine to have the hash persist if you go back, as long as you aren't redirected again (which you wouldn't be, since this isn't in any kind of listener).


```suggestion
```

src/features/postblock/index.js
```diff
@@ -4,26 +4,79 @@ import { registerMeatballItem, unregisterMeatballItem } from '../../utils/meatba
...
+      currentTarget.disabled = true;
+    },
+  }, 'show it');
+
+  const unblockButton = dom('button', null, {
+    click: () => {
+      unblockPost(rootID);
+      warningElement.remove();
+    },
+  }, 'unblock it');
```
AprilSylph: I think we shouldn't have an unblock button here... but we _should_ have an unblock button in the post's meatballs menu once revealed, similar to NotificationBlock.

AprilSylph (changes requested):

src/features/postblock/index.js
marcustyphoon: I suppose that would mean just having the show button remove/hide the whole warning, probably?

AprilSylph: Yep!

marcustyphoon:
Ah, this PR straight-up merge conflicts with #2132, doesn't it. So if I merge this PR, I'll then be immediately deleting some of the code in it in that PR. Fun!

marcustyphoon:
Feel free to bikeshed 03fd22f.

AprilSylph (approved):
After four years in development, hopefully it will have been worth the wait.

src/features/postblock/index.css
```diff
@@ -4,22 +4,52 @@
...
+}
+
+.xkit-postblock-hidden-post-controls button:focus-visible {
+  outline: 2px solid var(--content-ui-focus);
+  outline-offset: 2px;
 }
 
-.xkit-postblock-hidden-post-controls button:disabled {
-  color: var(--blog-link-color-40);
+:root[style*="--font-family-modern"] .xkit-postblock-hidden-post-controls {
```
marcustyphoon, line 51: @AprilSylph is this Palettes for Tumblr compat?

AprilSylph: Yep!

marcustyphoon:
It might be a liiiittle hard to see on some palettes? But, like, eh. Do I really want to bikeshed the bikeshedding? No, not really. (And we can always change it later.)

PR #739, opened 2022-09-01
New Script: Delay Posting

marcustyphoon:
Honestly, I dunno if I would actually add this. But I was gonna try writing the code regardless of your opinion about whether it's a good idea because I wanted to see if I could, so.

#### User-facing changes
- Basically, it's undo send. Adds adds a button to the post form that automatically sets the form state to schedule the post a user-configurable time period in the future.

Presumably if this were a real script it would make sense to integrate with quick reblog too.

#### Technical explanation
I was going to say something about using private APIs, but honestly, that term doesn't seem to go far enough to describing calling functions from the React internals.

~~Oh, also, this currently messes up the mode dropdown slightly since the date/time picker doesn't show up unless you click "scheduled."~~

(And yes, you could totally just do this by clicking the post form elements programmatically instead of poking the React internals with a stick.)

#### Issues this closes
n/a

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -4,6 +4,7 @@
   "classic_search",
   "cleanfeed",
   "collapsed_queue",
+  "delay_posting",
   "hide_avatars",
   "limit_checker",
   "mass_deleter",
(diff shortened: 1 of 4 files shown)


marcustyphoon:
Oh, this doesn't actually work in its current form because the post time is set when you press the button. I am very smart.

marcustyphoon:
I do think this idea is kind of cute, but I can't think of a way to do it without intercepting the event handler on the post button, and XKit 7's Editable Reblogs doing that kind of makes me want to go delete the extension's code off my IDE and go live in the woods :P Closing for now.

PR #749, opened 2022-09-08
New tweak: disable L-to-like

### LINKED ISSUE #535: Feature: Disable L button liking shortcut
### Concept

I can't stand accidentally liking posts with the L key, and I don't use it intentionally, so it would be great to be able to disable this shortcut! This seems reasonable to me as a tweak.

Of course, this could also be a toggle on Tumblr's side, so I'll submit it there as well.

marcustyphoon:
#### User-facing changes
- Adds a tweak that disables the L-to-like keyboard shortcut.

#### Technical explanation
The event listener that handles page-wide keyboard shortcuts are attached directly to the document element (`document.addEventListener("keydown", this.onKeyDown)`), so this prevents the L key event from reaching it even when no element is selected (`document.activeElement` is the body element).

I could inject a function that calls `disableHotkeys('KeyL')` and `enableHotkeys('KeyL')` on the big app component in the Redpop internals instead, which is sort of cuter but way more complicated and prone to breakage for no reason.

#### Issues this closes
resolves #535

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -89,6 +89,11 @@
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
       "default": false
+    },
+    "disable_like_shortcut": {
+      "type": "checkbox",
+      "label": "Disable the L-to-like keyboard shortcut",
+      "default": false
     }
   }
 }
--- a/src/scripts/tweaks/disable_like_shortcut.js
+++ b/src/scripts/tweaks/disable_like_shortcut.js
@@ -1,0 +1,8 @@
+const stopLKey = event => {
+  if (event.code === 'KeyL') {
+    event.stopPropagation();
+  }
+};
+
+export const main = async () => document.body.addEventListener('keydown', stopLKey);
+export const clean = async () => document.body.removeEventListener('keydown', stopLKey);


marcustyphoon:
The fun way:

```js
import { inject } from '../../util/inject.js';

const controlHotkeys = (keys, mode) => {
  const funcName = `${mode}Hotkeys`;

  const baseContainerElement = document.querySelector('#base-container');
  const reactKey = Object.keys(baseContainerElement).find(key => key.startsWith('__reactFiber'));
  let fiber = baseContainerElement[reactKey];

  while (fiber !== null) {
    if (fiber.stateNode?.[funcName] !== undefined) {
      fiber.stateNode[funcName](keys);
      return;
    } else {
      fiber = fiber.return;
    }
  }
};

/**
 * Disables or re-enables Tumblr's built-in hotkeys.
 *
 * @param {string[]} keys - Array of key codes to enable/disable (e.g. 'KeyL')
 * @param {'disable'|'enable'} mode - Whether to disable or re-enable the hotkey
 * @returns {Promise<void>} Resolves when finished
 */
const toggleHotkeys = (keys, mode) => inject(controlHotkeys, [keys, mode]);

export const main = async () => toggleHotkeys(['KeyL'], 'disable');
export const clean = async () => toggleHotkeys(['KeyL'], 'enable');
```

AprilSylph:
I don't plan to review this. It would be fun to have completely configurably keyboard shortcuts, but just the shortcut for liking is too niche on its own to justify its place in Tweaks.

PR #761, opened 2022-09-14
Tweaks: Increase unsticky tab bar tweak consistency

### LINKED ISSUE #759: Tweaks: Unsticky tab bar tweak nonfunctional (again)
### Platform

MacOS 10.15.7 x86

### Browser

Firefox 105.0b9

### Addon version

v0.20.10

### Details

The dashboard tab bar is still sticky again; the `.tabsHeader` style rule apparently takes precedence over `keyToCss('tabsHeader')`.

Edit: Oh, I should specify: this is inconsistent. I guess it must be a race condition with which CSS gets inserted first?

Screen Shot 2022-09-14 at 1 05 51 PM

marcustyphoon:
#### User-facing changes
- The unsticky tab bar tweak should not sometimes fail.

#### Technical explanation
I don't actually know how to test this, as usually it works fine as is, and I don't know how to get the XKit Rewritten rule applied before the Redpop one or how that could possibly happen.

#### Issues this closes
resolves #759

### DIFF
--- a/src/scripts/tweaks/unsticky_tab_bar.js
+++ b/src/scripts/tweaks/unsticky_tab_bar.js
@@ -2,7 +2,7 @@ import { keyToCss } from '../../util/css_map.js';
 import { buildStyle } from '../../util/interface.js';
 
 const styleElement = buildStyle(`
-  ${keyToCss('tabsHeader')} { position: static !important; transform: none !important; }
+  ${keyToCss('tabsHeader')} { position: static !important; }
   ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')} { top: 69px !important; }
 `);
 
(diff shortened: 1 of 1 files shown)


src/scripts/tweaks/unsticky_tab_bar.js
```diff
@@ -2,7 +2,7 @@ import { keyToCss } from '../../util/css_map.js';
 import { buildStyle } from '../../util/interface.js';
 
 const styleElement = buildStyle(`
-  ${keyToCss('tabsHeader')} { position: static; transform: none; }
+  ${keyToCss('tabsHeader')} { position: static !important; transform: none !important; }
```
AprilSylph, line 5: We can drop the `transform`, I removed the need for that property to get applied.

AprilSylph (approved):

AprilSylph:
Oh yeah, this is unreleased, huh. I can tell you why we need `!important` here: Tumblr's own CSS for this component can be missing on the initial pageload, for example if you load the activity page first. When it becomes necessary, the stylesheet gets added... after our `<style>` elements, overriding any rules of equal specificity.

The more complete fix for this is to append all our styles to the document root rather than document head, as we currently do with our static styles.

PR #766, opened 2022-09-27
Trim reblogs: User-configurable trimming

### LINKED ISSUE #352: Trim Reblogs: allow leaving >1 trail item
Some `input[type="number"]` styling in `content_scripts/modals.css` will be needed.

The "Trim this post?" confirmation modal should have its message changed to allow the user to customise the number of trail items kept. The input should always have `min="1"` and the `max` should be 1 less than the trail length on the post being actioned on. This input should be validated.

---

this is on the Feature Progress project board already but issues make april go faster. https://github.com/issues/assigned is my favourite to-do list app

marcustyphoon:
Updated concept:

Screen Shot 2022-09-28 at 12 50 40 AM

#### User-facing changes
- Displays each trail item in the "trim this post?"  modal as the user avatar and a short text preview, with a checkbox for each removable item. This allows the user to trim exactly the trail items they want.

I considered a preference option that would leave the first item unchecked by default if there are more than two trail items. I could imagine roleplayers using this if they use initial posts to title their threads.

---

Original concept:

Had an idea, had to try it out. 

Edit: Ooh, there could also be a scissors icon on a dashed horizontal line after the last excluded—okay yeah this is getting out of hand. 

#### User-facing changes
- Adds a form to the "trim this post?" menu with a numerical textbox input that allows the user to select how may trail items are kept and how many are removed. The input is validated.
- Overcomplicates things by adding a preview below the form showing the user avatar of and a short text preview of each trail item, crossing out trail items which will be removed by the current setting. Can remove this.

#### Technical explanation
- [x] confirm that the default for the "content" API response should be an empty array and/or revert and fix this if it should not

#### Issues this closes
resolves #352


AprilSylph:
Oh hey, this is actually halfway to how I was imagining a revamp.

Rather than having a number, I was thinking we could just display a short preview of the trail (now I see why you were musing about using `npf.js`... nah, we should just filter to text and display it with `white-space: nowrap; text-overflow: ellipsis;`) with each trail item as a checkbox, so you can just untick the trail items you want gone.

marcustyphoon:
...oh. Checkboxes are a thing. I am very smart :D

marcustyphoon:
Question, @AprilSylph: Do we want to allow users to trim the entire reblog trail (essentially implementing half of Editable Reblogs, for people to manually edit in responses at will), do we want to force the most recent trail item to be included, or should it be a preference?

AprilSylph:
I'm thinking to allow trimming the whole reblog trail.

AprilSylph (approved):
Appears to work, will merge and make my own style improvements.

PR #767, opened 2022-09-27
Trim Reblogs: allow legacy trimming with warning

marcustyphoon:
#### User-facing changes
- Trim Reblogs puts up an informational message but lets you continue if you try to trim a legacy post.

I just did four legacy post trim tests and all of them worked fine the first time, so it seems silly to totally lock this out.

Honestly, if this works well enough, maybe it doesn't need a second modal and it can just be added to the confirmation modal?

#### Technical explanation
`await new Promise` can serve as a cutesy way to stop a logic flow while enabling a resume button, but there are certainly more normal ways.

#### Issues this closes
n/a


AprilSylph (approved):
Feels fine to me. Hopefully will improve adoption.

PR #768, opened 2022-09-27
create softNavigate utility

marcustyphoon:
¯\\\_(ツ)\_/¯ 

#### User-facing changes
- 

#### Technical explanation
Creates a utility that steals the useNavigate router hook from from Tumblr's internals, enabling XKit Rewritten to simulate arbitrary soft navigation.

### DIFF
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -1,6 +1,6 @@
 import { filterPostElements } from '../util/interface.js';
 import { timelineObject } from '../util/react_props.js';
-import { apiFetch } from '../util/tumblr_helpers.js';
+import { apiFetch, softNavigate } from '../util/tumblr_helpers.js';
 import { primaryBlogName } from '../util/user.js';
 import { keyToCss } from '../util/css_map.js';
 import { onNewPosts } from '../util/mutations.js';
@@ -59,7 +59,9 @@ const addIcons = function (postElements) {
     const isMutual = await mutuals[blogName];
     if (isMutual) {
       postElement.classList.add(mutualsClass);
-      postAttribution.prepend(icon.cloneNode(true));
+      const iconClone = icon.cloneNode(true);
+      iconClone.addEventListener('click', () => softNavigate(`/${blogName}`));
+      postAttribution.prepend(iconClone);
     } else if (showOnlyMutuals) {
       postElement.classList.add(hiddenClass);
     }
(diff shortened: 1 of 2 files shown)


AprilSylph:
POG?

AprilSylph:
Oh I see, right of course there are still some non-functional components that can't use hooks and so need `navigate` as a prop... So this would depend on us finding a part of the React tree that has such a component in it and hoping it doesn't get converted to functional.

marcustyphoon:
Yeah, I haven't looked at whether funding component hooks are possible to grab externally. It looked like, iirc, the buttons to compose a post, the search form, part of the messaging system, and... I forget what the other one was... had it in props.

marcustyphoon:
I was not successful in finding another way to do this if `navigate` is eventually removed from all props. I guess we have to just hope `componentDidCatch` never gets a hook version if we want to use this*. Fun fact, though: `document.querySelector('title')` also works, though I'm not actually sure why the relevant component has a WithRouter on it.

*this is a joke, pls error boundary hook im begging

AprilSylph:
It would probably be more useful to just pitch exposing `navigate` on the global `tumblr` object. I can't imagine that triggering route loads can be any more dangerous than, you know, what we do already with making fully-authenticated API requests.

marcustyphoon:
Yeah, that's certainly the practical upshot. I don't think I would want to rely on this otherwise unless the fallback hard navigation/open-in-new-tab wasn't really a downside.

AprilSylph:
I spent about an hour today trying to extract `navigate` out of its hook and into `window.tumblr` but either I need to change how the object is built (we're using a non-React class, for some reason...) or I need to learn a trick to steal hook outputs and dump them. Memory references to already-rendered components in the React code itself sadly does not allow you to read their props, unless there's some method that wouldn't be exposed by logging the whole thing to the console. (Maybe some method on the prototype I'm not aware of...?)

marcustyphoon:
Yep, I see. The function we want wouldn't exist before hydration anyway, I assume, and if I'm skimming the code correctly, `window.tumblr` gets attached independent of that. Presumably it would have to be moved into a side effect on a component that's inside the `<Router>`?

(I noticed that React Router's `history` object gets created independently, given that `window.tumblr.on()` and `window.tumblr.off()` call `history.listen`; it looks like simply exposing `history.push` wouldn't work, though. And even if it did, that sounds pretty bad.)

PR #770, opened 2022-09-29
implement sidebar notification drop-in animation

AprilSylph:
#### User-facing changes
- XKit notifications in the sidebar are now shown newest-first (from the top) instead of oldest-first
- XKit notifications in the sidebar have a drop-in effect when inserted
- XKit notifications in the sidebar have a fall-away effect before being removed

#### Technical explanation
it is currently 11:50pm

#### Issues this closes
none

@marcustyphoon tell me if this looks cool or not please

### DIFF
--- a/src/content_scripts/notifications.css
+++ b/src/content_scripts/notifications.css
@@ -1,3 +1,23 @@
+@keyframes placed {
+  from {
+    transform: rotate3d(1, 0, 0, 90deg) translateZ(calc(var(--client-height) / 2));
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Ooh, fancy.
- I personally ~halved the duration on the place animation
- do you think it would be weird to override the flex direction to column in the disable layout animations accesskit option? (I can't think of a way to make it no sort of weird otherwise.)

PR #771, opened 2022-09-30
implement fast community label button

marcustyphoon:
playing around with a fast community label button

it got out of hand

proposal: along with something like this, we would essentially clone mass privater

cute prepended to `rightContent` also

#### User-facing changes
-

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -4,6 +4,7 @@
   "classic_search",
   "cleanfeed",
   "collapsed_queue",
+  "community_labeler",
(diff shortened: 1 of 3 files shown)


AprilSylph:
Hmm, I could see this being useful, but I think I'd prefer it to be more like Quick Tags, where the button is its own clear thing and then opens a little popup with options.

Also, naming suggestion: "Quick Flags"

marcustyphoon:
Ooh, nice.

And sure, I can go grab my other branch.

PR #783, opened 2022-10-07
New AccessKit feature: Alt text reminder

marcustyphoon:
#### User-facing changes
- Adds an AccessKit feature that highlights the alt text button in the post form on images that don't have user-set alt text.

I pretty much always post a post, look at the post on my dash, and go, "ah, right, alt text" and add it. Sometimes the post has already been reblogged by then, at which point it's too late.

#### Technical explanation
~~Creates its own MutationObserver, due to the need for `attributeFilter: ['alt']`; when you add alt text to an image in the beta post editor React is smart and only updates the attribute.~~


AprilSylph:
I should pitch this to be added to the editor proper. We're likely not going to add anything else to that meatballs menu, so a dedicated ALT button that calls attention to itself when there's no image description just... makes sense.

marcustyphoon:
*rubs hands together* Excellent, my master plan is proceeding as I have foretold! (Basically.)

marcustyphoon:
Note: This can take advantage of `:has()` and become a single stylesheet.

marcustyphoon:
Fun fact: At time of posting, the default alt text of an image added to the editor when the UI language is set to spanish is lowercase. `window.tumblr.languageData` doesn't have this variation, as the same key ("Image") corresponds to both case variations ("Imagen" / "imagen") and of course a javascript key-value object can only have one of them.

PR #793, opened 2022-10-29
New tweak: Hide after-post popup

marcustyphoon:
### Description

Adds a tweak that hides the `#after-post-actions` popup elements that appear after creating or reblogging a post using the full beta post editor form.

The small popup that appears after using the native Tumblr quick reblog is unaffected.

Technical details:
 - This toast container is portaled outside of the `#root` element, so it was not detected by our postModifications utility until I expanded its scope to include document.body. This didn't seem to have a negative impact.
 - I set an inline style; this could just as easily be a class.

Unsure of:
- UI copy improvements.

### Testing steps

Enable tweaks and this new tweak.

Create and post a new post using the beta post editor. There should be no popup and no confetti after the form is submitted.

Reblog a post using the beta post editor. There should be no purple popup after the form is submitted.

Fast-reblog a post using the native fast reblog. There should still be a small popup indicating success. Repeat this with quick queue.

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -89,6 +89,11 @@
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
       "default": false
+    },
+    "hide_after_post_popup": {
+      "type": "checkbox",
+      "label": "Hide the the popups that appear after posting",
+      "default": false
     }
   }
 }
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -92,7 +92,7 @@
     },
     "hide_after_post_popup": {
       "type": "checkbox",
-      "label": "Hide the the popups that appear after posting",
+      "label": "Hide the new post celebration banner",
       "default": false
     }
   }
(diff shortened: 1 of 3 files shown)


src/scripts/tweaks.json
```diff
@@ -89,6 +89,11 @@
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
       "default": false
+    },
+    "hide_after_post_popup": {
+      "type": "checkbox",
+      "label": "Hide the the popups that appear after posting",
```
AprilSylph, line 95:
```suggestion
      "label": "Hide the new post celebration banner",
```

AprilSylph:
I don't think this is necessary now that the element in question is now far less intrusive.

marcustyphoon:
@AprilSylph There does seem to be some expectation that we implement this (I think I saw 2-3 posts wondering why it's taking so long). Opinions?

AprilSylph:
Hmm, I'd be open to it since it's been this long... but also, it's been this long and nobody's created an issue for it.

If you personally want to do it, go ahead. Just keep in mind that complaining on Tumblr is not a proper channel for feature requests :P

PR #808, opened 2022-11-07
Quick Tags: Automatically change popup position

### LINKED ISSUE #805: Quick Tags: add option to change popup position
### Concept

Request for a Quick Tags option to, upon pressing the QT button, show the menu of tag bundles above the button instead of under it.

When using Quick Tags in the New Post Editor, especially while editing a draft of a length near or larger than the browser window's own, the tag bundles menu usually shows hardly a line and a half before being clipped by the window, forcing the user to scroll down in order to properly navigate the menu. Setting it above the button works more immediately and comfortable (if maybe it is a bit visually messy).

QT1

(This example was done by setting #quick-tags-post-option's transform to translate(50%, -115%) instead of (50%, 12px)

marcustyphoon:
### Description

Here's one way to resolve #805: it makes the quick tags popup automatically choose the "above" location only if placement in the "below" location would overflow the viewport, with no user preference.

I don't think this is necessarily better than the user preference option, but I coded it to see how it felt.

`getBoundingClientRect()` is used rather than an `IntersectionObserver`, as this logic only needs to be run synchronously when the user presses the quick tags button. 

### Testing steps

- Enable quick tags and create enough tag bundles for the quick tags popup to have meaningful size.
- Compose a post in the beta post editor.
- Click the quick tags button. The popup should appear in the normal, below-the-button position. Dismiss it.
- Increase the height of the post form by adding content.
- Click the quick tags button. The popup should appear above the quick tags button. It should not overflow off of the screen.

### DIFF
--- a/src/scripts/quick_tags.css
+++ b/src/scripts/quick_tags.css
@@ -27,6 +27,10 @@
   transform: translate(50%, 12px);
 }
 
+#quick-tags-post-option.above {
+  transform: translate(50%, calc(-100% - 12px - 22px));
+}
+
 @media (max-width: 650px) {
   #quick-tags {
     top: 50%;
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -121,11 +121,22 @@ const togglePopupDisplay = async function ({ target, currentTarget }) {
   currentTarget[appendOrRemove](popupElement);
 };
 
+const appendWithoutViewportOverflow = (element, target) => {
+  element.classList.remove('above');
+  target.appendChild(element);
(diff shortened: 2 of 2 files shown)


AprilSylph:
I really like this idea. Could it be expanded to the normal popup as well (outside of the editor)?

AprilSylph (approved):

marcustyphoon:
Drawback of this: you can't see what tags you added when the popup covers the tag area.

Do I want to make this automatic behavior a preference? Not... really, but I am sort of tempted.

nobodysuspectsthebutterfly:
Another drawback unfortunately - the popup above covers the queue/draft/post/post privately switch.
image

PR #817, opened 2022-11-14
Popup: Toggle script preference open toggling script

### LINKED ISSUE #815: Open script preference on script enable
### Concept

I have seen reports that new users are confused when they enable a script and it does not have the expected result because they are not aware that scripts have options. One option to prevent this is to open the options pane when an extension is toggled on (though it is kind of annoying). 

Other options would be to add text describing where you can click (awkward) or changing the UI to make it more obvious (I have no ideas for this currently).

marcustyphoon:
### Description

This toggles the open/closed state of a script's preferences when toggling it on/off. This ensures that new users will be aware of the existence of script preferences as soon as they toggle on their first script.

There are scenarios where this is slightly inconvenient (toggling more than one script at once necessitates some scrolling unless you go bottom to top), but that seems fairly niche.

This PR currently closes the preference pane when you toggle a script off as well as opening the preference pane when you toggle a script on; this felt more natural to me than only performing this action on enable.

There is some risk of users not realizing you can open the preference pane without clicking the toggle now, but that seems even more niche than the problem this solves and does not actually impede functionality.

Resolves #815.

### Testing steps

- Open the extension preference pane and toggle a script on. Its preference pane should open.
- Open the extension preference pane, open a script's preference pane, and toggle it on. It should stay open.
- Open the extension preference pane, open a script's preference pane, and toggle it off. Its preference pane should close.
- Open the extension preference pane and toggle a script off. It should stay closed.

### DIFF
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -17,6 +17,8 @@ const writeEnabled = async function ({ currentTarget }) {
   const detailsElement = currentTarget.closest('details');
   let { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
+  detailsElement.open = checked;
+
   if (checked) {
     enabledScripts.push(id);
     detailsElement.classList.remove('disabled');


AprilSylph (approved):
Feels very natural, I like it. A nice point for iteration would be to exclude scripts with no preferences from this behaviour.

PR #824, opened 2022-11-15
Use `dom()` utility in Quick Reblog/Quick Tags

marcustyphoon:
### Description

This refactors Quick Reblog and Quick Tags using the dom() utility. It should have no functional effect.

### Testing steps

No specific tests; use Quick Reblog and Quick Tags and expect no functional changes.


AprilSylph (approved):
seems to work :+1:

todo for me: write a better `dom()` that only takes two arguments
(i tried to do this already, but changing how `dom()` works requires updating usages of `dom()` and testing that sucked...)
(if you're curious, the first argument would still be the tag name, and the second argument would still map key-values to attribute names & values, but it would also map key-values to event listeners if the value is a function, and rather than a separate argument for children... i would just special-case the key name `children`. if it's good enough for react function components, it's good enough for me.)
(i haven't thought about the naming much though, i literally only just now had the idea of making it a separate function so that usages could be updated incrementally, any ideas? `h` seems obvious but perhaps misleading; maybe `el` would be better?)

marcustyphoon:
I tried part that in my userscript, I'm pretty sure. Seemed to work. I think the classic JSX transform kept children separate, though, and I was trying to emulate it.

I picked html kind of arbitrarily when I wrote mine, but, eh. No idea really.

```js
const html = (component, props = {}, children = []) => {
  const element = props?.xmlns
    ? document.createElementNS(props.xmlns, component)
    : document.createElement(component);

  props &&
    Object.entries(props).forEach(([nameOrType, value]) => {
      if (typeof value === 'function') {
        element.addEventListener(nameOrType.substring(2), value);
      } else {
        element.setAttribute(nameOrType, value);
      }
    });
  children && element.replaceChildren(...children);

  element.normalize();
  return element;
};
```

AprilSylph:
huh, new JSX transform does it unified the way i was thinking. i've literally never seen that code but it's exactly what i was thinking of.

nice.

marcustyphoon:
> huh, new JSX transform does it unified the way i was thinking.

yeah, I think if you wrote it out it the nesting would be a catastrophe though I think. ~~unless we actually write a jsx transform, write jsx, and run the whole thing through esbuild. and then we could switch to tsx.~~ /joke

> i've literally never seen that code but it's exactly what i was thinking of.

yeah I kind of have a [whole](https://transienturl.tumblr.com/post/697140713219080192/was-polishing-this-like-its-a-full-fledged) [thing](https://transienturl.tumblr.com/post/696996106604871680/tumblr-haxjs). it's sort of a tire fire and I didn't keep the gist up to date when I started using it for weird personal stuff like a shortcut to the xkit search page so.

PR #825, opened 2022-11-16
Sidebar util: Condense multi-row sidebars

marcustyphoon:
### Description

Tag Tracking+'s sidebar takes up a lot of vertical room on my display because each tracked tag is so tall, but I think the button height feels right for the scripts that have one button to open a UI popup.

This automatically vertically condenses sidebar lists with more than 2 items. (No idea if that's the ideal number. Discuss! If n = 1, there is also an easier way to do this; see reverted commit.)

### Testing steps

- Enable Tag Tracking+ (with the sidebar option) and e.g. Limit Checker. Compare their sidebar items' styling.
- Track more/fewer tags to make your number of tracked tags more/less than 2. Observe the new styling.

### DIFF
--- a/src/content_scripts/sidebar.css
+++ b/src/content_scripts/sidebar.css
@@ -43,6 +43,10 @@
 }
 
 .xkit-sidebar-item li {
+  padding: 0 10px;
+}
+
+.xkit-sidebar-item li:only-child {
   padding: 5px 10px;
 }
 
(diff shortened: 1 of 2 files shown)


AprilSylph:
I think I would prefer the existing uniformity, but I'll keep this open in case I change my mind after living with it for a while (I track nine tags on my main account).

marcustyphoon:
Note: This is effectively broken by the current implementation of #834, which would have inconsistent behavior with either version of this code when the option is enabled to dynamically hide sidebar list items.

PR #827, opened 2022-11-16
New Features: Quick Flags; Mass Labeler

### LINKED ISSUE #1435: Method to community label a reblog
### Concept

As discussed in a thread with our good friend echo, the tumblr API allows one to community label a reblog even if it doesn't have any content (using it not as an indication that the reblog's added content is mature, but instead as a way for followers and downstream rebloggers to have the content labelled without upstream versions also being labelled), but the post form does not. The community label system generally has a fair bit of nuance, which I won't get into more for brevity and relevance reasons.

Anyway, it could be neat if XKit Rewritten allowed one to do this. This could either be part of the Quick Reblog form if one enables a preference (although there's not a ton of room?), which I could build if we think that's the right path, or it could just be done as an action after the user performs the reblog, which I already built (#827).

marcustyphoon:
### Description

This creates a script to quickly toggle the mature content ~~community~~ labels on your ~~original posts~~ posts and reblogs.

It implements a Quick Tags-style dropdown that lets the user label an individual post and a bulk-edit sidebar option and modal form to easily retroactively flag every post with certain tags or the user's entire blog.

Continuation of #771.

- [x] import #839 with slight modifications
- ~~skip attempts to edit legacy posts (https://discord.com/channels/104051306309644288/327155782934200320/1074380420361224272)~~ no longer needed
- [x] use the mass post editor endpoints
- [x] update UI logic for virtual scroller (probably simplify)
- [x] check error handling code
- [x]  #1450

This PR can be split into two (quick reblogs / mass labeler) or three (two stacked PRs on a util implementation PR) on request; it's quite easy for me to do so feel free to do so. While it's not under active review I find it less maintenance to have it as a single PR to reduce work keeping it up to date.

### Testing steps

Quick Flags: Enable the feature and confirm that...

- Quick Flags buttons are added to editable post/post draft footers with all relevant footer-related account flags
- Quick Flags buttons are **not** added to ask/submission footers with all relevant footer-related account flags
- Quick Flags buttons open UI when clicked that accurately reflects post community label status
- Quick Flags can be used to:
  - add the mature label to an unlabelled post
  - add a category to an unlabelled post, which will also add the mature label
  - remove a category from a labelled post
  - remove the mature label from a labelled post, which will also remove all categories
- After using Quick Flags, Quick Flags buttons open UI when clicked that accurately reflect new post community label status
- After using Quick Flags, an indicator is shown if the new post community label status differs from that shown by the Tumblr UI
- After using Quick Flags, an indicator is **not** shown if the new post community label status **does not** differ from that shown by the Tumblr UI, e.g. if using Quick Flags to set back the original post state
- Indicators persist if posts are scrolled far off the screen and back with endless scrolling enabled
- An error message appears and Quick Flags checkboxes do not become out of sync if the `bulkCommunityLabel` API fetch fails (disconnected network, etc)

Mass Labeler: Enable the feature and confirm that...

- A Mass Labeler sidebar item appears on the "channel" pages' "posts" views
- The Mass Labeler form appears and has the current blog selected when opened
- Mass Labeler adds content labels to posts
- Mass Labeler adds content labels to only posts with the selected tags/in the selected time range
- Mass Labeler adds only the chosen content labels. For example, if the "violence" category is added, at the same time, to a post with the "drug/alcohol" label and a different post with the "sexual themes" label, each post should only have its label and the added label.
- An error message appears if the `bulkCommunityLabel` API fetch fails (disconnected network, etc)


src/scripts/quick_flags.js
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  const postId = postElement.dataset.id;
+  const {
+    blog: { uuid }
+  } = await timelineObject(postElement);
+
+  const {
+    response: {
+      content = {},
```
AprilSylph, line 119:
```suggestion
      content = [],
```

marcustyphoon:
I was going to update this to use the mass post editor endpoint (https://www.tumblr.com/changes/715595761003429888/tuesday-april-25th-2023) to perform the edits in batches of 100, but the page actually uses a legacy SVC endpoint at the moment, which we can't realistically make use of.

src/util/mega_editor.js
```diff
@@ -56,3 +56,46 @@ export const megaEdit = async function (postIds, options) {
...
+  formKey ??= await fetch('https://www.tumblr.com/neue_web/iframe/new/text').then(response => {
+    if (response.ok) {
+      return response.text();
+    } else {
+      throw Object.assign(new Error(response.status), { response });
+    }
+  }).then(responseText => {
+    const responseDocument = (new DOMParser()).parseFromString(responseText, 'text/html');
+    return responseDocument.getElementById('tumblr_form_key').getAttribute('content');
+  }).catch(console.error);
```
AprilSylph, line 81: let's abstract this to its own function rather than repeating it

marcustyphoon:
The mass post editor continues to overwrite rather than add, unlike what its UI suggests. Using its endpoint correctly would therefore require sorting posts by their current labels, which I actually am happy to code.

This is, on the one hand, a lot of coding effort (sure) and testing/maintenance effort (problematic) for a feature that I would be surprised if many people used.

On the other hand, I think this existing is good for the completeness of the ecosystem; if we really want to encourage people to community label their content comprehensively, better tools to do that seem like good messaging.

AprilSylph (changes requested):
I think this needs to be split into two features. "Quick Flags" should be the feature with the control buttons on individual posts, while the mass labeller should be, er, probably yeah that. "Mass Labeller".

marcustyphoon:
- [x] [rename all "community label" ui to "content label"](https://www.tumblr.com/changes/753647297666203648/tuesday-june-18th-2024)
- [x] #1013 I guess? sure
- [x] #1536
- [x] #1803 (if merged)
- [x] #1807

marcustyphoon:
- [ ] address #1610

marcustyphoon:
- [x] fix popup state not being set on open (??? how did I not notice this)
- [x] new footer updates

marcustyphoon:
Oh, hey.

I think I probably finally found the right handler so that we could inject a function to request that a certain post in the timeline be updated from the server (look for `stateNode.refreshPostInTimeline()`, a function with no arguments). Unfortunately, it's on the edit button, which... Tumblr just removed from the post body. Presumably we could click the meatballs button, then use the edit button in the meatballs menu to request a refresh, but that's pretty absurd.

Possibly more plausible, but considerably more involved, is obtaining the `onEditTimelineObject` function that it calls by grabbing it from... uh, some withContext HOC? and figuring out how to tell it to make the correct change.

Both of these are, obviously, way more than we generally do with injected functions, which means they're pretty fragile to Tumblr code changes and might be difficult or impossible to fix if they did break, which seem like reasonable reasons to continue favoring less elegant functionality that can be done without injecting, like the warning element thingy currently implemented here.

Somewhat related: #1901's "let's put back the edit and delete buttons" thing; I think in that case we'll need to just open the meatballs menu and click it, though (delete could maybe be done by simulating the `d` hotkey, but there's no edit hotkey, so).

marcustyphoon:
Hey, want to see some shit that absolutely should not work but does? https://github.com/AprilSylph/XKit-Rewritten/compare/master...AprilSylph:XKit-Rewritten:marcustyphoon/quick-tags-really-refresh

marcustyphoon:
Hm, interesting: updatePostOnPage is not fully functioning here; there are probably some state hooks in the post element somewhere. Editing a post with Tumblr's post editor breaks less than updatePostOnPage does*, though, so in theory I should be able to figure out what it's doing that we're not.

*it still breaks a bit, which I reported

PR #839, opened 2022-11-19
New Script: Mass Tip Toggler

marcustyphoon:
> pssssst hey staff members
> 
> I would love to know how the tipping enabled/disabled flag is internally stored on posts, since I can't figure it out
> 
> there is a mode for default-off where you can override it to on and a mode for default-on that can be overrided to off, but how do these interact? is there separate storage for each mode? is there a single key with "default" / "force enable" / "force disable" options?
> 
> – https://www.tumblr.com/transienturl/701323560556331008/

Looking for clarification on how the tipping flag actually works (https://github.com/tumblr/docs/issues/93) to understand if this is even a good idea, but this prototype is functional.

Fully respecting the "default" option that seems to exist internally is kind of hard in mass-edit mode; I don't know offhand how to get a blog's tipping toggle default or see if a post matching it is actually set to default or force-mode.

~~As I'm sure you would expect, it's just the mass privater PR with some quick tags and mass deleter code (and a few transformations that I needed to code anyway to potentially have a working version for community label manipulation).~~

 - [ ] skip attempts to edit legacy posts (https://discord.com/channels/104051306309644288/327155782934200320/1074380420361224272)
---


marcustyphoon:
- [ ] skip trying to edit non-original posts (what am I doing)

marcustyphoon:
- [ ] deal with the virtual scroller breaking my status messages

marcustyphoon:
https://www.tumblr.com/changes/749295377342251008/tipping-is-going-away-on-june-1-2024

PR #842, opened 2022-11-19
proof of concept: editable reblogs

marcustyphoon:
Before you ask: no. Barring a miracle, we're not doing this. 

### Description

Screen Shot 2022-11-19 at 1 47 56 PM

Demonstrates some of the ability to do what Editable Reblogs did in legacy XKit: on the press of a button, this opens a reblog form with the real Tumblr trail missing and the previous post contents copied into the editable part of the form.

Well, not really. Because there's no real good way to format this kind of thing right now, as NPF doesn't allow most formatting to be inside an indented block, so it's pretty hard to make it clear where one fake trail item ends and the next begins.

Right now this also removes `layout` data, which blows up photosets that are intended to have images side by side, and has no way of selecting which blog you want to reblog to, both of which would be fixable but are outside the scope of the proof of concept I wanted to create.

### Testing steps

Don't.

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -4,6 +4,7 @@
   "classic_search",
   "cleanfeed",
   "collapsed_queue",
+  "editable_reblogs",
   "hide_avatars",
   "limit_checker",
   "mass_deleter",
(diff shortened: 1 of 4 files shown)


marcustyphoon:
Immediate close, as intended. Some of this code might be useful for copy-pasting later though!

PR #844, opened 2022-11-20
New Script: Header Blogs

marcustyphoon:
Feel free to close. But hey, just in case.

### Description

Implements the blog-icons-in-the-header functionality from legacy XKit's Header Options.

- [ ] Highlight currently selected blog
- [ ] Maybe add a hover text thingy, but also maybe not, who cares
- [x] Add count preference probably
- [x] soft navigation? :3


src/scripts/header_blogs.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "relatedTerms": [ "Header Options" ],
+  "preferences": {
+    "shapeMode": {
+      "label": "Avatar Shape",
+      "type": "select",
+      "options": [
+        { "value": "circle", "label": "Circle" },
+        { "value": "square", "label": "Square" }
+      ],
```
AprilSylph, line 17: why not respect the per-blog shape setting? (`theme.avatarShape` on the blog object)

marcustyphoon: Oh yeah, could do. Aside: I've always found it a bit weird that the circle option doesn't apply to the dashboard. Does make sense, though, I think.

AprilSylph: The square option doesn't apply to messaging either. Two things I want to fix if Product Design will allow me to.

src/scripts/header_blogs.json
```diff
@@ -6,5 +6,12 @@
...
     "background_color": "#000000"
   },
-  "relatedTerms": [ "Header Options" ]
+  "relatedTerms": [ "Header Options" ],
+  "preferences": {
+    "maxBlogs": {
+      "type": "text",
+      "label": "Maximum blogs",
+      "default": "100"
+    }
```
AprilSylph, line 15: pretty sure you can use `flex-wrap: wrap; overflow: hidden;` to simply fill the available space cleanly

marcustyphoon: I am! But I think many people like putting like 3-5 blogs up there even if they have many more. (Legacy XKit defaulted to 3.) I personally don't want all 20+ blogs my main has access to up there if I fullscreen on a big monitor.

PR #866, opened 2022-11-24
new experimental script: title counts

marcustyphoon:
**edit:** currently broken; soft navigation replaces the title element

Experimenting with editing the tab title text with the user's notification count instead of their unseen posts count.

As of this initial commit, unburyPollerContext inexplicably gives me outdated pollerContext values, so the count lags behind by like a minute. Can't figure out why this is.

More general discussion: should XKit Rewritten implement what Tab Titles did (not that I know exactly what that is offhand), have essentially this functionality to change the count (if I can get it to work and we either view pollerContext extraction as stable, get it from the DOM and live with it not working in mobile view, or have it added to `window.tumblr`), or just have an option to remove the count (which I would just make a tweak)?

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -26,6 +26,7 @@
   "tag_replacer",
   "tag_tracking_plus",
   "themed_posts",
+  "title_counts",
   "timeformat",
   "timestamps",
   "trim_reblogs",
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Closing to reduce PR clutter.

PR #868, opened 2022-11-25
Add localization features

char-lock:
### Description

Implements use of localization files for strings in XKit Rewritten and sets the default locale to generic English. This allows for future translation and localization efforts to allow for multilingual usage of the extension.

* Created utility functions to pull localization information regardless of the browser.
* Added localization support for every string in the extension popup.
* Added localization support for individual script descriptions.

### Testing steps

In general, the behaviour should be largely unchanged from the original extension, save the new ability to add localization profiles. If the user locale is English, everything will look and act the same as it did previously.

### DIFF
--- a/src/util/localization.js
+++ b/src/util/localization.js
@@ -9,10 +9,7 @@
  */
 export function getLocalizedMessage (messageId, defaultMessage) {
   const localMessage = browser.i18n.getMessage(messageId);
-  if (typeof localMessage !== 'undefined' && localMessage !== '') {
-    return localMessage;
-  }
-  return defaultMessage;
+  return localMessage || defaultMessage;
 }
 
 /**
--- a/src/util/localization.js
+++ b/src/util/localization.js
@@ -1,0 +1,43 @@
+/**
+ * Returns the message in its localized format. Essentially an alias
+ * for the *.i18n.getBrowser method, but it performs the check for
+ * Chromium vs Firefox internally.
+ *
+ * @param {string} messageId - Identifier for the string
+ * @param {string} defaultMessage - Message to use if the browser can't be identified.
+ * @returns {string} Either the localization for the messageId if the browser can be identified, otherwise the defaultMessage.
+ */
+export function getLocalizedMessage (messageId, defaultMessage) {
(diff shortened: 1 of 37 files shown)


src/util/localization.js
```diff
@@ -0,0 +1,43 @@
...
+      return browser.i18n.getMessage(messageId);
+    }
+    // Must be Chrome. Use chrome.
+    return chrome.i18n.getMessage(messageId);
+  } else if (typeof browser !== 'undefined') {
+    // Must be Edge. Use browser.
+    return browser.i18n.getMessage(messageId);
+  }
+  // Unknown browser. Return the default message.
+  return defaultMessage;
```
AprilSylph, line 24: We use `browser-polyfill` so we can just use `browser` instead of performing all these checks (assuming we load it in every place we need this util, which we should already do but there might be gaps)

char-lock: Gotcha, gotcha -- this was mostly just a safety check more than anything else. I'll remove the check and just go straight for browser, then.

AprilSylph: (if you run into any problems with `browser` not existing on Chrome inside scripts loaded as modules, you should be able to just import the file for its side-effects, too)

char-lock: Should be good to go, now. Swapped out the check to just make sure we return a default message if there's no localization available.

src/util/localization.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param {string} messageId - Identifier for the string
+ * @param {string} defaultMessage - Message to use if the browser can't be identified.
+ * @returns {string} Either the localization for the messageId if the browser can be identified, otherwise the defaultMessage.
+ */
+export function getLocalizedMessage (messageId, defaultMessage) {
+  const localMessage = browser.i18n.getMessage(messageId);
+  if (typeof localMessage !== 'undefined' && localMessage !== '') {
+    return localMessage;
+  }
+  return defaultMessage;
```
AprilSylph, line 15:
```suggestion
  return localMessage || defaultMessage;
```

Issue #869, opened 2022-11-25
Hide Seen: Remove seen posts entirely from display


char-lock:
### Concept
While the Hide Seen script dims seen posts as one would want, I have found it much more beneficial to my Tumblr experience to have the posts hide upon a reload of the page.

### Proposed Solution
I implemented an additional preference to a custom version of XKit Rewritten that proved very nice to my overall experience -- pretty simply, if the plugin encounters a post it registers as seen, and the preference is enabled, it deletes the post element entirely, which also assists in protecting memory resources from piling up. If the user wants to see the posts again, you just disable the preference and reload.

I used my custom version of this feature for a few weeks now, and it has proved pretty much bug free, but I wanted to get general consensus on whether this should be implemented to the wider plugin features.

AprilSylph:
> it deletes the post element entirely

never do this. React will crash if it tries to remove an element that doesn't exist on the page.

I would be open to having this be implemented so long as it doesn't break endless scrolling, or only works with endless scrolling disabled.

marcustyphoon:
From a technical standpoint I would probably lean towards what we do in scripts like Show Originals, setting the css `display: none` property on the `article` element, since actual removal breaks j/k scrolling and doesn't - oh hi April. Anyway, yeah, the memory benefits aren't huge I think and we try to make everything reversible without a refresh.

char-lock:
> > it deletes the post element entirely
> 
> never do this. React will crash if it tries to remove an element that doesn't exist on the page, or has otherwise been moved from its original parent.
> 
> I would be open to having this be implemented so long as it doesn't break endless scrolling, or only works with endless scrolling disabled.

I haven't experience any issues in runtime, and endless scrolling works flawlessly.

char-lock:
> From a technical standpoint I would probably lean towards what we do in scripts like Show Originals, setting the css `display: none` property on the `article` element, since actual removal breaks j/k scrolling and doesn't - oh hi April. Anyway, yeah, the memory benefits aren't huge I think and we try to make everything reversible without a refresh.
> 
> Also check out Show Originals' `lengthenTimeline` function, which ensures the user can scroll to load more posts if all of the current ones are hidden!

I tried this in an original implementation, and it eventually caused an extreme amount of lag from just the sheer amount of posts being hidden. Albeit that took a while, it was still noticeable.

char-lock:
So that it's less nebulous, the proposed implementation can be found here:
[...]

PR #870, opened 2022-11-25
Seen Posts: Hide seen posts completely

char-lock:
### Description

As per #869, adds a preference to the Seen Posts script to allow users to hide seen posts on the dashboard using a CSS style.

### Testing steps

User should be able to browse as usual, but when their dashboard updates with new posts, any posts they have viewed will be hidden, leaving only new posts and old ones that have not been viewed.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "XKit Rewritten",
-  "version": "0.21.0",
+  "version": "0.21.1",
 
   "short_name": "XKit",
   "author": "April Sylph",
--- a/src/scripts/seen_posts.css
+++ b/src/scripts/seen_posts.css
@@ -2,3 +2,7 @@ body:not(.xkit-seen-posts-only-dim-avatar) .xkit-seen-posts-seen:not(:hover),
 body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article > :first-child img {
   opacity: 0.5;
 }
+
+body.xkit-seen-posts-hide .xkit-seen-posts-seen {
+  display: none;
+}
--- a/src/scripts/seen_posts.css
+++ b/src/scripts/seen_posts.css
@@ -4,5 +4,5 @@ body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article >
 }
 
 body.xkit-seen-posts-hide .xkit-seen-posts-seen article {
-  display: none;
+  display: none !important;
 }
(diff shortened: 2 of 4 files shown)


src/scripts/seen_posts.css
```diff
@@ -2,3 +2,7 @@ body:not(.xkit-seen-posts-only-dim-avatar) .xkit-seen-posts-seen:not(:hover),
 body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article > :first-child img {
   opacity: 0.5;
 }
+
+body.xkit-seen-posts-hide .xkit-seen-posts-seen {
+  display: none;
+}
```
marcustyphoon, line 8:
```suggestion
body.xkit-seen-posts-hide .xkit-seen-posts-seen article {
  display: none;
}
```

Hiding only the article element ensures that k-to-scroll-up still works; without this, if the post above the one you're pressing k while looking at is hidden with this extension, Tumblr's code will be unable to determine its location and will scroll to the top of the page.

src/scripts/seen_posts.css
```diff
@@ -3,6 +3,6 @@ body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article >
   opacity: 0.5;
 }
 
-body.xkit-seen-posts-hide .xkit-seen-posts-seen article {
+body.xkit-seen-posts-hide .xkit-seen-posts-seen {
```
marcustyphoon, line 6: Not sure why you reverted this!

src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "XKit Rewritten",
-  "version": "0.21.0",
+  "version": "0.21.1",
```
marcustyphoon, line 4: We don't do version bumps of the extension along with script updates!

char-lock: That was an accident! Sorry about that.

char-lock: I don't even remember why I did that/pushed it. Ough.

char-lock:
It's been long enough on this PR that I don't much remember the work I did on it. Should I just go ahead and close it and reopen whenever I go through and fix the issues, just to save the headache?

AprilSylph:
@char-lock up to you! Sorry that we (I) let this sit so long.

marcustyphoon:
Yep, either way. I could take it over as well; it's similar enough to Show Originals.

char-lock:
> Yep, either way. I could take it over as well; it's similar enough to Show Originals.

If you want to take it over, please feel free!

I appreciate y'all's work on the plugin and administrative handling, by the way!

PR #877, opened 2022-11-28
New feature/script: Mutual Activity

alleycatboy:
### Description

This feature brings something I've seen many people request on the site: filtering your activity feed to only show your mutuals. 

Blogs that get many interactions from non-mutuals have to scroll and scroll to find interactions from their mutuals; they find that they don't even notice when their mutuals @ them or replied to their posts.

### User-facing changes 
This adds a toggle button to the activity page that only shows from their feed, and optionally shows grouped notifications.

### Technical explanation
Filtering of non-mutuals is done through toggling the style element inherent to "following" and "rollup" notifications.

Filtering of grouped/rollup notifications is done by checking the state of the Filtering dialog modal and toggling the checkbox only if it is out of sync with the user preferences. 

The script uses a registered pageModifications on the Activity page body to check if it should run or not. 

On the dialog modal has a `display: none` and the Filters button is disabled while checking the state to avoid user interference , though the time window for user interference is extremely small. Additionally, it checks if the user already has the dialog open or not.

While the "Mutuals only" filter is enabled, the "Group similar notifications" option is disabled and is telegraphed to the user that the preference is controlled by the script. 
 
When the page is reloaded or navigated to, the previous state of the toggle button is recalled from `browser.storage`.

When filtering is enabled, a phantom scroll event is triggered on the feed container to trigger loading of more notifications. 

### Testing steps

Simply enable the extension and click on the "Mutuals only" toggle button to activate and deactivate.

Here is the default action:

firefox_mT2E4UE9Sp

When grouped notifications are **enabled**:
firefox_FIu7Sr8O2r

The filter dialog when group notifications are **disabled**:
firefox_9qR4HGlE58

The filter dialog when the "Mutuals only" toggle is **disabled**:
firefox_h4VTyBSGmq

Behavior has also been tested in various color pallets, different screen sizes, and Firefox mobile.


marcustyphoon:
Ooh, fancy.

- `keyToCss()` is fast enough to be effectively free to call; feel free to use it inline. Extracting selectors as constants is good when it increases clarity, but doesn't need to be done to the maximum extent possible, when it can reduce clarity!
- I see you're running the code to open the filter dialog, do something, and then close it without any `await` statements in between them; this means that the entire action is performed within the same javascript event loop task. There is thus no user input processing and no page render possible between the first and last task, and thus the glass-container-hide-and-show and filter-button-disable-and-enable can be removed without any change to functionality!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,210 @@
...
+const CHECKBOX_STATE_SELECTOR = `${ROLLUP_CHECKBOX_SELECTOR} + ${keyToCss(
+  'box'
+)}`;
+const GLASS_CONTAINER_SELECTOR = '#glass-container';
+const GLASS_CONTAINER_DIALOG_SELECTOR = keyToCss('glass');
+const APPLY_FILTERS_SELECTOR = `${GLASS_CONTAINER_SELECTOR} ${keyToCss(
+  'button'
+)}${keyToCss('default')}`;
+const CLOSE_FILTER_DIALOG_SELECTOR = '[aria-label="Close"]';
+const ACTIVITY_BODY_SELECTOR = '.usqcu';
```
marcustyphoon, line 24: This won't be reliable, unfortunately; we'll need a more explicit way of detecting when to run!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,210 @@
...
+    {
+      className: `${MUTUAL_ACTIVITY_CLASS} toggle-button`,
+      type: 'checkbox',
+      name: MUTUAL_ACTIVITY_CLASS,
+      checked: isActivated
+    }
+  );
+
+  mutualActivity.appendChild(mutualActivityLabel);
+  mutualActivity.appendChild(mutualActivityToggleButton);
```
marcustyphoon, line 169: While you certainly don't have to use it, the `dom()` helper function in `/util/dom.js` can make this sort of thing very succinct!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+};
+
+export const main = async () => {
+  pageModifications.register(ACTIVITY_BODY_SELECTOR, createToggleButton);
+};
+
+export const clean = async () => {
+  disableFilter();
```
marcustyphoon, line 205: I think this might be a slight bug? It'll save your current preference as if you'd switched the toggle to off, right?

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,210 @@
...
+    openFilterDialog(glassContainer, filterButton);
+  }
+
+  toggleRollupCheckbox(getRollupCheckboxState() !== showGroupedNotifications);
+  closeFilterDialog(glassContainer, filterButton);
+};
+
+const setIsActivated = async (isActivated) => {
+  browser.storage.local.set({ [IS_ACTIVATED_STORAGE_KEY]: isActivated });
+};
```
marcustyphoon, line 108: It doesn't wind up mattering here, but if you wanted this function to resolve only after the `browser.storage.local.set` call, you would either have to await it or return it!

```js
const setIsActivated = async (isActivated) => {
  await browser.storage.local.set({ [IS_ACTIVATED_STORAGE_KEY]: isActivated });
};
```
```js
const setIsActivated = (isActivated) => 
   browser.storage.local.set({ [IS_ACTIVATED_STORAGE_KEY]: isActivated });
```
```js
const setIsActivated = async (isActivated) => 
   browser.storage.local.set({ [IS_ACTIVATED_STORAGE_KEY]: isActivated });
```

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,210 @@
...
+const enableFilter = () => {
+  document.head.append(nonMutualStyleElement);
+  const notificationSection = document.querySelector(
+    NOTIFICATION_SECTION_SELECTOR
+  );
+  const scrollEvent = new Event('scroll', {
+    bubbles: true
+  });
+
+  notificationSection.dispatchEvent(scrollEvent);
```
marcustyphoon, line 122: Neat trick. Do you know if this works if the second "page" of notifications is also entirely hidden, or will it stop then? Kind of hard to test, I guess.

In Show Originals, I set `min-height: 100vh` (as tall as the browser viewport) on timeline elements where the script is active to ensure that, if all else fails, the user can manually scroll down and trigger post loading; might be appropriate here.

src/scripts/mutual_activity.js
marcustyphoon: uh. that's. not the line I wanted to comment on, github

alleycatboy: I can swap that out with the notification body selector, as it's unique. Would that be reliable?

src/scripts/mutual_activity.js
alleycatboy: I'll look into it, it definitely looks less verbose!

src/scripts/mutual_activity.js
alleycatboy: Yep, it'll preserve that. Would it be appropriate to clear the preferences upon clean?

src/scripts/mutual_activity.js
alleycatboy: I've tested that, it'll keep going till either the appropriate # of notifications are displayed or the end of the feed is reached!

src/scripts/mutual_activity.js
alleycatboy: Good catch! Let's change that to keep good form.

alleycatboy:
> * `keyToCss()` is fast enough to be effectively free to call; feel free to use it inline. Extracting selectors as constants is good when it increases clarity, but doesn't need to be done to the maximum extent possible, when it can reduce clarity!

I agree! I try to keep magic strings within the executing and repeating code to an absolute minimum as it makes refactoring a breeze. 

> * I see you're running the code to open the filter dialog, do something, and then close it without any `await` statements in between them; this means that the entire action is performed within the same javascript event loop task. There is thus no user input processing and no page render possible between the first and last task, and thus the glass-container-hide-and-show and filter-button-disable-and-enable can be removed without any change to functionality!

Nice catch! That was a holdover from an earlier method I was attempting with `await` on those functions. Just tested it and removed those functions.

Overall, very constructive review! I'll push changes with your suggestions.

src/scripts/mutual_activity.js
alleycatboy: Using `dom()` now, much cleaner!

src/scripts/mutual_activity.js
marcustyphoon: Oh, I see what you're doing: making the toggle state revert to false in clean(), but remembering your last setting when you close/open/navigate on Tumblr?

alleycatboy: It should do that inherently by the `browser.storage.local.get()`, but I removed `disableFilter()` and made it and explicitly remove the storage key for when the script is disabled.

marcustyphoon: No, I gotcha. I had just previously thought the storage being cleared when you toggle the script off was unintentional, but I see you meant to do that.

I think it probably makes sense either way!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,193 @@
...
+        checked: isActivated
+      },
+      {
+        input: toggleFilter
+      }
+    )
+  ]);
+  const activityBar = document.querySelector(FILTER_BUTTON_SELECTOR);
+
+  $(activityBar).wrap(`<span class="${FILTER_CONTAINER_CLASS}"></span>`);
```
marcustyphoon, line 162: You could touch the DOM structure less by not wrapping this and making the `mutualActivity` element fill the available area (`flex: 1`) and put its contents on the right (`justify-content: flex-end`).

Generally we try to avoid wrapping Tumblr's elements in anything, though I've done it in Collapsed Queue and it looks like the React framework that Tumblr's code uses can still figure out which element is which and nothing seems to break.

alleycatboy: Yep, messing with the DOM without React's knowledge generally leads to unexpected behavior; let's use `flex` instead!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,191 @@
...
+
+  ${ROLLUP_CHECKBOX_SELECTOR} {
+    pointer-events: none;
+    cursor: not-allowed;
+  }
+
+  ${CHECKBOX_STATE_SELECTOR} {
+    background-color: grey;
+    cursor: not-allowed;
+  }
```
marcustyphoon, line 48: This selector (computed to `[aria-labelledby="rollups"] + :is(.uNugm)` on my dash) winds up as the same specificity as `input:checked + .uNugm`, which is the selector for one of Tumblr's css rules that normally sets the background color of this element.

This results in inconsistent behavior, surprisingly, since it thus depends which style gets inserted first! See #761. (I assume the grey background is being applied for you, of course; for me, apparently, it isn't!)

Not sure if April has a way of preventing this problem from happening across the extension, as noted in the linked issue, but for now I guess setting !important on the color works, or manipulating the selector to have higher specificity.

src/scripts/mutual_activity.js
marcustyphoon: Typically, when registering a callback function that adds a DOM element, we register the selector that our element will be placed in/after/before! That way, the pageModifications engine will supply you with the element in question as the callback parameter and you don't have to query for it again. In this case, since you're finding `FILTER_BUTTON_SELECTOR` to add your button, you can register that, then accept e.g. `[filterButton]` as the createToggleButton function parameter.

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,191 @@
...
+    ]
+  );
+
+  $(FILTER_BUTTON_SELECTOR).before(mutualActivity);
+};
+
+const removeToggleButton = () => {
+  document.querySelector(`span.${MUTUAL_ACTIVITY_CLASS}`)?.remove();
+  $(FILTER_BUTTON_SELECTOR).unwrap(`span.${FILTER_CONTAINER_CLASS}`);
+};
```
marcustyphoon, line 164: no longer used!

src/scripts/mutual_activity.js
alleycatboy: I see, I think April is correct in her hypothesis that the native CSS is being lazy-loaded. I'll use both finer specificity in the selector and add an `!important` on it. 

Additionally I've changed it to use `rgba(var(--black), 0.40)` so it keeps the same look-and-feel across different themes, and it's the same variable used across the site for disabled actions.

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,189 @@
...
+
+const toggleFilter = async ({ target }) => {
+  target?.checked ? enableFilter() : disableFilter();
+};
+
+const isOnActivityPage = () =>
+  document.querySelector(FILTER_BUTTON_SELECTOR) !== null;
+
+const createToggleButton = async () => {
```
marcustyphoon, line 126:
```suggestion
const createToggleButton = async ([filterButton]) => {
```

Now you can take an array of page modifications in the callback!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,189 @@
...
+      Object.assign(document.createElement('input'), {
+        className: `${MUTUAL_ACTIVITY_CLASS} toggle-button`,
+        type: 'checkbox',
+        name: MUTUAL_ACTIVITY_CLASS,
+        checked: isActivated
+      })
+    ]
+  );
+
+  $(FILTER_BUTTON_SELECTOR).before(mutualActivity);
```
marcustyphoon, line 157:
```suggestion
  filterButton.before(mutualActivity);
```

src/scripts/mutual_activity.js
alleycatboy: NICE! Now we can keep all the jQuery out of this script.

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,189 @@
...
+const FILTER_BUTTON_SELECTOR = `${STATS_CONTAINER_SELECTOR} > ${MINI_HEADING_SELECTOR} ${BUTTON_SELECTOR}`;
+const ROLLUP_CHECKBOX_SELECTOR = '[aria-labelledby="rollups"]';
+const CHECKBOX_STATE_SELECTOR = `${ROLLUP_CHECKBOX_SELECTOR} + ${keyToCss(
+  'box'
+)}`;
+const GLASS_CONTAINER_DIALOG_SELECTOR = keyToCss('glass');
+const APPLY_FILTERS_SELECTOR = `#glass-container ${keyToCss(
+  'button'
+)}${keyToCss('default')}`;
+const CLOSE_FILTER_DIALOG_SELECTOR = '[aria-label="Close"]';
```
marcustyphoon, line 24: Selecting based on aria-labels requires use of the `translate()` utility (e.g. ``` `[aria-label="${translate('Close')}"]\` ```) to ensure that they'll work on non-English Tumblr!

src/scripts/mutual_activity.js
```diff
@@ -0,0 +1,189 @@
...
+import { dom } from '../util/dom.js';
+import { buildSvg } from '../util/remixicon.js';
+
+const NOTIFICATION_SECTION_SELECTOR = keyToCss('notifications');
+const NOTIFICATION_SELECTOR = keyToCss('notification');
+const FOLLOWED_SELECTOR = keyToCss('followed');
+const ROLLUP_SELECTOR = keyToCss('rollup');
+const MINI_HEADING_SELECTOR = keyToCss('miniHeading');
+const STATS_CONTAINER_SELECTOR = keyToCss('statsContainer');
+const BUTTON_SELECTOR = keyToCss('button');
```
marcustyphoon, line 14: If it were me I would inline these, though that's personal preference.

alleycatboy: On second evaluation, a lot of these were holdovers from previous methodologies. Let's pare these down so only the reused selectors are constant-ized.

alleycatboy: On second evaluation, a lot of these were holdovers from previous methodologies. Let's pare these down so only the reused selectors are constant-ized!

alleycatboy: Removing the ones that are single use and don't contribute to readability.

AprilSylph:
Obsoleted by #1303.

PR #878, opened 2022-11-29
Themed Posts: bugfix for deleted blogs, new option to control deleted blog theming

### LINKED ISSUE #874: Themed Posts breaks on posts containing deleted blogs
### Platform

Windows 10

### Browser

Chrome 107.0.5304.107

### Addon version

v0.21.0 (Beta)

### Details

Themed Posts doesn't work when `Theme every reblog trail item individually` is enabled and some of the posts in the trail are from deleted blogs

---

[example post](https://www.tumblr.com/im-mage/702142046276861952) (screenshots taken from timeline since the post view doesn't use the palette)

current behavior: 
image

expected behavior:  
either this: (using outer post's colors)
image

or this: (using the default post colors from the current palette)
image

(screenshots taken with these settings):
```json
{
  "enabledScripts": [
    "themed_posts"
  ],
  "themed_posts.preferences.enableOnPeepr": true,
  "themed_posts.preferences.reblogTrailTheming": true
}
```

---

I've got a fix for this mostly done (that's what the expected behavior screenshots are taken from), ~~I'll open a PR once it's done~~ #878

adrianmgg:
### Description

Fixes #874, and introduces a new option to control the theming behavior for deleted/blacklisted posts (described in #874 under "expected behavior").

### Testing steps

1. import these settings
    ```json
    {
      "enabledScripts": ["themed_posts"],
      "themed_posts.preferences.blacklistedUsernames": "",
      "themed_posts.preferences.enableOnPeepr": false,
      "themed_posts.preferences.missingPostMode": "palette",
      "themed_posts.preferences.reblogTrailTheming": true
    }
    ```
1. either:
   - a) scroll down to a post containing a deactivated blog
   - or b) scroll down to a post containing multiple blogs, and add one user somewhere in the middle of the trail to the themed posts blacklist
1. set your palette to something easily distinguishable from any of the themes in that post (in my screenshots I used the built in `Low-Contrast Classic`)
1. the deleted/manually blacklisted blog should have the colors of the palette.
   image
1. change `Themed Posts`->`Disabled/deleted blog theming` to `inherit reblogging blog's theme`
1. the deleted/manually blacklisted blog should now have the colors of the reblogging user's theme
   image


src/scripts/themed_posts.js
```diff
@@ -65,15 +67,19 @@ const processPosts = async function (postElements) {
     postElement.dataset.xkitThemed = blog.name ?? '';
 
     if (reblogTrailTheming) {
+      const blogNameTrail = trail.map(item => item?.blog?.name);
+      if (content.length > 0) {
+        blogNameTrail.push(blog?.name);
+      }
       [...postElement.querySelectorAll(reblogSelector)].forEach((reblog, i) => {
-        reblog.dataset.xkitThemed = trail[i]?.blog?.name ?? '';
+        reblog.dataset.xkitThemed = blogNameTrail[i] ?? '';
```
marcustyphoon, line 75: I think I see what you're doing... this ensures that you theme the added content block, if there is one, with the post colors because you've used "has the data attribute set to nothing" as the selector for the fallback color for broken/blacklisted blogs?

adrianmgg: pretty much, yep! though fwiw the fallback selector is actually "has the data attribute set to anything", since that works with how the blacklisted blogs were already being handled (data attribute is still set to their username, but no css rule is generated)

AprilSylph (approved):
Sorry for the late review, almost a year later I'm still trying to balance having a full-time job with not abandoning the projects that got me hired in the first place!

LGTM :+1:

Issue #838, opened 2022-11-19
Quick Tags (and Trim Reblogs and Quick Flags PR): tipping status lost


marcustyphoon:
### Platform

MacOS 10.15.7 x86

### Browser

Ungoogled Chromium v107.0.5304.110

### Addon version

v0.21.0

### Details

Adding a tag to an original post with Quick Tags turns off tipping on the post; we don't pass through `canBeTipped` and I guess undefined is falsey.

This presumably technically isn't an issue in Trim Reblogs, since you can't have tipping on a non-original post and you can only run Trim Reblogs on those, but more generally, an audit of passthrough fields that may be missing from our post-editing code seems wise.

> For editing a post, all of the request parameters from the NPF Post Creation route are expected (depending on if it's an original post or reblog), along with the Post's ID in the query path. If you are editing a scheduled post, make sure to include its publish_on value.
> https://github.com/tumblr/docs/blob/master/api.md#postspost-id---editing-a-post-neue-post-format

can_be_tipped
community_label_categories
has_community_label
hide_trail
interactability_reblog
placement_id (and possibly earned_id?)
send_to_twitter

#827 #839

marcustyphoon:
proposal:
  
```js

import { filterPropsForEditing } from './util/something.js'

const {
    response: {
      tags = [],
      ...rest
    }
  } = await apiFetch(`/v2/blog/${uuid}/posts/${postId}`);

  tags.push(...tagsToAdd);

  try {
    const { response: { displayText } } = await apiFetch(`/v2/blog/${uuid}/posts/${postId}`, {
      method: 'PUT',
      body: {
        ...filterPropsForEditing(rest),
        tags: tags.join(','),
      }
    });

marcustyphoon:
gutenblr:

```js
i()(this, "getPostData", (()=>{
  const {editablePost: e, editorValue: t, tags: n} = this.state
    , {trail: r} = this.props
    , s = this.blockEditorRef.current?.getContent()
    , o = s?.content ?? (0,
  Yt.uM)(t)
    , i = o.findIndex(we.KM)
    , a = i > -1;
  a && o.splice(i, 1);
  const l = s?.layout ?? e?.layout?.filter(we.af) ?? [];
  if (!s) {
      const e = (0,
      Yt.lZ)(t)
        , n = e.display.some((e=>e.blocks.length > 1));
      (a || n) && (a && (e.truncateAfter = i - 1),
      l.push(e))
  }
[...]

PR #879, opened 2022-11-29
Create NPF edit parameter utility

marcustyphoon:
This primarily serves to ensure that Quick Tags has no side effects. Makes sense to do anyway, though, I think.

Does this need default values to handle missing data?

### Description

As proposed in #838, this creates a utility function that encapsulates the logic for, "I want to edit a post using the NPF API endpoints; what do I start with so that I don't change anything besides what I want to change?" It takes the JSON response from the `/posts/{post-id}` GET request, removes irrelevant fields, and applies a few transformations, which Quick Tags and Trim Reblogs both had implemented separately.

It ensures that there is one place to add future properties that should be passed through or future transformations that should be added.

The functional change here is passing through `can_be_tipped` (which is necessary) and `has_community_label`, and `community_label_categories` (which do not seem necessary).

For discussion: According to my testing, the list of passed-through properties could actually be shorter; for example, `source_url` is optional; should this err on the side of passing the maximum amount of data in case the Tumblr backend is ever changed, or the minimum amount to save network bytes?

Also, note that this relies on the snake_case conversion in apiFetch. Keeping everything as camelCase for as long as possible is the least confusing, IMO.

### Testing steps

- Enable tipping on your blog.
- Load XKit Rewritten without this PR.
- Set the tipping status on a test post to the opposite of your blog's "Add tip button to your posts by default" preference.
- Add a tag to the post with Quick Tags. Refresh the page. The tipping status should change to the default.
- Load XKit Rewritten with this PR.
- Repeat the above steps. Quick Tags should function and the tipping status should stay as you set it.

Also, test Trim Reblogs to ensure it functions normally; it should be no different than before.

### DIFF
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -6,7 +6,7 @@ import { notify } from '../util/notifications.js';
 import { registerPostOption, unregisterPostOption } from '../util/post_actions.js';
 import { getPreferences } from '../util/preferences.js';
 import { timelineObject, editPostFormTags } from '../util/react_props.js';
-import { apiFetch } from '../util/tumblr_helpers.js';
+import { apiFetch, createEditRequestBody } from '../util/tumblr_helpers.js';
 
 const symbolId = 'ri-price-tag-3-line';
 const buttonClass = 'xkit-quick-tags-button';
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Honestly maybe it's just cleaner to do this? Being able to copy-paste snake_case from the Tumblr API docs is a bit of a fringe benefit I guess. I thought it was going to match exactly but `is_private` gets omitted, etc.

```js
/**
 * Create an NPF edit request body.
 * 
 * @see https://github.com/tumblr/docs/blob/master/api.md#postspost-id---editing-a-post-neue-post-format
 * @see https://github.com/tumblr/docs/blob/master/api.md#posts---createreblog-a-post-neue-post-format
 *
 * @param {object} postData - camelCased /posts/{post-id} GET request response JSON
 * @returns {object} editRequestBody - camelCased /posts/{post-id} PUT request body parameters
 */
export const createEditRequestBody = postData => {
  const { 
    content,
    layout,
    state,
    publishOn,
    date,
    tags,
    sourceUrl,
    sourceUrlRaw,
    slug,
    interactabilityReblog,
    
    canBeTipped,
    hasCommunityLabel,
    communityLabelCategories
  } = postData;

  return {
    content,
    layout,
    state,
    publishOn,
    date,
    tags: tags.join(','),
    sourceUrl: sourceUrlRaw ?? sourceUrl,
    slug,
    interactabilityReblog,
    
    canBeTipped,
    hasCommunityLabel,
    communityLabelCategories
  };
};
```

AprilSylph (approved):
Enabling tipping would be a pain for me, but I can see `can_be_tipped: false` getting sent in the PUT request for Quick Tags. Trim Reblogs also appears to still work correctly!

PR #880, opened 2022-11-29
New Tweak: Respect user's theme when in blog-view.

alleycatboy:
### Description

This tweak is great for users with dark color pallets that wish to not have blinding light when viewing other user's blogs with lighter themes. 

I got the idea from this Suggestion Box comment [here](https://github.com/AprilSylph/XKit-Rewritten/discussions/646#discussioncomment-3087855) on how they'd like to still see the theme while keeping posts readable.

This allows for more comfortable post viewing while still seeing the splash of flair from 

Note: this is only for the posts themselves. The surrounding theme of the blog is preserved. 

Additional note: this respects all the color pallets (True Blue, Goth Rave, Cybernetic, etc.)

### Inquiry

Since this doesn't truly affect the dashboard, simply the **blog-view**, is it appropriate to be included in Tweaks? Tweaks is defined to the users as "Miscellaneous **dashboard** options". 

Would it be more appropriate for this to be a stand-alone script?

Note: this is not the same behavior as Themed Posts, as this is the user's pallet and not the target blog's theme.

Note: this is not the same behavior as disabling "Use blog colors when viewing blogs" as mentioned [here](https://github.com/AprilSylph/XKit-Rewritten/discussions/646#discussioncomment-3055765). It preserves the surrounding theme. 

firefox_BU0WVMVOcm

### Technical Explanation
All this does is prevent the CSS variables of the blog from overriding the user's color pallet on posts by setting the color variables to `inherit !important;` and appends the style to the document head. 

### Testing steps

Just enable "Respect the user's theme on blog pages" under Tweaks! 

It should only change the theme of **posts** in blog-view.

### DIFF
--- a/src/scripts/tweaks/respect_theme.js
+++ b/src/scripts/tweaks/respect_theme.js
@@ -1,0 +1,32 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const BLOG_POST_BODY_SELECTOR = keyToCss('scrollContainer', 'isModal');
+
+const styleElement = buildStyle(`
+ ${BLOG_POST_BODY_SELECTOR} {
+    --black: inherit !important;
+    --white: inherit !important;
+    --white-on-dark: inherit !important;
+    --navy: inherit !important;
+    --red: inherit !important;
+    --orange: inherit !important;
+    --yellow: inherit !important;
+    --green: inherit !important;
+    --blue: inherit !important;
+    --purple: inherit !important;
+    --pink: inherit !important;
+    --accent: inherit !important;
+    --secondary-accent: inherit !important;
+    --follow: inherit !important;
+    --override-posts--color-primary-link: inherit !important;
+  }
+`);
+
+export const main = async function () {
+  document.head.append(styleElement);
+};
+
+export const clean = async function () {
+  styleElement.remove();
+};
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -7,7 +7,7 @@
     "background_color": "#ecc84b"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#tweaks",
-  "relatedTerms": [ "Blacklist", "Separator" ],
+  "relatedTerms": ["Blacklist", "Separator"],
   "preferences": {
     "slim_filtered_screens": {
       "type": "checkbox",
@@ -88,6 +88,11 @@
     "no_where_were_we": {
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
+      "default": false
+    },
+    "respect_theme": {
+      "type": "checkbox",
+      "label": "Respect the user's theme on blog pages",
       "default": false
     }
   }


src/scripts/tweaks/respect_theme.js
```diff
@@ -0,0 +1,32 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const BLOG_POST_BODY_SELECTOR = keyToCss('scrollContainer', 'isModal');
```
marcustyphoon, line 4: This works on the modal version of the blog view that pops up if you click a blog, but not the embedded one that appears if you directly open a blog view page (e.g. open a new tab, go to `https://www.tumblr.com/staff`).

We had to solve this problem before and put the result in `blogViewSelector` in interface.js, so you can import and use that!

alleycatboy: I'm looking at it now and it appears to work on that page, does it look different for you?

firefox_9i4F0bizVP

I think staff may have changed how these selectors work?

marcustyphoon: Interesting. Yeah, mine has an `isServer` (translated) class.

alleycatboy: That method works for Themed Posts as they're selecting each post individually and appending CSS for each blog name. 

Applying the overriding the styles to `inherit !important` via `[style*="--blog-title-color"] *` won't yield results here as it will just inherit the parent, `:is(.jxu4l, .DCCfo, .kXP4L, .ZI59C, .feCof)`. At that point it loses scope of the `:root` inherited styles.

marcustyphoon: Oh, yeah, wait, my mistake; this is a different error with another extension; checking on that.

And whoops, I see; I thought that variable was set to `[style*="--blog-title-color"]`. What I meant was that that works in place of `keyToCss('scrollContainer', 'isModal')`!

By the way, `keyToCss('a', 'b')` matches any element that contains either a or b, not elements that match both.

alleycatboy: > Interesting. Yeah, mine has an `isServer` (translated) class.

Curiously enough, mine has `isServer` as well, but I think it gets scooped up by `scrollContainer` (translated) class. 

I'll add `isServer` to the `keyToCss` and see if that gets coverage. Let me know if that changes anything!

marcustyphoon: For clarification, these should both work:

```js
const BLOG_POST_BODY_SELECTOR = `${keyToCss('scrollContainer')}${{keyToCss('isServer', 'isModal'})`;
```

```js
const BLOG_POST_BODY_SELECTOR = '[style*="--blog-title-color"]';
```

alleycatboy: > And whoops, I see; I thought that variable was set to `[style*="--blog-title-color"]`. What I meant was that that works in place of `keyToCss('scrollContainer', 'isModal')`!

Ah, that makes more sense. Since both work, I'll just use the string of `[style*="--blog-title-color"]` since that doesn't have to iterate through the map.

alleycatboy: Do you think we should add `[style*="--blog-title-color"]` as an export to `interface.js`? I could see that being used again in the future.

marcustyphoon: Sounds good. And... huh. I have no idea why I thought it wasn't working on that page now; I can't get it to fail again (with your original PR).

alleycatboy: > Sounds good. And... huh. I have no idea why I thought it wasn't working on that page now; I can't get it to fail again (with your original PR).

Now ain't that extension development in a nutshell! 😭

PR #882, opened 2022-11-30
AccessKit Disable Gifs: Fix hover-to-animate in post editor trail

marcustyphoon:
Is this the most pointless PR of all time? I think it probably is.

### Description

Ensures that the Disable Gifs behavior of unfreezing an entire photoset when any of it is hovered is applied to the trail items shown in the post editor when reblogging.

### Testing steps

Find a post with multiple gifs in a photoset (I search for "gif"). Press reblog on it. Mouse over the photoset in the trail within the editor. All of the gifs in the photoset should animate.


marcustyphoon:
this doesn't even work. what am I doing.

AprilSylph (approved):

PR #889, opened 2022-12-01
Mutual Checker: Icons in blog cards; icon tooltips

marcustyphoon:
<details>
<summary>resolved</summary>
<p>Right now this doesn't completely work; I don't understand the CSS that's causing a specific friend's URL to vanish completely (presumably because it's long and has no hyphen) and I'm not making any progress. I require a CSS wizard.</p>

Screen Shot 2022-12-01 at 2 29 19 PM
</details>

### Description

This adds mutual icons to the blog cards that appear when hovering over a blog avatar in most places, which I find really useful. For example, Mutual Checker only adds icons to the user who posted a post, not to users who contributed trail items; with this PR one can mouse over them to see if they're the person you're thinking of or not. (I would like to add an "is following you" icon type as well; implementing in #992)

This also adds on-hover tooltips to mutual icons; see #890.

Code changes performed to do this:
- Moved static stylesheet styles to `buildStyle`
- Renamed the style element used for the only-mutual-notifications feature added by #1303
- Refactor: Extracted am-I-following-this-user and am-I-mutuals-with-this-user into their own functions
- Used the `blogData` react-props-extracting utility function to get the target user blog data when possible in the resulting `getIsFollowing` function
- Replaced the global `icon` variable/element that gets cloned when needed with a `createIcon` function in order to implement icon tooltips and icon color varying in blog cards


alleycatboy:
> Right now this doesn't completely work; I don't understand the CSS that's causing a specific friend's URL to vanish completely (presumably because it's long and has no hyphen) and I'm not making any progress. I require a CSS wizard.

Figured out a way around this! 

We can prepend the icon to the parent of the parent of blogCardLink, I've tried this at several node levels and this prevents the link from overflowing to blank. 

```javascript
blogCardLink.parentElement.parentElement.prepend(icon.cloneNode(true));
```

We would then have to change the display of the parent to flex and also set the height of the SVG. 

```javascript
  ${keyToCss('popoverHolder')} ${keyToCss('blogCardBlogLink')} {
    display: flex;
  }

  ${keyToCss('popoverHolder')} ${keyToCss('blogCardBlogLink')} svg.xkit-mutual-icon {
    height: 1.5em;
  }
`);
```

I've tried and verified that it produces this:
firefox_dnSZ9YPjPw

marcustyphoon:
Hmm, it does indeed! Targeting the anchor element is probably the cleanest way to implement that, and allows for easily stealing the right color.

```js
pageModifications.register(`${keyToCss('blogCardBlogLink')} > a`, processBlogCardLinks);
```

```js
const clonedIcon = icon.cloneNode(true);
clonedIcon.setAttribute('fill', blogCardLink.style.color);
blogCardLink.before(clonedIcon);
```

alleycatboy:
> Edit: Ah, this breaks one day a year.

Not if we curry `processBlogCardLinks`!

```javascript
const processBlogCardLinks = aprilFools => blogCardLinks =>
// other stuff
!aprilFools && clonedIcon.setAttribute('fill', blogCardLink.style.color);
```

```javascript
pageModifications.register(`${keyToCss('blogCardBlogLink')} > a`, processBlogCardLinks(aprilFools));
```

marcustyphoon:
Generally in this codebase we just define preference variables in the module scope, relying on main() having been called and having set them before any other functions in the module, but that does work here too!

alleycatboy:
Ah, then would setting `aprilFools` to `let` and moving it out of main to the top work?

marcustyphoon:
Note: a network request can be skipped the first time you mouse over a blog avatar that you do not follow by using the `blogData` react props helper function from #910 to get the following status of a blog card:

```diff
+ import { blogData } from '../util/meatballs.js';

  const getIsFollowing = async (blogName, element) => {
-   const { blog } = await timelineObject(element) ?? {};
+   const blog = await blogData(element) ?? (await timelineObject(element))?.blog;

  if (following[blogName] === undefined) {
    if (blogName === blog?.name) {
```

This really isn't very many network requests IMO, so super minor.

marcustyphoon:
(Mention me here and I'll bother fixing the merge conflicts in this PR.)

marcustyphoon:
I feel like there should be a better way to do this, but... I kind of have to rewrite a bunch of stuff to get this to merge correctly, so I could either make all of those changes in the merge commit (obviously stupid) or just sort of merge commit garbage and make a rather random diff in the next commit? Eh, oh well.

src/scripts/mutual_checker.js
```diff
@@ -50,72 +74,103 @@ const addIcons = function (postElements) {
...
   const today = new Date();
   const aprilFools = (today.getMonth() === 3 && today.getDate() === 1);
 
-  icon = dom('svg', {
+  const icon = dom('svg', {
     xmlns: 'http://www.w3.org/2000/svg',
     class: mutualIconClass,
     viewBox: '0 0 1000 1000',
-    fill: aprilFools ? '#00b8ff' : 'rgb(var(--black))'
+    fill: color || 'rgb(var(--black))'
```
AprilSylph, line 153: 1. Do we ever expect `color` to be an empty string here?
2. I feel like we can keep the `aprilFools` ternary, and just chain it with the fallback

```suggestion
    fill: aprilFools ? '#00b8ff' : (color ?? 'rgb(var(--black))')
```

src/scripts/mutual_checker.js
```diff
@@ -50,72 +74,103 @@ const addIcons = function (postElements) {
...
+    ]),
     dom('path', { xmlns: 'http://www.w3.org/2000/svg', d: aprilFools ? aprilFoolsPath : regularPath })
   ]);
-
-  onNewPosts.addListener(addIcons);
-  if (showOnlyMutualNotifications) {
-    document.documentElement.append(styleElement);
-    onNewNotifications.addListener(processNotifications);
-  }
+  aprilFools && icon.setAttribute('fill', '#00b8ff');
```
AprilSylph, line 160:
```suggestion
```

src/scripts/mutual_checker.js
```diff
@@ -50,72 +74,103 @@ const addIcons = function (postElements) {
...
+  blogCardLinks.forEach(async blogCardLink => {
+    const blogName = blogCardLink.querySelector(keyToCss('blogLinkShort'))?.textContent || blogCardLink?.textContent;
+    if (!blogName) return;
+
+    const followingBlog = await getIsFollowing(blogName, blogCardLink);
+    if (!followingBlog) return;
+
+    const isMutual = await getIsFollowingYou(blogName);
+    if (isMutual) {
+      blogCardLink.before(createIcon(blogName, blogCardLink.style.color));
```
AprilSylph, line 101: Ah, I see. Well, we can fix this by using `getComputedStyle` instead, which will always return a valid value...

```suggestion
      blogCardLink.before(createIcon(blogName, getComputedStyle(blogCardLink).color));
```

...or is `getComputedStyle` too expensive? 😛

AprilSylph (commented):
hey I found these bikes, you got a shed for them?

src/scripts/mutual_checker.js
marcustyphoon: I think what I was supposed to do here was set a default argument for the function, tbh.

src/scripts/mutual_checker.js
marcustyphoon: I don't think I was worried about it not working! But this is more resilient, so why not.

AprilSylph:
Works, and I have no more code nits, but I'm not happy about the vertical alignment of the icon in blog cards...

It looks like the `<svg>`'s bounding box is just... wrong? It's bigger than the actual icon. Needs to be cleaned up in a vector editor; I'd do it in Figma right now, but it's 11:57pm so I really shouldn't. (Goodnight.)

AprilSylph:
oh my GODDDD it's not the SVG's bounding box that's the problem at all it's just the fucking text having `position: relative; top: 3px;` that's causing the alignment problem!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

AprilSylph:
wait no. the real problem is... that the SVG element... resizes incorrectly when a `height` is set explicitly? It behaves differently when setting `width: 1.4em` even though the viewbox is square.

I hate this.

marcustyphoon:
There's something in the back of my mind about having encountered this before and maybe fixed it but I can't recall what the details were... and it's annoying to investigate because the element only appears on hover. Might be time to override removeChild to make it easier to examine again, sigh.

AprilSylph:
It's pretty easy to inspect, actually. Just open up the debugger, have it focused so that F8 will pause execution, hover the thing, pause execution. On Firefox at least the inspector still works completely and changes to CSS are reflected visually.

I have a fix almost ready to push, but now I'm trying to fix normal icons too... the alignment has been bothering me for forever, but if I do it too good it looks like part of the icon gets cut off, which I am trying to mitigate now.

AprilSylph:
Actually nevermind, the in-post icons are fine actually. I thought I saw an elevated level perfection in the fixed blog card alignment, but it was illusory.

AprilSylph:
Nevermind on that nevermind. I have seen a clean 18px square mutual icon and I like it a lot.

marcustyphoon:
> It's pretty easy to inspect, actually. Just open up the debugger, have it focused so that F8 will pause execution, hover the thing, pause execution. On Firefox at least the inspector still works completely and changes to CSS are reflected visually.

Oh, fun, I figured out why when I tried that it never worked: I use the debugger popped out in another window, which of course prevents you from simultaneously focusing it and mousing over the page to make the popup appear. I solved this with a userscript, though I actually picked a hotkey combination that also does something on Tumblr, so maybe change it, hypothetical viewer:

```js
document.addEventListener('keydown', event => {
  if (event.code === 'KeyP' && event.shiftKey) {
    event.stopPropagation();
    debugger;
  }
}, { useCapture: true });
```

Anyway, I still find this method inconvenient, as you can't (or I haven't figured out how to) navigate to an element by right clicking it, and sure I've gotten pretty good at keyboard-navigating through the redpop DOM, but it seems silly to have to :D

AprilSylph (approved):

AprilSylph:
The element inspector also has a button that lets you pick an element to inspect via left-click. It's the one shaped like a pointer.

marcustyphoon:
Yeah, and it doesn't work when, oh no wait, yes it does 🤦

marcustyphoon:
Me: "are you sure it doesn't look worse this way"
Also me: \*using Palettes for Tumblr with both a font-face and font-size override*
¯\\\_(ツ)\_/¯

AprilSylph:
Side-by-side for fun:

Before | After
-|-
image | image

...hm, maybe the improvement was an illusion after all. I can't spot a difference here. (I spot it in the element inspector, though!)

marcustyphoon:
Right, so: the blog text in the post header has a normal font size of 14px (0.875rem) and a line height of 18px (1.125rem). We add the SVG inline and size it to match the line height of 18px (1.125rem), which it does. However...



The blog text in the blog card has a normal font size of 16px (1rem), a line height of 24px (1.5rem), and is shifted downward from its regular position in a flex container by 3px (hardcoded), for Reasons I guess.

Because we insert the SVG as the child of a flex container, it will get shrunk unless we override that. Your PR sets a width of 24px (1.5rem), but I observe an actual 18.531px width in practice. That's basically the same as setting `flex: none;` and not overriding the 1.125rem value. But this only vertically aligns correctly with the default font size, as it's basically a coincidence.

Setting the icon height to actually match the 1.5rem line height and shifting it downward by 3px with `position: relative` presumably makes the vertical alignment consistent with any font size value, but that's obviously way too big and also clips. 

Presumably the perfect CSS makes the icon the right size (maybe 18px, maybe closer to 18*16/14 = 20.57px so it's proportional to the font, whatever), positions it so it aligns with a 1.5rem-height box (no matter the font size), then shifts it down by 3px. Not sure this is actually doable declaratively, since text isn't centered vertically in its line.

This is one way to do that but it's... stupid?

```diff
   ${keyToCss('blogCardBlogLink')} svg.xkit-mutual-icon {
-    width: 1.5rem;
+    position: relative;
+    top: calc(3px + 0.15rem);
+    flex: none;
+    align-self: start;
   }
```

marcustyphoon:
Wait no, me stupid, this works.

```diff
   ${keyToCss('blogCardBlogLink')} svg.xkit-mutual-icon {
-    width: 1.5rem;
+    position: relative;
+    top: 3px;
+    flex: none;
+    align-self: start;
+    height: 1.5rem;
   }
```

AprilSylph:
Unfortunately because of the `overflow: hidden` on the blog name wrapper, the line-height-aligned icon gets snipped off at the bottom at the default font size. (Forgot to take a screenshot; whoops!) HOWEVER because the effect is only visual, and the SVG box _is_ perfectly aligned with the text box, we can just use `border-box` sizing and some top/bottom padding to reduce the content box to the text size, which fixes the visual issue.

Also tested and still aligned at weird-but-not-crazy base font sizes (12px to 22px).

AprilSylph (dismissed):

src/scripts/mutual_checker.js
```diff
@@ -1,28 +1,60 @@
...
+  }
+
+  [data-timeline="/v2/timeline/dashboard"] [${hiddenAttribute}] article {
+    display: none;
+  }
+
+  ${keyToCss('blogCardBlogLink')} {
+    display: flex;
+    align-items: end;
+  }
```
marcustyphoon, line 44: oh god wait. we're the ones. setting align-self.

marcustyphoon:
> the line-height-aligned icon gets snipped off at the bottom at the default font size.

Interesting, I'm not able to reproduce this!

src/scripts/mutual_checker.js
```diff
@@ -1,28 +1,60 @@
...
+  ${keyToCss('blogCardBlogLink')} svg.xkit-mutual-icon {
+    position: relative;
+    top: 3px;
+
+    flex: none;
+    align-self: start;
+
+    box-sizing: border-box;
+    height: 1.5rem;
+    padding: 0.25rem 0;
```
marcustyphoon, line 55: This results in a visual icon height of 1rem, less than the 1.125rem size of the icon in post headers; is that intentional?

AprilSylph: Sort of... I thought "if we can't match the line height we should match the text instead", but, matching the size of the non-blogcard icon did not occur to me at the time.

AprilSylph:
```suggestion
    height: 1.5rem;
    padding: 0.1875rem 0;
```

AprilSylph (approved):
I cannot imagine that there is a way to improve upon this. It just looks right.

image

marcustyphoon:
> I require a CSS wizard.

Spoiler alert: I required a CSS wizard.

Issue #791, opened 2022-10-28
Add a method to trigger the Quick Reblog pop-up on touchscreens


salazarbarrera:
### Concept

**Concept and motivation**

In mobile, when one taps and holds an image or <a> tag, Chromium and Firefox-based browsers alike display the context menu. At the same time, pressing and holding an element, will mimic a hovering cursor. Both things combined means that every time the Quick Reblog pop up is invoked, a context menu appears above it, slowing down the process and defeating its purpose in mobile browsers.

Adding an alternative way to invoke the popup would solve this problem. Maybe deleting the href and overriding the onclick listener associated to all the reblog buttons may do the trick, but I'm not sure if disabling the official click behaviour is the way to go.

*Things I tried*

I've been messing around with js and css injection to check if I could find a workaround without changing XKit Rewritten, but it seems adding a new option within the extension is the only way. In particular, I tried to disable the context menu, but it seems to be immortal in Android Firefox, and disabling the pointer events via css also disables the hover event. Also, I noticed that the reblog buttons and the new post button, share the same class.

AprilSylph:
The hover event doesn't happen on touchscreens at all. We would need to disable the context menu and add some sort of long-press listener (I don't think that's an actual event in JavaScript).

marcustyphoon:
Might be able to preventDefault() the touch event and then use the touch api to detect the long press. Might take a look at it at some point.

salazarbarrera:
Maybe we could close this issue. Not because of Xkit rewritten itself, but because web mobile Tumblr already incorporated something similar.

AprilSylph:
Between that and #893, yes, I think we can consider this resolved.

PR #893, opened 2022-12-04
Quick Reblog: Prevent context menu interference on touchscreens

marcustyphoon:
Browsers:
 - Kiwi Browser, Android: tested, **works**
 - Firefox nightly, Android: tested, **doesn't work reliably** 
 - Safari, iOS via packaging this extension with XCode: untested; probably does not work (see below)
 - Chromium, Windows w/ touchscreen: tested, **works**
 - Firefox, Windows w/ touchscreen: tested, **works**

### Description

This prevents the context menu that appears when long-pressing the reblog button using a touchscreen from popping up, allowing that gesture to be used to open the quick reblog menu.

Provides a resolution for #791 given that you aren't using Firefox for Android. (If you are, it seems to... sort of help? There's a specific duration that you can do the long press that seems to work.)

### Testing steps

- Enable Quick Reblog on a device with a touch screen.
- Long-press an active reblog button. The Quick Reblog popup should appear, and no right-click context menu should appear, if you are using a device/browser combination that this supports.


marcustyphoon:
> [Bug] `preventDefault` doesn't work for "contextmenu" events on anchor elements

https://github.com/mozilla-mobile/fenix/issues/14067

Ah, yep, there it is.

marcustyphoon:
I'm not clear from https://w3c.github.io/uievents/#event-type-contextmenu whether the `contextmenu` event ought to include the additional fields that `PointerEvent` supplies on top of `MouseEvent`, but Chromium appears to, Firefox appears not to (but adds a non-standard `mozInputSource` field), and Safari 15.6.1 appears not to as well.

Anyway, this is necessary to distinguish in the event handler between a mouse right-click, which we don't want to have any effect on, and a touch device long-press, which we want to prevent.

There are a whole bunch of mouse, pointer, and touch events going on close together, so maybe there's a way to be a bit fancier with a bunch more code, but at the moment I don't think it would enable this to work in more meaningful situations (Fenix seems straight-up bugged; we don't currently build for Safari).

AprilSylph (approved):

Issue #636, opened 2022-06-16
Chrome: Scripts load inconsistently on very-fresh tabs(?)


marcustyphoon:
### Platform

MacOS 10.14.6 x86

### Browser

Edge 102.0.1245.41

### Addon version

f318547

### Details

I am mostly at a loss here.

On a completely inconsistent basis, I sometimes open Tumblr to find that notificationblock is not active until a hard refresh. I finally got this to occur after adding logging before and after the dynamic import statements in main.js and at the top of every utility function, and discovered that it's actually any script that imports meatballs.js, as it for some reason is not resolved.

<img width="1265" alt="Screen Shot 2022-06-16 at 10 08 35 AM" src="https://user-images.githubusercontent.com/8336245/174127526-08b8096d-7319-4fc7-8d7c-13998ee83fcc.png">

<img width="1536" alt="Screen Shot 2022-06-16 at 10 15 36 AM" src="https://user-images.githubusercontent.com/8336245/174128965-f3a77707-dc59-4fc7-be3b-a693e72b8882.png">

I have never had this problem with a different set of scripts, but I can't figure out what would be different about the meatballs utility based on its dependencies. I also don't know if this occurs in production or if it's just a weird artifact of my local install, but I have no reproduction method yet so I can't test that really.

marcustyphoon:
Wait... mutual checker didn't load in this screenshot either.

Uh, what?

AprilSylph:
The current commit isn't released, so it by definition would not occur in production right now. I don't know enough about your local install, but if you hard-reset to master and reload the extension in the browser, you should get the exact same experience as if it were released.

If you can figure out how to reliably reproduce it (even if it is just "hard-refresh Tumblr x times until it happens"; I narrowed down a `fakeImport` bug that way), please rewind to 98af3c5e1706d82b647d6885a1aaef7fdf78cb6a and see if it still occurs.

I'm pausing all other development until this is resolved. If 225f7fec28fc50dd2155101269a3dc813c7b354c needs to be reverted I want a clean revert.

Also, to help me test this, do you mind sharing the config you're using?

AprilSylph:
Okay, if 225f7fec28fc50dd2155101269a3dc813c7b354c can be ruled out as the cause, that's good. Ignore my comment about other development being paused.

AprilSylph:
[...]

PR #897, opened 2022-12-06
fast startup experiments

marcustyphoon:
### Description

Committing some attempts to get XKit Rewritten to load more quickly on startup (partially to mitigate the impacts of a potential fix for #636). I don't think these made a perceptible difference, so so far I think this is a failure(?), but I haven't, like, screen recorded or anything yet.

Fundamentally, I don't think you can force a big import waterfall with a bunch of top level awaits to run much earlier. I do manage to run the main waterfall of script imports right after hydration finishes by fetching the list of scripts to run beforehand, but I can't get the API fetches for the user and css map utilities to run that early, so they're still limiting factors.

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -90,6 +90,11 @@
 
     browser.storage.onChanged.addListener(onStorageChanged);
 
+    // load scripts sequentially to avoid chromium load failures
+    // for (const name of installedScripts.filter(name => enabledScripts.includes(name))) {
+    //   await import(getURL(`/scripts/${name}.js`));
+    // }
+
     installedScripts
       .filter(scriptName => enabledScripts.includes(scriptName))
       .forEach(runScript);
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -77,9 +77,40 @@
     return installedScripts;
   };
 
+  const getInstalledUtils = async function () {
+    const url = getURL('/util/_index.json');
+    const file = await fetch(url);
+    const installedUtils = await file.json();
+
+    return installedUtils;
+  };
+
+  const createPreloadLinkElement = (name, directory) =>
+    Object.assign(document.createElement('link'), {
+      href: getURL(`/${directory}/${name}.js`),
+      rel: 'preload',
+      as: 'script',
+      crossOrigin: 'anonymous'
+    });
+
   const init = async function () {
-    const installedScripts = await getInstalledScripts();
-    const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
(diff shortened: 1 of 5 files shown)


marcustyphoon:
Aha. Actually, one can quite easily preload all of the resources before hydration. Unfortunately, this spams my chromium console with hundreds of `The resource <URL> was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.` messages even though according to the performance monitoring and network monitoring tools I definitely am using the preloaded resources (besides the css/json for scripts that don't use them, of course).

Demo:

```js
    const scriptPreloadLinks = installedScripts
      .filter(name => enabledScripts.includes(name))
      .flatMap(name => {
        const scriptPreload = document.createElement('link');
        scriptPreload.href = getURL(`/scripts/${name}.js`);
        scriptPreload.rel = 'preload';
        scriptPreload.as = 'script';
        scriptPreload.crossOrigin = 'anonymous';

        const jsonPreload = document.createElement('link');
        jsonPreload.href = getURL(`/scripts/${name}.json`);
        jsonPreload.rel = 'preload';
        jsonPreload.as = 'fetch';
        jsonPreload.crossOrigin = 'anonymous';

        const stylePreload = document.createElement('link');
        stylePreload.href = getURL(`/scripts/${name}.css`);
        stylePreload.rel = 'preload';
        stylePreload.as = 'style';

        return [scriptPreload, jsonPreload, stylePreload];
      });

    const allUtils = [
      'control_buttons',
      'dom',
      'language_data',
      'modals',
      'post_actions',
      'remixicon',
      'user',
      'crypto',
      'inject',
      'meatballs',
      'mutations',
      'preferences',
      'sidebar',
      'css_map',
      'interface',
      'mega_editor',
      'notifications',
      'react_props',
      'tumblr_helpers'
    ];
    const utilPreloadLinks = allUtils.map(utilname => {
      const preloadLink = document.createElement('link');
      preloadLink.href = getURL(`/util/${utilname}.js`);
      preloadLink.rel = 'preload';
      preloadLink.as = 'script';
      preloadLink.crossOrigin = 'anonymous';
      return preloadLink;
    });

    document.head.append(...scriptPreloadLinks, ...utilPreloadLinks);

    await waitForDocumentReady();
```

marcustyphoon:
Weirdly, preloading all of the installed enabled script files and all of the utility script files as per the above snippet (with or without also preloading the json/css files that cause the console spam) appears to so far have solved #636 for me without any changes to the actual runScript code.

marcustyphoon:
Implementing as #945.

Issue #894, opened 2022-12-04
PostBlock: Block posts instantly (minor)


marcustyphoon:
### Concept

PostBlock should be able to use the storage architecture that Seen Posts uses as of #523 to process posts instantly when soft navigating. This seems a bit ironic, as the comment in that post describing how the implementation ought to work describes it as "PostBlock-like."

marcustyphoon:
Now that I think about it, this could also be done by adding a data attribute to every post with its root id and making a css selector that hides the root ids the user has blocked (the way notificationblock does it).

I don't think it really matters which way one does it. This way would perform slightly better when blocking or unblocking a post, I guess (no need to touch every post with classList.add/remove), but does it matter?

alleycatboy:
> Now that I think about it, this could also be done by adding a data attribute to every post with its root id and making a css selector that hides the root ids the user has blocked (the way notificationblock does it).
> 
> I don't think it really matters which way one does it. This way would perform slightly better when blocking or unblocking a post, I guess (no need to touch every post with classList.add/remove), but does it matter?

I like this idea, I think it could really cut down on how many times the DOM is touched.

alleycatboy:
Also it seems like there's an awful lot of processing going on in checking and adding/removing classes that may or may not already be there for every single post on the dashboard.

marcustyphoon:
Yep. Of course, it presumably makes the CSS engine have to do more work, which has to be done a lot more often then blocking/unblocking a post, which is done, you know, basically never :D But I'm constantly surprised at how unreasonably fast CSS layout engines seem to be.

PR #898, opened 2022-12-06
Enhancement: PostBlock blocks instantly.

alleycatboy:
### Description

Addresses issue #894 reported by @marcustyphoon by using the same caching as Seen Posts to block posts during soft navigation.

### Testing steps

Enable PostBlock script and use as normal.


src/scripts/postblock.js
```diff
@@ -50,14 +52,25 @@ const blockPost = async rootID => {
...
-  registerMeatballItem({ id: meatballButtonId, label: meatballButtonLabel, onclick: onButtonClicked });
+  registerMeatballItem({
+    id: meatballButtonId,
+    label: meatballButtonLabel,
+    onclick: onButtonClicked
+  });
+
+  ({ [storageKey]: blockedPostRootIDs = [] } = await browser.storage.local.get(
+    storageKey
+  ));
```
marcustyphoon, line 73: This obviously should cause no problems in this case, but as a pattern I think it's probably best to await setup functions before registering effects, so that none of the code can run with uninitialized variables.

src/scripts/postblock.js
```diff
@@ -9,22 +9,24 @@ const meatballButtonId = 'postblock';
...
-      postElement.classList.remove(hiddenClass);
-    }
-  });
+      if (blockedPostRootIDs.includes(rootID)) {
+        postElement.classList.add(hiddenClass);
+      } else {
+        postElement.classList.remove(hiddenClass);
+      }
+    });
+  }
```
marcustyphoon, line 29: I see you switched from async-await to .then() - any particular reason?

alleycatboy: After a some speed testing I've found that as the user accumulates more blocks and scrolls the dashboard, the time to block each post sequentially with the thread blocking `await` increases rapidly, mainly due to the response time of `timeline` object.

Using `then` splits off the scope until the response is ready, and reduces the time to block on average by 1ms and the reduction increases as the user scrolls further blocks more.

The real gains come when unblocking the posts. Using `then` removes them in constant time, about 4.11ms every single time.  Using `await` begins at 9.79ms and increases by about 0.2ms per per post. 

And that's just with not too extensive dash scrolling, the numbers keep going up as I go further.

marcustyphoon: ...Hm. Interesting. Was your test with async-await still using the await inside the forEach, or was that including your change to for-of?

I'll have to go try this and verify if I'm correct, but my understanding is that:

```js
// awaits simultaneously
const exampleOne = (posts) => {
  posts.forEach(async (post) => {
    before();
    await someFunction(post);
    after();
  });
};

// awaits simultaneously; identical behavior to first example unless you transpile earlier than es2017
const exampleTwo = (posts) => {
  for (const post of posts) {
    before();
    someFunction(post).then(() => after());
  }
};

// awaits sequentially; slower then the other two ways if someFunction resolves in a new task
const exampleThree = async (posts) => {
  for (const post of posts) {
    before();
    await someFunction(post);
    after();
  }
};
```

`await` doesn't "block" the thread per se; it's essentially syntactic sugar for `then()` but it preserves function scope/context.

Something else must be up, though; even if you were testing the "third" way here with a sequential await in a loop, our `timelineObject` implementation resolves in the same event loop task, so I wouldn't expect it to take a measurably different amount of time in that case. That's really interesting. Are you testing in Firefox or a Chromium-based browser?

marcustyphoon: Hm, those numbers look weird to me. What did you measure? I'm observing that step taking like 1ms, so I can't produce a multiple-millisecond time difference (Chromium). Maybe Firefox behaves differently; I've seen optimizations go the other way on it before.

> When the results aren't used, splitting off can provide a "more asynchronous" experience.

Not exactly sure what you mean by this! If the results aren't used at all and you don't care about sequencing, you can use neither await nor then, correct? 

> In either case, do you think it would be overkill to catch the potential exceptions thrown by timelineObject? The function throws up the exceptions in inject, but it seems like that isn't a realistic concern. Unless it's already being handled?

I don't think we generally explicitly handle `timelineObject` exceptions unless there's a specific action we' like to take in that case besides silently failing. I do try to keep in mind how large the effect of an individual error would be, of course (to avoid things like #874).

If we wanted those failures not to be silent, the best way IMO would be to attach error and unhandledrejection listeners to `window`, as that would only have to be done once (i.e. #472).

marcustyphoon: Both! Strange. `timelineObject` is cached, so it should be basically instant as long as you've already seen the posts (inside a promise, of course, so if you time a `then` or `await` you'll still be measuring the duration of everything already in the microtask queue when you make the call; [one can have a fair amount of fun with this](https://gist.github.com/marcustyphoon/b8e08636501c478f05efe695d4badcbd)).

You said you were testing blocking and unblocking, which would mean the task doing this would have been triggered by the storage mutation callback running onStorageChanged, which... shouldn't have any other asynchronous code in it, so the microtask queue should be empty. Oh well, no idea why it would be different!

> Thoughts on using a `Promise.all(filterPostElements(postElement , {...}).map(async postElement => {`?

This is equivalent to `filterPostElements(postElement , {...}).forEach(async postElement => {` if you don't have any code you need to run after all of the callbacks resolve, right? That would just discard the callback's return values instead of collecting them.

alleycatboy: > This is equivalent to `filterPostElements(postElement , {...}).forEach(async postElement => {` if you don't have any code you need to run after all of the callbacks resolve, right?

True! Let's stick with the `forEach` instead of the `for...of` as well!

alleycatboy:
Now using CSS selectors instead of directly adding/removing classes as suggested [here](https://github.com/AprilSylph/XKit-Rewritten/pull/898/commits/8ebcf10ac27cb382f0b0ee84fac3c1353930e13b).

marcustyphoon (changes requested):
The GitHub review UI apparently doesn't let me make a code suggestion when there's a CI warning in the "files changed" interface, but as per that warning, you can combine the two imports from the same file.

src/scripts/postblock.js
```diff
@@ -2,30 +2,29 @@ import { filterPostElements } from '../util/interface.js';
...
-    if (blockedPostRootIDs.includes(rootID)) {
-      postElement.classList.add(hiddenClass);
-    } else {
-      postElement.classList.remove(hiddenClass);
+const processPosts = (postElements) =>
+  filterPostElements(postElements, { includeFiltered: true }).forEach(
+    async (postElement) => {
+      const { rebloggedRootId } = await timelineObject(postElement);
+      const rootID = rebloggedRootId || postElement.dataset.id;
+      postElement.dataset.targetRootId = rootID;
```
marcustyphoon, line 25: These need to be removed in `clean()`!

src/scripts/postblock.js
```diff
@@ -2,30 +2,29 @@ import { filterPostElements } from '../util/interface.js';
...
-
-  filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
-    const postID = postElement.dataset.id;
-    const { rebloggedRootId } = await timelineObject(postElement);
-
-    const rootID = rebloggedRootId || postID;
+const styleElement = buildStyle();
+const buildCss = () =>
+  `:is(${blockedPostRootIDs
+    .map((rootId) => `[data-target-root-id="${rootId}"]`)
```
marcustyphoon, line 17: Here we should target the `article` elements within the post elements, not the post elements themselves, to avoid breaking J/K scrolling.

marcustyphoon (commented):
`postblock.css` can be removed as part of this, as it is no longer used! Addtionally:

PR #900, opened 2022-12-08
Quote Replies: Show all buttons when using touchscreen

### LINKED ISSUE #895: Quote Replies: Show all buttons when using touchscreen
### Concept

Right now, quote replies is presumably pretty annoying to use when using a touchscreen.

The way I would probably solve this is by adding an event listener that adds an e.g. xkit-has-touchscreen class to the document body on the first touch event, and overriding the only-on-hover css based on that. (Maybe this could be removed on a long debounce?)

marcustyphoon:
### Description

This shows the Quote Replies button whether or not the user is hovering over a notification if they're using a device without the ability to hover or if they have made use of the touchscreen since the extension was turned on.

I didn't worry about the fact that Tumblr's block buttons don't have this behavior, but still take up some space, so the spacing does look a bit odd.

I did try this on a Surface device without a keyboard attached, and the media query works in that case, so I think one could make a reasonable argument that the media query alone is sufficient and the on-touch behavior change is unnecessary, and users with touchscreen laptops with trackpads can just use the trackpads. I could go either way on this.

Resolves #895

### Testing steps

- Enable Quote Replies and navigate to the notifications page on a touchscreen-only mobile device. Quote Replies buttons should appear on replies.
- ~~Enable Quote Replies and navigate to the notifications page on a laptop with a touchscreen. Use the touchscreen to interact with the page. Quote Replies buttons should appear on replies.~~ functionality removed


AprilSylph (changes requested):
Better to assume the feature's absence and then test for its existence rather than vice-versa

```diff
  button.xkit-quote-replies svg {
    width: 21.5px;
    height: 21.5px;

    fill: rgb(var(--blue));
-   opacity: 0;
-   transform: scale(0);
    transition: all .25s ease-out .4s;
  }

+ @media (hover: hover) {
+   button.xkit-quote-replies svg {
+     opacity: 0;
+     transform: scale(0);
+   }
+
+   :is(:hover, :focus-within) > button.xkit-quote-replies svg {
+     opacity: 1;
+     transform: scale(1);
+   }
+ }
```

src/scripts/quote_replies.css
```diff
@@ -25,7 +23,19 @@ button.xkit-quote-replies:disabled svg {
...
+  :is(:hover, :focus-within)>button.xkit-quote-replies svg {
+    opacity: 1;
+    transform: scale(1);
+  }
+}
+
+body.xkit-quote-replies-touch button.xkit-quote-replies svg {
   opacity: 1;
   transform: scale(1);
 }
```
AprilSylph, line 41:
```suggestion
```

src/scripts/quote_replies.js
```diff
@@ -9,6 +9,7 @@ import { apiFetch } from '../util/tumblr_helpers.js';
 
 const storageKey = 'quote_replies.currentResponseId';
 const buttonClass = 'xkit-quote-replies';
+const touchClass = 'xkit-quote-replies-touch';
```
AprilSylph, line 12:
```suggestion
```

src/scripts/quote_replies.js
```diff
@@ -101,6 +102,8 @@ const quoteReply = async ({ id, summary, name, uuid, timestamp }) => {
 
 const showError = exception => notify(exception.body?.errors?.[0]?.detail || exception.message);
 
+const addTouchClass = () => document.body.classList.add(touchClass);
```
AprilSylph, line 105:
```suggestion
```

src/scripts/quote_replies.js
```diff
@@ -114,11 +117,15 @@ export const main = async function () {
   if (responseId !== undefined && /^\/blog\/.+\/drafts/.test(location.pathname)) {
     document.querySelector(`[href*="/edit/"][href$="/${responseId}"]`)?.click();
   }
+
+  document.body.addEventListener('touchstart', addTouchClass, { once: true });
```
AprilSylph, line 121:
```suggestion
```

src/scripts/quote_replies.js
```diff
@@ -114,11 +117,15 @@ export const main = async function () {
...
 };
 
 export const clean = async function () {
   pageModifications.unregister(processNotifications);
+  document.body.removeEventListener('touchstart', addTouchClass);
```
AprilSylph, line 126:
```suggestion
```

src/scripts/quote_replies.js
```diff
@@ -114,11 +117,15 @@ export const main = async function () {
...
   $(`.${buttonClass}`).remove();
+  document.body.classList.remove(touchClass);
```
AprilSylph, line 128:
```suggestion
```

AprilSylph (commented):
I can see what you're trying to do with the event listener, but... it's pretty safe to assume that if a trackpad is being recognised as the primary pointing device by the browser, that it's probably being used as such. Do you have a hybrid device that you're testing this with? Is that assumption flawed?

src/scripts/quote_replies.css
```diff
@@ -25,7 +23,19 @@ button.xkit-quote-replies:disabled svg {
...
 
-:is(:hover, :focus-within) > button.xkit-quote-replies svg {
+@media (hover: hover) {
+  button.xkit-quote-replies svg {
+    opacity: 0;
+    transform: scale(0);
+  }
+
+  :is(:hover, :focus-within)>button.xkit-quote-replies svg {
```
AprilSylph, line 32:
```suggestion
  :is(:hover, :focus-within) > button.xkit-quote-replies svg {
```

marcustyphoon:
>  I think one could make a reasonable argument that the media query alone is sufficient and the on-touch behavior change is unnecessary, and users with touchscreen laptops with trackpads can just use the trackpads. I could go either way on this.

That's what I was trying to describe here! The touch event listener only serves to facilitate someone who has a traditional laptop with a touchscreen but prefers to use the touchscreen over the trackpad, which seems niche. (I don't know if this would apply to "iPad with a magic keyboard attached," but we don't build for Safari anyway.)

marcustyphoon:
This appears not to work in Kiwi Browser 107 on my phone even though it does in Firefox Nightly. I assume that's probably a Kiwi Browser bug but it's a huge pain to investigate on my 2016-era device.

PR #901, opened 2022-12-08
Quote Replies: Enable quoting mentions

### LINKED ISSUE #861: Quote Replies: Enable quoting mentions
### Concept

Right now quote replies functions on notifications about a reply on one of your posts, but not on notifications about replies on other posts that @ mention you (`note_mention`). This would be convenient!

marcustyphoon:
Now that I've thought about it more, I guess I can see how this is sort of weird to do in practice in some cases? I dunno. 

### Description

Enables Quote Replies on notes about replies that mention you on posts which are not yours, in addition to its current function that works on replies to your own posts whether they mention you or not.

This does not enable the script on notes about posts in which you're mentioned in the actual post (those, you can just reblog).

Resolves #861.

### Testing steps

- Find a note about a mention (purple at sign). Use Quote Replies on it.
- Find a note about a mention on a sideblog. Use Quote Replies on it.


src/scripts/quote_replies.js
```diff
@@ -89,10 +93,20 @@ const quoteReply = async ({ id, summary, name, uuid, timestamp }) => {
...
+    format => format.type === 'mention' && userBlogNames.includes(format.blog?.name)
+  )?.blog;
+  if (type === 'note_mention' && !yourMentionedBlog) {
+    throw new Error('Reply not found.');
+  }
+
+  const yourUuid = isReply ? uuid : yourMentionedBlog.uuid;
+  const yourName = isReply ? name : yourMentionedBlog.name;
+
+  const { response: { id: responseId, displayText } } = await apiFetch(`/v2/blog/${yourUuid}/posts`, { method: 'POST', body: { content, state: 'draft', tags } });
```
AprilSylph, line 106: I don't think these changes are necessary. Quote Replies should always create drafts on the blog you're viewing activity for, but this logic looks for the first mention of any of your blogs, and then changes the target accordingly. Why?

If a reply mentions both my main blog and my sideblog, I don't care about what order those mentions are in; if I'm viewing my main blog's activity and quote that reply, I expect to be acting as my main blog. Likewise, if I am viewing my sideblog's activity, I expect to be acting as my sideblog.

AprilSylph (changes requested):

src/scripts/quote_replies.js
AprilSylph: We already check if the reply's timestamp matches the activity item's timestamp for handling deleted replies up on L76, too.

marcustyphoon: If I recall correctly, in this code, in this location, if you are quote replying a mention, `uuid` isn't your blog's: it's the blog that posted the post which the reply was on, which you may not even own.

AprilSylph: Ah, that's annoying.

marcustyphoon: Yeah, I think that's it. Want me to rename it to e.g. `postAuthorUuid` for clarity?

AprilSylph: I think I'd prefer removing the prop renames from L38-41

marcustyphoon: Oh, yep, much better.

src/scripts/quote_replies.js
```diff
@@ -89,10 +87,20 @@ const quoteReply = async ({ id, summary, name, uuid, timestamp }) => {
     ...tagReplyingBlog ? [reply.blog.name] : []
   ].join(',');
 
-  const { response: { id: responseId, displayText } } = await apiFetch(`/v2/blog/${uuid}/posts`, { method: 'POST', body: { content, state: 'draft', tags } });
+  const yourMentionedBlog = reply.content[0]?.formatting?.find(
+    format => format.type === 'mention' && userBlogNames.includes(format.blog?.name)
+  )?.blog;
```
AprilSylph, line 92: I still would like to account for the possibility of multiple mentioned blogs belonging to the same account. If we can't do it via the notification's props, surely we could go higher in the React tree to extract the endpoint it was fetched with?

AprilSylph (changes requested):

src/scripts/quote_replies.js
marcustyphoon: Oh, activity pages are per-blog. Right—no idea why I thought there was just the one. Will check on that.

marcustyphoon: Well, that was substantially less convoluted.

AprilSylph (approved):

PR #910, opened 2022-12-12
Meatballs util: Enable meatball items in blog menus

marcustyphoon:
### Description

This creates `registerBlogMeatballItem`/ `unregisterBlogMeatballItem`, allowing scripts to add meatball items to the meatball menus in blog cards and the blog view header.

- [x] Update this if #908 is merged.

### Testing steps

Unrevert the demo script commit. Use the meatball menus in a blog card, in the modal blog view header, and in the non-modal blog view header and confirm that they work and adapt to the name of the selected blog.


marcustyphoon:
Worth noting, I guess, that I didn't put much thought into this API; perhaps it's better to add configuration to the register and unregister functions so scripts can register once for both post and user meatball menus if the menu items they're registering are substantially similar? Thinking about it now, I think one might want to make the labels different in that case, though. So actually, eh, this seems fine.

src/util/meatballs.js
```diff
@@ -88,6 +118,51 @@ const addMeatballItems = meatballMenus => meatballMenus.forEach(async meatballMe
...
-});
+};
+
+const addBlogMeatballItem = async meatballMenu => {
+  const __blogData = await blogData(meatballMenu);
+
+  $(meatballMenu).children('[data-xkit-meatball-button]').remove();
+
+  Object.keys(blogMeatballItems).sort().forEach(id => {
+    const { label, onclick, postFilter } = blogMeatballItems[id];
```
AprilSylph, line 129:
```suggestion
    const { label, onClick, blogFilter } = blogMeatballItems[id];
```

src/util/meatballs.js
```diff
@@ -88,6 +118,51 @@ const addMeatballItems = meatballMenus => meatballMenus.forEach(async meatballMe
...
+
+    const meatballItemButton = dom('button', {
+      'data-xkit-blog-meatball-button': id,
+      hidden: true
+    }, {
+      click: onclick
```
AprilSylph, line 135:
```suggestion
      click: onClick
```

src/util/meatballs.js
```diff
@@ -88,6 +118,51 @@ const addMeatballItems = meatballMenus => meatballMenus.forEach(async meatballMe
...
+        labelResult.then(result => { meatballItemButton.textContent = result; });
+      } else {
+        meatballItemButton.textContent = labelResult;
+      }
+    } else {
+      meatballItemButton.textContent = label;
+    }
+
+    if (postFilter instanceof Function) {
+      const shouldShowItem = postFilter(__blogData);
```
AprilSylph, line 154:
```suggestion
    if (blogFilter instanceof Function) {
      const shouldShowItem = blogFilter(__blogData);
```

src/util/meatballs.js
```diff
@@ -42,10 +44,38 @@ export const unregisterMeatballItem = id => {
...
+ * Add a custom button to blogs' meatball menus in blog cards and the blog view header.
+ *
+ * @param {object} options - Destructured
+ * @param {string} options.id - Identifier for this button (must be unique)
+ * @param {string|Function} options.label - Button text to display. May be a function accepting the blog data of the post element being actioned on.
+ * @param {Function} options.onclick - Button click listener function
+ * @param {Function} [options.postFilter] - Filter function, called with the blog data of the menu element being actioned on. Must return true for button to be added. Some blog data fields, such as "followed," are not available in blog cards.
+ */
+export const registerBlogMeatballItem = function ({ id, label, onclick, postFilter }) {
+  blogMeatballItems[id] = { label, onclick, postFilter };
```
AprilSylph, line 57: 1. I've regretted the casing of `onclick` for a while now
2. `postFilter` doesn't make sense if we're not filtering posts :P

```suggestion
 * @param {Function} options.onClick - Button click listener function
 * @param {Function} [options.blogFilter] - Filter function, called with the blog data of the menu element being actioned on. Must return true for button to be added. Some blog data fields, such as "followed", are not available in blog cards.
 */
export const registerBlogMeatballItem = function ({ id, label, onClick, blogFilter }) {
  blogMeatballItems[id] = { label, onClick, blogFilter };
```

(I thought about suggesting `postFilter` &rarr; `filter` but I probably had a reason for not going for `filter` for post meatball items, maybe to avoid confusion with `Array.prototype.filter`?)

AprilSylph (changes requested):
didn't test the demo but code looks good, though I have some nitpicks about variable naming. The suggestions are backwards because I fumbled the selected lines for the first one and didn't notice until I'd done the others, whoops.

There's a high likelihood that I missed some usages anyway.

src/util/meatballs.js
marcustyphoon: I'll merge then take a look!

marcustyphoon: Ah, I see; changing to `onClick` in the current meatballs utils would require a big diff.

AprilSylph (approved):

PR #916, opened 2022-12-15
Painter: Add legacy thread option

### LINKED ISSUE #896: Indicate legacy/new post type in meatballs menu
### Concept

I couldn't think of a good way to do this as an always-visible UI element ~~(both for UI and performance reasons)~~, but it would be great to provide people with a way to see whether a thread is legacy or new during the legacy post editor transition, and putting it in the meatballs menu solves both problems nicely.

a) We don't currently have a way to customize the text shown in a meatball menu item, particularly if the function to determine what text needs to be shown is asynchronous (which it would be here; we want the ~~root post's`shouldOpenInLegacy`~~ `isBlocksPostFormat` property, I believe, ~~which requires an API call~~). I could make `options.dynamicLabel` or whatever easily enough though.

b) Would this go in Trim Reblogs, a tweak... or just happen as part of having XKit Rewritten installed? We have a bunch of scripts that don't function on legacy posts, not just Trim Reblogs, so in some sense the semantic answer to me is the last one, but that's obviously inconsistent with many principles of this extension.

marcustyphoon:
### Description

Eh. I dunno.

This is the easiest way I could think of to let heavy editable/trim reblogs users see what threads are still legacy at a glance/in advance. Not particularly happy with it since the information is very trim reblogs specific (it's not like XKit Rewritten doesn't work on legacy threads at all) so putting it randomly in an unrelated extension sucks for both of them. This is reviewable if you think it's a good idea though.

I also tried some other things; see #967.

Resolves #896 (well, is one way to do so).


marcustyphoon:
Finding this useful for knowing when quick tags will fail.

AprilSylph:
I think this is unnecessary following the merge of #969!

marcustyphoon:
I don't think of the Quick Tags case as super relevant; that was IMO a straight-up bug, as we can easily handle both thread types and so the user shouldn't need to know the difference in that case.

I judge the necessary-ness of this entirely on the existence of #767 and the removal of the legacy post editor killing Editable Reblogs. Is that removal complete now? If so, the Trim Reblogs user has a way to get that information and the Editable Reblogs user doesn't exist, and exactly what one needs to explain to a member of the RP community is no longer situational if they don't have a choice of editor in the first place.

AprilSylph:
The legacy editor is now only available for editing existing legacy posts, so without weird workarounds like using XKit 7 Quick Tags to turn an NPF-compatible reblog into an NPF-incompatible one, Editable Reblogs is dead.

Issue #899, opened 2022-12-07
Quick Reblog: Ctrl + Enter to post


alleycatboy:
### Concept

When using Quick Reblog I often find it inconvenient to have to reach for my mouse after quickly typing in tags or text to reblog. Adding Ctrl + Enter as a hotkey for posting (possibly also other hotkeys for queuing and drafting) would remove a lot of friction when posting.

alleycatboy:
Actually, tabbing around works fine 😅

Closing this issue.

marcustyphoon:
I dunno, I kind of want this!

PR #917, opened 2022-12-16
 Quick Reblog: Add keyboard navigation

alleycatboy:
### Description

This PR addresses #899

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -9,6 +9,22 @@ import { notify } from '../util/notifications.js';
 import { translate } from '../util/language_data.js';
 import { dom } from '../util/dom.js';
 
+const inputEvents = (event) => {
+  event.stopPropagation();
+
+  if (enableKeyNav) {
+    if (event.key === 'Enter' && event.ctrlKey) {
+      reblogButton.click();
+    }
+
+    if (event.key === 'Escape') {
+      const { parentNode } = popupElement;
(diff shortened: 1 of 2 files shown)


src/scripts/quick_reblog.js
```diff
@@ -9,14 +9,30 @@ import { notify } from '../util/notifications.js';
 import { translate } from '../util/language_data.js';
 import { dom } from '../util/dom.js';
 
+const inputEvents = (event) => {
+  event.stopPropagation();
+
+  if (enableKeyNav) {
+    if (event.key === 'Enter' && event.ctrlKey) {
```
marcustyphoon, line 16: For consistency on Macs, this should probably include `metaKey` as an alternative!

AprilSylph, line 16:
```suggestion
    if (event.key === 'Enter' && (event.ctrlKey || event.metaKey)) {
```

src/scripts/quick_reblog.json
```diff
@@ -71,6 +71,11 @@
         { "value": "10000", "label": "10,000 posts" }
       ],
       "default": "100"
+    },
+    "enableKeyNav": {
+      "type": "checkbox",
+      "label": "Enable keyboard navigation",
+      "default": true
```
AprilSylph, line 78: I don't see the need to make this a preference. Let's just build it in.

AprilSylph (changes requested):

PR #918, opened 2022-12-16
Tweaks: Remove show full notes tweak

marcustyphoon:
### Description

Since the full note count is now shown the vast majority of the time (only collapsing when there's not enough space e.g. in masonry views), this ~~renames~~ removes the show all notes tweak.

Needs wiki update.


AprilSylph:
Removing it is probably better.

AprilSylph (approved):

PR #920, opened 2022-12-18
demo: custom events for synchronous(ish) sandbox messaging

marcustyphoon:
> MV2 could add a script element with code in textContent and it would run synchronously, which is particularly useful for extracting or modifying values of JS variables in the same event loop task via **the synchronous DOM events like `CustomEvent`** or the upcoming new API to communicate between the main world and the content scripts bug 1054624
> - https://bugs.chromium.org/p/chromium/issues/detail?id=1207006

Saw this and my eyes bugged out of my head.

### Description

This demonstrates the use of custom events in the DOM to pass data between the web extension sandbox/"isolated world" and the page context/"main world."

This serves the same purpose as our current method of using a mutation observer to watch for a DOM manipulation. As far as I can tell, it has no major advantages/disadvantages versus that method, but it's interesting (and feels IMO more semantically correct than momentarily serializing callback results into a data attribute.).

One advantage that it does appear to have is that "detail" data attached to a custom event does not need to be serializable to JSON; I put a `Map()` in there and it worked fine. (~~Haven't played around with this much; if you can put DOM elements in there that would be very interesting indeed.~~ As expected, it's StructuredClone; circular references work but DOM elements don't.)

One drawback that's not relevant in the current codebase is that our current code allows a fully synchronous code path for injected functions that are synchronous. We don't make meaningful use of this at the moment anyways since `inject` is defined as an async function and scripts must treat it as asynchronous regardless.

(also, it might be a fraction of a millisecond slower for some reason, but I'm not going to put a lot of work into trying to measure that for sure - seems fine to me)

Both the old and new code will resolve a synchronous injected function within the same event loop invocation (i.e. before the page can render), preserving "instant" page modification functionality.

### DIFF
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -15,45 +15,41 @@ export const inject = async (func, args = [], target = document.documentElement)
   const async = func instanceof AsyncFunction;
 
   const script = dom('script', { nonce }, null, [`{
-    const { dataset } = document.currentScript;
+    const scriptElement = document.currentScript;
     const ${name} = ${func.toString()};
     const returnValue = ${name}(...${JSON.stringify(args)});
     ${async
       ? `returnValue
-          .then(result => { dataset.result = JSON.stringify(result || null); })
-          .catch(exception => { dataset.exception = JSON.stringify({
-            message: exception.message,
-            name: exception.name,
-            stack: exception.stack,
-            ...exception
-          })})
+          .then(result => scriptElement.dispatchEvent(
+            new CustomEvent('xkitinjection', { detail: { result } })
+          ))
+          .catch(exception => {
+            const e = {
+              message: exception.message,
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Aw. I was trying to troubleshoot why a 403'ing mass post editor endpoint request didn't return anything from `inject()` when testing #711's error handling, which of course is because the `fetch` returns a non-serializable `Response` object. Got excited that this would solve that cleanly...

...but `Response` objects aren't allowed to pass through here either. Meh.

marcustyphoon:
- [ ] fix a `Response` object crashing the callback handler (`detail` becomes null if its contents are not serializable).

marcustyphoon:
Closing to reduce clutter; I can always find this easily from its linked PR.

PR #921, opened 2022-12-18
break glass in case of emergency

marcustyphoon:
### Description

There may be a better way to do this, but it does work.

This moves all of the injected function bodies into another content script and injects that content script into the page context using the `world` flag in Chromium's dynamic content script feature (current Firefox does not appear to have this). The sandboxed content script containing all of XKit Rewritten's actual code communicates with the unsandboxed one via DOM `CustomEvent`s (see #920) in both directions, requesting an injected call and receiving a response.

I wasn't sure how to pass the target element (if any) to the injected functions, since `document.currentScript` doesn't work here; at the moment I just appended it to the argument list, which is kind of awkward.

Oh, this also bumps the `manifest_version` key in the manifest file. Not sure what that does.

hypothetical todo: ensure that the unsandboxed content script has finished loading before proceeding in main.js


AprilSylph:
we can still inject scripts from content scripts as normal, the injected scripts just need to be their own file (and referenced by web accessible URL). something something content security policy

AprilSylph:
https://gist.github.com/AprilSylph/b84392939f92a5df4e4743d026ca3074

marcustyphoon:
I saw that method mentioned in another thread, yeah! Seems viable. I'm not sure if it's doable while keeping instant processing (didn't try) and I couldn't think of a way not to duplicate the boilerplate "send back the result" code in every file, but I didn't play with it at all yet.

marcustyphoon:
- [ ] fix a `Response` object crashing the callback handler on destructure attempt (`detail` becomes null if its contents are not serializable). See comment on #920.

marcustyphoon:

PR #923, opened 2022-12-21
Mutual Checker: Fix misaligned reblog icons in mobile view

### LINKED ISSUE #919: Mutual Checker: Icons inserted incorrectly in mobile view
### Platform

MacOS 10.15.7 x86

### Browser

Ungoogled Chromium v108.0.5359.98

### Addon version

master

### Details

When posts from mutuals are loaded onto the timeline by scrolling down via infinite scrolling, their mutual icons are appearing in the same unintended way that they do when the mode is switched as a side effect of 772bfe5930d2c2f335759c18a8c8abc9a939cdd1. I assume the detection code there is running during a time window when redpop has not finished updating the DOM to the mobile styles or something.

marcustyphoon:
### Description

This fixes mutual checker icons looking weird in the mobile/tablet view by simply reverting the previous fix, which no longer appears necessary.

Resolves #919

### Testing steps

Load posts by a mutual with a narrow viewport. Mutual icons should not appear as they do in the screenshot in the linked issue.
Repeat when loading posts with infinite scrolling, when loading the page with a wide viewport and resizing to narrow, in the blog view modal, in the blog view non-modal, etc.

### DIFF
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -59,9 +59,7 @@ const addIcons = function (postElements) {
     const isMutual = await mutuals[blogName];
     if (isMutual) {
       postElement.classList.add(mutualsClass);
-      getComputedStyle(postAttribution).getPropertyValue('display') === 'flex'
-        ? postAttribution.prepend(icon.cloneNode(true))
-        : postAttribution.before(icon.cloneNode(true));
+      postAttribution.prepend(icon.cloneNode(true));
     } else if (showOnlyMutuals) {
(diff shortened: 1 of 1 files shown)


alleycatboy (commented):
Tested and working perfectly!

I think the reason April's hotfix suddenly stopped working is that Staff changed the `postAttribution` container to always be an `inline-block` instead of a `flex` on mobile. 

I wonder, did Staff have it as an `inline-block` before that hotfix was needed?

Thanks for the fix, it was starting to drive me nuts!

AprilSylph (approved):

Issue #867, opened 2022-11-25
Bigger quick tags window


Starsynal:
### Concept

Hello. I use quick tags a lot and for a lotta different things which has to lead to me having to scroll a lot when quick reblogging. Could you add an option for custom set height to the quick tags box that opens when using quick reblog? https://i.gyazo.com/5430b77af97e1a6f84009a1c643a8f8b.png This is a pic of what I'm talking about in case wording is bad.

AprilSylph:
- see https://github.com/AprilSylph/XKit-Rewritten/pull/926#issuecomment-3502976206

PR #926, opened 2022-12-30
Quick Reblog: Add dialog height option for Quick Tags Integration.

alleycatboy:
### Description

This adds an option in Quick Reblog's preferences to increase or decrease the height of the Quick Tags Integration dialog. 

This addresses the issue raised in #867.

It parses the the preference from Quick Reblog and sets it as a CSS variable in `quickTagsList` component. This makes it available on every subsequent render of the Quick Reblog popup dialog without having to check every single time. 

The variable changes the height of the dialog from 1 to 20 units, each unit being the height of the button which is the same height as each tag element.

Default (4):
default size

Size 10:
size 10

Size 1:
size 1

### Testing steps

 
To test, go to the Quick Reblog preferences and set a height different than the default of 4. The change should be reflected immediately.

### DIFF
--- a/src/scripts/quick_reblog.js
+++ b/src/scripts/quick_reblog.js
@@ -48,6 +48,7 @@ let showBlogSelector;
 let rememberLastBlog;
 let showCommentInput;
 let quickTagsIntegration;
+let quickTagsIntegrationSize;
 let showTagsInput;
 let showTagSuggestions;
 let reblogTag;
@@ -274,6 +275,21 @@ const renderQuickTags = async function () {
   });
 };
 
+const parseQuickTagsSize = (size) => {
+  if (!Number.isInteger(size)) {
+    return 4;
+  }
+
+  return Math.min(20, Math.max(1, size));
+};
+
+const setQuickTagsSize = (size) => {
+  quickTagsList.style.setProperty(
+    '--quick-tags-size',
+    parseQuickTagsSize(size)
+  );
+};
+
 const updateQuickTags = (changes, areaName) => {
   if (areaName === 'local' && Object.keys(changes).includes(quickTagsStorageKey)) {
     renderQuickTags();
@@ -298,6 +314,7 @@ export const main = async function () {
     rememberLastBlog,
     showCommentInput,
     quickTagsIntegration,
+    quickTagsIntegrationSize,
     showTagsInput,
     showTagSuggestions,
     reblogTag,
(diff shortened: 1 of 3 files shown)


src/scripts/quick_reblog.js
```diff
@@ -340,6 +357,7 @@ export const main = async function () {
   $(document.body).on('mouseenter', reblogButtonSelector, showPopupOnHover);
 
   if (quickTagsIntegration) {
+    setQuickTagsSize(Number.parseInt(quickTagsIntegrationSize));
```
marcustyphoon, line 360: I would probably move this parseInt call into the rest of the parsing logic, since it seems related.

src/scripts/quick_reblog.js
```diff
@@ -274,6 +275,21 @@ const renderQuickTags = async function () {
   });
 };
 
+const parseQuickTagsSize = (size) => {
+  if (!Number.isInteger(size)) {
+    return 4;
+  }
```
AprilSylph, line 281:
```suggestion
const parseQuickTagsSize = (sizeString) => {
  const size = Number.parseInt(sizeString);
  if (Number.isNaN(size)) return 4;
```

src/scripts/quick_reblog.js
AprilSylph:
```suggestion
    setQuickTagsSize(quickTagsIntegrationSize);
```

AprilSylph (changes requested):

skircheis:
This is lovely, but would it be possible to also provide an "expand to fit" option? Say, if the config variable is set to `0`, it acts as if set to `length(tagBundles)`.

marcustyphoon:
That's pretty much equivalent to removing the maximum value from this PR and the user setting an extremely high number, which may be simpler.

skircheis:
Would that also shrink the list if `length(tagBundles) < N`? Expand to fit, but no more. If so, sure.

marcustyphoon:
Yep! That's what it does currently.

skircheis:
Oh, I have had so many bundles for so long I don't recall ever seeing that. Then ignore me being silly, if there's no limit and I can just set it to 813 or whatever that's great.

AprilSylph:
I was going to take this over given alleycatboy's absence from GitHub, but:
1. Seems that reviewer edits are not allowed on this PR's branch, making it impossible for me to push to; this could be worked around by merging this into a different dev branch, but also...
2. The enhancement idea kinda conflicts with the idea I've got for solving #1611—I want to essentially use a copy of the Quick Tags buttons as a better way to suggest tags from the post, if that option is enabled. That already adds a bunch of height, and then there will need to be section headings to separate the two lists as well.

So, I'm instead going to close this PR, and its linked issue as "not planned".

PR #934, opened 2023-01-14
Fix Chrome script loading inconsistency

### LINKED ISSUE #636: Chrome: Scripts load inconsistently on very-fresh tabs(?)

marcustyphoon:
### Description

At the expense of initial extension load speed, this ensures that XKit will not silently fail to load some scripts on slow computers in certain scenarios in chromium-based browsers; see the linked issue for details.

It does so by importing script files one at a time on their initial import instead of all at once, which seems to prevent some problem with large import waterfalls. (Subsequent imports do not trigger this.)

Functionality-wise, this PR is a strictly worse version of #897, which fixes the issue without a load speed drawback using link preloading. That code is, however, significantly more complicated to understand, both in a PR review sense and a "having the code make sense when a future contributor reads it" sense.

Resolves #636.

### Testing steps

I don't expect anyone to be able to consistently replicate the problem this solves, as I didn't find a great method, so testing is probably confined to ensuring that XKit continues to function correctly.

To test the actual fix, merge #756 into master on a slow PC, enable ~all of the scripts, and try to observe a failure. Then, merge #756 into this PR with the same scenario and observe that there are no failures.

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -85,6 +85,10 @@
     const installedScripts = await getInstalledScripts();
     const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
+    for (const scriptName of installedScripts.filter(scriptName => enabledScripts.includes(scriptName))) {
+      await import(getURL(`/scripts/${scriptName}.js`));
+    }
+
     installedScripts
       .filter(scriptName => enabledScripts.includes(scriptName))
       .forEach(runScript);
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -85,6 +85,7 @@
     const installedScripts = await getInstalledScripts();
     const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Superseded by #945.

PR #936, opened 2023-01-20
New Script: XNeko

marcustyphoon:
Is this a good use of two and a half hours? No.

### Description

Ports XNeko from XKit 7 without changing any of its behavior. You know... probably.

Yes, yes, this may not count as a serious extension, but a lot of people in the tags seem to enjoy it and will be sad if the MV3 requirement ever hits before the heat death of the universe.


marcustyphoon:
There is at least one bug with the state machine (`stopped = false` seems like a hack to fix it) that prevents a kitty yawn from ever appearing, but you know, someone else can fix that if they want to.

marcustyphoon:

PR #945, opened 2023-01-29
Improve loading speed via script file preload

marcustyphoon:
### Description

(rewritten to take further testing into account)

Right now, XKit Rewritten's initial load on a fresh page begins after redpop's hydration finishes and comprises a number of sequential asynchronous tasks. This improves the loading speed of the extension by starting those tasks earlier when they don't depend on the redpop code having executed, reducing the time in which the loaded Tumblr page can be seen without XKit's modifications when the page is navigated to, hard-navigated, or refreshed.

 - The list of installed extensions must be determined by fetching the script index file and querying extension storage; these depend on nothing, so they can be fired at `document_start`.
 - Script and utility files must then be fetched; to avoid running side effects too early and creating a deep nested import waterfall, they are preloaded using `<link rel="preload">` elements, which reliably simultaneously fetch the js files in question from disk without executing them. (This seemed to improve the silent failure to load scripts discussed in #636, but as discussed in the below thread, it does not fully do so and a comprehensive fix for that will be necessary.) This is performed on `interactive`, as per the suggested change.
 - Scripts are then executed once React hydration completes. (The `waitForReactLoaded` change is just a refactor because it took me so long to figure out what it was doing :D)

There are still wait periods post-hydration while scripts that have preferences fetch their json manifests and while scripts' stylesheets are loaded, of course; if we cared about this, having the main script index file indicate whether a script has preferences or a css stylesheet file would allow those to be preloaded, too, but we really don't. (Also, scripts with individual JS files for each function like tweaks could be link preloaded etc etc whatever.)

### Testing steps

~~I don't expect anyone to be able to consistently replicate the problem this solves, as I didn't find a great method, so testing is probably confined to ensuring that XKit continues to function correctly.~~

~~To test the actual fix, merge https://github.com/AprilSylph/XKit-Rewritten/pull/756 into master on a slow PC, enable all of the scripts, and try to observe a failure. Then, merge https://github.com/AprilSylph/XKit-Rewritten/pull/756 into this PR with the same scenario and observe that there are no failures.~~

load it; it go fast


src/util/_index.json
```diff
@@ -0,0 +1,21 @@
...
+  "mutations",
+  "notifications",
+  "post_actions",
+  "preferences",
+  "react_props",
+  "remixicon",
+  "sidebar",
+  "tumblr_helpers",
+  "user"
+]
```
AprilSylph, line 21: We're not using this array anywhere else, so I think we should move it to `preload.js`; the only reason that `src/scripts/_index.json` exists is to avoid duplicating it beween `main.js` and the control panel

src/manifest.json
```diff
@@ -38,6 +38,22 @@
...
+      "matches": [
+        "*://www.tumblr.com/*"
+      ],
+      "exclude_matches": [
+        "*://www.tumblr.com/login",
+        "*://www.tumblr.com/register",
+        "*://www.tumblr.com/register?*",
+        "*://www.tumblr.com/privacy/*"
+      ],
+      "run_at": "document_start",
```
AprilSylph, line 51: We probably shouldn't touch the DOM until the page is in the `interactive` state

```suggestion
      "run_at": "document_end",
```

AprilSylph (changes requested):

src/manifest.json
```diff
@@ -38,6 +38,22 @@
...
+      "js": [
+        "lib/browser-polyfill.min.js",
+        "content_scripts/preload.js"
+      ]
+    },
```
AprilSylph, line 56: 🤔 Do separate content scripts for the same addon on the same page get different sandboxes? It would be cool to be able to load all our libraries here...

src/manifest.json
marcustyphoon: Oh... hm. I will have to see if that actually works!

src/manifest.json
marcustyphoon: I did it like this just because of the differing `run_at` values; if they're both going to be at `document_end` I don't see why we can't just combine them.

marcustyphoon: (Also, yep, they do; that works if we don't change the `run_at` value and just delay the preload)

src/manifest.json
```diff
@@ -38,6 +38,22 @@
...
+    {
+      "matches": [
+        "*://www.tumblr.com/*"
+      ],
+      "exclude_matches": [
+        "*://www.tumblr.com/login",
+        "*://www.tumblr.com/register",
+        "*://www.tumblr.com/register?*",
+        "*://www.tumblr.com/privacy/*"
+      ],
```
AprilSylph, line 50: We don't need these exclusions just for preloading - they're in place to prevent passive API calls (think `user.js`) from triggering infinite redirects

```suggestion
```

AprilSylph (changes requested):

src/util/_index.json
marcustyphoon: Sure, I'm indifferent either way. Loading an extra JSON file for no actual reason seems silly; on the other hand, having it there makes it clearer that it should be updated if another util file is added.

src/manifest.json
marcustyphoon: So far it seems to work.

AprilSylph: We might have to change the boot script to run at `document_idle` for this to be reliable, assuming the scripts we specify for preload behave normally (i.e. have to be loaded before the document's state is considered `complete`)

Honestly, I thought we were using `document_idle` already since I don't imagine hydration can ever occur before the document `load` event, but I didn't ever test that, so I might be wrong.

marcustyphoon: Ah, I remember now: in the fully optimized version, I started fetching the script index JSON at `document_start` and could delay stuff until `interactive` using a `waitForDocumentReady()` function. Still wouldn't be 100% guaranteed reliable, I suppose.

Hydration does occur before that, as far as I know, yeah.

AprilSylph: For the sake of reliability I'm willing to entertain the idea of actually delaying boot until the preloading is finished. Would be interesting to see just how much of a difference it makes to boot at `document_idle` with scripts preloaded vs `document_end` without preloading.

src/util/_index.json
AprilSylph: Maybe we should just create a `src/util/README.md` instructing contributors to update the preload file.
Maybe we should have a README for _every_ directory...

(These would both be out of scope for this PR; right now my verdict is "more synchronous code, please")

src/manifest.json
marcustyphoon: image

Man, this whole thing is hard to illustrate without getting out the ol' ms paint or whatever and marking up screenshots. (Screenshot is illustrative only.)

Short version: I poked some things a bunch. With the current unmodified main.js code, `init()` waits on its own script index JSON load after waitForReactLoaded, which puts the import() statements and waterfall later than `document_idle` anyway. Of course, running main.js at `document_idle` would cause that JSON load to begin at that point. And, uh... I dunno, some other things are also true; I may try to summarize this better later :P 

Stepping back, I would randomly ballpark the total difference in time between hydration and xkit load from any of these possibilities at, like, I dunno, 400ms? Do we care?

AprilSylph: I don't care about 400ms. We could probably regain that with further optimisations, but I wouldn't feel bad if we figure out that we can't.

marcustyphoon:
Okay, as nightpool mentioned... somewhere, I honestly think that setting run_at just doesn't really work that well. MDN says:

> `"document_end"` | Corresponds to `interactive`.

When I put a content script there, though, it gets run way after the `readystatechange` event that corresponds with `document.readyState === interactive`. Maybe that's when the browser starts loading the script file? If so, I mean, that's objectively reasonable, I guess, but also, why suggest that that's when the script will actually get executed, then?

Putting a `readystatechange` listener at the top of `init()`, on the other hand, times everything reliably and consistently as expected.

marcustyphoon:
Note for posterity: adding `type: 'application/json',` seems to fix the console spam when preloading JSON files for getPreferences. However, since we have no way of knowing if scripts have any preferences at all, this still results in loading a bunch of JSON files for scripts that will never reference them, and those files will spam the console.

marcustyphoon:
Random tidbits:

>  I don't imagine React hydration can ever occur before the document load event, but I didn't ever test that, so I might be wrong.

I checked for fun; hydration fires after a [stylesheet load event](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#stylesheet_load_events).

I remember someone said something complicated about when exactly a stylesheet counts as loaded (probably on MDN somewhere), but I forget where. If more research needed to be done, though, there's a direction to look in.

> https://www.tumblr.com/transienturl/709284574500438016

Poked a bit more and yep: it seems like the same synchronous event loop invocation contains both the `readyStateChange` **interactive** event handlers and the `DOMContentLoaded` event handlers... but it takes more than half a second (parsing redpop code), the `readyStateChange` handlers are at the beginning, and the `DOMContentLoaded` handlers are at the end, so which event you use to time a resource fetch makes a huge difference!

If the thing one was trying to do was a DOM manipulation, it wouldn't make a difference, of course; the page only gets rendered at the end of all of that parsing, so the time at which the user will see the result is the same. But when the work you're doing is on another thread, the order can matter!

marcustyphoon:
Fascinating result from loading the extension into Safari 15 on MacOS: the extension consistently fails to load unless the imports really are not all at once, e.g. the old:

```js
for (const scriptName of installedScripts.filter(scriptName => enabledScripts.includes(scriptName))) {
  await runScript(scriptName);
}
```

Apparently there is a Safari bug with multiple simultaneous fetches to a module that uses top-level await. It doesn't seem like the same problem as the Chrome one, but does share a lot of similarities to it? Annoyingly, "make a test copy of XKit Rewritten that does not use top-level await" isn't really a viable testing option.

https://www.tumblr.com/transienturl/709927672316329984/reference-error-cannot-access-uninitialized

marcustyphoon:
Confirmed that the description of that bug is correct; Safari loads correctly with the unnecessary top level await in Tag Tracking+ removed and no multiple simultaneous imports of the same file with top-level await by prepending:

```js
await Promise.all(['css_map', 'language_data', 'user'].map(async name => {
  const scriptPath = getURL(`/util/${name}.js`);
  await import(scriptPath);
}));
```

(Identical result 16.3 on Ventura on M1 Mac; 15.6.1 on Catalina on Intel.)

Kind of a pain.

marcustyphoon:
The above message about Safari appears to apply to Chrome as well. It's much harder to trigger, but having legacy XKit loaded at the same time helps.

So, to recap, as far as I can tell:
 - Two scripts importing the same js file at the same time, when the js file has a top-level await, can be problematic depending on timing. This seems to be a webkit bug, affecting both Safari and Chrome on iOS.

Thus we could:
 - Import the top level await utils before loading scripts. The list would effectively be a magic number; we/contributors could forget to update it.
 - Import the top level await utils before loading scripts and use an eslint rule to warn about top level awaits, disabling it on certain files.
 - Import all of the utils before loading scripts (wasted cpu cycles)
 - Don't import scripts simultaneously (much larger loading speed drawback)

In addition to this, we could improve loading speed by:
 - Implementing a timing improvement in main.js that fires events exactly when they should happen
 - Doing that, and also preloading util and script js files. This makes the change when viewed as a whole a loading speed improvement regardless of which option is picked in the first section, I think, probably, don't quote me on that.

AprilSylph:
So, where are we at now? Preloading doesn't solve the issue across the board? I think you said that on your blog, but you didn't mention it here.

Maybe we _should_ just disallow top-level await in util files. Just exporting promises and allowing the features to top-level await the results is almost as good as exporting results of promises.

marcustyphoon:
Oh, sorry, to clarify:

> Preloading actually doesn't solve the issue in Chrome?

Yes. (After I installed XKit 7, I got occurrences with the preloading patch loaded.)

> Maybe we should just disallow top-level await in util files. Just exporting promises and allowing the features to top-level await the results is almost as good as exporting results of promises.

For user info, I definitely think it's fine either way. Not being able to use `keyToCss` and `translate` in top-level buildStyle calls sucks though.

AprilSylph:
It would look a little silly, but think of it as clown makeup for this WebKit bug.

marcustyphoon:
```js
// util:

const cssMapPromise = inject(async () => window.tumblr.getCssMap());

export const getKeyToClasses = cssMapPromise.then(
  cssMap =>
    /**
     * @param {...string} keys - One or more element source names
     * @returns {string[]} An array of generated classnames from the CSS map
     */
    function (...keys) {
      return keys.flatMap(key => cssMap[key]).filter(Boolean);
    }
);

export const getKeyToCss = getKeyToClasses.then(
  keyToClasses =>
    /**
     * @param {...string} keys - One or more element source names
     * @returns {string} - A CSS :is() selector which targets all elements that match any of the given source names
     */
    function (...keys) {
      const classes = keyToClasses(...keys);
      return `:is(${classes.map(className => `.${className}`).join(', ')})`;
    }
);

// script:

import { getKeyToCss } from '../util/css_map.js';
const keyToCss = await getKeyToCss;
```

```js
// util:

export default inject(async () => window.tumblr.getCssMap()).then(cssMap => {
  /**
   * @param {...string} keys - One or more element source names
   * @returns {string[]} An array of generated classnames from the CSS map
   */
  const keyToClasses = (...keys) => keys.flatMap(key => cssMap[key]).filter(Boolean);

  /**
   * @param {...string} keys - One or more element source names
   * @returns {string} - A CSS :is() selector which targets all elements that match any of the given source names
   */
  const keyToCss = function (...keys) {
    const classes = keyToClasses(...keys);
    return `:is(${classes.map(className => `.${className}`).join(', ')})`;
  };

  return { keyToClasses, keyToCss };
});

// script:
import cssMapPromise from '../util/css_map.js';
const { keyToCss } = await cssMapPromise;
```

Eugh.

marcustyphoon:

PR #960, opened 2023-02-08
New Tweak: Stacked badges

marcustyphoon:
Q: why
A: it was a more fun use of half an hour than trying to instrument an intentionally-delayed domcontentloaded event okay

### Description

Adds a tweak that stacks the "important internet checkmarks" unless you hover over them.

~~Slightly out of date~~ hover animation demo:

### Testing steps

- Confirm that badges are stacked, can be hovered, animate correctly, and still have popups when hovered.
- Confirm that stacked badges appear correctly in the regular and community header.


marcustyphoon:
Note that this is now broken by the (very nice! good job!) `tooManyCheckmarks` css used to achieve a lesser but somewhat similar effect by redpop if someone has a ton.

 - [ ] override this

```css
.EIL5u .tDT3c {
    top: 2px;
    left: calc(2% + 95%*var(--checkmarks-index)/var(--checkmarks-total));
    position: absolute;
}
```

src/scripts/tweaks/badge_stack.js
```diff
@@ -5,7 +5,7 @@ import { pageModifications } from '../../util/mutations.js';
 let styleElementMaxBadges = 0;
 const styleElement = buildStyle();
 
-const wrapper = keyToCss('badgesContainer');
+const wrapper = `:is(article ${keyToCss('badgesContainer')})`;
```
AprilSylph, line 8: ...what is this `:is()` doing?

AprilSylph: oh wait, I didn't expand the whole file, I see.

marcustyphoon:
https://developer.mozilla.org/en-US/docs/Web/CSS/transition-behavior

Hey, look, a new CSS thingy. Firefox nightly only at the moment, but probably coming soon. Useful for that old "scroll to bottom button needs to change `visibility` after a delay to match the scroll to top button" situation, possibly? Can't remember how or if we solved that.

marcustyphoon:
- [x] do something about multiple rows in new header

PR #962, opened 2023-02-08
Tweaks: Fix "where were we" tweak consistency

### LINKED ISSUE #961: Tweaks: Hide "where were we" tweak inconsistent
### Platform

MacOS 10.15.7 x86

### Browser

Ungoogled Chromium v109.0.5414.119

### Addon version

actually 0.20.11 but it should happen in master I believe

### Details

It's just #761 again; the CSS load order makes a `display` property override this tweak's `display: none`.

marcustyphoon:
### Description

Uses !important to ensure race conditions do not break the "where were we" hiding tweak when soft navigating to the dashboard.

Resolves #961.

### DIFF
--- a/src/scripts/tweaks/no_where_were_we.js
+++ b/src/scripts/tweaks/no_where_were_we.js
@@ -1,7 +1,7 @@
 import { keyToCss } from '../../util/css_map.js';
 import { buildStyle } from '../../util/interface.js';
 
-const styleElement = buildStyle(`${keyToCss('wrapper')} ${keyToCss('newPostIndicator')} { display: none; }`);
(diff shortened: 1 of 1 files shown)


AprilSylph:
Wouldn't the global fix for this to append our inline styles to the end of the `document.documentElement`? (I think that's the only option available to us right now to ensure same-specificity rules override Tumblr's stylesheets, anyway.)

marcustyphoon:
Happy to PR that. I see that Stylus uses it. I guess it's one of those "technically the spec doesn't have anything about style tags outside of head but it obviously works so let's roll with it" things? I didn't even know you could do it at all; MDN really ought to mention it.

PR #963, opened 2023-02-10
Ensure style elements override Tumblr styles

### LINKED ISSUE #961: Tweaks: Hide "where were we" tweak inconsistent

marcustyphoon:
### Description

This changes the location where dynamic style elements are inserted to the end of the `document.documentElement` (`<html>` element) instead of inside `<head>`, such that `!important` declarations should not be required to reliably resolve race conditions between Tumblr's styles and ours.

Resolves #961. Obsoletes #962.

### DIFF
--- a/src/scripts/accesskit/disable_animations.js
+++ b/src/scripts/accesskit/disable_animations.js
@@ -22,5 +22,5 @@ canvas#fire-everywhere {
 }
 `);
 
-export const main = async () => document.head.append(styleElement);
+export const main = async () => document.documentElement.append(styleElement);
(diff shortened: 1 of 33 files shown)


marcustyphoon:
note to self: update your new-features PRs if this gets merged

AprilSylph (approved):

PR #966, opened 2023-02-12
Quick Reblog: Increase default remembered post storage

marcustyphoon:
### Description

> the 100 default is probably just for the sake of browser.storage.sync, which we do not use anymore, so yes please set it higher

### DIFF
--- a/src/scripts/quick_reblog.json
+++ b/src/scripts/quick_reblog.json
@@ -67,7 +67,7 @@
       "label": "Remember the last ",
       "options": [
         { "value": "100", "label": "100 posts" },
-        { "value": "1000", "label": "1000 posts" },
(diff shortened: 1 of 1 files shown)


src/scripts/quick_reblog.json
```diff
@@ -70,7 +70,7 @@
         { "value": "1000", "label": "1000 posts" },
```
AprilSylph, line 70: while we're here, might as well make this label consistent

```suggestion
        { "value": "1000", "label": "1,000 posts" },
```

AprilSylph (approved):

AprilSylph:
oh wait changes remove approvals. whoops

AprilSylph (approved):
again

PR #969, opened 2023-02-13
Quick Tags: Work correctly on non-NPF posts

### LINKED ISSUE #940: Quick Tags feature spits out just errors
### Platform

MacOs Monterey 12.5.1

### Browser

Mozilla Firefox 108.0.2

### Addon version

0.21.0

### Details

Hi, I just installed XKit and used the Quick Tags feature – worked great, except I listed the tags with hashes, so in my posts all tags looked like ##tag. I removed the hashes from the list, and now the QT doesn't work at all. I thought this might do with the use of slashes in the tags, but I removed all of those and still it doesn't work. Put in a new list of tags, completely random – again; worked once, and then nothing. Then I added the original bundle of hashtags to XKit, tried the QT – worked once, then error galore. Help?

Backup:

`{ "enabledScripts": [ "quick_tags", "anti_capitalism" ], "quick_tags.preferences.answerTag": "", "quick_tags.preferences.autoTagAsker": false, "quick_tags.preferences.originalPostTag": "", "quick_tags.preferences.tagBundles": [ { "title": "test", "tags": "day, window, sky, crap, donna williams, levity, carnival, britney, asshole, demo, bill, andrew, tate, category, awesome, library, picnic, trash, cocacola, trashcan, velocity, oval, tube, intestine" } ] }`

### LINKED ISSUE #965: Quick Tags: Work correctly on non-NPF posts
### Concept

Right now, non-text-only legacy posts can't be tagged using Quick Tags.

This should be fixable by implementing an alternative API call using the mass post editor endpoint if the post has `isBlocksPostFormat: false`. (This isn't technically ideal, I believe; text-only legacy posts will have this flag but can be edited using the regular NPF edit endpoints. But both endpoints should do the same thing as far as I know, right? So it shouldn't matter.)

### LINKED ISSUE #1224: quick tags not functioning
### Platform

windows 10

### Browser

google chrome

### Addon version

v0.21.15

### Details

quick tags is not functioning. i can add new bundles or type out tags but they are not saved upon hitting enter/clicking the quick tag. text pops up saying "Uh-oh! That didn't work. Please try again." but it still does not work. have updated extension.

marcustyphoon:
### Description

This adds an alternative API endpoint call to Quick Tags on posts that we cannot ensure the NPF edit endpoint will work on. It uses the mass post editor endpoint, which should work on both legacy and NPF posts.

For discussion about which posts are NPF-compatible, see [discord.com/channels/104051306309644288/327155782934200320/1074286553733533738](https://discord.com/channels/104051306309644288/327155782934200320/1074286553733533738).

Resolves #940 (most likely reason). Resolves #965. Resolves #1224.


marcustyphoon:
- [ ] see if the mega editor endpoint changes tipping

src/scripts/quick_tags.js
```diff
@@ -146,26 +147,33 @@ const togglePostOptionPopupDisplay = async function ({ target, currentTarget })
...
+        body: {
+          ...createEditRequestBody(postData),
+          tags: tags.join(',')
+        }
+      });
+
+      notify(displayText);
+    } else {
+      await megaEdit([postId], { mode: 'add', tags: tagsToAdd });
+      notify(`Edited on ${blogName} (legacy)`);
```
AprilSylph, line 175: will need to change the `catch` logic to account for this, do we even get a JSON response from the mass post editor?

marcustyphoon: See #971.

marcustyphoon:
This next commit addresses the above comment and #1046, though I'm not sure if this is the ideal way to do so (vs, for example, making the mega editor error wrapped in `{ body: contentsGoHere }`.

marcustyphoon:
Besides the virtual scroller stuff, this is IMO a good priority, as the virtual scroller broke legacy XKit's Quick Tags and that seems like a great ~~excuse to have people switch~~ ahem, reason to ensure we have a good alternative.

src/scripts/quick_tags.js
```diff
@@ -159,15 +160,20 @@ const addTagsToPost = async function ({ postElement, inputTags = [] }) {
...
```
AprilSylph, line 175: could make this copy more accurate

```suggestion
      notify(`Edited legacy post on ${blogName}`);
```

marcustyphoon:
Now that I think about it, it's probably better not to mention legacy posts in the copy at all, considering that there is some ambiguity there.

AprilSylph:
Eh, it's accurate. If we're displaying that message then it's definitely a legacy post, even if it doesn't display for all legacy types.

AprilSylph (approved):

PR #970, opened 2023-02-13
Fix `createEditRequestBody` helper

marcustyphoon:
### Description

what am I even doing this is an embarrassment

### DIFF
--- a/src/util/tumblr_helpers.js
+++ b/src/util/tumblr_helpers.js
@@ -68,8 +68,10 @@ export const createEditRequestBody = postData => {
     interactabilityReblog,
 
     canBeTipped,
-    hasCommunityLabel,
-    communityLabelCategories
+    communityLabels: {
+      hasCommunityLabel,
+      communityLabelCategories
+    }
   } = postData;
 
   return {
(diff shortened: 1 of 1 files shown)


src/util/tumblr_helpers.js
```diff
@@ -68,8 +68,10 @@ export const createEditRequestBody = postData => {
     interactabilityReblog,
 
     canBeTipped,
-    hasCommunityLabel,
-    communityLabelCategories
+    communityLabels: {
+      hasCommunityLabel,
+      communityLabelCategories
```
AprilSylph, line 73: not seeing `communityLabels.communityLabelCategories` in timeline objects, but am seeing `communityLabels.categories`

```suggestion
      categories: communityLabelCategories
```

AprilSylph (changes requested):

src/util/tumblr_helpers.js
marcustyphoon: > what am I even doing this is an embarrassment

mhmm yes good

AprilSylph: you're fine. remember: a mistake approved is a mistake shared. we're equally responsible

AprilSylph (approved):

PR #974, opened 2023-02-14
Quote Replies: Prevent post-open failure

### LINKED ISSUE #710: Quote Replies: Possible failure to open draft post
### Platform

MacOS 10.14.6 x86

### Browser

Edge 104.0.1293.54

### Addon version

v0.20.7

### Details

Haven't managed to reproduce it, but I did just encounter a situation where Quote Replies opened its new tab on the drafts page, but the draft it created didn't have the post in question until a refresh.

Not sure if this could be mitigated by either doing a fetch after the post requests that creates the draft to see if it's there yet, or by forcing a page refresh in the drafts tab if the post isn't there yet (and the information stored in local storage indicates that the script has just been activated, and we've only tried this refresh once, and etc).

marcustyphoon:
todo: rebase this; it is currently on top of #901. 

This adds a check to ensure that Quote Replies' created draft exists on the server before attempting to open it. (Without this, we still await the apiFetch creating it, which... seems like it should be enough, honestly? Is that true? Anyway, I did have it fail once, as mentioned in the linked issue.)

Resolves #710.


marcustyphoon:
Closing for "probably not necessary," see linked issue.

PR #977, opened 2023-02-16
feat(Scroll to Bottom): Function in modal blog view

marcustyphoon:
### Description

> It doesn't add functionality to work in the modal blog view; that's more complicated.
> – #976.

This adds scroll to bottom functionality in the mobile blog view. To do so, it adds a duplicate of the round "Back to top" button that's specific to the modal blog view container, and adjusts the scrolled element and observed element in `startScrolling` depending upon the presence or absence of the modal.

It also replaces the `checkForButtonRemoved` pageModification callback with a check during the `scrollToBottom` function.

### Testing steps

todo: write all of the tests that this stops scrolling when you would expect it to


marcustyphoon:
Issue with this: because the button color is now customized by the user theme (?), the button color will desync when navigating to a different blog in the modal.

- [ ] address this

AprilSylph (changes requested):
Nice, it works.

Some details I care about, though:
- [x] Showing both the regular scroll to top button and the blog view one at the same time is kinda weird. Let's hide the regular button when the blog view modal is open.
- [ ] Only showing the blog view scroll to bottom button when the blog view scroll to top button is visible doesn't make sense to me. Let's match the behaviour of the regular one, where it is shown regardless of scroll position.
- [x] The active `--yellow` can't be expected to always contrast against the button's background colour... could we... change the button's background colour to `--black` when scrolling?
    - I was initially thinking we use the same colour as the BG but apply an invert colours filter on the SVG, but that's probably really weird-looking.

marcustyphoon:
> Showing both the regular scroll to top button and the blog view one at the same time is kinda weird. Let's hide the regular button when the blog view modal is open.

Oh, definitely. Not immediately sure how to do that, so if anyone has any ideas, hit me up.

> Only showing the blog view scroll to bottom button when the blog view scroll to top button is visible doesn't make sense to me. Let's match the behaviour of the regular one, where it is shown regardless of scroll position.

You know, I'm not actually sure how to do that either?

marcustyphoon:
Note: This can probably take advantage of `:has()`.

marcustyphoon:
> Only showing the blog view scroll to bottom button when the blog view scroll to top button is visible doesn't make sense to me. Let's match the behaviour of the regular one, where it is shown regardless of scroll position.

Yeah, I don't think this is really possible without changing the Tumblr code's handling of the `scrollToTopEnabled` prop to render-but-hide the back to top button instead of conditionally rendering both it and its container.

marcustyphoon:
note to self: nontrivial merge commit there; make sure logic still makes sense with the scroll in an rAF

PR #983, opened 2023-02-16
No Recommended: Split recommended blogs sidebar preference

### LINKED ISSUE #605: No Recommended: Split recommended blogs sidebar preference
### Concept

I kind of like having the recommended related blogs in the blog view sidebar, but not having the generic recommended blogs in the dashboard sidebar. No Recommended could have separate options for these (and they can inherit their states from the current single preference.)

marcustyphoon:
Really digging into the bag today I guess.

### Description

This splits the No Recommended option to hide recommended blogs into two: one that works in the blog view modal and one that does everything else.

(Technically there are more than two types of recommended blog carousels—the one in the tagged page is different than the one beside the dashboard, and the one on the search page is actually a fourth kind that we don't even hit right now. Also there's one on the logged-out page that gets hit by the code for the tagged page at the moment. But a) treating them differently is a level of granularity that I'd say is somewhat silly; and b) I don't know if anyone really notices those anyway.

The big ones, IMO, are the dashboard one because it is immediately visible and takes up space we want for XKit sidebar elements, and the blog view modal one because without it there's often nothing at all there.)

Important note: this tweaks the code for inherited preferences so that the previous preference doesn't get auto-deleted. I would be fine with reverting that and having the blog view option here not inherit, too; users who dislike seeing the newly non-hidden element in the blog view modal will presumably find their way to the No Recommended preference pane on their own, right? And then maybe people will complain about the empty blog view modal sidebar less often, idk.

Resolves #605.

### Testing steps

Inheritance:
 - Load the main branch via `npm start`
 - Enable the "hide recommended blogs in the sidebar" preference (and no recommended)
 - Switch to this PR branch
 - Open Tumblr
 - Ensure that both recommended blog options are checked, having inherited the preference

Functionality:
- Disable both options
- Go find a blog view modal with a recommended blog sidebar
- Ensure that only the correct option hides the sidebar
- Repeat with the dashboard sidebar
- Repeat with a tagged page sidebar
- (a search page sidebar won't work)


marcustyphoon:
Oh, huh.

Resolves #1285, apparently.

AprilSylph (approved):

PR #845, opened 2022-11-20
Panorama: Custom maximum post width; multiple bug fixes

marcustyphoon:
Note: enabling "hide whitespace" when viewing the diff is recommended.

### Description

Adds a preference letting the user customize the maximum post width when Panorama is active. I really, really wanted the preference to actually be the width of the posts themselves rather than e.g. just setting the width of the outside container.

I experimented with and learned a ton about the box model and intrinsic widths and stuff while researching this... and used none of it, because I was totally overcomplicating things thinking the highest-up-the-tree place to set the timeline width was a layer lower than it is and couldn't interact with the `container` flex, uh, container. Oops.

No idea what to set the default width to, honestly.

Resolves #814,  discussion #391, discussion #853

**Edit:** Additionally, this applies styling to some elements that were missing them (native audio blocks and link embeds), correctly applies Panorama expansion to the tag and search pages when they're not in "masonry" views, and hotfixes the embedded blog view exploding; see the later commits.

Resolves half of #937, the tag page portion; blog view panorama is more complicated and I'll address it in a future PR.

Resolves #985.

### Testing steps

 - enable Panorama
 - adjust the new preference; try a blank value, a number (treated as pixels), and a pixel value. (CSS functions like `calc(100vw - 400px)` should work as well).

additionally:

- ensure Panorama works on an audio post and a link post
- ensure Panorama works on e.g. https://www.tumblr.com/tagged/xkit?sort=recent and https://www.tumblr.com/search/xkit/recent (in "list" view)
- ensure Panorama does nothing on e.g. https://www.tumblr.com/tagged/xkit?sort=recent and https://www.tumblr.com/search/xkit/recent in "masonry" view
- hard navigate to https://www.tumblr.com/april and ensure that it doesn't look like this:

PR #984, opened 2023-02-17
Panorama: Widen ALL the things

### LINKED ISSUE #937: Panorama on Tag Tracking
### Concept

I'm in love with the panorama option, but if I click a tracked tag or a blog from my dashboard, the posts go back to normal width. Would it be possible to extend this feature?

marcustyphoon:
todo: rebase after #845 

### Description

This stores some code experiments to make Panorama work in the blog view (modal and embedded; various screen widths).

- [x] figure out how to limit the width of the blog view modal timeline when it has no sidebar

Resolves #937.

### DIFF
--- a/src/scripts/panorama.js
+++ b/src/scripts/panorama.js
@@ -1,39 +1,56 @@
 import { keyToCss } from '../util/css_map.js';
 import { buildStyle } from '../util/interface.js';
+import { getPreferences } from '../util/preferences.js';
 
 const container = `${keyToCss('bluespaceLayout')} > ${keyToCss('container')}`;
 const reblog = `${keyToCss('post')} ${keyToCss('reblog')}`;
 const videoBlock = keyToCss('videoBlock');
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Re: the last 3 commits:

Besides a small annoyance where certain widths can squish the timeline width slightly below 540px, I would call the functionality in this state fairly complete in a "handle every scenario; damn the complexity" sense for the 9 different scenarios resulting from [regular timeline; blog view modal; blog view embedded] x [mobile width range; single-column blog view width range; wide range] with width preferences respected.

I think after deduplication of most of the CSS rules and with a couple of hours of refactoring, this could become mostly-comprehensible, but I doubt it's worth the time. Nevertheless, this is the way my brain works: determine and implement the maximum functionality set, then discard and simplify as appropriate. I'm pretty confident a reviewable PR that people won't notice any missing functionality from should be reasonably easy.

marcustyphoon:
Outdated as of #1291.

PR #992, opened 2023-02-21
feat(Mutual Checker): Following icons in blog cards

marcustyphoon:
### Description

This is really what I actually wanted the Mutual Checker blog card thing for: the ability to add an icon to blog cards when a user follows you but you don't follow them back as well as when they're a mutual. Oftentimes I'm in a thread somewhere and I see a blog that I think I might recognize, but that I'm not super familiar with as I don't follow them, and being able to see that they're a follower quickly is very convenient.

Obviously this doesn't work if the blog in question is a sideblog, but most of the time when I want this, it works great.

Why not do this for post headers as well? It would be an API request for every unique user whose post you see. Mutual Checker generally only fires API requests for each unique user **who you follow** whose posts you see, limiting impact.

### Testing steps

- Confirm that mutual icons are added to post headers and blog cards, that their colors are correct, and that they have tooltips ~~(except that the tooltips on post headers don't currently work, see #1779)~~.
- Confirm that following icons are added to blog cards of users that follow you, that their colors are correct, and that they have tooltips.


src/features/mutual_checker/index.js
```diff
@@ -151,18 +151,26 @@ export const main = async function () {
...
+        dom('title', { xmlns: 'http://www.w3.org/2000/svg' }, null, [
+          translate('Mutuals'),
+        ]),
+        dom('path', { xmlns: 'http://www.w3.org/2000/svg', d: path }),
+      ]
+    : [
+        dom('title', { xmlns: 'http://www.w3.org/2000/svg' }, null, [
+          translate('{{blogNameLink /}} follows you!').replace('{{blogNameLink /}}', blogName),
+        ]),
+        dom('use', { xmlns: 'http://www.w3.org/2000/svg', href: '#ri-user-shared-line' }),
```
AprilSylph: This doesn't work if you don't have Quick Tags or Trim Reblogs enabled.

Maybe we should be using the icons that Tumblr used to use for the Mutuals/Following activity badges? Which ones were those again?

AprilSylph (changes requested):

src/features/mutual_checker/index.js
marcustyphoon: "Used to" implying that they differed from the current ones, I assume? I vaguely remember this. Were they, like, kinda squareish...?

marcustyphoon: `managed-icon__following` isn't it, is it?

AprilSylph: "Used to" as in they don't have icons at all anymore:

image

AprilSylph: I see we still have the icon IDs in Tweaks:

https://github.com/AprilSylph/XKit-Rewritten/blob/f8d496ad0ae57d081b8e2dfa1b1b56b88d8409e0/src/features/tweaks/subtle_activity_mutuals.js#L54-L56

I wonder if the design system icons have equivalents we can update to. The pre-DS icons could be on the chopping block at some point.

AprilSylph: Mutuals icon equivalent: `#managed-icon__ds-users-filled-24`
(Or, intended replacement: `#managed-icon__ds-user-mutual-filled`)

Following icon equivalent: `#managed-icon__ds-user-following-filled`

marcustyphoon: Ah, you mean the ones that are still currently used in—

—ah, damn, I had forgotten we insert them in that tweak.

src/features/mutual_checker/index.js
```diff
@@ -151,18 +152,26 @@ export const main = async function () {
...
+        dom('title', { xmlns: 'http://www.w3.org/2000/svg' }, null, [
+          translate('Mutuals'),
+        ]),
+        dom('path', { xmlns: 'http://www.w3.org/2000/svg', d: path }),
+      ]
+    : [
+        dom('title', { xmlns: 'http://www.w3.org/2000/svg' }, null, [
+          translate('{{blogNameLink /}} follows you!').replace('{{blogNameLink /}}', blogName),
+        ]),
+        buildSvg('ri-user-shared-line').firstElementChild,
```
AprilSylph: As per the other conversation:

```suggestion
        buildSvg('ri-user-follow-line').firstElementChild,
```

AprilSylph (approved):
Ready to merge after one last thing...

src/features/mutual_checker/index.js
marcustyphoon: Ah, so—still using a remix icon icon, but one with a check mark instead of an arrow, matching the iconography with Tumblr's icon that represents a following blog even though it's not currently used anywhere(?), so that it will be consistent if it is.

I think the check mark is slightly less clear as meaning "this user follows you," but neither one of them is actually particularly clear and it's thus not a huge difference.

AprilSylph: Yep—if we have to choose one, we should go with the one that doesn't contradict Tumblr. There's lots of ways[^1] for the Tumblr icon to still be part of one's picto-vocabulary[^2], so we can exploit that to make our icon choice seem more intuitive to more people, without it actually being any more intuitive than the alternative.

[^1]: e.g. simply remembering the icon from when it was used by Tumblr previously
  e.g. seeing the icon in older screenshots, circulated via reblogs today
  e.g. using an older mobile app version that still uses the icon
  e.g. using the Tweaks option that inserts the icon
[^2]: Yes, I just made up this term; is there an actual term for this that I'm forgetting?

AprilSylph: (Obviously, this line of thinking only works when the two choices are similarly [un]intuitive in a vacuum.)

marcustyphoon: > picto-vocabulary
> Yes, I just made up this term; is there an actual term for this that I'm forgetting? 

I asked an AI model for you, and it just said "Noah's Fish." (Was it the worst AI I could find in a few minutes of googling? That's entirely unimportant and I don't know why you would ask that.)

PR #1002, opened 2023-02-22
Notify on errors

marcustyphoon:
### Description

Pops up a notification on most errors instead of logging them to the console silently.

Like #472, but without the dev-only check. This was already rejected, which I'm fine with; PRing to keep track of it because I need to merge it into dev sometimes.


marcustyphoon:
```js
main().catch(console.error);
```

Come to think of it, given that main() is async, this catch handler is kind of pointless, no? Either way, something will get logged to the console and the code will continue (as there is no await statement). And typically it seems to me like logging an error gives you less information than letting the unhandled rejection log itself.

PR #1006, opened 2023-02-24
Seen Posts: Ignore header/footer in seen post detection

marcustyphoon:
### Description

This ensures that posts are not marked as seen if only a bit of the header becomes visible.

(Yes, I know `hiddenObserverMiddlePoint` exists, but I have multiple problems with it, such that you can skip it with J/K and it breaks the integration we have with our other post-hiding scripts. Playing with making my own now, actually.)


AprilSylph (dismissed):
didnt test; code looks reasonable

AprilSylph (approved):
i would love if clean merges didn't auto-dismiss reviews. i am approving the same diff.

marcustyphoon:
To be fair, it wasn't a clean merge! (I have no idea if a clean merge actually does dismiss, actually. I'll have to test that.)

PR #1008, opened 2023-02-26
No Recommended: Stop hiding when all posts are recommended

marcustyphoon:
### Description

This turns No Recommended's hide recommended posts tweak off once more than (an arbitrary) 10 posts in a row are recommendations.

This prevents the endless-scrolling-enabled dashboard from continually loading new content if part of one's dash is entirely recommended posts, which appears to be happening on my account if I scroll back enough (and might be true on a new user's dash as well). This saves server load and memory, of course.

~~I think this is fine to do without any UI copy updates. A fancier version could apply a CSS style to the element with `precedingHiddenPosts(postElement) === 10` that gives it a `before` with a message about too many recommended posts to hide.~~ actually you know what here

### Testing steps

If your dash does the same thing as mine: scroll far enough down your dashboard that all of the posts are recommended and confirm that the UI appears as shown.

### DIFF
--- a/src/scripts/no_recommended/hide_recommended_posts.js
+++ b/src/scripts/no_recommended/hide_recommended_posts.js
@@ -8,7 +8,23 @@ const unHiddenClass = 'xkit-no-recommended-posts-many';
 const timeline = /\/v2\/timeline\/dashboard/;
 const includeFiltered = true;
 
-const styleElement = buildStyle(`.${hiddenClass}:not(.${unHiddenClass}) article { display: none; }`);
+const styleElement = buildStyle(`
+.${hiddenClass}:not(.${unHiddenClass}) article {
+  display: none;
+}
+
+:not(.${unHiddenClass}) + .${unHiddenClass}::before {
+  content: 'Too many recommended posts to hide!';
+
+  display: block;
+  text-align: center;
+  padding: 25px 20px;
+  border-radius: 3px;
+  background-color: rgba(var(--white-on-dark), 0.13);
+  color: rgba(var(--white-on-dark), 0.6);
+  font-weight: 700;
+}
+`);
 
 const precedingHiddenPosts = ({ previousElementSibling: previousElement }, count = 0) => {
   if (!previousElement) return count;
--- a/src/scripts/no_recommended/hide_recommended_posts.js
+++ b/src/scripts/no_recommended/hide_recommended_posts.js
@@ -1,13 +1,22 @@
-import { buildStyle, filterPostElements } from '../../util/interface.js';
+import { buildStyle, filterPostElements, postSelector } from '../../util/interface.js';
 import { onNewPosts } from '../../util/mutations.js';
 import { timelineObject } from '../../util/react_props.js';
 
 const excludeClass = 'xkit-no-recommended-posts-done';
 const hiddenClass = 'xkit-no-recommended-posts-hidden';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I can't remember how Limit People is supposed to work exactly, but this technique could probably work there too? possibly?

AprilSylph:
> I can't remember how Limit People is supposed to work exactly, but this technique could probably work there too? possibly?

Basically, though with a configuration option for how many posts by the same blog is considered acceptable, and with the option to expand the hidden posts.

marcustyphoon:
Started thinking about how I would do that and realized that if it were me I would want it to interface with hide liked posts, hide seen posts, show originals, etc, which would have a complicated state machine both to work and to not have unpredictable behavior if you e.g. turned off hide seen posts or changed the show originals mode. Well, story for another time.

src/scripts/no_recommended/hide_recommended_posts.js
```diff
@@ -1,13 +1,41 @@
...
+  text-align: center;
+  line-height: 1.5em;
+}
+`);
+
+const precedingHiddenPosts = ({ previousElementSibling: previousElement }, count = 0) => {
+  if (!previousElement) return count;
+  if (!previousElement.matches(postSelector)) return precedingHiddenPosts(previousElement, count);
+  if (previousElement.classList.contains(hiddenClass)) return precedingHiddenPosts(previousElement, count + 1);
+  return count;
```
AprilSylph, line 37: I'd like comments on these. Recursion always takes me a minute to wrap my head around...

```suggestion
  // If there is no previous sibling, stop counting
  if (!previousElement) return count;
  // If the previous sibling is not a post, skip over it
  if (!previousElement.matches(postSelector)) return precedingHiddenPosts(previousElement, count);
  // If the previous sibling is hidden, count it and continue
  if (previousElement.classList.contains(hiddenClass)) return precedingHiddenPosts(previousElement, count + 1);
  // Otherwise, we've hit a non-hidden post; stop counting
  return count;
```

AprilSylph (changes requested):

src/scripts/no_recommended/hide_recommended_posts.js
marcustyphoon: Yeah, for sure. Normally I try to avoid recursion entirely (see: `waitForReactLoaded`:D), but I didn't think of a clearer solution in this case.

src/scripts/no_recommended/hide_recommended_posts.js
```diff
@@ -1,13 +1,45 @@
...
+:not(.${unHiddenClass}) + .${unHiddenClass}::before {
+  content: 'Too many recommended posts to hide!';
+
+  display: block;
+  padding: 25px 20px;
+  border-radius: 3px;
+  margin-bottom: var(--post-padding);
+
+  background-color: rgba(var(--white-on-dark), 0.13);
+  color: rgba(var(--white-on-dark), 0.6);
```
AprilSylph, line 25: This might be the most annoying change I ever request but... Tumblr never uses 60% opacity colours... only 65%

```suggestion
  color: rgba(var(--white-on-dark), 0.65);
```

AprilSylph (approved):

marcustyphoon:
Not sure if we should consider doing things like this for other post-hiding extensions too; it kind of depends. Like, for "hide seen posts" I could see it being annoying for people trying to get farther back in their dash or whatever.

PR #1012, opened 2023-03-01
User util: use redpop's user info

marcustyphoon:
### Description

Rather than doing our own API request to get user info right after redpop does a request for the same thing, this steals it from appContext. This improves initial load speed (especially with legacy xkit loaded, for [weird reasons](https://www.tumblr.com/transienturl/710563203197861888/no-idea-if-anyone-would-know-this-but-if-you)) and technically reduces server load by some minute amount.

Worth bothering with? Probably no.

### DIFF
--- a/src/util/user.js
+++ b/src/util/user.js
@@ -1,11 +1,24 @@
-import { apiFetch } from './tumblr_helpers.js';
+import { inject } from './inject.js';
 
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Closing to reduce clutter.

PR #1013, opened 2023-03-02
Trim Reblogs: always show trim button

### LINKED ISSUE #958: Trim Reblogs: always show trim button
It would probably be more intuitive to disable the trim button on editable but untrimmable posts, rather than not add it at all.

The appropriate styles to make disabled XKit control buttons match Tumblr's disabled control buttons will need to be added for this.

marcustyphoon:
Seeing a lot of user reports that Trim Reblogs is broken and I would really like to know if it's a button add failure or a misunderstanding about which posts should be trimmable. Hence:

### Description

As described in the linked issue, this disables the trim button on editable but untrimmable posts, rather than not adding it at all.

This uses the native dom `disabled` attribute on the button element and moves the event handlers on control buttons to the button element itself.

Resolves #958.

### Testing steps

- Confirm that e.g. Quick Tags works as expected.
- Try to press the Trim Reblogs button on an original post, a reblog with no added content and only one trail item, and a post with enough content to trim.

### DIFF
--- a/src/content_scripts/control_buttons.css
+++ b/src/content_scripts/control_buttons.css
@@ -28,6 +28,11 @@
   fill: rgba(var(--black), 0.65);
 }
 
+.xkit-control-button:disabled svg {
+  fill: rgba(var(--black), 0.4);
+  cursor: not-allowed;
+}
+
 @media (max-width: 990px) {
   .xkit-control-button-container {
     margin-left: 0;
--- a/src/content_scripts/control_buttons.css
+++ b/src/content_scripts/control_buttons.css
@@ -28,9 +28,12 @@
   fill: rgba(var(--black), 0.65);
 }
 
+.xkit-control-button:disabled {
+  cursor: not-allowed;
+}
+
 .xkit-control-button:disabled svg {
   fill: rgba(var(--black), 0.4);
-  cursor: not-allowed;
 }
 
 @media (max-width: 990px) {
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -124,13 +124,15 @@ const appendWithoutViewportOverflow = (element, target) => {
   }
 };
 
-const togglePopupDisplay = async function ({ target, currentTarget }) {
+const togglePopupDisplay = async function ({ target, currentTarget: controlButton }) {
   if (target === popupElement || popupElement.contains(target)) { return; }
 
-  if (currentTarget.contains(popupElement)) {
-    currentTarget.removeChild(popupElement);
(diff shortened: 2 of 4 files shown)


marcustyphoon:
(oh, and yes, I checked what happens if a user removes the disabled attribute and clicks the button; nothing bad occurs.)

marcustyphoon:
Oh, major problem with this:

 - [ ] fix quick tags popup closing when you press space

I guess you have focus of the button element once you click it and the spacebar clicks it again or something...? Not sure yet, help appreciated

AprilSylph:
> I guess you have focus of the button element once you click it and the spacebar clicks it again or something...?

Are we not stopping keydown propagation in the Quick Tags field? We should be doing that for every `<input type="text">` we add to the page

marcustyphoon:
We are; ~~I don't understand what's happening.~~ Apparently you need to preventDefault() they keyup event??????????

Possibly related:

- https://bugzilla.mozilla.org/show_bug.cgi?id=816437
- https://github.com/emberjs/ember-test-helpers/issues/1054

(testing in chrome, though)

marcustyphoon:
Oh, I bet I know what's happening: because I attached the handlers to the button element without changing the handlers themselves, and they reference `currentTarget`, the popups are being placed inside the `button` element instead of inside the container. Pressing spacebar inside the popup is thus pressing spacebar with the focus inside a button.

AprilSylph:
oh right, yeah, definitely do not put popups inside button elements. that's no good on multiple levels.

src/content_scripts/control_buttons.css
```diff
@@ -28,6 +28,11 @@
   fill: rgba(var(--black), 0.65);
 }
 
+.xkit-control-button:disabled svg {
+  fill: rgba(var(--black), 0.4);
+  cursor: not-allowed;
+}
```
AprilSylph, line 34:
```suggestion
.xkit-control-button:disabled {
  cursor: not-allowed;
}

.xkit-control-button:disabled svg {
  fill: rgba(var(--black), 0.4);
}
```

AprilSylph (changes requested):
Works well with the exception that I can sneak a peak of `cursor: pointer` when hovering the very edge of a disabled button. We should attach `cursor: not-allowed` to the element that is actually disabled.

src/content_scripts/control_buttons.css
marcustyphoon: Ooh, good catch! I didn't think to check that in mobile view.

AprilSylph (approved):

PR #1014, opened 2023-03-02
Implement simple load failure warning

marcustyphoon:
### Description

In a few lines of code, this implements a simple warning display when not all scripts have been successfully loaded after an arbitrary delay (5 seconds). It is very difficult to detect this failure state otherwise, as no error occurs.

The warning message vanishes if scripts do eventually load, preventing possible false positives if loading takes an excessive amount of time (reopening a browser with many windows can cause this, for example); the message should be chosen not to definitively imply that the extension is in a permanent error state.

This would make sense to merge along with a fix for the "haunted chromium" bug, to see if the fix worked. (Right now I kind of favor the ```await Promise.all(['css_map', 'language_data', 'user'].map(name => import(getURL(`/util/${name}.js`))));``` oneliner, just so we have something released.)

#756 is a much more convoluted version of this that identifies which script(s) failed to load.

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -85,9 +85,13 @@
     const installedScripts = await getInstalledScripts();
     const { enabledScripts = [] } = await browser.storage.local.get('enabledScripts');
 
-    installedScripts
+    const loaded = installedScripts
       .filter(scriptName => enabledScripts.includes(scriptName))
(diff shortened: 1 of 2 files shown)


AprilSylph:
I don't intend to merge this given that the issue causing load failures was mitigated.

PR #1015, opened 2023-03-05
Trim Reblogs: always show trim button (alternate)

### LINKED ISSUE #958: Trim Reblogs: always show trim button

marcustyphoon:
Seeing a lot of user reports that Trim Reblogs is broken and I would really like to know if it's a button add failure or a misunderstanding about which posts should be trimmable. Hence:

### Description

As described in the linked issue, this disables the trim button on editable but untrimmable posts, rather than not adding it at all.

This does not use the native dom `disabled` attribute on the button element, just simulating it with a class and not attaching handlers.

Resolves #958.

### Testing steps

- Confirm that e.g. Quick Tags works as expected.
- Try to press the Trim Reblogs button on an original post, a reblog with no added content and only one trail item, and a post with enough content to trim.

### DIFF
--- a/src/content_scripts/control_buttons.css
+++ b/src/content_scripts/control_buttons.css
@@ -28,6 +28,11 @@
   fill: rgba(var(--black), 0.65);
 }
 
+.xkit-control-button-disabled svg {
+  fill: rgba(var(--black), 0.4);
+  cursor: not-allowed;
+}
+
 @media (max-width: 990px) {
   .xkit-control-button-container {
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Closing, as the non-alternate version was merged!

PR #1017, opened 2023-03-05
demo: enforce no top level await in utilities; refactor user.js

marcustyphoon:
### Description

this still breaks intellisense on user.js exports unless you do ridiculous things with it, come on

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,6 +10,9 @@
     "plugin:jsdoc/recommended",
     "plugin:import/recommended"
   ],
+  "plugins": [
+    "mozilla"
+  ],
   "parserOptions": {
     "ecmaVersion": 2022,
     "sourceType": "module"
@@ -17,6 +20,14 @@
   "rules": {
     "import/no-cycle": "error"
   },
+  "overrides": [
+    {
+      "files": ["src/util/*.js"],
+      "rules": {
+        "mozilla/reject-top-level-await": "error"
+      }
+    }
+  ],
   "ignorePatterns": [
     "lib/"
   ]
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,7 @@
   },
   "overrides": [
     {
-      "files": ["src/util/*.js"],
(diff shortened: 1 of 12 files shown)


AprilSylph:
to fix intellisense, can you define a custom jsdoc type and then annotate the function as returning that?

AprilSylph:
also, this PR title (and current contents) has me worried that you're doing too much at once - I think 5746d0dc1ae5da9375e356177653308ec7a4e817 is PR-sized by itself

marcustyphoon:
Was more trying to demo what I was thinking and setting up a test to see if it actually works. I don't know what path you want to take right now; are you still thinking about doing this transformation on all 7 relevant utility files wherever they are imported? I don't think I would do that.

AprilSylph:
what's the alternative?

marcustyphoon:
Among other things, pre-importing utils that contain top-level await, as shown in this PR. ```await Promise.all(['css_map', 'language_data'].map(name => import(getURL(`/util/${name}.js`))));```

AprilSylph:
I'm not a fan of that because that means potentially importing utils that aren't used by any enabled features.

AprilSylph:
I will admit that it is a less invasive modification though... is there a PR for doing just that? I'll accept it as an immediate solution, then we can discuss longer-term solutions afterwards.

marcustyphoon:
Yep. That's what I was illustrating/proposing here: the use of special case logic for css map and language specifically, as they are nontrivially used in other utils. There are certainly ways to configure the extension such that no `window.tumblr` utilities are required for all enabled features to function, but they seem quite rare to me.

From the standpoint of the design reflecting this framing, it feels reasonable to me to obtain a reference to these core APIs during the boot process. I couldn't figure out a way to pass this down to dynamically imported js files without setting a property on `window` in the initialization script and figuring out how to make intellisense work on a a global, however, which is obviously awkward. (Apparently you need to use typescript syntax for VS Code to accept it; also, globals = eew, etc.)

> is there a PR for doing just that? I'll accept it as an immediate solution, then we can discuss longer-term solutions afterwards.

Will do!

PR #1024, opened 2023-03-08
Quote Replies: Activate in notification dropdown

### LINKED ISSUE #862: Quote Replies: Activate in notification dropdown
### Concept

Quote Replies works on the activity page, but not in the activity dropdown. I see the necessary React data in both places, so I assume this should be fairly easy to implement? Or was there a technical reason why this isn't already doable?

marcustyphoon:
### Description

Adds Quote Replies buttons to the activity dropdown.

idk man. css or whatever. there's not a lot of space in there

Resolves #862.

### Testing steps

Use Quote Replies on the activity page. Confirm that it still works as expected.
Use Quote Replies from the notification dropdown. Confirm that it works as expected.
Comment the `@media (hover: hover)` CSS rule section or use a touchscreen and confirm that nothing explodes and the Quote Replies buttons are still visible.

### DIFF
--- a/src/scripts/quote_replies.css
+++ b/src/scripts/quote_replies.css
@@ -29,7 +29,7 @@ button.xkit-quote-replies-dropdown {
 }
 
 @media (hover: hover) {
-  button.xkit-quote-replies:not(.xkit-quote-replies-dropdown) svg {
+  button.xkit-quote-replies svg {
     opacity: 0;
     transform: scale(0);
   }
--- a/src/scripts/quote_replies.css
+++ b/src/scripts/quote_replies.css
@@ -23,24 +23,19 @@ button.xkit-quote-replies:disabled svg {
   transition-property: none;
 }
 
+button.xkit-quote-replies-dropdown {
+  align-self: flex-start;
+  margin: 10px 0 0;
+}
+
 @media (hover: hover) {
-  button.xkit-quote-replies-activity svg {
+  button.xkit-quote-replies:not(.xkit-quote-replies-dropdown) svg {
     opacity: 0;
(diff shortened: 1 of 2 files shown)


marcustyphoon:
- [ ] don't process the notification counters on blog avatars in the blog selector dropdown inside the activity dropdown (causing a silent error)

See, this is why error messages on uncaught promise rejections are useful!

AprilSylph:
I think rather than any of the proposed CSS, I would just do...

```css
button.xkit-quote-replies-dropdown {
  align-self: flex-start;
  margin: 10px 0 0;
  transform: none;
}
```

and remove the `-activity` classname. would commit myself, but am sleepytired. good night

marcustyphoon:
Would you leave the fade-unless-hovered effect or leave it?

AprilSylph:
Between leaving it and leaving it 🤔 I would leave it

(I prefer the consistent behaviour of having the button hide itself regardless of location)

marcustyphoon:
Hah, whoops.

AprilSylph (approved):

AprilSylph:
> Comment the @media (hover: hover) CSS rule section or use a touchscreen and confirm that nothing explodes and the Quote Replies buttons are still visible.

This is the funniest way to learn that Chromium's responsive design mode doesn't have a simple touch simulation toggle. I've been listing "[enable/disable] touch simulation" as a testing step on Redpop PRs for months, I guess I'll try to say "simulate a touchscreen device" instead from now on.

marcustyphoon:
I was going to say you were right, but then I checked: press the three dots -> "add device type!"

PR #1027, opened 2023-03-10
Quote Replies: Soft navigation and new tab modes

### LINKED ISSUE #710: Quote Replies: Possible failure to open draft post

### LINKED ISSUE #857: Quote Replies: Use soft navigation
### Concept

Quote Replies can use `window.tumblr.navigate` to load the editor with its draft.

marcustyphoon:
### Description

This makes Quote Replies open its created draft in the current tab via soft navigation by default, while leaving the previous behavior of opening it in a new tab over the drafts page as a preference. It fixes (hopefully) the possible inconsistency with the open-in-new-tab mode by ensuring that the drafted post appears in the drafts endpoint before opening a tab on the drafts page.

Resolves #710, resolves #857.

This is the "max functionality" PR. We could, of course, only implement the soft navigation method. Justification for why I find this (or #746) interesting:

> One interesting side effect of the fact that we're actually saving a draft (and not hacking the post form button to switch back to "submit") is that, with the current functionality, the user can click the post button without thinking about it and actually just save their draft when they meant to post. In that specific case, being on the drafts page is actually an advantage, as it is then easy to correct this.
> 
> We have the ability to keep this behavior by soft navigating to the drafts page and then soft navigating to the post in question... but this seems mostly silly, as it clobbers the user's current page context for something that is only useful if the user makes a mistake.
> 
> I do, though, think that it would be an improvement to have an "open drafted reply in new tab" checkbox:
> 
> * false: soft navigate directly to the draft
> * true: open a new tab on the user's draft page; in that new tab, soft navigate to the draft
> 
> This doesn't take any additional code, really; we already have the code to take the correct action when the new tab is opened.
> 
> Also, I think that changing the button status programmatically makes even more sense now. Gutenblr has automatic saving to drafts, now, therefore the functionality that we would be implementing if we did so is now essentially the same as the ideal of "opening the post form with the right text in it."
> 
> Assuming that you closed it for fragility and complexity reasons, though, those are of course still valid and I share the concerns I am projecting on to you for the duration of this sentence.


AprilSylph:
Didn't you say that using `navigate` to open the post form prevents failure? If so, the new tab behaviour could be changed from clicking the edit button (which relies on the draft being in the timeline) to using `navigate` there as well (which does not).

marcustyphoon:
Oh, yeah, that's true. I wonder if saving the edited draft refreshes the timeline and makes the missing post appear in that scenario (i.e. so could press post on it). I would guess that it probably does but I could imagine ways the react code could work where it does not.

 - [ ] do that

AprilSylph (changes requested):
Taking this out of my review queue until the new tab mode also uses `navigate`

marcustyphoon:
I was going to say "I'll get around to it at some point but I want to properly test what happens if the draft existing is delayed," but I guess this is technically a bugfix PR so maybe I should prioritize it a bit more.

marcustyphoon:
> I wonder if saving the edited draft refreshes the timeline and makes the missing post appear in that scenario (i.e. so you could press post on it).

Nope, it doesn't. That's unfortunate.

marcustyphoon:
Okay, so we can remove `waitForDraft` with the only minor drawback being that it's possible for a user with the new tab mode setting enabled to quote a reply, have the reply open, save the reply, and then wind up on a drafts channel page that doesn't have the draft they just saved until they refresh/navigate away and back. Or, we could leave it.

AprilSylph (changes requested):
I don't like the notification appearing under the post form glass. Let's only use `notify` in the new tab mode, and only as a fallback.

Also, forgot this in the previous Quote Replies PR... can we update the description to drop the "from the Activity page" now that that's no longer accurate? "Reply to reply notifications" would be good, I think.

src/scripts/quote_replies.js
```diff
@@ -97,29 +98,38 @@ const quoteReply = async (tumblelogName, notificationProps) => {
...
+  if (newTab) {
+    await browser.storage.local.set({ [storageKey]: currentDraftLocation });
+
+    const openedTab = window.open(`/blog/${tumblelogName}/drafts`);
+    if (openedTab) {
+      return;
+    } else {
+      notify(displayText);
+      browser.storage.local.remove(storageKey);
+    }
```
AprilSylph, line 113:
```suggestion
    if (openedTab === null) {
      browser.storage.local.remove(storageKey);
      notify(displayText);
    }
```

src/scripts/quote_replies.js
```diff
@@ -97,29 +98,38 @@ const quoteReply = async (tumblelogName, notificationProps) => {
...
   }
+
+  navigate(currentDraftLocation);
```
AprilSylph, line 116:
```suggestion
  } else {
    navigate(currentDraftLocation);
  }
```

AprilSylph (changes requested):
essentially ignoring the preference as a fallback doesn't feel great, if we're going to have a new tab mode at all I think we should keep the current flow; the browser typically notifies the user if a popup was blocked with the option to allow it, so along with the notification this gives them a path to get the behaviour they opted into

marcustyphoon:
(test comment ignore this sorry)

AprilSylph (approved):

PR #1032, opened 2023-03-19
Open in Tabs: Fix in blog card modals/blog view modals

### LINKED ISSUE #1019: Open in Tabs: Not functional in blog card modals/blog view modals
### Platform

MacOS 13.2.1

### Browser

Chrome 110.0.5481.177 arm64

### Addon version

master

### Details

Putting aside which URL Open in Tabs actually navigates to, as reported in #751, links to blogs in blog cards and the blog view modal are not opened in tabs at all. (Calling stopPropagation from a delegated jQuery `on` handler, for whatever reason, doesn't cancel redpop's soft navigation.)

AlecBlance:
### Description

This fixes #1019 by applying pageModifications targeting `a[role="link"][target="_blank"]` and attaching click event listener. 

### Testing steps

- While Open in Tabs is enabled, access the modals as shown the in the images below. 
- Click on the any of the links (e.g. blog name)
- On-dashboard blog view won't show but instead it will open a new tab as intended. 

image
image

### DIFF
--- a/src/scripts/open_in_tabs.js
+++ b/src/scripts/open_in_tabs.js
@@ -1,10 +1,10 @@
-import { pageModifications } from '../util/mutations.js';
-
 const linkSelector = 'a[role="link"][target="_blank"]';
 
-const onClickExternalLink = event => event.stopPropagation();
-
-const processLinks = links => links.forEach(link => link.addEventListener('click', onClickExternalLink));
+const onDocumentClick = event => {
+  if (event.target.matches(`${linkSelector}, ${linkSelector} *`)) {
+    event.stopPropagation();
+  }
+};
 
 const onClickBlogViewLink = event => {
   event.stopPropagation();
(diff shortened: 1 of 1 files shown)


AprilSylph:
I'm pretty sure this could be done without using `pageModifications` or multiple event listeners - just one event listener attached to the document root, listening to the [capturing phase](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#usecapture) and stopping the event's propagation immediately so that it never even reaches the document body.

Unfortunately, there's no way to do this with jQuery, so we have to do the target filtering manually:
```js
const linkSelector = 'a[role="link"][target="_blank"]';
const onDocumentClick = event => {
  if (event.target.matches(linkSelector)) {
    event.stopPropagation();
  }
};

export const main = () => document.documentElement.addEventListener('click', onDocumentClick, { capture: true });
export const clean = () => document.documentElement.removeEventListener('click', onDocumentClick, { capture: true });
```

AprilSylph (approved):
Works well! Good catch with the selector code.

And of course, thank you for contributing to the project!

PR #1038, opened 2023-03-22
Seen Posts: Fix faded checkmarks in mobile view

### LINKED ISSUE #986: Seen Posts: Faded checkmarks in mobile view
### Platform

MacOS 10.15.7 x86

### Browser

Ungoogled Chromium v110.0.5481.100

### Addon version

all-my-prs; master

### Details

The code to fade user avatars when Seen Posts is in avatar-only mode also fades checkmarks when posts are in the mobile view (<990px width).

```css
/* etc */ article > :first-child img {
    opacity: 0.5;
}
```

In mobile mode, the first child of the article element is the entire header, so this targets all images in it.

_(Yes, the screenshot isn't from master, but I did check)_

AlecBlance:
### Description

This fixes #986 by adding a media rule with the max-width of 990px, and applying `opacity: 1;` to the other images besides the first image (avatar). 

### Testing steps
- Turn on Seen Posts > Dim avatars
- View a post with checkmarks 
- Refresh and locate the same post
- Change the window size to 990px or lower for the mobile view 

Before:
current bug

After: 
intended function

### DIFF
--- a/src/scripts/seen_posts.css
+++ b/src/scripts/seen_posts.css
@@ -1,12 +1,6 @@
 body:not(.xkit-seen-posts-only-dim-avatar) .xkit-seen-posts-seen:not(:hover),
-body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article > :first-child img {
+body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article > :first-child  > :first-child img {
   opacity: 0.5;
-}
-
-@media (max-width: 990px) {
-  body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article > :first-child > :not(:first-child) img {
-    opacity: 1;
-  }
 }
 
 body.xkit-seen-posts-hide .xkit-seen-posts-seen article {
--- a/src/scripts/seen_posts.css
+++ b/src/scripts/seen_posts.css
@@ -3,6 +3,12 @@ body.xkit-seen-posts-only-dim-avatar .xkit-seen-posts-seen:not(:hover) article >
   opacity: 0.5;
 }
 
+@media (max-width: 990px) {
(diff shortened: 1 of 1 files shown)


marcustyphoon:
My guess is this can be done more cleanly by making the targeted faded element location more specific! For example, the "unsticky tab bar" script that you've seen makes use of the `keyToCss` helper to target Tumblr's translated `avatar` css class (as of this writing, `RYkKH`, but it occasionally changes) when it targets avatar elements. This would require the CSS in question to be added as a dynamic style element instead of as a static stylesheet, of course, which you can see in the same script!

https://github.com/AprilSylph/XKit-Rewritten/blob/35ac5557d13831ac6f1314606c04f375e21e89cf/src/scripts/tweaks/unsticky_tab_bar.js#L1-L10

To see documentation about the Tumblr CSS map, if you aren't familiar, you can check out https://github.com/tumblr/docs/blob/master/web-platform.md! Additionally, a bunch of us have scripts that we paste into the browser console to translate everything on the page to make it easy to write selectors of this kind; here's mine: 

https://gist.github.com/marcustyphoon/22a2554e35ae59b84fd0365cd49c4b28

(I actually just run a user script that does this all the time, but that's perhaps a bit extreme :D)

AlecBlance:
Thanks for the suggestion! I looked it at again and I think I can make it more specific. Instead of adding `keyToCss` to target the Tumblr's translated `avatar` css class, I altered the css a bit by adding another `> :first-child`. 

```css
/* etc */ article > :first-child  > :first-child img {
  opacity: 0.5;
}
```

This targets the same ancestor div of the `avatar` class. Let me know what you think. I'll implement your suggestion if this doesn't fit. Thanks again! 

Desktop: 
```html
<article>
    <div>            
        <div>       <!-- Ancestor div -->
            ...
            <img>   <!-- Avatar image -->
        </div>
    </div>
    <header>
    ...
</article>
```
Mobile: 
```html
<article>
    <header>
        <div>       <!-- Ancestor div -->
            ...
            <img>   <!-- Avatar image -->
        </div>
    </header>
    ...
</article>
```

AprilSylph (approved):

Issue #1035, opened 2023-03-19
New AccessKit feature: normal width scroll bar


marcustyphoon:
### Discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/1021

<div type='discussions-op-text'>

<sup>Originally posted by **krc** March  7, 2023</sup>
Tumblr set the scrollbar width to narrow, I find it really hard to use like that. I have a bookmarklet that works to make it wide, but I have to click it every time I go to tumblr or refresh.

</div>

<div type='discussions-op-text'>

<sup>Originally posted by @marcustyphoon in https://github.com/AprilSylph/XKit-Rewritten/discussions/1021#discussioncomment-5261864</sup>

Awesome!

I think this could be a reasonable addition to AccessKit, maybe? Kind of weird for it not to affect Chrome/Edge/Opera/etc, but I guess "in Firefox" isn't too long to add to the label.

</div>

PR #1040, opened 2023-03-26
[AccessKit] New script: normal width scroll bar in Firefox

michellepanzironi:
### Description
  This PR aims to resolve [#1035]( https://github.com/AprilSylph/XKit-Rewritten/issues/1035) by applying `scrollbar-width: auto !important` to the root, the body, and the scrollbar-container element in blog view.  

  Originally discussed in #1021. 

### Testing steps
In Firefox, open any Tumblr page and the scrollbars default to `scrollbar-width: thin` instead of the standard `auto` width.

Feature Off:
image

Feature On:
image


AprilSylph (changes requested):
Welcome to the codebase, and thanks for picking this up!

Left a few suggestions for improvements, otherwise looks good.

src/scripts/accesskit.css
```diff
@@ -50,6 +50,11 @@ body.accesskit-no-user-colours article span[style^="color"] {
   color: inherit !important;
 }
 
+body.accesskit-normal-width-scrollbar,
+html.accesskit-normal-width-scrollbar,
+body.accesskit-normal-width-scrollbar [data-testid="scroll-container"] {
```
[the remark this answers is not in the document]
michellepanzironi: Makes sense, we do want this to affect all the scrollbars.

The reason I specified was because `body.accesskit-normal-width-scrollbar div` creates a visible scrollbar on the horizontal scroll on the dashboard tab selector (see below) which is otherwise is set to `none`. 

I will need to specify something to exclude that. Would it be appropriate to try to use the `aria-label=scroll` properties to select the dashboard tabs instead? 

With:
image

Without:
image

michellepanzironi:
@AprilSylph Thank you for your comments. I'm happy to contribute to XKit!

src/scripts/accesskit.css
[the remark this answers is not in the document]
michellepanzironi: Or is this actually desired behavior for this script as is? 🤔

michellepanzironi:
@AprilSylph I added this to prevent the scrollbar from showing up on the dashboard selector tabs at the top 

```
body.accesskit-normal-width-scrollbar main > div:nth-child(4) div div {
  scrollbar-width: none !important;
}

**Comment by AprilSylph** [comment]
That's not stable either... again, using the same selector that Tumblr uses is always the best bet. Here, it means you can't use the static stylesheet, and will need to create a style element in the script and make use of the `keyToCss` function to target the element. I use this script to expose source names on Tumblr: https://gist.github.com/AprilSylph/511b809ffec1f4e3b8bcb4af93c1f8cd

**Comment by AprilSylph** [comment]
In this case, you'll want to target `keyToCss('timelineHeaderNavInner')`.

**Comment by michellepanzironi** [comment]
@marcustyphoon yes, it seems to work without the `!important`s. Will remove.

**Comment by AprilSylph** [comment]
@marcustyphoon the static CSS in this PR actually doesn't use `!important`, the specificity is just higher...

- Specificity of `.WVvBo` (the rule with `scrollbar-width: none`): 0 ID, 1 CLASS, 0 TYPE
- Specificity of `body.accesskit-normal-width-scrollbar div`: 0 ID, 1 CLASS, 2 TYPE

Your line of thinking is good, though... we could cheat the specificity rules by using `:where()` to make the rule intentionally less specific! (`:where()` [always has a specificity of zero](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#difference_between_is_and_where).)

File: src/scripts/accesskit.css [hunk 2]
```diff
@@ -50,6 +50,11 @@ body.accesskit-no-user-colours article span[style^="color"] {
   color: inherit !important;
 }
 
+body.accesskit-normal-width-scrollbar,
+html.accesskit-normal-width-scrollbar,
+body.accesskit-normal-width-scrollbar div {
```
AprilSylph, line 55: I think this should make the generated style element unnecessary?
Specificity of the second rule should be 0, 0, 1 if I've counted right

```suggestion
.accesskit-normal-width-scrollbar:is(html, body),
:where(.accesskit-normal-width-scrollbar) div {
```

AprilSylph: updated the suggestion and comment to make sure we actually have enough specificity to not rely on CSS insertion order 😅

src/scripts/accesskit/normal_width_scrollbar.js
```diff
@@ -0,0 +1,21 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const className = 'accesskit-normal-width-scrollbar';
+const styleElement = buildStyle(`
+  .${className}
+  ${keyToCss('timelineHeaderNavInner')} { 
+    scrollbar-width: none; 
+}`
+);
```
AprilSylph, line 10:
```suggestion
const className = 'accesskit-normal-width-scrollbar';
```

src/scripts/accesskit/normal_width_scrollbar.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const main = async () => {
+  document.documentElement.classList.add(className);
+  document.body.classList.add(className);
+  document.documentElement.append(styleElement);
```
AprilSylph, line 15:
```suggestion
```

src/scripts/accesskit/normal_width_scrollbar.js
```diff
@@ -0,0 +1,21 @@
...
+};
+export const clean = async () => {
+  document.documentElement.classList.remove(className);
+  document.body.classList.remove(className);
+  styleElement.remove();
```
AprilSylph, line 20:
```suggestion
```

AprilSylph (commented):
Tested it out, and can confirm that the static CSS is now doing the job by itself! Hooray for specificity hacks :partying_face:

PR #1049, opened 2023-04-09
update live carousel selector

### LINKED ISSUE #1047: Updated Tumblr Live carousel is not hidden
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 111.0.5563.146

### Addon version

v0.21.4

### Details

I designed the Tumblr Live carousel hiding tweak specifically to hide the first iteration of the carousel code, which imported its content from Tumblr's livestreaming partner. Now that the carousel is properly integrated with the site, the tweak is currently nonfunctional.

marcustyphoon:
### Description

This updates the CSS selector used to hide the Tumblr live carousel. I don't actually think the iframe selector is necessary, but I didn't save the HTML of the old element, so better safe than sorry.

Resolves #1047.

### Testing steps

Enable Tweaks and the tweak to hide the Live carousel on an account with the new, iframe-less carousel. Confirm that it works.

### DIFF
--- a/src/scripts/tweaks/no_live.js
+++ b/src/scripts/tweaks/no_live.js
@@ -12,7 +12,7 @@ const processFrames = frames =>
 
 export const main = async function () {
   pageModifications.register(
-    '[data-timeline="/v2/timeline/dashboard"] iframe[src^="https://api.gateway.tumblr-live.com/"]',
+    `[data-timeline="/v2/timeline/dashboard"] :is(iframe[src^="https://api.gateway.tumblr-live.com/"], ${keyToCss('liveMarqueeContainer')})`,
     processFrames
   );
   document.documentElement.append(styleElement);


nightpool (approved):

src/scripts/tweaks/no_live.js
```diff
@@ -12,7 +12,7 @@ const processFrames = frames =>
 
 export const main = async function () {
   pageModifications.register(
-    '[data-timeline="/v2/timeline/dashboard"] iframe[src^="https://api.gateway.tumblr-live.com/"]',
+    `[data-timeline="/v2/timeline/dashboard"] :is(iframe[src^="https://api.gateway.tumblr-live.com/"], ${keyToCss('liveMarqueeContainer')})`,
```
nightpool, line 15: the nested `:is` here is a little awkward and unnecessary but not sure it's worth monkeying around with.

marcustyphoon: which one? sorry, not sure what you're referring to

nightpool: the generated code here is going to be something like `[data-timeline="/v2/timeline/dashboard"] :is(iframe[src^="https://api.gateway.tumblr-live.com/"], :is(.xxxx, .xxyy))`, because keyToCss already comes with an `:is` wrapper

marcustyphoon: indeed it does! `:is` is great for selector composition.

this could be removed by modifying the css map utilities to pass through non-translatable css strings (i.e. `keys.flatMap(key => cssMap[key] ?? key)`) and then doing `[data-timeline="/v2/timeline/dashboard"] ${keyToCss('iframe[src^="https://api.gateway.tumblr-live.com/"]', 'liveMarqueeContainer')}`, though it doesn't confer a performance benefit and imo isn't as semantically clear.

Issue #1058, opened 2023-04-29
add mechanism for marking features as deprecated


AprilSylph:
Feature manifests should accept a boolean `"deprecated"` property.

When this value is `true`, it should be hidden in the control panel, unless:
- the feature is currently enabled, or
- any of the feature's preferences have values present in storage

This will have the effect of making the feature unavailable only to users who have never used the feature on their current installation.

Timestamps should be the first feature to be deprecated.

marcustyphoon:
Inconsistency: script preferences are only saved to storage when you change them. Thus, this will behave differently if a user:

- Enables Timestamps
- [some time passes]
- Updates to an XKit Rewritten version in which Timestamps is deprecated
- Disables Timestamps

or if the user instead:

- Enables Timestamps
- Toggles a checkbox, decides they don't like it, toggles it back
- [some time passes]
- Updates to an XKit Rewritten version in which Timestamps is deprecated
- Disables Timestamps

In the first case, Timestamps will vanish from the preference pane even though the user has used it before and very well might have been using it minutes ago.

This could be solved by pushing a version that stores, I dunno, a `hasEverBeenEnabledScripts` array, waiting a bit, and then pushing the deprecation feature.

AprilSylph:
Ah damn, I forgot that scripts don't save their defaults to storage on first usage anymore.

I don't think this inconsistency super matters though. Ultimately the goal is to phase the feature out of use, and conditionally continuing to provide access to it is primarily just a way to not "break" existing setups.

If anything, to resolve this inconsistency, I would rather go harder into hiding the feature(s) and ignore the presence of preferences entirely.

marcustyphoon:
I do think ignoring preferences makes sense. (I was going to say something about, like, "what if we someday deprecate quick tags; I don't want users to lose access to their tag bundle data," but since we export preferences to JSON I actually have no qualms about that.)

I also think that disabling a script, dismissing the preference popup, opening the popup, and having that script no longer there feels a little... the kind of thing I would write a GitHub issue about. I hold this opinion very loosely, especially because it's Timestamps and who cares, but I pushed a version of #1190 that shows disabled deprecated scripts if and only if only if the user had them enabled at the moment of deprecation; that seems like a possibility to me.

PR #1063, opened 2023-05-06
Meatballs util: Add XKit labels to meatball menu items

marcustyphoon:
### Description

> We’re working to add the ability to “mute” posts on the web, which will prevent future push notifications and activity items from being created about that post, until the post is unmuted. After that’s out, we’re also extending the mute functionality on all platforms to allow temporary mutes of notifications about a post.
– https://www.tumblr.com/changes/716500434236342272/friday-may-5th-2023

Due to the imminent implementation of native notification "muting," it is likely to be confusing to have multiple menu items that do essentially the same thing. This adds a label to NotificationBlock's menu item disambiguating this.

(Not sure if you plan to simply remove the script? Deleting people's lists of notifcation-blocked posts sucks, but we definitely don't want another "I use the XKit feature instead of the native one" situation.

Also, this got me thinking about putting an XKit icon on all XKit-added meatball items, again.

Also, dear whoever decided to call the native feature "muting," I have... a complaint.)

### DIFF
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -9,7 +9,7 @@ import { userBlogNames } from '../util/user.js';
 
 const storageKey = 'notificationblock.blockedPostTargetIDs';
 const meatballButtonBlockId = 'notificationblock-block';
-const meatballButtonBlockLabel = 'Block notifications';
+const meatballButtonBlockLabel = 'Block notifications (XKit)';
 const meatballButtonUnblockId = 'notificationblock-unblock';
 const meatballButtonUnblockLabel = 'Unblock notifications';
 const notificationSelector = keyToCss('notification');
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -9,7 +9,7 @@ import { userBlogNames } from '../util/user.js';
 
 const storageKey = 'notificationblock.blockedPostTargetIDs';
 const meatballButtonBlockId = 'notificationblock-block';
-const meatballButtonBlockLabel = 'Block notifications (XKit)';
+const meatballButtonBlockLabel = 'Block notifications';
 const meatballButtonUnblockId = 'notificationblock-unblock';
 const meatballButtonUnblockLabel = 'Unblock notifications';
 const notificationSelector = keyToCss('notification');
--- a/src/util/meatballs.js
+++ b/src/util/meatballs.js
@@ -1,20 +1,12 @@
 import { keyToCss } from './css_map.js';
 import { dom } from './dom.js';
-import { buildStyle, postSelector } from './interface.js';
+import { postSelector } from './interface.js';
 import { pageModifications } from './mutations.js';
 import { inject } from './inject.js';
 import { blogData, timelineObject } from './react_props.js';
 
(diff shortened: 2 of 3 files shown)


AprilSylph:
I think prepending every custom meatball menu item label with `[XKit]` would be a good idea.

Also, I plan to deprecate the script rather than outright remove it (#1058). Probably gonna pick that issue up for myself unless you beat me to it.

marcustyphoon:
Yeah, I was thinking about the deprecation method too. I would be tempted to disable the ability to block notifications on new posts, even, but that's probably more trouble than it's worth.

nightpool:
it probably wouldn't be too hard to offer some sort of auto import
functionality, even today since the API appears to be finished.

src/scripts/notificationblock.js
```diff
@@ -9,7 +9,7 @@ import { userBlogNames } from '../util/user.js';
 
 const storageKey = 'notificationblock.blockedPostTargetIDs';
 const meatballButtonBlockId = 'notificationblock-block';
-const meatballButtonBlockLabel = 'Block notifications';
+const meatballButtonBlockLabel = 'Block notifications (XKit)';
```
AprilSylph, line 12:
```suggestion
const meatballButtonBlockLabel = '[XKit] Block notifications';
```

src/scripts/notificationblock.js
```diff
@@ -9,7 +9,7 @@ import { userBlogNames } from '../util/user.js';
...
 const meatballButtonUnblockId = 'notificationblock-unblock';
 const meatballButtonUnblockLabel = 'Unblock notifications';
```
AprilSylph, line 14:
```suggestion
const meatballButtonUnblockLabel = '[XKit] Unblock notifications';
```

AprilSylph (changes requested):
really the direction I would like this PR to go is for the meatballs util itself to prepand `[XKit] ` to every label, but failing that:

marcustyphoon:
Sure thing! Made that change. Additionally, I tried adding actual XKit Rewritten icons in #1064 and #1065, which I find cleaner. (I also tried messing with the text labels a bit, like `[XKit] ` vs `XKit: ` and setting `font-weight: normal`; none felt great to me.)

src/util/meatballs.js
```diff
@@ -1,13 +1,21 @@
...
 import { blogData, timelineObject } from './react_props.js';
 
 const postHeaderSelector = `${postSelector} article > header`;
 const blogHeaderSelector = `[style*="--blog-title-color"] > div > div > header, ${keyToCss('blogCardHeaderBar')}`;
 
+const styleElement = buildStyle(`
+.xkit-meatball-button::before {
+  content: "[XKit] ";
+  white-space: pre-wrap;
+}
```
AprilSylph, line 15: this can just go in the meatballs stylesheet if we're not interpolating anything

AprilSylph (changes requested):

src/util/meatballs.js
marcustyphoon: oh, right

AprilSylph (approved):

PR #1065, opened 2023-05-07
Meatballs util: Add color-matched XKit icons to meatball menu items

marcustyphoon:
### Description

This adds XKit icons before meatball menus, using CSS image masking to allow the icon to use the `black` palette color. This requires a prefix for Chromium-based browsers.


marcustyphoon:
I'm going to be totally honest and say I have no idea why this happens. Oddity with flex children being display:inline, or something?

- [ ] fix this

PR #1067, opened 2023-05-09
Meatballs util: Fix visual weirdness with long menu items

### LINKED ISSUE #1066: Meatballs util: visual weirdness with long menu items
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 112.0.5615.137

### Addon version

all of my non-draft PRs (yes, this is not—as far as I know—easily observable in the current release since none of our meatballs items are long enough, but it is with font size adjustments, such as from Palettes for Tumblr)

### Details

I love pseudo elements.

Requires a rewrite of #1063 (inline pseudo or just adding the label to the DOM instead; preference?) or tweaks to #1065 if it replaces it.

marcustyphoon:
### Description

Removes display: flex from meatballs items so the `before` pseudo element can be displayed inline.

resolves #1066


AprilSylph (approved):
deleting code > writing code

PR #1071, opened 2023-05-14
NotificationBlock: UI to show blocked posts

marcustyphoon:
Okay, so, this is currently a catastrophe, but.

### Description

This implements a UI for NotificationBlock's preference pane similar to PostBlock's, including additions in #719. That is, it displays a list of posts that one has blocked notifications for, including the ability to click to navigate to the posts in question so that one can migrate to the upcoming native "notification muting" feature instead.

This requires looking up which blog each notification-blocked post is from, since we didn't previously record that. This performs said lookup by trying to load every unknown post on every one of the user's blogs as a background task in `main()`, setting `false` if all lookups fail. ~~This means that a post whose notifications are blocked that is on a blog the user is not currently logged into will never get a link created for it, which kind of sucks, but the alternative (API spam on every page load or keeping track in storage of every UUID which has been tested for each unknown post ID) are messy.~~

~~This currently adds an additional dictionary matching blog UUIDs to blog names so that the preference UI can show which blog a notification-blocked post is on, but I'll probably remove that as it's quite complex and not really necessary.~~

**Edit:** In the case where a user has blocked notifications from a root post that they did not create before the creation of this PR, this will always fail. I can't currently think of a way around this: Tumblr intentionally does not expose looking up arbitrary posts by ID, correct?

### Testing steps

Oof. There is a lot of logic to test here, including importing blocked posts from before this PR as well as blocking posts with this PR merged, and being/not being logged into an account with access to the blocked post in question when clicking a linked post in the preference pane.


marcustyphoon:
Some updates: for simplicity, the next commits remove the blog names dictionary (removing the ability to display the user blog each post is from in the preference pane) and performs the UUID lookup when the user clicks the link. This ensures that lookups will work correctly if the same XKit Rewritten profile is logged into multiple blogs.

This relegates the on-extension-load lookup to be an optimization that will be run one time for each post the first time the user loads a page with this PR merged.

PR #1084, opened 2023-06-03
New Tweak: Restore post links in attribution header

### LINKED ISSUE #1077: Add a link to the previous post in a reblog chain
### Concept

Previously, Tumblr's "X reblogged Y" header included a link to Y's post which was reblodded. Tumblr's latest update removed this, changing it to the link to Y's blog instead. Not sure whether this should be under i.e. Tweaks, or if it should be an entire new feature (so that if Tumblr officially adds it back, it can be removed/deprecated, c.f.  #1058).

inklesspen:
### Description
Tumblr recently removed the links to the actual post (and the reblogged post, if any) from the header, and [apparently intends](https://www.tumblr.com/changes/718768407758225408/tuesday-may-30th-2023) to keep this change.

However, the links are helpful for many people, so now there is a tweak which restores them.

Fixes #1077.

(Note: the tweak does not attempt to remove the helpful links when disabled. Of course, they will be "removed" on next page reload.)

### Testing steps

1. Enable "Restore links to individual posts in the post header" in the Tweaks module.
2. Examine the links on the blog names in the post attribution header.

I wasn't entirely sure which arguments to use with `filterPostElements`, so I used "Hide Liked Posts" as a model. Feel free to suggest a better way.

### DIFF
--- a/src/scripts/tweaks/restore_attribution_links.js
+++ b/src/scripts/tweaks/restore_attribution_links.js
@@ -1,5 +1,4 @@
 import { onNewPosts } from '../../util/mutations.js';
-import { filterPostElements } from '../../util/interface.js';
 import { keyToCss } from '../../util/css_map.js';
 import { timelineObject } from '../../util/react_props.js';
 
@@ -7,7 +6,7 @@ const postAttributionSelector = keyToCss('postAttribution');
 const reblogAttributionSelector = keyToCss('reblogAttribution');
 
 const processPosts = async function (postElements) {
-  filterPostElements(postElements).forEach(async postElement => {
+  postElements.forEach(async postElement => {
     const { postUrl, rebloggedFromUrl } = await timelineObject(postElement);
     const postAttribution = postElement.querySelector(postAttributionSelector);
     const postAttributionLink = postAttribution?.querySelector('a');
--- a/src/scripts/tweaks/restore_attribution_links.js
+++ b/src/scripts/tweaks/restore_attribution_links.js
@@ -1,0 +1,33 @@
+import { onNewPosts } from '../../util/mutations.js';
+import { filterPostElements } from '../../util/interface.js';
+import { keyToCss } from '../../util/css_map.js';
+import { timelineObject } from '../../util/react_props.js';
+
+const timeline = /\/v2\/timeline\/dashboard/;
+
+const postAttributionSelector = keyToCss('postAttribution');
+const reblogAttributionSelector = keyToCss('reblogAttribution');
+
+const processPosts = async function (postElements) {
+  filterPostElements(postElements, { timeline }).forEach(async postElement => {
+    const { postUrl, rebloggedFromUrl } = await timelineObject(postElement);
+    const postAttribution = postElement.querySelector(postAttributionSelector);
+    const postAttributionLink = postAttribution?.querySelector('a');
+    if (postAttributionLink !== undefined) {
+      postAttributionLink.href = postUrl;
+    }
+    const reblogAttribution = postElement.querySelector(reblogAttributionSelector);
+    const reblogAttributionLink = reblogAttribution?.querySelector('a');
+    if (reblogAttributionLink !== undefined && rebloggedFromUrl !== undefined) {
+      reblogAttributionLink.href = rebloggedFromUrl;
+    }
+  });
+};
+
+export const main = async function () {
+  onNewPosts.addListener(processPosts);
+};
+
+export const clean = async function () {
+  onNewPosts.removeListener(processPosts);
+};
--- a/src/scripts/tweaks/restore_attribution_links.js
+++ b/src/scripts/tweaks/restore_attribution_links.js
@@ -3,13 +3,11 @@ import { filterPostElements } from '../../util/interface.js';
 import { keyToCss } from '../../util/css_map.js';
 import { timelineObject } from '../../util/react_props.js';
 
-const timeline = /\/v2\/timeline\/dashboard/;
-
 const postAttributionSelector = keyToCss('postAttribution');
 const reblogAttributionSelector = keyToCss('reblogAttribution');
 
 const processPosts = async function (postElements) {
-  filterPostElements(postElements, { timeline }).forEach(async postElement => {
+  filterPostElements(postElements).forEach(async postElement => {
     const { postUrl, rebloggedFromUrl } = await timelineObject(postElement);
     const postAttribution = postElement.querySelector(postAttributionSelector);
     const postAttributionLink = postAttribution?.querySelector('a');
--- a/src/scripts/tweaks/restore_attribution_links.js
+++ b/src/scripts/tweaks/restore_attribution_links.js
@@ -2,20 +2,18 @@ import { onNewPosts } from '../../util/mutations.js';
 import { keyToCss } from '../../util/css_map.js';
 import { timelineObject } from '../../util/react_props.js';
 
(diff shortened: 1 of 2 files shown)


src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -0,0 +1,33 @@
+import { onNewPosts } from '../../util/mutations.js';
+import { filterPostElements } from '../../util/interface.js';
+import { keyToCss } from '../../util/css_map.js';
+import { timelineObject } from '../../util/react_props.js';
+
+const timeline = /\/v2\/timeline\/dashboard/;
```
nightpool, line 6: Why only apply this tweak to the dashboard? I think it's helpful everywhere posts are shown.

inklesspen: Like I said, I do not entirely understand the intended use of the arguments to `filterPostElements`, so I followed "Hide Liked Posts" as a model.

If you want to suggest a different set of arguments, please feel free.

AprilSylph: Omitting `timeline` entirely should make the callback action on all posts.

inklesspen: Done.

inklesspen:
Usernames in the post "trail" could also be made into links to the target post… For instance, in [this post](https://tumblr.com/inklesspen/719106819163897856), we could have not just the username and reblog source in the header, but the "neil-gaiman" in the post trail be a link (in this case to https://neil-gaiman.tumblr.com/post/717333013893087232).

Should this be folded into the same tweak, or a separate tweak?

src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -3,13 +3,11 @@ import { filterPostElements } from '../../util/interface.js';
...
 import { timelineObject } from '../../util/react_props.js';
 
-const timeline = /\/v2\/timeline\/dashboard/;
-
 const postAttributionSelector = keyToCss('postAttribution');
 const reblogAttributionSelector = keyToCss('reblogAttribution');
 
 const processPosts = async function (postElements) {
-  filterPostElements(postElements, { timeline }).forEach(async postElement => {
+  filterPostElements(postElements).forEach(async postElement => {
```
AprilSylph, line 10: with no arguments, this becomes synonymous with...

```suggestion
  postElements.forEach(async postElement => {
```

inklesspen: Does it? Looks like filterPostElements is still doing some work:

```
postElements = postElements.map(element => element.closest(postSelector)).filter(Boolean);
```

But if that work is not necessary here, then I can drop the function call.

inklesspen: Ah, I see, postSelector is already used in the onNewPosts logic.

nightpool:
> Usernames in the post "trail" could also be made into links to the target post… For instance, in [this post](https://tumblr.com/inklesspen/719106819163897856), we could have not just the username and reblog source in the header, but the "neil-gaiman" in the post trail be a link (in this case to https://neil-gaiman.tumblr.com/post/717333013893087232).

Since this link is already accessible (by clicking on the reblog trail item header, just not on any specific blog), I'm not sure if it's necessary or is worth a tweak. It's not adding a new feature that doesn't exist, just changing things "back to how they used to be". I would suggest leaving it out for now until we see how this feature shakes out (Tumblr staff have mentioned multiple times that this change is just pt. 1 of a larger change to reblog trail items)

inklesspen:
> Since this link is already accessible

It's accessible, but requires extra steps… Still, as you say, worth waiting for now.

src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -0,0 +1,30 @@
+import { onNewPosts } from '../../util/mutations.js';
+import { keyToCss } from '../../util/css_map.js';
+import { timelineObject } from '../../util/react_props.js';
+
+const postAttributionSelector = keyToCss('postAttribution');
+const reblogAttributionSelector = keyToCss('reblogAttribution');
```
AprilSylph, line 6:
```suggestion
const postAttributionLinkSelector = `${keyToCss('postAttribution')} a`;
const reblogAttributionLinkSelector = `${keyToCss('reblogAttribution')} a`;
```

src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -0,0 +1,30 @@
...
+
+const processPosts = async function (postElements) {
+  postElements.forEach(async postElement => {
+    const { postUrl, rebloggedFromUrl } = await timelineObject(postElement);
+    const postAttribution = postElement.querySelector(postAttributionSelector);
+    const postAttributionLink = postAttribution?.querySelector('a');
```
AprilSylph, line 12:
```suggestion
    const postAttributionLink = postElement.querySelector(postAttributionLinkSelector);
```

src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -0,0 +1,30 @@
...
+    if (postAttributionLink !== undefined) {
+      postAttributionLink.href = postUrl;
+    }
+    const reblogAttribution = postElement.querySelector(reblogAttributionSelector);
+    const reblogAttributionLink = reblogAttribution?.querySelector('a');
```
AprilSylph, line 17:
```suggestion
    const reblogAttributionLink = postElement.querySelector(reblogAttributionLinkSelector);
```

AprilSylph (commented):
it _looks_ like we can combine some of these `querySelector` calls

inklesspen:
> it _looks_ like we can combine some of these `querySelector` calls

Had to also change some of the `undefined` comparisons to `null`, because `querySelector` returns an element or `null`.

AprilSylph (changes requested):
The last thing this needs is proper click behaviour for the updated links. We have an `onClickNavigate` function that can be imported and attached to hopefully override Tumblr's navigation, though you may have to attach it with [`{ capture: true }`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options), and the function itself may need to be modified to stop propagation (which I don't think should have any adverse effects elsewhere).

If this is too tall an order, I'll try to give it a go myself tonight. Thanks for working on this!

AprilSylph:
Ah, right, these links aren't blog view URLs, so just plugging in `onClickNavigate` doesn't work at all...

AprilSylph:
Thanks for the contribution! This'll go out in the next release, which I guess I'll cut... tomorrow?

inklesspen:
Thanks! I wasn't sure exactly what you meant about the navigation integration (and it seemed to be working well in my local copy) but I'll check out the additional commits.

AprilSylph:
If you just modify the `href` of a link created by the Tumblr web app, clicking it normally (i.e. with no modifier keys held down) will still navigate within the app to the original link location. There's also a few cases (or at least, there were) where the `href` and navigation target are different out of the gate, e.g. a link to `https://staff.tumblr.com` will navigate to `https://www.tumblr.com/staff`.

Intercepting the click event before it reaches its target, and performing our own navigation instead is necessary to override the default behaviour. Otherwise, we would be changing the attribution links to `https://staff.tumblr.com/post/719408723730251776` but a normal click would still go to `https://www.tumblr.com/staff` instead of `https://www.tumblr.com/staff/719408723730251776`.

PR #1086, opened 2023-06-06
Anti-Capitalism: block cruelSummer takeover

marcustyphoon:
### Description

This hides the bottom elements from the "cruel summer" takeover ad, including the start menu and "reminder: turn your computer off before midnight on 12/31/99" animated sticker GIF.

~~In Chromium-based browsers,~~ it also hides the "1999" dashboard tab added by this takeover. ~~I used a `:has()` selector to do this; it is possible via adding our own class so that it will also work in Firefox, but this is difficult to implement without bugs because changing tabs updates the classlists on the tab item wrappers.~~

### Testing steps

Enable anti-capitalism. Confirm that the elements in question no longer appear.

### DIFF
--- a/src/scripts/anti_capitalism.js
+++ b/src/scripts/anti_capitalism.js
@@ -8,7 +8,7 @@ const listTimelineObjectInnerSelector = keyToCss('listTimelineObjectInner');
 
 const styleElement = buildStyle(`
 .${hiddenClass},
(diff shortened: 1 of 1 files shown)


nightpool (approved):
Tested locally, LGTM!

nightpool:
Kind of sad to take the whole start menu away without any way to add it back. I guess that's my only remaining complaint here (compared to the XKit one where at least we have a config option)

PR #1107, opened 2023-06-29
Vertical nav layout: Update sidebar utility

marcustyphoon:
Speedrun, let's go.

### Description

Updates the sidebar utility for the upcoming vertical nav layout and the current updated mobile navigation drawer.

~~The significant code change here is that because we need to prioritize the sidebar over the navigation section when both exist on the page (in the wrong DOM order), I re-query the candidate elements in the pageModifications callback in priority order instead of using the element list given by pageModifications itself.~~ never mind. I can read.

Question: Were XKit's sidebar items supposed to be at the top or the bottom of the mobile navigation drawer? I actually don't remember.

### Testing steps

Load XKit Rewritten on an account with the new vertical nav layout and an account without it. Enable limit checker and confirm that the sidebar item appears at the bottom of the mobile navigation drawer or in the correct location in the right sidebar (above radar):

 - in the dashboard; mobile and desktop views (direct navigation and resize)
 - on an explore page
 - in the inbox
 - on each kind of blog "channel" page
 - on a search page
 - on a tagged page
 - on a "secret dashboard" like https://www.tumblr.com/timeline/trending
 - on a "blogpack" secret dashboard like https://www.tumblr.com/timeline/blogpack?blogs=april,transienturl

Confirm that the sidebar items do not appear:
 - on a settings page in the desktop view (exception: #1115)
 - in the "embedded blog view"

Do it again with ad-free on/off (I did not fully do this; only one of my accounts has ad-free)

### DIFF
--- a/src/util/sidebar.js
+++ b/src/util/sidebar.js
@@ -86,7 +86,7 @@ const desktopContainerSelector = `aside ${keyToCss('desktopContainer', 'summary'
 // Sidebar on most other desktop pages
 const sidebarItemSelector = keyToCss('sidebarItem', 'sidebarContent');
 // Mobile drawer
-const navItemSelector = keyToCss('navItem');
+const navItemSelector = `${keyToCss('drawerContent')} ${keyToCss('navItem')}`;
 
 const sidebarItemSelectors = [desktopContainerSelector, sidebarItemSelector, navItemSelector];
 
--- a/src/util/sidebar.js
+++ b/src/util/sidebar.js
@@ -88,12 +88,8 @@ const sidebarItemSelector = keyToCss('sidebarItem', 'sidebarContent');
 // Mobile drawer
 const navItemSelector = `${keyToCss('drawerContent')} ${keyToCss('navItem')}`;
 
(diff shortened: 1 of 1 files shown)


AprilSylph:
> Question: Were XKit's sidebar items supposed to be at the top or the bottom of the mobile navigation drawer? I actually don't remember.

They were originally in the middle, between the nav items and the blogs list, With the blogs list now under the Account submenu, they should go at the bottom in the top level instead.

AprilSylph:
Question for you: are you also testing these changes against an account that doesn't have the feature preview?

Rather than leaving "testing steps" blank it would be nice if you listed the steps _you_ took to test the changes.

marcustyphoon:
Loosely, yes. Sorry—should have left this as draft until I got an answer to that question!

marcustyphoon:
Adding a note here that this doesn't currently work on https://www.tumblr.com/timeline/blogpack?blogs=april,transienturl with ad-free because it has no sidebar items. Fixable on vertical nav by targeting the footer.

AprilSylph (approved):

PR #1108, opened 2023-06-29
Vertical nav layout: Update unsticky tab bar tweak

marcustyphoon:
### Description

Updates the unsticky tab bar tweak for the upcoming vertical nav layout.

Presumably the `newDesktopLayout` CSS class will eventually go away, at which point this will break again?

### DIFF
--- a/src/scripts/tweaks/unsticky_tab_bar.js
+++ b/src/scripts/tweaks/unsticky_tab_bar.js
@@ -8,6 +8,7 @@ const styleElement = buildStyle(`
     transition: none !important;
   }
   ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')} { top: 69px !important; }
+  ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')}${keyToCss('newDesktopLayout')} { top: 14px !important; }
 `);
 
 export const main = async () => document.documentElement.append(styleElement);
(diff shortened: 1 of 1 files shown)


AprilSylph:
It won't go away immediately, and I'll be able to tell you when the feature is ramped to 100% of users 😁

src/scripts/tweaks/unsticky_tab_bar.js
```diff
@@ -8,6 +8,7 @@ const styleElement = buildStyle(`
     transition: none !important;
   }
   ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')} { top: 69px !important; }
+  ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')}${keyToCss('newDesktopLayout')} { top: 14px !important; }
```
AprilSylph, line 11: Could combine these into one selector with a different approach

```suggestion
  ${keyToCss('post')} ${keyToCss('stickyContainer')} > ${keyToCss('avatar')} {
    --dashboard-tabs-header-height: 0px !important;
  }
```

AprilSylph (changes requested):

src/scripts/tweaks/unsticky_tab_bar.js
marcustyphoon: oh, damn, yeah

AprilSylph (approved):

PR #1110, opened 2023-06-29
Timeformat: Prevent line break in relative time segment

marcustyphoon:
### Description

As a minor readability/aesthetic optimization of longer Timeformat strings shown in posts that are rendered to be narrow, such as in the radar or in masonry views, this uses non-breaking spaces in the optional relative time segment.

(would like to buy non-breaking en space for the option to put the dot on the next line, pls)


AprilSylph:
We do match the user's language for this string... are there any languages in which the relative time is too long to display on one line without overflowing?

Unfortunately, there doesn't seem to be a character that is "a space that is non-breaking unless it _really_ has to break".

Maybe the most robust thing would be to use different attributes for the absolute and relative timestamps (displaying the absolute timestamp with `::before` and the relative timestamp with `::after`) in order to set `display: inline-block` on just the relative timestamp...?

marcustyphoon:
Implemented! I have no idea if it can actually happen "naturally" or not, but I just set the font size in Palettes for Tumblr to be extremely large and it seemed to work just how we want.

marcustyphoon:
> (would like to buy non-breaking en space for the option to put the dot on the next line, pls)

I tried this with the new method and I think it looked worse.

src/scripts/timeformat.css
```diff
@@ -2,11 +2,17 @@
...
 }
 
-[data-formatted-time][title]::after {
+[data-formatted-relative-time]::after {
+  content: attr(data-formatted-relative-time);
+  font-size: .78125rem;
+  display: inline-block;
+}
+
+[data-formatted-time]::before, [data-formatted-time][title]::after {
```
AprilSylph, line 16: The usage of `[title]` is so that the cursor only changes if there is text to show on hover, so should be included in both selectors

```suggestion
[data-formatted-time][title]::before,
[data-formatted-time][title]::after {
```

AprilSylph (changes requested):

src/scripts/timeformat.css
marcustyphoon: Oh, whoops, that's a typo!

AprilSylph (approved):

PR #1125, opened 2023-07-08
New tweak: Hidden following/mutuals indicators

### LINKED ISSUE #1122: Turn Off Showing Mutuals/Followers in Notifications
### Concept

I wanted to know if there's a way to turn off notifications that show if someone you're mutuals with or following you or you're following have interacted with your posts? When I open my notifications, I don't like seeing every single relationship I have with every person there. I don't want to think about who I'm following, who's following me, or who I'm mutuals with to show up next to the notification. It's anxiety inducing. Any way there can be an option to turn it off/get rid of it?

marcustyphoon:
### Description

~~This adds tweaks that remove the bright colors from, and outright hide, the following/mutuals indicators on activity notifications.~~

This adds a tweak that hides the following/mutuals indicators on activity notifications.

Resolves #1122.

### Testing steps

Confirm that the new tweaks remove the bright coloration from or hide the following/mutuals indicators on activity notifications in the activity popover modal.
Confirm that the new tweaks remove the bright coloration from or hide the following/mutuals indicators on activity notifications on the activity page.


src/scripts/tweaks/subtle_activity_mutuals.js
```diff
@@ -0,0 +1,12 @@
+import { keyToCss } from '../../util/css_map.js';
+import { buildStyle } from '../../util/interface.js';
+
+const styleElement = buildStyle(`
+${keyToCss('activity')} ${keyToCss('generalLabelContainer')} {
+  color: RGB(var(--black));
+  background-color: RGB(var(--black), 0.07);
```
AprilSylph, line 7:
```suggestion
  background-color: RGBA(var(--black), 0.07);
```

AprilSylph:
Do we really want two new tweaks for these elements? I think if you want them visible then you want them to stand out.

marcustyphoon:
Ah, I should have figured you'd ask.

I was going to go start laying out my arguments for yes, but I realized while doing so that the "subtle" tweak doesn't really work/do anything on a bunch of the palettes, and also that the person I had seen giving feedback on the color being difficult to read was talking about something else. So, eh.

***I*** want to see them but find the green and purple quite unpleasant, so I am personally using it at the moment, but especially given the chance of user confusion with the current iteration of it, I definitely see the argument against it.

src/scripts/tweaks.json
```diff
@@ -58,6 +58,16 @@
       "default": false,
       "inherit": "tweaks.preferences.subtle_follow"
     },
+    "subtle_activity_mutuals": {
+      "type": "checkbox",
+      "label": "Make following/mutuals indicators on notifications more subtle",
+      "default": false
+    },
```
AprilSylph, line 65:
```suggestion
```

src/scripts/tweaks/subtle_activity_mutuals.js
```diff
@@ -0,0 +1,12 @@
...
+
+const styleElement = buildStyle(`
+${keyToCss('activity')} ${keyToCss('generalLabelContainer')} {
+  color: RGB(var(--black));
+  background-color: RGBA(var(--black), 0.07);
+}
+`);
+
+export const main = async () => document.documentElement.append(styleElement);
+export const clean = async () => styleElement.remove();
```
AprilSylph, line 12: I still don't know what happens if you accept a code suggestion that removes the entire file contents. Does this result in a blank file? Is GitHub smart enough to outright delete it instead?

```suggestion
```

AprilSylph (changes requested):

src/scripts/tweaks/subtle_activity_mutuals.js
marcustyphoon: Hm, let's find out!

AprilSylph (approved):

PR #1128, opened 2023-07-09
New Tweak: Subtle following/mutuals indicators

marcustyphoon:
<details>
<summary>outdated</summary>

Just for funsies for now. No idea how to get this to actually work properly.
</details>

### Description

Adds a tweak that, rather than showing or completely hiding the following/mutuals icons in activity, makes them only appear as icons until the user hovers over them.

As this is a CSS vanity project, this has some unnecessary complexity to implement a smooth show-the-label animation. Specifically, putting a transition on the width property requires a hardcoded final width (grabbed via `getBoundingClientRect`) rather than just setting `width: unset`. It would be quite easy to remove this and make the label instantly appear; mention me if that's desired.

### Testing steps

Enable the tweak and observe its functionality on the activity page and activity modal.


marcustyphoon:
"screw it, whatever. if one is already manipulating text nodes one may as well rely on `getBoundingClientRect` too"

src/scripts/tweaks/subtle_activity_mutuals.js
```diff
@@ -0,0 +1,67 @@
...
+import { dom } from '../../util/dom.js';
+
+const labelSelector = keyToCss('followingBadgeContainer', 'mutualsBadgeContainer');
+
+const spanClass = 'xkit-tweaks-subtle-activity-span';
+
+const styleElement = buildStyle(`
+.${spanClass} {
+  display: inline-block;
+  overflow-x: clip;
```
AprilSylph, line 13: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x#browser_compatibility

we'd need a small bump in the addon manifest for this to be reliable

- `"minimum_chrome_version"`: `"89"` &rarr; `"90"`

(don't we use `overflow: clip` in other places already? if so: WHOOPS)

AprilSylph (changes requested):

src/scripts/tweaks/subtle_activity_mutuals.js
AprilSylph: (yes we do! fuck!!!!!)

https://github.com/AprilSylph/XKit-Rewritten/blob/e7af3451d5132f2e9165d422ccc528a6ae4b2a4c/src/scripts/shorten_posts.css#L13

marcustyphoon: Apparently we thought of this! https://github.com/AprilSylph/XKit-Rewritten/pull/176#issuecomment-874297815 I'm not sure I remember how to install Chrome 89 in order to see what this code does when that rule is unsupported, though :D

AprilSylph (approved):
so nice, so smooth

PR #1134, opened 2023-07-10
New Tweak: Force the default dashboard tab to "Following"

### LINKED ISSUE #1158: automatically take me to following tab
### Concept

I made a new blog and unfortunately i cant change the "for you" page as my default tab, i don't know how hard it would be but it would be very nice for new users for their to be an option to automatically take them to their following tab!

marcustyphoon:
Is Staff planning to add a feature that makes your first dashboard tab customizable **and the default** (or that remembers your last-selected dashboard tab or something) any time soon? If yes, this is obviously pointless. I, however, was not sure if that was the implication of

> We are also working on making dashboard tabs even more customizable, including adding the ability to choose which tab appears first.

### Description

Adds a tweak that forces navigation to tumblr.com and tumblr.com/dashboard to actually point to the "Following" dashboard tab instead of the "For You" dashboard tab, as it does for accounts created after May 8th, 2023.

I will probably publish an equivalent to this as a userscript (already written) and then no-op it when it is no longer needed if we don't want to add it to XKit and the behavior is not expected to be implemented natively soon.

Resolves #1158

### Testing steps

Load the new tweak on a blog created after May 8th, 2023.

 - Open a tab at tumblr.com
 - Open a tab at tumblr.com/dashboard
 - Navigate to the dashboard from a non-dash page via clicking the Tumblr logo
 - Navigate to the dashboard from a non-dash page via clicking the home icon
 - Navigate to the dashboard from a different dashboard tab via clicking the Tumblr logo
 - Navigate to the dashboard from a different dashboard tab via clicking the home icon

Also, perform the above actions in the mobile layout.

In each case, it should appear that the "Following" tab is the target of said navigation actions.

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -102,6 +102,11 @@
       "type": "checkbox",
       "label": "Hide the Tumblr Live carousel",
       "default": false
+    },
+    "following_tab": {
+      "type": "checkbox",
+      "label": "Force the \"Following\" dashboard tab to be the default",
+      "default": false
     }
   }
 }
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -105,7 +105,7 @@
     },
     "following_tab": {
       "type": "checkbox",
-      "label": "Force the \"Following\" dashboard tab to be the default",
+      "label": "Force the default dashboard tab to \"Following\"",
       "default": false
     }
   }
--- a/src/scripts/tweaks/following_tab.js
+++ b/src/scripts/tweaks/following_tab.js
@@ -1,0 +1,25 @@
+import { navigate } from '../../util/tumblr_helpers.js';
+
+const attemptRedirect = () => {
+  if (
+    ['/dashboard', '/'].includes(location.pathname) &&
+    document.querySelector('main > [data-timeline^="/v2/tabs/for_you"]')
+  ) {
+    navigate('/dashboard/following');
+  }
+};
+const observer = new MutationObserver(attemptRedirect);
+
+export const main = async function () {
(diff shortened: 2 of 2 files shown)


nightpool:
going to test this tonight

nightpool:
Can you explain the performance optimization commit? I'm not sure I understand how it improves performance

marcustyphoon:
Went to write an answer to that and realized that I messed it up (yes it did improve it; no it did not improve it by the amount desired). Fixing...

marcustyphoon:
Take two:

Sure!

querySelector calls stop searching once they find an element that matches the whole selector—which, notably, means that they don't short-circuit if such an element doesn't exist, and will check every element on the page against the rightmost selector in their query string. On a Tumblr page, this can easily be hundreds of thousands of elements.

The tweaked code finds the timeline element quickly no matter how many posts are loaded (querySelector starts at the top of the DOM tree) and no matter what the currently-logged-in account's default dashboard tab is, putting a ceiling on execution time at a small fraction of a millisecond.

This could be optimized more by stealing more code out of mutations.js (rAF throttle), but it's plenty fast enough and this tweak will be obsolete soon anyway.

marcustyphoon:
This could also include click handlers on the home button and tumblr logo that go straight to the following tab to skip an extra api request, now that I think of it. But also, like, how many people will actually use this between when it hypothetically gets released and when it hypothetically gets retired (and/or optimized, if this isn't on the redpop roadmap)? Probably not worth the extra testing and review?

AprilSylph (approved):
Don't have an account that is capable of testing this, but code looks good.

PR #1154, opened 2023-07-19
New Tweak: Hide the home/following notification badge

marcustyphoon:
### Description

Similar to the the tweak that hides the notification badge for activity, this adds one that hides the notification badge for new posts on your chronological dashboard (following feed/home button). I've wanted this for years, people assume the activity one does this already, and it's particularly relevant as people are using it as a specific criticism of the vertical nav layout.

I am actually... confused? because I currently only see the badge in question appearing in the vertical nav layout? But a) I remember wanting this for ages so it must have been a thing, and b) the linked discussion definitely implies that it was a thing very recently?

Resolves #1132.

### Testing steps

- Confirm that the badges in both the vertical nav layout's "home" nav item and "following" dashboard tab bar tab are hidden by this tweak.
- Confirm that the badge in the mobile layout's "home" nav item... would be if it existed(?????)
- Confirm that the badge in the non-vertical-nav-layout's "home" header button... would be if it existed(?????)

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -93,6 +93,11 @@
       "label": "Hide the activity notification badge",
       "default": false
     },
+    "hide_following_notification_badge": {
+      "type": "checkbox",
+      "label": "Hide the home/following notification badge",
+      "default": false
+    },
     "no_where_were_we": {
       "type": "checkbox",
       "label": "Hide the \"Now, where were we?\" button",
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -93,14 +93,14 @@
       "label": "Hide my follower count where possible",
       "default": false
     },
-    "hide_activity_notification_badge": {
-      "type": "checkbox",
-      "label": "Hide the activity notification badge",
-      "default": false
-    },
     "hide_following_notification_badge": {
       "type": "checkbox",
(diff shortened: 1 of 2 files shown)


AprilSylph (changes requested):
1. This tweak should precede the "Hide the activity notification badge" since the Home button precedes the Activity button
2. Naming nitpicks
    - We should match casing when referring to stuff like button labels, for clarity (do a markless quotation)
    - Even though it is technically also a "notification" badge it doesn't notify about notifications, which makes me want to avoid using the word notification (wow that sounds insane when I spell it out like that)

---
    
- `"Hide the Home/Following unread badge"`
- `"Hide the Activity notification badge"`

marcustyphoon:
Is the name of this tweak actually accurate? I'm not familiar with the progression of where and in what A/B test branches these badges appear.

Of course, if this is still up in the air it makes sense to me to leave it as-is; the code does hide both no matter whether they exist or not and we can always remove the unnecessary reference later.

AprilSylph:
> Is the name of this tweak actually accurate?

I think right now, badges only appear on Home, but maybe they still appear on Following for accounts whose default tab is For You. I dont have such an account to test with and I'm too lazy to make one.

AprilSylph (approved):

PR #1171, opened 2023-07-29
New AccessKit feature: Remove unusual spacing

marcustyphoon:
"This should be easy," I say, immediately before spending 6 hours on a still only partially working draft.

Interesting decisions include:

 - does one use `white-space: pre-line`, which handles a lot of the whitespace collapsing for you but removes leading indentation, or try to go without it and correctly manually parse e.g. `  hello <strong> world</strong>` (simple regex will not see two spaces there)

### DIFF
--- a/src/scripts/accesskit.json
+++ b/src/scripts/accesskit.json
@@ -39,6 +39,11 @@
       "label": "Remove user-set fonts from text in posts",
       "default": false
     },
+    "remove_unusual_spacing": {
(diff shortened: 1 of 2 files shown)


Zavionspain (commented):
> thanks for the suuport 😊🫂🙏 #1174 

>

marcustyphoon:
Cleanup: Closed because, eh. Reopenable.

Issue #1178, opened 2023-08-06
Cleanfeed: filter posts with specified tags


enchanted-sword:
### Concept

An option to treat posts with specified tags as mature would be helpful since a lot of people still use some variation of "#nsfw" or "#nsft" to mark mature posts without community labels. While tag filtering does exist, implementing it alongside the option to mark specified blogs as mature would add an element of consistency as well as allowing people to simply hide media in posts specified as mature while still using the "hide filtered posts entirely" tweak.

marcustyphoon:
I both like this idea and think that one could make the argument that if something is tagged nsfw/nsft and does not have the "mature" community label (and thus does not get its media hidden by cleanfeed), it's supposed to have one and thus one should mark it with the "missing label" button. (Got to follow those iOS store rules, unfortunately.)

enchanted-sword:
Of course, but if you're using Cleanfeed you probably want the media hidden before you flag it and not after.

Issue #982, opened 2023-02-16
Quick Tags: Remove leading hash characters from user inputs


marcustyphoon:
### Concept

As mentioned in #940 and elsewhere, occasionally people type `#whatever` into inputs requesting a tag. Tumblr does not allow leading hash characters in tags (...mostly), so this isn't valid input and we should quietly remove the hash characters. I prefer to actually make them vanish (the way we replace straight quotes with smart quotes in-place) over stripping them only during processing.

PR #1185, opened 2023-08-10
CleanFeed: Add option to treat specified tags as mature

enchanted-sword:
### Description
Resolves [#1178](https://github.com/AprilSylph/XKit-Rewritten/issues/1178)
Adds an option to treat specified tags as mature in the same way specified blogs can be treated as mature

### Testing steps
Add a tag to the text field in the CleanFeed options and view a post with said tag

### DIFF
--- a/src/scripts/cleanfeed.js
+++ b/src/scripts/cleanfeed.js
@@ -23,14 +23,17 @@ const processPosts = postElements => filterPostElements(postElements).forEach(as
 
   const { blog: { name, isAdult }, communityLabels, trail, tags } = await timelineObject(postElement);
 
-  if (isAdult || communityLabels.hasCommunityLabel || localFlaggedBlogs.includes(name) || localFlaggedTags.some(t => tags.includes(t))) {
+  if (isAdult ||
+      communityLabels.hasCommunityLabel ||
+      localFlaggedBlogs.includes(name) ||
+      localFlaggedTags.some(t => tags.map(tag => tag.toLowerCase()).includes(t))) {
     postElement.classList.add(hiddenClass);
     return;
   }
 
   const reblogs = postElement.querySelectorAll(reblogSelector);
   trail.forEach((trailItem, i) => {
-    if (trailItem.blog?.isAdult || localFlaggedBlogs.includes(trailItem.blog?.name || localFlaggedTags.some(t => trailItem.tags?.includes(t)))) {
+    if (trailItem.blog?.isAdult || localFlaggedBlogs.includes(trailItem.blog?.name)) {
       reblogs[i].classList.add(hiddenClass);
     }
   });
@@ -38,8 +41,8 @@ const processPosts = postElements => filterPostElements(postElements).forEach(as
 
 export const main = async function () {
   ({ blockingMode, localBlogFlagging, localTagFlagging } = await getPreferences('cleanfeed'));
-  localFlaggedBlogs = localBlogFlagging.split(',').map(username => username.trim());
-  localFlaggedTags = localTagFlagging.split(',').map(tag => tag.trim());
+  localFlaggedBlogs = localBlogFlagging.toLowerCase().split(',').map(username => username.trim());
+  localFlaggedTags = localTagFlagging.toLowerCase().split(',').map(tag => tag.replace(/\#/gi, '').trim());
 
   styleElement.textContent = localFlaggedBlogs
     .map(username => `[title="${username}"] img[alt="${translate('Avatar')}"] { filter: blur(20px); }`)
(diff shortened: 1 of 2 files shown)


marcustyphoon:
I don't think we can rename the `localFlagging` preference parameter without ensuring that the user's current preference is migrated by making use of the "inherit" preference key, as seen in the last few entries of `no_recommended.json` and in `hide_mini_follow` in `tweaks.json`.

enchanted-sword:
> I don't think we can rename the `localFlagging` preference parameter without ensuring that the user's current preference is migrated by making use of the "inherit" preference key, as seen in the last few entries of `no_recommended.json` and in `hide_mini_follow` in `tweaks.json`.

Ah, makes sense, I'll set it to inherit the value of `cleanfeed.preferences.locaFlagging`.

src/scripts/cleanfeed.js
```diff
@@ -10,33 +10,36 @@ const styleElement = buildStyle();
...
-  if (isAdult || communityLabels.hasCommunityLabel || localFlaggedBlogs.includes(name)) {
+  if (isAdult || communityLabels.hasCommunityLabel || localFlaggedBlogs.includes(name) || localFlaggedTags.some(t => tags.includes(t))) {
     postElement.classList.add(hiddenClass);
     return;
   }
 
   const reblogs = postElement.querySelectorAll(reblogSelector);
   trail.forEach((trailItem, i) => {
-    if (trailItem.blog?.isAdult || localFlaggedBlogs.includes(trailItem.blog?.name)) {
+    if (trailItem.blog?.isAdult || localFlaggedBlogs.includes(trailItem.blog?.name || localFlaggedTags.some(t => trailItem.tags?.includes(t)))) {
```
marcustyphoon, line 33: Does this line function? That I know of, trail items don't have tag information.

marcustyphoon:
Any opinions on, sort of related to #982, the idea of allowing both `a tag` and `#a tag` in the input box by stripping any leading hash characters when processing the preference input?

src/scripts/cleanfeed.js
enchanted-sword: Whoops, I don't know why I added that. With the amount of work I put into my theme you'd think I would remember how NPF works 🤦‍♀️

enchanted-sword:
I did think about that, and my reasoning for not adding it was that having the default values `nsfw, nsft` with no hash prefix would indicate that the values don't need to have a hash prefix, but I suppose it wouldn't hurt to include it anyways. In the same vein, it might be a good idea to set the preference value strings to lowercase.

src/scripts/cleanfeed.js
```diff
@@ -35,8 +40,9 @@ const processPosts = postElements => filterPostElements(postElements).forEach(as
 });
 
 export const main = async function () {
-  ({ blockingMode, localFlagging } = await getPreferences('cleanfeed'));
-  localFlaggedBlogs = localFlagging.split(',').map(username => username.trim());
+  ({ blockingMode, localBlogFlagging, localTagFlagging } = await getPreferences('cleanfeed'));
+  localFlaggedBlogs = localBlogFlagging.toLowerCase().split(',').map(username => username.trim());
+  localFlaggedTags = localTagFlagging.toLowerCase().split(',').map(tag => tag.replace(/#/gi, '').trim());
```
marcustyphoon, line 45: In a simple case like this you can also use `.replaceAll('#', '')`; this may not be true in all codebases, but our browser support list is more than new enough! (https://caniuse.com/mdn-javascript_builtins_string_replaceall)

marcustyphoon (dismissed):
Functionality looks good to me! Not sure if anyone has opinions about the default contents of the tags field; I could imagine having one but I can't think of anything.

marcustyphoon:
Did that commit change anything, out of curiosity? `localBlogFlagging.split(',').toLowerCase()` would have been a bug, calling toLowerCase on an array, but `localBlogFlagging.toLowerCase().split(',')` seems fine.

(Not that it matters if it didn't; whichever way feels more clear is better!)

marcustyphoon (approved):

enchanted-sword:
> Did that commit change anything, out of curiosity? `localBlogFlagging.split(',').toLowerCase()` would have been a bug, calling toLowerCase on an array, but `localBlogFlagging.toLowerCase().split(',')` seems fine.

I think I spent too long looking at stylesheets yesterday lol, for some reason I looked at it again and thought I had tried to use `toLowerCase()` on an array so I changed it After this gets merged I was planning on working on #1188 so I can always just standardize it then.

marcustyphoon:
Yeah, to be fair, the fact that the previous way made us both have to think about (I started thinking about implicit type coercion) s a good sign that the new way is a clarity improvement :D

Thanks for contributing!

PR #1190, opened 2023-08-12
Implement feature deprecation

### LINKED ISSUE #1058: add mechanism for marking features as deprecated

marcustyphoon:
### Description

resolves #1058

As per the linked issue and the discussion in its comments, this adds a "deprecated" property to script manifests that will hide them unless they're enabled at the time of deprecation (still ironing out details as I write this), and deprecates Timestamps.

### Testing steps

- Load a branch without this PR.
- Optionally, enable and then disable and/or play with the settings of Timestamps.
- Switch to a branch with this PR.
- Confirm that Timestamps disappears from XKit's preferences.

Restart with clean storage. (I ctrl-c'd `npm start`; importing `{}` into the backup interface doesn't currently wipe your storage.)

- Load a branch without this PR.
- Enable Timestamps.
- Switch to a branch with this PR.
- Confirm that Timestamps still appears in XKit's preferences.
- Disable Timestamps.
- Confirm that Timestamps still appears in XKit's preferences.

### DIFF
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -137,12 +137,13 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, deprecated } = await file.json();
+    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, deprecated = false } = await file.json();
 
     const scriptTemplateClone = document.getElementById('script').content.cloneNode(true);
 
     const detailsElement = scriptTemplateClone.querySelector('details.script');
     detailsElement.dataset.relatedTerms = relatedTerms;
+    detailsElement.dataset.deprecated = deprecated;
 
     if (enabledScripts.includes(scriptName) === false) {
       detailsElement.classList.add('disabled');
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -151,7 +151,7 @@ const renderScripts = async function () {
       detailsElement.classList.add('disabled');
 
       if (deprecated && !specialAccess.includes(scriptName)) {
-        detailsElement.hidden = true;
+        continue;
       }
     }
 
--- a/src/browser_action/render_scripts.js
+++ b/src/browser_action/render_scripts.js
@@ -131,7 +131,6 @@ const renderScripts = async function () {
   scriptsDiv.textContent = '';
 
   const installedScripts = await getInstalledScripts();
-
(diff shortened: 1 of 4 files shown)


marcustyphoon:
I had the idea of labelling deprecated scripts, but there's already a parenthesized label being used for disabled scripts. It's possible to do a label only when the script is enabled, but do you think this is getting too cute?

https://github.com/marcustyphoon/XKit-Rewritten/commit/cd065007284d488985b2583c85c67118e830b1c0

AprilSylph:
> I had the idea of labelling deprecated scripts, but there's already a parenthesized label being used for disabled scripts. It's possible to do a label only when the script is enabled, but do you think this is getting too cute?

No, I like it. I was considering the same thing.

src/browser_action/render_scripts.js
```diff
@@ -25,10 +25,11 @@ const writeEnabled = async function ({ currentTarget }) {
     detailsElement.classList.remove('disabled');
   } else {
     enabledScripts = enabledScripts.filter(x => x !== id);
+    previouslyEnabledScripts.includes(id) || previouslyEnabledScripts.push(id);
```
AprilSylph, line 28: Two comments:
- I don't think we should keep an entire history of which features have been enabled. We only need to keep track of deprecated features. Seems like this would be trivial with the `data-deprecated` attribute from cd065007284d488985b2583c85c67118e830b1c0.
- Consequently, I think this key should be renamed... I'm thinking `specialAccess` right now (admittedly mostly for cool factor; if you can think of something similarly terse but more descriptive, do feel free to use that instead!)

src/browser_action/render_scripts.js
```diff
@@ -127,23 +131,28 @@ const renderScripts = async function () {
...
 
     const detailsElement = scriptTemplateClone.querySelector('details.script');
     detailsElement.dataset.relatedTerms = relatedTerms;
+    detailsElement.dataset.deprecated = deprecated;
 
     if (enabledScripts.includes(scriptName) === false) {
       detailsElement.classList.add('disabled');
+
+      if (deprecated && !specialAccess.includes(scriptName)) {
+        detailsElement.hidden = true;
```
AprilSylph, line 154: Any reason to keep the element in the DOM?

Personally I would just skip rendering it:

```suggestion
        continue;
```

marcustyphoon: Not particularly! I regularly remove the `display: none` css on deprecated XKit 7 scripts to install them for testing purposes, but in this repo it's quite easy to just add a script name to `enabledScripts` and I assume a user dedicated/knowledgeable enough to do one will just as easily do the other.

AprilSylph (approved):

PR #1193, opened 2023-08-13
Popup: Add standalone control panel link

marcustyphoon:
Just an idea I had.

### Description

| | |
| - | - |
|  |  |

This adds a link to the XKit Rewritten control panel that opens it in a new tab, useful for e.g. copying lots of tag bundles from XKit 7.

Obviously (to us), what this provides is a duplicate of the feature in the browser's extension preferences page to open the control panel embedded there, but a) most people don't know about that, and maybe more people will find this version, and b) Chromium's version of the embedded preference pane kinda sucks. Yes, yes, Firefox best bae etc.

Bonus points for the ability to use inspect element on the preference pane without the shadow DOM getting in the way.

### DIFF
--- a/src/browser_action/links.css
+++ b/src/browser_action/links.css
@@ -1,7 +1,5 @@
 #links {
   position: relative;
-
-  margin: 1em 0;
   padding: 0 1em;
 }
 
--- a/src/browser_action/popup.css
+++ b/src/browser_action/popup.css
@@ -57,11 +57,23 @@ body {
   user-select: none;
 }
 
+body.standalone {
+  width: 100%;
+  max-width: 632px;
+  margin: 1ch auto;
+
+  outline: 1px solid rgb(var(--active-grey));
+}
+
 #embedded-banner {
   padding: 4px 8px;
   border-bottom: 1px solid rgb(var(--active-grey));
 
   text-align: center;
+}
+
+body:not(.embedded) #embedded-banner {
+  display: none;
 }
 
 nav {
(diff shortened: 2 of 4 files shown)


AprilSylph:
isn't there a method for opening the preferences pane...?

AprilSylph:
also, this is kinda already a feature, just a hidden one:

image

marcustyphoon:
Oh, yeah, it's certainly possible to access a non-popup version without this. All this does is add a link in a location where someone might go looking for it. (Well: a link and CSS that makes the standalone web view look fuller-sized like the Firefox extension preference view, but not ultra-stretched.)

(Also: interesting. It would—or, I guess, did—not occur to me to treat a navigation button like a link and expect it to randomly have an href to the current page. If there was still a header on the top with "XKit Control Panel" and it'd had `cursor: pointer` or something, I'd certainly have tried that. My method was `[right click the button -> select options if in chromium; just left click if in firefox] -> right click -> view source -> edit the front of the URL off of the view source page`.)

marcustyphoon:
> isn't there a method for opening the preferences pane...?

Do you mean `right click the button -> options` in chromium /// `right click the button -> manage -> preferences` in firefox?

AprilSylph:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/openOptionsPage

marcustyphoon:
Oh! You meant for the developer, gotcha. Interesting. This PR has different URL parameters for embedded vs in-a-new-tab, at the moment, but that's not really necessary.

marcustyphoon:
Drafting this because ostensibly the correct way to do it is to unify the "embedded" and "is in a browser window modes, but I cannot be bothered to figure out what this is supposed to mean at the moment:

```css
html, main {
  width: 375px;
  min-width: 100%;
  max-width: 100%;
}
```

AprilSylph:
if the popup width is controlled by the content width, the popup is 375px wide
if the popup width is controlled by something else, e.g. was inside the firefox overflow menu pre-whatever version the extensions button was added, the content width instead shrinks or grows to match the popup width

marcustyphoon:
welp. something is messed up about these merge commits, but I can't for the life of me figure out what.

marcustyphoon:
going to force push this, sorry!

PR #1197, opened 2023-08-15
Fix carousel-hiding scripts with "virtual scroller" experiment

### LINKED ISSUE #1195: "Hide recommended blogs", "Hide recommended tags", "Turn the carousel links into a separating line" options broken for some accounts
Some options under No Recommended & Tweaks are broken for users who have received the "virtual scroller" Tumblr feature:
- Hide recommended blogs between posts
- Hide recommended tags between posts
- Turn the Changes/Shop/etc. links carousel into a separating line

marcustyphoon:
### Description

Force-pushed to a subset of #1213.

This ensures that carousel hiding works whether or not the user has the "virtual scroller" experiment, which wraps each timeline object in a "cell" element, by implementing a helper function (`closestTimelineItem`—open to better names) which finds either the `cell` or `listTimelineObject` containing an element and using it to traverse the DOM correctly whether or not the user has the experiment. It changes from class names to data attributes for all content hiding, preventing a changed classname from the React code overwriting the hidden property on a cell.

resolves #1195

### DIFF
--- a/src/scripts/tweaks/caught_up_line.js
+++ b/src/scripts/tweaks/caught_up_line.js
@@ -1,13 +1,13 @@
 import { keyToCss } from '../../util/css_map.js';
-import { buildStyle } from '../../util/interface.js';
+import { buildStyle, closestTimelineItem } from '../../util/interface.js';
 import { pageModifications } from '../../util/mutations.js';
 
-const hiddenClass = 'xkit-tweaks-caught-up-line-title';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
as per https://github.com/AprilSylph/XKit-Rewritten/discussions/1207#discussioncomment-6744071 this will glitch the scrollbar less with a data attribute

PR #1213, opened 2023-08-16
Fix/improve content hiding with "virtual scroller" experiment

marcustyphoon:
### Description

As per discussion #1207, this fixes and improves content hiding on accounts with the "virtual scroller" experiment. This experiment wraps timeline items in `cell` wrappers and unmounts the wrapper contents when they are far off the screen.

This PR:

 - Implements a helper function (`closestTimelineItem`—open to better names) which finds either the `cell` or `listTimelineObject` containing an element, unifying logic between accounts with and without the experiment.
 - Places the content-hiding class on the `cell` when it exists, ensuring that content is still hidden when it is re-mounted on scroll, preventing scrollbar jumps.
 - Changes from class names to data attributes for all content hiding, preventing a changed classname from the React code overwriting the hidden property on a cell

Seen Posts, Shorten Posts, and Collapsed Queue are not addressed here; they'll require additional logic and I created separate issues for each.

Resolves #1195.

I migrated this over from a dev branch where I have the majority of my PRs merged and manually resolved the conflicts, so while I had tested each script before that process, they should probably go through some basic testing afterward. I also have not tested these with the virtual scroll experiment turned off yet.

PR #1215, opened 2023-08-17
Prevent visible UI flicker when processing new posts

### LINKED ISSUE #1200: Shorten Posts/Collapsed Queue: thrashing with "virtual scroller" Tumblr feature
### Platform

MacOS 13.5 arm64

### Browser

Chrome

### Addon version

master

### Details

We will have to unconditionally set posts' max height so that it is correct on the first render and doesn't cause multiple reflows.

marcustyphoon:
### Description

You know I had to do it to em

(actually, I don't know what that meme means or where it comes from. hopefully it's not, like, racist. anyway.)

This improves the behavior discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/1207#discussioncomment-6747508 somewhat by running `pageModifications` instantly instead of in a `requestAnimationFrame` callback if and only if the modification in question is the mount of a post (or whatever) into a virtual scroll experiment `cell` element. These mounts are currently (often?) run using `requestAnimationFrame`, which means that our trick of delaying `pageModifications` processing only within the current frame doesn't work and causes a render without our modifications for one frame.

It leaves the `requestAnimationFrame` trick (debounce, effectively) in place in other cases.

~~This would be more elegant if one could detect the part of the event loop the current code is running in; just doing this for posts is more of a symptomatic fix than a—hey wait a second.~~ (I pushed the second commit after writing that sentence).

Pretty much addresses #1200, though further optimization would be ideal (and relegates some of #1213 to a performance improvement PR rather than a fix PR), apparently, which was sort of an accident...? Although I should have, and I suppose did, see this coming:

>It seems reasonable that, in theory, XKit's current processing might be able to avoid these scrollbar shenanigans. Specifically: if you scroll up, a post renders in with its full height, the virtual scroll container rerenders (moving everything down), XKit processes the post and hides it, the cell height becomes 0, the resize observer fires, the virtual scroll container rerenders (moving everything back up)... and then the javascript event loop ends and the browser actually renders the page for the first time, then nothing actually changed re: scroll position.
> 
> I am not actually 100% clear on whether this is working or not, at the moment. I thought it wasn't, but

It wasn't, and this is why.

(I was going to draft this but it worked a lot better than I thought it would so I'm not sure now.)

### Testing steps

I have no specific tests for non-breakage.

To see the effect, turn Shorten Posts on with its minimum height, and with/without this PR, compare:
 - scrolling with your mouse cursor to an area far above/below your current scroll position: with this PR, posts should appear shortened instead of thrashing
 - using Scroll to Bottom

### DIFF
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -60,7 +60,7 @@ const onBeforeRepaint = () => {
 
   const addedNodes = addedNodesPool
     .splice(0)
-    .filter(addedNode => addedNode instanceof Element && addedNode.isConnected);
+    .filter(addedNode => addedNode.isConnected);
 
   if (addedNodes.length === 0) return;
 
--- a/src/util/mutations.js
+++ b/src/util/mutations.js
@@ -1,3 +1,4 @@
+import { keyToCss } from './css_map.js';
 import { postSelector } from './interface.js';
 const rootNode = document.getElementById('root');
 
@@ -81,6 +82,8 @@ const onBeforeRepaint = () => {
   }
 };
 
+const cellSelector = keyToCss('cell');
+
 const observer = new MutationObserver(mutations => {
   const addedNodes = mutations
     .flatMap(({ addedNodes }) => [...addedNodes])
@@ -88,10 +91,7 @@ const observer = new MutationObserver(mutations => {
 
   addedNodesPool.push(...addedNodes);
 
-  if ([
-    ...addedNodes.filter(addedNode => addedNode.matches(postSelector)),
-    ...addedNodes.flatMap(addedNode => [...addedNode.querySelectorAll(postSelector)])
-  ].length) {
+  if (addedNodes.some(addedNode => addedNode.parentElement?.matches(cellSelector))) {
     cancelAnimationFrame(timerId);
     onBeforeRepaint();
   } else if (repaintQueued === false) {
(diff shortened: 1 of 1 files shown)


src/util/mutations.js
```diff
@@ -56,9 +58,8 @@ export const onNewPosts = Object.freeze({
 const onBeforeRepaint = () => {
   repaintQueued = false;
 
-  const addedNodes = mutationsPool
+  const addedNodes = addedNodesPool
     .splice(0)
-    .flatMap(({ addedNodes }) => [...addedNodes])
     .filter(addedNode => addedNode instanceof Element && addedNode.isConnected);
```
AprilSylph, line 63: We don't need to repeat `addedNode instanceof Element` here, that filtering is already performed in the observer callback
```suggestion
    .filter(addedNode => addedNode.isConnected);
```

marcustyphoon: Oops—thought I had removed that!

marcustyphoon: (I considered changing `addedNode` to `addedElement`, which we could also do.

AprilSylph (approved):

marcustyphoon:
I'll have to remember to see if this is actually necessary once all of these features launch; as per https://www.tumblr.com/transienturl/725940392323842048 there's a specific line of code that causes this to be useful.

PR #1216, opened 2023-08-17
Create dev tools and css map tool

marcustyphoon:
### Description

I have, like, zero opinion about what any of this is called, what format it uses, whatever.

This adds a CSS map translation tool within a "dev tools" extension that won't be visible unless the user edits their preference JSON manually.

The notable choice we have here is whether to:
- use `pageModifications`, which has slightly less performance impact but will fail on elements that Tumblr updates the `classList` of (like when you switch dashboard tabs and the tab elements become highlighted/unhighlighted
- use a mutation observer with `attributeFilter: ['class']` and just querySelector the whole damn document, which is technically slower (though I run this as a userscript 100% of the time; the performance hit is not noticable)
- use a mutation observer with `attributeFilter: ['class']` and parse the `mutationRecord` (fancy~)

### Testing steps

No specific tests; confirm feature functionality.


AprilSylph:
third option plz :heart:

marcustyphoon:
Aw man.

I had this cute idea to add attribute change observation as an option for registering page modification handlers with our regular util, but that would probably only be a good idea if one could disable the mutation observer while we set classnames on stuff (and enable it again afterward) so that you don't get a cascading waterfall of `onBeforeRepaint` calls or whatever. But a) we usually do modifications asynchronously, so they're not actually inside `onBeforeRepaint`, and b) while I don't know of any way that a `pageModifications` handler could result in Redpop synchronously adding an element (which would then not get detected, if the observer was disabled), I'm not confident enough in that to potentially break XKit Rewritten for no actual reason*.

*besides that my alt text reminder PR would lose a few lines of code maybe

marcustyphoon:
image

marcustyphoon:
Note: in my personal CSS map tool I've also appended the number of classes that correspond to a translated string, so that I know to avoid making selectors with like 45 entries. We could do that here if desired.

Issue #1177, opened 2023-08-05
Re-add account stats to the account submenu


enchanted-sword:
### Concept

Among many other disastrous things, Tumblr is currently trialing a forced 'mobile' account stats layout that gets rid of the expandable account stats menu, preventing you from viewing or visiting your follower page/count, queued posts, drafts, blog members, post+, blaze, and mass post editor from the dashboard. For comparison, the same blog's HTML in the sidebar:

Desktop layout:
![image](https://github.com/AprilSylph/XKit-Rewritten/assets/81981111/06fc6875-f09b-43bf-8bb7-b75b2cb3e220)

Forced mobile layout:
![image](https://github.com/AprilSylph/XKit-Rewritten/assets/81981111/838563f1-b3c9-4167-9530-6e3a1e14bc28)

And while the `floatingButton` div remains across both layouts, the caret button itself is completely removed from the mobile layout.

enchanted-sword:
Reverting this change is relatively simple, since it only removed elements rather than adding or moving anything, and because the CSS styles for all of the removed elements are still present in the stylesheets so simply adding the elements with the proper classes is enough to style it to exactly how it should look.

marcustyphoon:
I find this one super funny because I read through more posts than I can count on one hand screaming about how the account menu in the left rail was *obviously* broken and redundant because the same items are shown in the right rail. Of course being able to see these for you main blog with one click instead of two would be nice, though.

marcustyphoon:
Oh: for clarity, do you think we should name this, like, "re-add account stats to the account menu" or something? I don't think there's anything "mobile" about the current layout (if anything, it's the opposite; the view you want to re-add is the mobile one, because on a mobile device there isn't any room to put the items in question on the right).

enchanted-sword:
I described it as the mobile layout because the class of the elements were specifically changed from `desktop` to `mobile`, but it's actually a bit insulting because the...actual mobile layout...doesn't hide the stats. Even on blogs that are part of the vertical nav beta.

![image](https://github.com/AprilSylph/XKit-Rewritten/assets/81981111/3a488184-e970-4c31-a1da-e1491025bf56)
![image](https://github.com/AprilSylph/XKit-Rewritten/assets/81981111/39655109-bc26-4e05-b68f-3232fd057417)

[...]

PR #1217, opened 2023-08-18
New Feature: Revert Vertical Navigation Layout

### LINKED ISSUE #1149: New tweak proposal: revert the new "vertical nav layout"
### Concept

Tumblr recently rolled out the new Vertical Nav Layout to a large amount of users (appears not to be the default quite yet), and lots of people, including myself, don't particularly like it. From what I can tell, this feature has been in AB testing for at least 3 weeks, and there are already other Issues here relating to it.

Would it be difficult to bring back the old layout with a Tweak, or does this require reimplementing it from scratch?

enchanted-sword:
### Description
Adds a new feature that allows toggling the vertical navigation layout update, with preferences for toggling specific fixes such as adding account stats to the account subnav, adding a searchbar to permalink pages, and moving the settings navitem to the account submenu.

Resolves #1149, #1177, and #1150.


enchanted-sword:
Very rough draft, throwing this up now because I went to test it and it keeps throwing `TypeError: error loading dynamically imported module: moz-extension://6bd15d38-8933-456a-abbe-4f4a6981f0f9/scripts/revert_vertical_nav.js` despite passing all checks and compiling fine.

enchanted-sword:
image
Title and description definitely seem a bit lengthy, open to more concise suggestions.

enchanted-sword:
I'm still getting stuck, so sorry for the mention, but @marcustyphoon do you have any idea why it would consistently fail to load the JS module?

marcustyphoon:
Hmm. I have yet to decide whether I want to pursue reviewing this (if it happens, I'll be the one taking ownership from the maintainer side, so to speak). Procedural considerations include:

- Can this be done without detaching any Tumblr-created layout element from the DOM? That is a precedent that XKit Rewritten largely sticks to, and one I would strongly like to enforce. I have not looked at the source code for [pixiel](https://www.tumblr.com/pixiel)'s user style in detail, but the fact that it is a user style naturally suggests that significant amounts of functionality can be done with those restrictions. (Physically moving the search bar via CSS is a fun one; pretty sure that's doable in a reliable and responsive way.)
- Can this be done without bulk-adding every CSS class from the map that matches a certain key to elements we create? This is obviously not ideal (as Staff reusing a class name internally could screw up our created UI), but the alternatives I can think of are "implement all CSS ourselves" (whew) or "carefully clone class names off of elements" which is also a lot of work. Not sure how AprilSylph feels about this one but I'm maybe hesitantly willing to allow it on unique-sounding names?

Since I haven't made up my mind on whether I want to commit to merging this yet and whether I think the limitations I would want to put on it before doing so are feasible, I don't want to encourage you to spend a ton of time on this unless you think it's fun or whatever*. Bear that in mind. 

(*Not a hypothetical; my unmerged PR list for legacy XKit is, imo, Fairly Impressive :P)

That being said: the error you're running into is caused by omitting the `.js` extensions from the imports; native browser ESM apparently doesn't resolve extensionless imports using an algorithm like Node or a build tool like webpack/esbuild would.

enchanted-sword:
> Can this be done without detaching any Tumblr-created layout element from the DOM? That is a precedent that XKit Rewritten largely sticks to, and one I would strongly like to enforce. 

I would imagine so, I managed to cut down on a lot of the moving of elements, but the ones that remain could probably be omitted with some clever CSS. I haven't taken a look at pixiel's user style yet, but from what I've heard it is rather impressive. 

> Can this be done without bulk-adding every CSS class from the map that matches a certain key to elements we create?

The most crude, but at the same time, simplest solution to this is to just limit the returned value to the first item of the matching array. I had to do this at least once anyways, because there are some cases where having the multiple classes causes styling bugs. 

> the error you're running into is caused by omitting the .js extensions from the imports

I knew I would be something stupid like that lol, VSCode autofilled it without the extensions and I guess I never noticed.

enchanted-sword:
image
If you say so....

enchanted-sword:
The basic functionality seems solid now, although I didn't test it super rigorously. I saw the occasional weird bug or flicker when turning the feature on for the first time after reloading the extension, but it wasn't consistent.

The additional features are still a work in progress. Weirdly enough, both the settings and searchbar preferences worked fine BEFORE I started on the bulk of the style changes, so I don't know what happened there. As for the account stats, it works just fine (although I might mess around more with getting the primary blog stats to always open with the subnav). However, the function I'm using to fetch the user stats is giving me some trouble (at least it stopped throwing errors). I'd say this is pretty good for a day's work, at least.

enchanted-sword:
It was all small issues thankfully, so everything works as intended now. I did some more rigorous testing, and it seems to hold up well even with going between the <1150px and >1150px layouts. I wrapped both the `main` and `clean` functions in a `requestAnimationFrame` function and now it runs with zero flashing of unstyled content.

enchanted-sword:
Still open to suggestions to improve on the name and description

mayfrogs:
Thanks for this, I was hoping the dashboard unfucker would consolidate with xkit at some point so I wouldn't need to have more than one extension installed just for tumblr.

> Still open to suggestions to improve on the name and description

Some suggestions:
"Old dashboard" - "Revert dashboard layout to pre-August 2023."
"De-twitter-fy" - "Make the layout look less like Twitter."
"Horizontal Navbar" - "Return the navigation to it's horizontal position."

I don't know if any of these are good, but I hope they can at least provide some inspiration 🙂

enchanted-sword:
> Can this be done without bulk-adding every CSS class from the map that matches a certain key to elements we create? This is obviously not ideal (as Staff reusing a class name internally could screw up our created UI), but the alternatives I can think of are "implement all CSS ourselves" (whew) or "carefully clone class names off of elements" which is also a lot of work. Not sure how AprilSylph feels about this one but I'm maybe hesitantly willing to allow it on unique-sounding names?

Haven't touched this PR in a while, but while testing a normal JS/CSS fix to re-add avatars, I discovered that some elements end up with different classes from the map based on the browser's perceived window size even in the >990px and >1150px layouts, so it turns out the only real solution is to instead give created elements custom classes and simply just add a CSS file linked to the module that styles them.

enchanted-sword:
I could maybe clone the something for the account subnav stats, but the searchbar seems more specific.

marcustyphoon:
Additional note that this UI would need to function with every menu item shown in order to have feature parity with the vertical navigation UI; not sure what the cleanest way to do that is (for those of us who exclusively use trackpads, a horizontal scroll works great, but for everyone else it basically doesn't at all).

PR #1221, opened 2023-08-18
Seen Posts: Fix functionality with "virtual scroller" experiment

### LINKED ISSUE #1211: Seen Posts: broken with "virtual scroller" Tumblr feature
### Platform

MacOS 13.5 arm64

### Browser

Chrome

### Addon version

master

### Details

Seen Posts' whole thing is almost certainly going to get messed up by the virtual scroller's post reinsertion. We'll have to mark the `cell` elements we've already processed and be sure not to do so multiple times.

marcustyphoon:
### Description

This applies the same modifications for post hiding (moving the class to the `cell` if present; using a data attribute) that fixed other scripts with the virtual scroller in #1213.

It also moves Seen Posts' `excludeClass` functionality to this method as well. This is important because double processing of posts has an unintended functional side effect with Seen Posts

Resolves #1211.

### Testing steps

I typically:
- unrevert the test code
- enable Show Originals on the original posts view (to test interoperability)
- enable Seen Posts and scroll a bit down my dash, confirming that posts I see are marked red by the test code but not dimmed
- soft navigate somewhere else (e.g. explore)
- soft navigate back and confirm that the posts I saw and marked are dimmed
- toggle Show Originals to the all posts view
- confirm that some of the reblog posts I had previously not seen in between the original posts that I did see are not dimmed

I did this with and without the virtual scroller flag, and also did a fresh run with the virtual scroller flag and scrolled up and down a bunch, confirming that no posts became dimmed on remount,

### DIFF
--- a/src/scripts/seen_posts.js
+++ b/src/scripts/seen_posts.js
@@ -3,7 +3,7 @@ import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
 import { keyToCss } from '../util/css_map.js';
 
-const excludeClass = 'xkit-seen-posts-done';
+const excludeAttribute = 'data-seen-posts-done';
 const timeline = '/v2/timeline/dashboard';
 const includeFiltered = true;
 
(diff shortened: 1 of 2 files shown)


src/scripts/seen_posts.js
```diff
@@ -52,11 +52,15 @@ const lengthenTimelines = () =>
 const dimPosts = function (postElements) {
   lengthenTimelines();
 
-  for (const postElement of filterPostElements(postElements, { excludeClass, timeline, includeFiltered })) {
+  for (const postElement of filterPostElements(postElements, { timeline, includeFiltered })) {
     const { id } = postElement.dataset;
+    const timelineItem = getTimelineItemWrapper(postElement);
+
+    if (timelineItem.getAttribute(excludeAttribute) === '') return;
```
AprilSylph, line 59: would prefer explicitly checking for existence of the attribute, rather than a specific value

```suggestion
    if (timelineItem.getAttribute(excludeAttribute) !== null) return;
```

AprilSylph (approved):

marcustyphoon:
oh god I screwed this up hold on

PR #1228, opened 2023-08-21
Standardize text preference processing

### LINKED ISSUE #1188: Standardize text preference processing across scripts
### Concept

As pointed out in discussion in #1185, CleanFeed's blog preference breaks if you write them as uppercase. This is also true for Hide Avatars, Show Originals, and Themed Posts. Painter's tag input does not suffer from case sensitivity and also strips hashtag characters (i.e. #982). Quick Reblog and Tag Replacer are probably inconsistent as well.

Doing a pass over all of these is probably a good idea. (Should tags, in general, be case sensitive? When using them as a match filter I assume the answer is no—Tumblr treats them as the same—but when adding tags to a post, case is preserved, so one has to be careful with that too.)

enchanted-sword:
### Description
Standardizes all preferences using text inputs to ignore case and strip hashtags.
Currently, Hide Avatars, Show Originals, and Themed Posts do not ignore case, and Quick Reblog, Tag Replacer, Painter, and Cleanfeed ignore case and strip hashtags.

Resolves #1188 and #982

### DIFF
--- a/src/scripts/hide_avatars.js
+++ b/src/scripts/hide_avatars.js
@@ -9,7 +9,7 @@ export const main = async function () {
 
   styleElement.textContent = hiddenAvatars
     .split(',')
-    .map(username => `[title="${username.trim()}"] img[alt="${translate('Avatar')}"] { display: none; }`)
+    .map(username => `[title="${username.trim().toLowerCase()}"] img[alt="${translate('Avatar')}"] { display: none; }`)
     .join('\n');
 
   document.documentElement.append(styleElement);
--- a/src/scripts/show_originals.js
+++ b/src/scripts/show_originals.js
@@ -119,7 +119,7 @@ export const main = async function () {
     whitelistedUsernames
   } = await getPreferences('show_originals'));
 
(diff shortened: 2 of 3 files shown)


enchanted-sword:
Currently, case sensitivity is standardized, but I haven't handled stripping hashtags yet. For that, I imagine an event listener will have to be added to `render_scripts.js` that replaces hashtag characters in the input area on input, similar to how Quick Reblog and Quick Tags replace quotes with smart quotes as mentioned in #982, something like:
```js
...
const stripTags = ({ target }) => {
  const { value } = target.value;
  target.value = value.replace(/#/g, '');
};
...
const renderPreferences = async function ({ scriptName, preferences, preferenceList }) {
...
 switch (preference.type) {
      case 'text':
      case 'textarea':
        preferenceInput.addEventListener('input', debounce(writePreference, 500));
        preferenceInput.addEventListener('input', stripTags);
        break;
      case 'iframe':
        break;
      default:
        preferenceInput.addEventListener('input', writePreference);
    }
...
};
```

PR #1064, opened 2023-05-06
Meatballs util: Add purple XKit icons to meatball menu items

marcustyphoon:
### Description

Is this weird? Is this cute? I dunno.

PR #1234, opened 2023-08-21
Control button util: Implement control button tooltips

### LINKED ISSUE #973: Add aria-labels + titles (+ popups?) to control buttons
### Concept

Right now, if one forgets what one of XKit's control buttons is, there's no hovertext to indicate this. This should be easy to fix. Also, accessibility, of course.

Edit: I just had a vision of popping up a notification when you register a control button without one of these accessibility attributes being set, but only in developer mode. We've already gone down that road though :D

marcustyphoon:
### Description

Adds aria-labels to our control buttons and duplicates Tumblr's on-hover tooltips.

Resolves #973.

### Testing steps

Enable Quick Tags and/or Trim Reblogs. Mouse over the control buttons with the viewport width both greater and less than 990px and with touchscreen emulation turned on and off.


marcustyphoon:
Right now this is an exact clone of Tumblr's DOM structure, which could probably be simplified. More importantly, though, it doesn't allow tooltips to be wider than about one word.

Things to figure out include:
 - One word tooltips or multiple word?
 - If multiple, should the words be on one line (and thus be wider than the native tooltips) or wrapped (looks bad)?
 - Should we mark that these are XKit-added in some way?

marcustyphoon:
Current commit screenshot, with graphics I stole from #1064/#1065:

AprilSylph:
I think an X inside a rounded rectangle, especially a white X on a black background, means something very different now.

The purple is the only thing the addon icon has going for it.

marcustyphoon:
Ah. Well... shit.

marcustyphoon:
Question is, should these be active in the ≤990px width mobile view? Tumblr's are not.

src/util/dom.js
```diff
@@ -13,7 +13,7 @@ export function dom (tagName, attributes = {}, events = {}, children = []) {
 
   attributes && Object.entries(attributes).forEach(([name, value]) => element.setAttribute(name, value));
   events && Object.entries(events).forEach(([type, listener]) => element.addEventListener(type, listener));
-  children && element.replaceChildren(...children);
+  children && element.replaceChildren(...children.filter(Boolean));
```
marcustyphoon, line 16: _Fascinating_ (IMHO) issue: this breaks if `children` is a string, which it currently is in this line:

https://github.com/AprilSylph/XKit-Rewritten/blob/2028df100f2c5c903179b094e94cb16cb1683939/src/scripts/tag_tracking_plus.js#L149

...but this has never been a problem before because **strings are iterable, replaceChildren takes an arbitrary number of arguments, and sequential text nodes are identical in practice to a single text node**, so in practice, `dom` can take a non-array string argument for children even though it isn't supposed to.

Do you think we should enforce this properly or just allow non-array arguments here explicitly?

marcustyphoon: quick look through the codebase:

https://github.com/AprilSylph/XKit-Rewritten/blob/2028df100f2c5c903179b094e94cb16cb1683939/src/scripts/trim_reblogs.js#L42

https://github.com/AprilSylph/XKit-Rewritten/blob/2028df100f2c5c903179b094e94cb16cb1683939/src/scripts/mass_deleter.js#L234-L238

https://github.com/AprilSylph/XKit-Rewritten/blob/2028df100f2c5c903179b094e94cb16cb1683939/src/scripts/mirror_posts.js#L31-L35

this one would crash if called:

https://github.com/AprilSylph/XKit-Rewritten/blob/2028df100f2c5c903179b094e94cb16cb1683939/src/scripts/limit_checker.js#L38-L42

AprilSylph: JS being type-unsafe is so fun...

I would love to add typechecking to our JS files but holy fuck will that be a lot of work (which should not be done all at once, before you get any ideas!!!!)

marcustyphoon: > add typechecking to our JS files

As in—theoretically—make the codebase typescript? (I don't know of another interpretation of this; just confirming)

AprilSylph: No, I'm very against there being any build steps. I mean just adding typechecking to our existing JS files.

It's been a while since I looked into it. I even had a go at it, before realising what a monumental task it would be. The bulk of the work would be making our JS type-safe.

marcustyphoon: Interesting—I'm not familiar with how that's done!

AprilSylph: https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html

marcustyphoon: Ah, I see. Lots of extracting variables so they can have jsdoc applied, I guess.

Seems like that would be a lot more characters than using typescript, which I guess is unsurprising.

AprilSylph: I think we should be able to rely on type inferrence enough that we should only need jsdoc for methods...?

I don't find myself explicitly typing variables in `.ts`/`.tsx` files like, ever, so I would hope that type-checking `.js` files is the same...

marcustyphoon: I poked at it a bit. Didn't look much, but yeah, probably. The pattern where we define variables in module scope and set them in main will be a bit of a pain since you have to type the declarations, but oh well.

AprilSylph: I feel like there will be more than a few cases where sane defaults would let us avoid typing even those, like initialising boolean preferences as all `false`, all string preferences as `''`...

marcustyphoon: That would work, yeah. Fewer characters, but also less semantic (it's a bit odd to specify a default when the code unconditionally ignores the value of the default, in my mind—then again, typescript would probably complain about use-before-define in the same case anyway, so).

marcustyphoon: (Anyway, in the meantime, do you prefer allowing singular values here or fixing any non-array values in the repository?)

AprilSylph: Let's make them all arrays.

AprilSylph:
Hmm, I don't really like custom tooltips in general. Simplest thing would just be to add `aria-label` and `title`.

marcustyphoon:
Sure: #1404.

PR #1239, opened 2023-08-23
update post author attribution targeting

### LINKED ISSUE #1237: Mutual Checker not working
### Platform

Windows 11

### Browser

Chrome 115.0.5790.173 (Official Build) (64-bit)

### Addon version

v0.22.0 (beta)

### Details

It's not showing me who my mutuals are

### LINKED ISSUE #1238: Restore links to individual posts in the post header broken
### Platform

Windows 10

### Browser

Microsoft Edge Version 116.0.1938.54 (Official build) (64-bit)

### Addon version

v0.22.0

### Details

Restored links to individual posts in the post header under "Tweaks" is broken again in some cases. When reblogged/drafted/queued from the original poster, it just shows a blank space, and then the source appears in the reblogged post header, but it once again only goes to the front page of their blog in the blog view, and not the direct link to the original post. See screenshot for what I mean.

You can click on the blank space which brings up the post so it can be reblogged, but only from the blog you originally reblogged/drafted/queued it from. I use the drafts on some of my sideblogs to save posts I want to reblog with my main blog later, so this is definitely an issue for me, and I know other people do this as well.

Screenshot:

AprilSylph:
### Description

Updates our selectors for post author attribution, which we have to be careful does not include reblog parent attribution. This fixes Mutual Checker and the post author part of the "Restore links to individual posts in the post header" tweak.

For Mutual Checker, because of how popover wrappers work, we need to insert the mutual icon _before_ the wrapper and not _inside_ it - otherwise, on click, the DOM modification (the icon) will be cleared. This necessitated a new method for walking up the DOM tree to find the outermost popover wrapper - there is always one nested inside another, for some reason.

Fixes #1237
Fixes #1238

### Testing steps

1. Load the modified addon in your browser of choice
2. Enable Mutual Checker
3. Enable Tweaks
4. Enable Tweaks &rarr; "Restore links to individual posts in the post header"
5. Assert that Mutual Checker shows icons on mutuals' posts, attached to the post author name (NOT the reblog author name)
6. Click a labelled mutual's blog name to open the blog view
7. Assert that the blog view opens to the specific post you clicked on, rather than their blog as a whole
8. Close the blog view
9. Assert that the blog name still has a mutual icon attached to it

### DIFF
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -32,7 +32,7 @@ const addIcons = function (postElements) {
     const postAttribution = postElement.querySelector(postAttributionSelector);
     if (postAttribution === null) { return; }
 
-    const blogName = postAttribution.textContent;
+    const blogName = postAttribution.textContent.trim();
     if (!blogName) return;
 
     if (following[blogName] === undefined) {
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -1,4 +1,4 @@
-import { getTimelineItemWrapper, filterPostElements } from '../util/interface.js';
+import { getTimelineItemWrapper, filterPostElements, getPopoverWrapper } from '../util/interface.js';
 import { timelineObject } from '../util/react_props.js';
 import { apiFetch } from '../util/tumblr_helpers.js';
 import { primaryBlogName } from '../util/user.js';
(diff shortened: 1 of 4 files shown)


AprilSylph:
cc @inklesspen

src/scripts/tweaks/restore_attribution_links.js
```diff
@@ -3,7 +3,7 @@ import { keyToCss } from '../../util/css_map.js';
 import { timelineObject } from '../../util/react_props.js';
 import { navigate } from '../../util/tumblr_helpers.js';
 
-const postAttributionLinkSelector = `header ${keyToCss('attributionHeaderText')} a`;
+const postAttributionLinkSelector = `header ${keyToCss('attribution')} > span:not(${keyToCss('reblogAttribution')}) a`;
```
inklesspen, line 6: I think this could be simplified to `header ${keyToCss('attribution')} ${keyToCss('blogLink')}`

inklesspen:
LGTM for restore_attribution_links

src/scripts/tweaks/restore_attribution_links.js
AprilSylph: > we have to be careful does not include reblog parent attribution

`blogLink` is also present on reblog parent attribution links. I don't like relying on their order in the DOM.

inklesspen: yeah okay that's fair

src/util/interface.js
```diff
@@ -6,8 +6,30 @@ export const blogViewSelector = '[style*="--blog-title-color"] *';
...
+  'targetWrapperInline'
+);
 
-export const getTimelineItemWrapper = element => element.closest(cellSelector) || element.closest(listTimelineObjectSelector);
+/**
+ * @param {Element} element Element within a timeline item
+ * @returns {Element | null} The timeline item wrapper
+ */
+export const getTimelineItemWrapper = element =>
+  element.closest(cellSelector) || element.closest(listTimelineObjectSelector);
```
marcustyphoon, line 21: The temptation to make a recursive "farthest" helper is real here :P

marcustyphoon: Er. Wrong selected line.

marcustyphoon (commented):
- Appears to work with and without the reblog redesign feature flag (unless mutual checker breaks on hydration; I can't test that easily since my one blog with the reblog redesign has no mutuals)
- While I of course don't love excluding the reblog attribution via `:not()`, this direction this code breaks in falls back on DOM order, which is nice. In theory we could check the name text using `timelineObject`, but that's probably not necessary.

src/scripts/mutual_checker.js
```diff
@@ -32,8 +32,7 @@ const addIcons = function (postElements) {
     const postAttribution = postElement.querySelector(postAttributionSelector);
     if (postAttribution === null) { return; }
 
-    const blogLink = postAttribution.querySelector('a');
-    const blogName = blogLink?.textContent;
+    const blogName = postAttribution.textContent;
```
marcustyphoon, line 35:
```suggestion
    const blogName = postAttribution.textContent.trim();
```

I've been burned by this too many times before (see: Blacklist in XKit 7).

marcustyphoon (approved):
LGTM!

AprilSylph:
time for a single-commit release then! thanks both for your eyes on this!

AprilSylph:
The release of v0.22.0 on AMO has been cancelled, and v0.22.1 is now in review on both stores.

PR #1249, opened 2023-08-29
Tag Tracking+: Fix race condition and improve preference change behavior with `onStorageChanged`

### LINKED ISSUE #1247: Tag Tracking+: Only one post marked per page load
### Platform

MacOS 13.5 arm64

### Browser

Chrome 115.0.5790.170

### Addon version

v0.22.1

### Details

Tag Tracking+ only drops the number of new posts in a tag by one when you load a tagged page.

This is probably because of the a race condition again (see #912). With the virtual scroller, it appears that `processPosts` is called once per post in rapid succession; I'm not sure exactly what this breaks, but the result appears to be what is described in the issue title.

(Yes, this one is my fault—insufficient testing.)

marcustyphoon:
### Description

This fixes the race condition causing Tag Tracking+ to only mark one post at a time as seen using NotificationBlock's storage architecture (keep a local copy of the storage in memory and don't rely on setting/getting local storage; using an `onStorageChanged` callback to grab any changes made by another instance of the extension).

Note: for all I know it's possible to get erroneous behavior with this method, too, albeit quite difficult in practice. In theory, if one did two back-to-back storage edits, which both asynchronously called `browser.storage.local.set`, this could maybe trigger `onStorageChanged` asynchronously twice, and if any code ran in between those two it would get the wrong data. I haven't seen `onStorageChanged` trigger twice though, so, eh. Anyway.

> This also—and I will make another PR that does not—resolves the fact that we now have an `onStorageChanged` handler and thus don't automatically call `clean().then(main)` on preference change by actually implementing a preference change handler that instantly cleans up and then re-adds the correct DOM elements according to the user's preferences without wiping the unread count cache and then slowly re-querying everything (visible on the sidebar elements). This has been a pet peeve of mine.
> 
> This, unfortunately, was a fairly large pain in the ass thanks to the sort-of-storing-data-in-the-DOM setup we currently have and the lack of easy access to the `<li>` elements in the sidebar items. I kind of hacked a prop into `buildSidebarRow` to enable this, which was needed for the scenario in which we are re-inserting the sidebar because the user changed the 
> 
> There's probably a better way to do this, like just always adding the sidebar element to the page and hiding it based on a classname on the body.

**Edit:** Ignore the quoted paragraphs.

This PR additionally replaces the behavior in which the entire script is restarted (`clean().then(main)`) on preference changes by implementing both of the the preference toggles entirely via data attributes which can be easily toggled instantly in `onStorageChanged`. This does mean inserting the whole sidebar element into the page even when it's disabled in preferences, but I don't think this is a big deal; Tumblr has like ten thousand DOM elements and this is like fifty.

Resolves #1247

### Testing steps

Test on an account with the virtual scroller experiment.

- Confirm that Tag Tracking+ clears the entire unread count of a tag when navigating to it
- Confirm that changing Tag Tracking+ preferences works, particularly when the initial load is still happening*

*not likely to be a problem with the data attribute method

I have not tested:

- Multiple tab coordination
- Behavior without the virtual scroller experiment

### DIFF
--- a/src/scripts/tag_tracking_plus.css
+++ b/src/scripts/tag_tracking_plus.css
@@ -16,14 +16,19 @@
   content: 'No tracked tags!';
 }
 
-#tag-tracking-plus.only-show-new[data-loading="true"] ul:not(:empty)::after {
+#tag-tracking-plus[data-only-show-new="true"][data-loading="true"] ul:not(:empty)::after {
   content: 'Loading...';
 }
 
-#tag-tracking-plus.only-show-new[data-loading="false"][data-has-new="false"] ul:not(:empty)::after {
+#tag-tracking-plus[data-only-show-new="true"][data-loading="false"][data-has-new="false"] ul:not(:empty)::after {
   content: 'No unread tags.';
 }
 
-#tag-tracking-plus.only-show-new li:not([data-new="true"]) {
+#tag-tracking-plus[data-only-show-new="true"] li:not([data-new="true"]) {
+  display: none;
+}
+
+[data-tag-tracking-plus-show-sidebar="false"] #tag-tracking-plus,
+[data-tag-tracking-plus-show-search="false"] .xkit-tag-tracking-plus-search-count {
   display: none;
 }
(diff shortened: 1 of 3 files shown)


AprilSylph (approved):
Don't have the virtual scroller, but code LGTM and no regressions found in testing. Did test with multiple tabs open, seems to work fine.

PR #1251, opened 2023-08-30
Popup: Implement info boxes for e.g. nonfunctional scripts

marcustyphoon:
### Description

This adds a field to script manifests that allows them to specify a small text snippet to add as a note above the script's preferences, e.g. to inform the user that the script may be out of date even if they have the latest extension release.

I'm not sure what to do re: colors since as far as I can tell the XKit Rewritten preference palette doesn't match the Firefox one anymore and I tried to git blame the Firefox source for like 3 minutes before giving up.

### Testing steps

n/a


marcustyphoon:
Ooh. Or these could actually be preference items, with `"type": "text"`, or whatever.

Ehh, nah, I can't think of an actual advantage to that.

src/browser_action/render_scripts.js
```diff
@@ -139,7 +139,7 @@ const renderScripts = async function () {
   for (const scriptName of [...orderedEnabledScripts, ...disabledScripts]) {
     const url = getURL(`/scripts/${scriptName}.json`);
     const file = await fetch(url);
-    const { title = scriptName, description = '', icon = {}, help = '', relatedTerms = [], preferences = {}, deprecated = false } = await file.json();
+    const { title = scriptName, description = '', note = '', icon = {}, help = '', relatedTerms = [], preferences = {}, deprecated = false } = await file.json();
```
AprilSylph, line 142: this line is getting quite long... can we do one-property-per-line here instead, pretty please?

src/browser_action/configuration.css
```diff
@@ -150,6 +150,19 @@ label[for="filter"] {
   font-weight: normal;
 }
 
+.script .note {
+  padding: 1ch 2ch;
+  margin: 0;
+
+  background-color: var(--in-content-box-info-background);
+  text-align: center;
+  font-size: 0.875rem;
```
AprilSylph, line 159: since I'm already submitting a nit review now...
prefer misc styles to be alphabetised

```suggestion
  font-size: 0.875rem;
  text-align: center;
```

AprilSylph (commented):
otherwise LGTM, although I've half a mind to suggest we simply remove `panorama` from the feature list entirely while it's broken...

marcustyphoon:
Yeah, that's definitely also an option :D

AprilSylph (approved):

PR #1262, opened 2023-09-02
NotificationBlock: Add note about "Mute Notifications"

### LINKED ISSUE #1173: NotificationBlock: Mention native option in UI
### Concept

Tumblr has an option to turn notifications off natively now, but it works differently.

The NotificationBlock modal popup should explain this.

marcustyphoon:
### Description

This adds a note to the NotificationBlock interface about the new "Mute Notifications" native feature.

It also—and I'm fine with not doing this if it seems confusing—changes the terminology throughout the NotificationBlock UI to "hide," as the native feature is much closer to what "block" actually means than what the script does. It does not update the name of the script itself, although we could do that too (it would still start with "Notification").

Copy improvement suggestions are welcome.

Resolves #1173.

### Testing steps

n/a

### DIFF
--- a/src/scripts/notificationblock.js
+++ b/src/scripts/notificationblock.js
@@ -9,9 +9,9 @@ import { userBlogNames } from '../util/user.js';
 
 const storageKey = 'notificationblock.blockedPostTargetIDs';
 const meatballButtonBlockId = 'notificationblock-block';
-const meatballButtonBlockLabel = 'Block notifications';
+const meatballButtonBlockLabel = 'Hide notifications';
 const meatballButtonUnblockId = 'notificationblock-unblock';
-const meatballButtonUnblockLabel = 'Unblock notifications';
+const meatballButtonUnblockLabel = 'Unhide notifications';
 const notificationSelector = keyToCss('notification');
 
 let blockedPostTargetIDs;
@@ -48,8 +48,8 @@ const onButtonClicked = async function ({ currentTarget }) {
   const shouldBlockNotifications = blockedPostTargetIDs.includes(rootId) === false;
 
   const title = shouldBlockNotifications
-    ? 'Block this post\'s notifications?'
-    : 'Unblock this post\'s notifications?';
+    ? 'Hide this post\'s notifications?'
+    : 'Unhide this post\'s notifications?';
   const message = shouldBlockNotifications
     ? [
         'Notifications for this post will be hidden from your activity feed.',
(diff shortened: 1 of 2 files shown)


AprilSylph:
I think I prefer keeping the existing terminology. Putting myself in the shoes of someone who has tried native notification muting and found it not to do what I want, something that promises to _block_ notifications would be much more compelling than something that promises to _hide_ notifications.

marcustyphoon:
Sure thing. Done!

Interesting, though. It's not very obvious to me why that would be the case!

AprilSylph:
I think it's fairly well-established in the social media lexicon that a "Block" is more aggressive than a "Mute" or a "Hide". If muting isn't enough, I want to block!

marcustyphoon:
Ah, gotcha, that's where we're differing: I'm not thinking of this script as more aggressive than the native feature (as it doesn't prevent the notification badge going up and doesn't apply to mobile), while you are. Is what this script brings over/in addition to the native feature just the fact that it's retroactive, or are there notifications that slip through the native one but are "blocked" by this?

AprilSylph:
Just the fact that it's retroactive. I'm thinking in terms of a post breaking containment being like a breaking incident:
- Preventing new notifications is a mitigation. The attack is stopped, but the damage is visible.
- Hiding all notifications is a fix. The damage has been undone, and things can continue as if the attack never happened.

...I'll admit I forgot that the notification badge exists, though. It's always been broken on my account.

AprilSylph:
I won't formally request changes for this because I am actively losing consciousness, but I think the info blurb should explain the difference between the two features in the context of the blocking action, like...

> Unlike Tumblr's option to "Mute notifications", this will not prevent notifications for this post from being created, so they will still increment your unread notification count.

Also, for bonus points, hide the blurb entirely if the post's notifications are already muted. Probably would need an API call? Do we have a cacheless `timelineObject` getter? ouh i feel weird goodnight

marcustyphoon:
> the info blurb should explain the difference between the two features in the context of the blocking action, like...

Agreed!

- [ ] do that

> hide the blurb entirely if the post's notifications are already muted.

Also agreed!

- [ ] that

> Probably would need an API call? Do we have a cacheless `timelineObject` getter?

This one's kind of interesting. Menu items get their `timelineObject` data by default from a query we run in `addPostMeatballItem` when you click the menu, which in some sense is always fresh... except for the fact that that is the menu in which the user would be turning on the feature in question. So if the user opens the menu, mutes the post natively, then clicks on postblock, our data will be outdated; if the user opens the menu, mutes the post, closes the menu, opens the menu, and clicks on postblock, our data will be correct.

I think just doing an API request is simplest here. But yeah, we did remove the unmemoized version of `timelineObject` and it does become a little weird in certain circumstances.

AprilSylph (approved):

Issue #1271, opened 2023-09-08
Legacy post edit continue button not responding.


AcePhoenix32:
### Platform

Windows 10

### Browser

Chrome 64 GB

### Addon version

v0.22.1

### Details

Can't edit reply on tumblr post

Note: Legacy post
This thread was originally created, or at some point was edited, using the legacy post editor or a previous XKit version.

On these threads, Trim Reblogs may work normally, have no effect, or require a repeat of the trim action to completely remove the desired trail items.

When pressing the continue button, there is no response

enchanted-sword:
Could be caused by Tumblr removing the legacy editor as of today.

marcustyphoon:
I don't think that would be related, since we're using entirely NPF API calls and the failure is on a fetch. Not impossible, though?

AprilSylph:
It's not broken across the board; I can still continue to trimming a legacy Link post on my test blog.

marcustyphoon:
Yeah, same. We need to add error handling to the API request so there's some indication what's happening (#1272)—this could, for all we can tell without seeing the javascript console output, be a bug in e.g. `createPreviewItem` that only happens on that particular post structure rather than the API call 404ing.

Issue #847, opened 2022-11-20
Tag Tracking+: add error handling to `refreshCount`


AprilSylph:
Expect the `apiFetch` to potentially throw; if it does, set the count string to `\u26A0\uFE0F` (⚠) to communicate the failure to the user.

Possible partial solution to #846, as this should prevent `refreshCount` from ever throwing and interrupting the `refreshAllCounts` loop.

marcustyphoon:
I guess the shortest solution to this would be replacing Promise.all with Promise.allSettled! (Not saying we should do only that, of course, just noting it because I haven't used that function before and before now it wouldn't have crossed my mind.)

PR #1276, opened 2023-09-10
Create and use unified error handler

marcustyphoon:
### Description

Problems:
 - Not all of our API requests (apiFetch, megaEdit) are wrapped in defined, consistent error handling code.
 - The friendly apiFetch error messages from `exception.body?.errors?.[0]?.detail` don't tell the user or us-the-people-reading-the-user's-bug-report what API request actually failed, and the user can't copy them down if they're just in a temporary little notification popup.
 - There is some chance that an API request gives back data that crashes our code and it's hard to tell if that happened or if the API request itself failed.
 - Catching errors can mean they're not in the browser console for analysis of the whole error, not just the part(s) we print.

This:
 - Creates a single error handler that prints both the friendly "detail" text given by a bad API response (whether it is from apiFetch or megaEdit) and the error message detailing what happened (no matter if it is a network failure or a js code exception). It also logs the error to the javascript console, indicating that it's XKit Rewritten so it's easier to distinguish from other stuff in there.
 - Wraps the handler around each location where we call apiFetch or megaEdit a single time due to a user action. (In loops, we already have success/failure counter logic.)
 - ~~Implements a lighter, notify-only handler where we call apiFetch "in the background" (Quick Tags on-post-form-open; inside the Tag Tracking+ main function).~~

This should give us more visibility into e.g. #1271.

Not addressed in this:
- Tag Tracking+: I did not implement #847
- Timestamps: ignored; deprecated
- _edit:_ Instances where we call apiFetch "in the background" (Quick Tags on-post-form-open; inside the Tag Tracking+ main function)

### Testing steps

As shown in the reverted commit, I forced every apiFetch to 404 and used each script that is modified here (except mass unliker). I then reverted that and used every script that is modified here (except mass unliker) to confirm basic functionality.

This was not particularly exhaustive. I did try inserting `const { a } = null` into a function with a `showErrorModal` handler to confirm that it prints non-network errors.


src/scripts/quick_tags.js
```diff
@@ -104,6 +105,11 @@ const processPostForm = async function ([selectedTagsElement]) {
   }
 };
 
+const showError = exception => {
+  console.error(exception);
+  notify(exception.body?.errors?.[0]?.detail || exception.message);
```
nightpool, line 110: we're still using notifytt here?

src/scripts/tag_tracking_plus.js
```diff
@@ -168,9 +169,14 @@ export const onStorageChanged = async (changes, areaName) => {
...
 
+const showError = exception => {
+  console.error(exception);
+  notify(exception.body?.errors?.[0]?.detail || exception.message);
+};
+
 export const main = async function () {
-  const trackedTagsData = (await apiFetch('/v2/user/tags')) ?? {};
-  trackedTags = trackedTagsData.response?.tags?.map(({ name }) => name) ?? [];
+  const trackedTagsData = await apiFetch('/v2/user/tags').catch(showError);
```
nightpool, line 178: i feel like a few of these should just not trigger any error at all, the annoyance here of showing a transient error to the user seems like it outweighs the benefits.

marcustyphoon: Hm, yeah, that might be true. When it's on extension initialization, at least, presumably it won't be clear the error is referring to anyway. Probably not worth implementing a whole error path that calls `addSidebarItem` with an errored state.

src/scripts/quick_tags.js
marcustyphoon: I thought it might be kind of weird to open a modal popup if adding the quick tags button to the post form fails, since the user didn't do any XKit-specific interaction to trigger that code, and the post form still works fine without the quick tags button.

marcustyphoon:
todo: update for #711

marcustyphoon:
Something to investigate at some point: what happens if there's a failure attempting to fetch the form key for the mega-edit utility? We can probably throw a slightly more useful error there.

marcustyphoon:
Logic could be added in the future to give helpful error messages on certain HTTP status codes, like suggesting that the user might be logged out or (if this winds up being a thing) that they might need to log out and log back in/clear cookies/etc.

src/util/modals.js
```diff
@@ -40,3 +40,17 @@ export const hideModal = () => {
...
+
+export const showErrorModal = exception => {
+  console.error('XKit Rewritten error:', exception);
+
+  showModal({
+    title: 'Something went wrong.',
+    message: [
+      [exception.body?.errors?.[0]?.detail, exception.errors?.[0]?.detail, exception.message]
+        .filter(Boolean)
+        .join('\n\n')
```
AprilSylph, line 52: Hmm, do we want to show more than one message...? Logging the exception to the console feels "good enough" for debugging purposes...

```suggestion
        .find(Boolean)
```

marcustyphoon: Hm, that sounds reasonable enough. Just a matter of picking the order correctly so that the useful one is surfaced, then.

I am personally strongly against only giving the user the "friendly" message, because it tells both them and the developer they report the issue to absolutely nothing, and the amount of time I've spent trying to fix XKit 7's Editable Reblogs that was just spent not getting any information about which error it was... okay fine yes I didn't need to do that and so I can't really use that as an argument, but.

Screenshot 2024-02-20 at 6 25 14 PM

AprilSylph: Hmm, right, the friendly messages are... very bare, huh. Nevermind, this is fine as-is.

marcustyphoon:
(did I mean to do that force push? no. no I did not.)

AprilSylph (approved):

PR #1278, opened 2023-09-10
CleanFeed: process unstretched media

### LINKED ISSUE #1277: Cleanfeed not hiding unstretched media
### Platform

Windows 10

### Browser

Firefox Nightly 119.0a1

### Addon version

v0.22.1

### Details

Cleanfeed occasionally fails to filter certain media in posts as intended (manually blurred here)
image

It isn't a problem with the tagged post filtering, as other posts with the same tag are filtered correctly, but rather a problem in `cleanfeed.css`.
image

The selector for media to be hidden specifically ignores `<figure>` elements with an existing `style` attribute, however in this case, the media is unstretched, and thus contains inline styling:
image

I am assuming there is a valid reason for why inline-styled media is ignored, but the selector might need to be adapted in some way to include unstretched media.

enchanted-sword:
### Description
Fixes Cleanfeed not filtering certain "unstretched" media by changing the selector that targets media.

### Testing steps
Find or make a post that is filtered by your Cleanfeed preferences, add inline styling to the media `<figure>` element. The element should still be hidden as normal. The filter should also still ignore anon avatars in asks.

Resolves #1277

### DIFF
--- a/src/scripts/cleanfeed.css
+++ b/src/scripts/cleanfeed.css
@@ -3,7 +3,7 @@
   position: relative;
 }
 
-.xkit-cleanfeed-filtered figure:not([style]):not(:hover)::after,
+.xkit-cleanfeed-filtered figure:not([aria-label]):not(:hover)::after,
 .xkit-cleanfeed-filtered [role="application"]:not(:hover)::after {
   position: absolute;
   top: 0;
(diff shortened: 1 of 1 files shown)


AprilSylph:
(Haven't tested yet) only anon avatars? The intention is to hide blog avatars in posts (or the part of the trail) that are considered potentially mature.

AprilSylph (approved):
Works a treat, thanks!

PR #1279, opened 2023-09-12
Quick Tags: Implement New XKit migration

marcustyphoon:
### Description

Along with https://github.com/new-xkit/XKit/pull/2135, this is one way of implementing a button that migrates data from New XKit to XKit Rewritten when pressed. It communicates via custom events from one extension's content script to the other. 

### Testing steps

- Load the corresponding New XKit PR, e.g. via the XKit editor.
- Create at least one unique New XKit tag bundle.
- Enable Quick Tags in XKit Rewritten.
- Press the button in the New XKit Quick Tags settings pane. Confirm that the tag bundle appears in XKit Rewritten.


AprilSylph (changes requested):
I'd like to see some sort of confirmation prompt on this side so we're not just giving storage-add access to literally anything running on Tumblr.

marcustyphoon:
Hm—now that I think about it, the success UI could be from XKit Rewritten and the failure UI could be from New XKit. No reason to do two prompts in a row.

marcustyphoon:
The New XKit control panel has `z-index: 21474836`.

This is, you know. This is cool.

marcustyphoon:
Worth noting that:
a) Maybe we could do something more general—button in New XKit that shoves over everything? That may not be ideal, though (if we add more stuff that can be imported later, that's a mess.
b) Maybe we could make the handler not inside the Quick Tags script itself, so that it could set Quick Tags to enabled when the user confirmed this import? That would be convenient for the "I am migrating from New XKit to this new thing" workflow, especially if we had more scripts with export functionality.

src/scripts/quick_tags.js
```diff
@@ -241,6 +241,60 @@ popupElement.addEventListener('click', processBundleClick);
...
+    if (toAdd.length) {
+      await new Promise(resolve => {
+        showModal({
+          title: 'Add tag bundles?',
+          message: [
+            `Would you like to add ${
+              toAdd.length > 1 ? `${toAdd.length} tag bundles` : 'a tag bundle'
+            }?`,
+            '\n\n',
+            `(${toAdd.map(({ title }) => title).join(', ')})`
```
AprilSylph, line 271: Ideal copy:

> Would you like to import the following [ x tag bundles | tag bundle ] from New XKit to XKit Rewritten?
>
> * tag bundle name
> * tag bundle name
> * etc.

src/scripts/quick_tags.js
```diff
@@ -241,6 +241,60 @@ popupElement.addEventListener('click', processBundleClick);
...
+          ]
+        });
+      });
+
+      tagBundles.push(...toAdd);
+      await browser.storage.local.set({ [storageKey]: tagBundles });
+
+      showModal({
+        title: 'Success',
+        message: `Added ${toAdd.length > 1 ? `${toAdd.length} tag bundles` : 'a tag bundle'}!`,
```
AprilSylph, line 285:
```suggestion
        message: `Imported ${toAdd.length > 1 ? `${toAdd.length} tag bundles` : 'a tag bundle'}!`,
```

src/scripts/quick_tags.js
```diff
@@ -241,6 +241,60 @@ popupElement.addEventListener('click', processBundleClick);
...
+        buttons: [modalCompleteButton]
+      });
+    } else {
+      showModal({
+        title: 'No new bundles!',
+        message: 'XKit Rewritten has these tag bundles already.',
```
AprilSylph, line 291:
```suggestion
        message: 'Your XKit Rewritten configuration has these tag bundles already.',
```

AprilSylph (changes requested):
It works! I think the copy can be improved here though. The parenthesised comma-separated list of tag bundles threw me off a bit in testing.

marcustyphoon:
Any idea how to make a ul look good in otherwise-centered text?

src/scripts/quick_tags.js
```diff
@@ -241,6 +241,60 @@ popupElement.addEventListener('click', processBundleClick);
...
+      );
+
+    if (toAdd.length) {
+      await new Promise(resolve => {
+        showModal({
+          title: 'Add tag bundles?',
+          message: [
+            `Would you like to import the following ${
+              toAdd.length > 1 ? `${toAdd.length} tag bundles` : 'tag bundle'
+            } from New XKit to XKit Rewritten??`,
```
AprilSylph, line 269:
```suggestion
            } from New XKit to XKit Rewritten?`,
```

AprilSylph: lol nice syntax highlighting there github

marcustyphoon: Whoops.

AprilSylph:
Hmm, don't we have styles for `<table>` elements in modals already?
A single-column table would also serve as a fairly good list, I think.

AprilSylph:
Nevermind. I completely misremembered what our modal table styles looked like. Maybe not.

AprilSylph:
uhhh something something `text-align: left; display: inline-block; margin: 0 auto;`...?

I might need to play with it myself.

marcustyphoon:
idea:

- [ ] ignore case in duplicate checking logic

AprilSylph (approved):

marcustyphoon:
Every single time I check if this is off-center, and every single time it's correct. Maybe we should set a minimum width on the buttons so the gap between them centers in this case :D

PR #1280, opened 2023-09-13
Tag Tracking+: add error handling to `refreshCount`

### LINKED ISSUE #847: Tag Tracking+: add error handling to `refreshCount`

marcustyphoon:
### Description

As per the linked issue, this catches errors thrown by the refresh function in Tag Tracking+ and displays "⚠" instead of a count if it does so.

Resolves #847.

### Testing steps

Un-revert the reverted commit and confirm that the script continues to update the non-errored tags.

### DIFF
--- a/src/scripts/tag_tracking_plus.js
+++ b/src/scripts/tag_tracking_plus.js
@@ -38,7 +38,7 @@ const refreshCount = async function (tag) {
         }
       }
     } = await apiFetch(
-      Math.random() > 0.5 ? `/v2/hubs/${encodeURIComponent(tag)}/timeline` : '/sjkdnksjn',
+      `/v2/hubs/${encodeURIComponent(tag)}/timeline`,
       { queryParams: { limit: 20, sort: 'recent' } }
     );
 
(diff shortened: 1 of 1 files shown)


src/scripts/tag_tracking_plus.js
```diff
@@ -26,37 +26,43 @@ let sidebarItem;
 const refreshCount = async function (tag) {
   if (!trackedTags.includes(tag)) return;
 
-  const savedTimestamp = timestamps[tag] ?? 0;
-  const {
-    response: {
-      timeline: {
-        elements = [],
-        links
+  let unreadCountString = '⚠';
```
AprilSylph, line 29: I think I actually prefer the full emoji...

```suggestion
  let unreadCountString = '⚠️';
```

AprilSylph (approved):

PR #1284, opened 2023-09-21
Migrate to Manifest V3

marcustyphoon:
### Description

So we know it can be done:

This is a functional MV3 port that works in both Chrome and Firefox without per-browser modification, i.e. slightly updated #921. There are a number of methods to do our main world script injection; this one uses bubbled custom events and puts all of the code in one file with a listener (simplest way to feature parity).

I had to append the target element as the last argument to the injected function unconditionally; this means injected functions can't have optional arguments. We could do other stuff: make it the first argument, change the signature to take an object of named fields, whatever.

The most interesting thing about this is that Firefox requires an explicit call to `cloneInto` when passing an object via `CustomEvent`, while Chrome does not; my understanding that the Firefox-specific Xray Vision stuff was all things that Chrome couldn't do and thus a cross-platform extension wouldn't need to use them. May make a https://github.com/w3c/webextensions issue about this or something, idk. 

### Testing steps

no specific tests

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -77,8 +77,18 @@
     return installedScripts;
   };
 
+  const initInjection = () => {
+    const { nonce } = [...document.scripts].find(script => script.getAttributeNames().includes('nonce'));
+    const script = document.createElement('script');
+    script.nonce = nonce;
+    script.src = browser.runtime.getURL('/injected.js');
+    document.documentElement.append(script);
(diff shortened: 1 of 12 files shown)


marcustyphoon:
probably still a problem, would need to address if merging:

- [ ] fix a `Response` object crashing the callback handler on destructure attempt (`detail` becomes null if its contents are not serializable).

PR #1287, opened 2023-09-23
Scroll to Bottom: Fix scrolling stopping after one page

### LINKED ISSUE #1286: Scroll to Bottom not working
### Platform

macOS Venture 13.0.1

### Browser

Firefox 117.0.1

### Addon version

0.22.2

### Details

The Scroll to Bottom feature is no longer working. It allows you to scroll through one set of posts, but once it's time to load another set of posts, it stops. This has been happening for the last 24 hours or so.

marcustyphoon:
### Description

This fixes Scroll to Bottom's loader selection code being out of date. The element containing the loader beneath a timeline now has CSS classes equivalent to `.container.condensed`.

Resolves #1286.

I don't love using "container" as a translated CSS class key because it has so many classnames associated with it, but oh well.

### Testing steps

Use Scroll to Bottom on a dashboard, on a channel page, on the followers page of a blog, and the activity page. Confirm that it does not stop scrolling after the first page and does stop scrolling when it runs out of content.

(This continues to not work on the following page, as Tumblr does not show a loading element there.)

### DIFF
--- a/src/scripts/scroll_to_bottom.js
+++ b/src/scripts/scroll_to_bottom.js
@@ -8,7 +8,7 @@ $(`[id="${scrollToBottomButtonId}"]`).remove();
 const activeClass = 'xkit-scroll-to-bottom-active';
 
 const loaderSelector = `
-${keyToCss('timeline', 'blogRows')} > ${keyToCss('loader')},
+${keyToCss('timeline', 'blogRows')} > ${keyToCss('loader', 'container')},
 ${keyToCss('notifications')} + ${keyToCss('loader')}
 `;
 const knightRiderLoaderSelector = `:is(${loaderSelector}) > ${keyToCss('knightRiderLoader')}`;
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):

AprilSylph:
> I don't love using "container" as a translated CSS class key because it has so many classnames associated with it, but oh well.

Hmm... if I were to rewrite this selector, it would be as...

```css
:is(.timeline, .blogRows) > :last-child > .knightRiderLoader
```

marcustyphoon:
Sounds good to me! 

> I don't, like, check against the Redpop source code when I make these suggestions.

I was going to say I obviously don't either, but I suppose that in a meaningful sense I can. Really, though, we're more protecting against future changes with the way we structure these—I'm not too worried about making something that works now; that part is easy enough.

AprilSylph (approved):

PR #1291, opened 2023-09-29
Panorama: Update for vertical navigation; implement custom width

### LINKED ISSUE #814: Panorama Custom Width
### Concept

As someone who wants a wider space to read tumblr posts without needing to zoom in, a customizable-width option to the panorama (either by pixels or percentage) would be very helpful.

### LINKED ISSUE #860: Panorama: Fix media embed (youtube) aspect ratios
### Concept

Right now, Youtube iframes are embedded with a hardcoded max-width based on post width (540px) and a hardcoded height calculate in pixels based on the video aspect ratio. Panorama overrides the max-width, but the height stays the same, resulting in the iframe not being tall enough.

Our supported browsers all support `aspect-ratio` (https://css-tricks.com/almanac/properties/a/aspect-ratio/), so we can fix this without the percentage vertical padding hack. My thought was adding a mutation listener targeting the youtube iframe, calculating its intended aspect ratio from the inline styles on the element, then overriding `height` to `unset !important` and setting `aspect-ratio`. I guess we don't want to overwrite the inline height property so this would have to be via a classname and a dynamic style element?

Unless, that is, all youtube embed iframes on the site are 16:9 regardless of their contents, which would make this very simple.

### LINKED ISSUE #985: Panorama: Visual weirdness in "embedded blog view"
### Platform

MacOS 10.15.7 x86

### Browser

Ungoogled Chromium v110.0.5481.100

### Addon version

master

### Details

Because the container element surrounding the blog view that is shown when you navigate directly to a blog view is very high up the DOM tree and indistinguishable from container elements Panorama wants to expand without `:has()` or javascript, there wasn't a super easy way to exclude it from getting messed with. Thus, this happens:

I thought I had already submitted this issue, but I don't see it. Already have a hotfix to override the problem back to normal-looking and draft code to actually make Panorama work on blog views.

### LINKED ISSUE #1026: Panorama: Squished `takeoverBanner` ad element
### Platform

Windows 10

### Browser

Chrome 110

### Addon version

a dev branch with a million changes; should be equivalent to #845

### Details

This element has a fixed width of 540px and should be overridden.

image

### LINKED ISSUE #1041: Panorama: Not working in Explore Dashboard
### Platform

Windows 11

### Browser

Chrome 111.0.5563.147 

### Addon version

master (0.21.4)

### Details

This happened because of the added parent div `mainColumContainer` that contains the `main` element. Originally, this isn't the case in Home dashboard.

```html
<!-- In Home -->
<div class="main">
    <main>  
```

```html
<!-- In Explore -->
<div class="main">
    <div class="mainColumContainer">
        <main>
```

The selector isn't targeting the `main` element since it looks for it as a child of `:first-child`. 

https://github.com/AprilSylph/XKit-Rewritten/blob/c2c34d71ab48bf15ee270ab13f162f9d1f2d6bde/src/scripts/panorama.js#L23-L25

image

### LINKED ISSUE #1109: Panorama: Not functional
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 114.0.5735.106

### Addon version

master

### Details

Panorama is currently nonfunctional due to the vertical nav layout changes.

marcustyphoon:
### Description

This reimplements Panorama after the vertical navigation UI changes, adds a preference for the maximum post column width, and fixes a number of elements that were not correctly widened in Panorama's previous iteration (some of which I never bug reported, like link elements).

~~(The default width, pre-reblog-redesign, corresponds to the 3-column masonry view on the explore page. Post reblog redesign it's totally arbitrary at the moment.)~~

~~It is somewhat awkward due to Redpop's use of the same wrapper elements around the main page content and right sidebar regardless of the contents of said main page content. Without the use of `:has`, it is very difficult to do behavior conditional on e.g. the settings page rather than the post column being the main page contents when using a pure CSS solution. I thought a little bit about whether `pageModifications` would be helpful here, but it's not immediately obvious to me what one would target and what one would toggle. I bet that's eventually the best path forward though.~~

~~At the moment, this can't handle the reblog redesign, so it should stay a draft until that experiment ends in either direction.~~

Resolves #814, resolves #985, resolves #1026, resolves #1109, resolves #1041, resolves discussion #391, resolves discussion #853, resolves discussion #1145.

Resolves #860.

Resolves half of #937, the tag page portion; blog view panorama is more complicated.

### Technical notes

The breakpoint value constants are in the end magic numbers, but they're broken out as they appear in the Tumblr CSS that dictates them, so it should be possible to follow them/edit them if and when Tumblr makes changes.

`onStorageChanged` is used just to replace the `clean().then(main)` storage change behavior, which caused pretty spectacular visual thrashing when typing in the preference box, with just running `main()`. A bit unorthodox, I know; the semantically correct version would basically just be putting everything in `main` in a differently-named function and calling it, though, which seems a bit silly.

### Testing steps

- Confirm that Panorama widens post columns based on the viewport width until they reach the width in pixels specified by the added preference.
- Confirm that media elements in posts are, depending on the preference, either widened to the post width (and heightened proportionally) or left with their normal width and centered. This includes:
	- full width single images
	- <300px width images (should be left alone)
	- images with source links
	- images in asks
	- photosets (should be widened when images are side-by-side; individual images should max out at 540px if the expand setting is disabled)
	- polls
	- native videos
	- native videos with vanilla videos enabled
	- audio elements
	- spotify embeds
	- youtube embeds
	- link cards with images
	- various ad containers

--

- Confirm that Panorama functions with endless scrolling enabled and with endless scrolling disabled.
- Confirm that j/k scrolling works on a widened timeline.
- Confirm that resizing the viewport horizontally, affecting post height, doesn't (completely) break post rendering on a widened timeline.

--

- Confirm that Panorama functions:
	- on the dashboard (only in the widest Tumblr desktop layout), including the "tumblr tv" tab
	- on the "channel" pages (including draft, queue and blaze)
	- in the inbox
	- on communities pages
	- in explore pages if and only if a single column of posts is selected
	- in search pages if and only if a single column of posts is selected
	- on collections pages if and only if a single column of posts is selected
	- on patio columns if and only if "wide column" is selected
- Note that Panorama doesn't function:
	- in the floating blog view modal
	- in the "embedded" blog view modal from direct navigation to a blog
	- in settings/blog settings
	- in tagged pages*
	- on https://www.tumblr.com/communities*
	- on individual community posts*
	- in communities settings*
	- on the likes page**
	- on the following/followers pages
	- in the "tumblr tv" ui

*probably would be active here in an ideal world
**I honestly think this is just a redpop bug; other "postColumn" elements are `<main>`s but the likes route uses a div for some reason


marcustyphoon:
- This affects the settings page; unsure how to prevent this.

marcustyphoon:
(Note that this only works when the screen is very wide; I'm sure certain people would like the post column to be widened in cases where that would necessitate collapsing the left and right rails, but a) no idea what Tumblr plans to do in that regard if anything and b) because of how media queries work I'm not sure it's really realistic to even attempt.)

marcustyphoon:
Note: This should be significantly easier, and can be somewhat more feature rich, by taking advantage of `:has()`.

marcustyphoon:
Now featuring: Tumblr Patio support! Optional preference not to expand media in posts that centers the media instead! Bug fixes for ad containers (yes, really)!

AprilSylph (commented):
**Question**: Is this meant to work on https://www.tumblr.com/communities? It doesn't seem to, currently.

---

EDIT: I should be reading the PR description more carefully.

AprilSylph (approved):
LGTM, up to you what to do with my post width options idea

marcustyphoon:
Hm, interesting thought—I wish we could poll users on that. (Well, no; user polls probably wouldn't really reflect usage in the same way an a/b or focus test would, I'd imagine, particularly if the poll was just a question without a demo.)

My guess is that most feature users would want a max width in between options that differed by that much (though there could be more of them). Typing pixel values doesn't actually seem like a great way to select a width, as it's referencing a value that has nothing to do with the actual user experience, but having a dropdown with numerical values basically has the same problem. Probably the ideal input on desktop devices, given enough space, would be a notched slider that updates the visible page in real time, allowing the user to immediately dial in the page appearance that looks correct to them, but we would have to write new code for that and I'm not sure how well it'd work on touchscreens (admittedly a place where it's more likely that this feature is either useless or doesn't need a maximum width value). Slider below a text input is also an option if laid out in a clear way; I do tend to do those.



From a "we try one or the other and then change our minds" perspective, it's relatively easy to go from a set of discrete options (however they're selected from) to a text input, and relatively annoying to go in the reverse direction, which all else being equal leans toward giving discrete options and seeing if anyone requests the reverse.

marcustyphoon:
https://github.com/marcustyphoon/XKit-Rewritten/compare/new-panorama...marcustyphoon:XKit-Rewritten:new-panorama-slider

Here's a quick(...ish) example of a slider input. (I can't figure out why in Firefox the iframe refuses to be less than 85px tall).

AprilSylph:
I think my preference for how we proceed here is:
1. Implement the `"type": "select"` version of the preference, to eliminate potential user error
2. Merge this
3. Explore converting the preference to a new type later, as a low-priority enhancement

marcustyphoon:
Sure, we can do that. What do you think should be the options list and default option?

marcustyphoon:
Hmm, you think people want posts that wide? They seem almost impossible to read when you get to the higher widths.

As far as exact media sizing goes, it probably makes sense to include 1024px (2048px images on @2x retina displays), since—

—uhhhhhh hold on, Tumblr doesn't actually serve higher resolution images in posts at all? Post cancelled, give me a minute.

marcustyphoon:
Okay! We're back. Redpop defines a `srcset` attribute with the entire set of responsive widths and then specifies a sizes attribute that can only fetch images of up to `540 * display scale ratio` pixels in width, so without modification it'll only use those ultra-high-res versions on >2x devices (some 4k windows laptops, plus-sized iphones). Setting `sizes` to `auto` fixes this.

Right, where was I? Uh... okay, retina displays. It seems to me to make sense to include the exact media sizes halved, which would thus exact match on 2x displays (which all modern Macs are; some Windows devices will be, though in my experience most use fractional scaling and iirc those won't have exact matching no matter what we do).

Splitting the second interval into three instead of two yields a mostly-256-spaced:

- 640px (exact media size 1x, 2x)
- 768px _or 832px or split the difference and say 800, whatever_
- 1024px (exact media size 2x)
- 1280px (exact media size 1x)
- 1536px
- 1792px
- 2048px (exact media size 1x)

AprilSylph:
:+1: Sounds good to me

marcustyphoon:
### Additional testing steps

Confirmed that:

- Post images have their `sizes` attribute changed to `auto`:
  - in the dashboard
  - in communities
- Post images have their `sizes` attribute changed back to the original value when Panorama is disabled
- Post images use appropriate increased-resolution source (check `$0.currentSrc`):
  - in the dashboard
  - in communities
  - in Chrome
  - in Firefox
  - on a 2x retina display
  - on a 1x retina display
  - in a photoset with two side-by-side images

AprilSylph (approved):
LGTM based on code review for changes made since my last approval!

I think erring on the side of caution when it comes to behavioural changes is fine.

PR #1294, opened 2023-09-30
Experiment: Create navigation utility

marcustyphoon:
### Description

This implements a wrapper around window.tumblr.on('navigation') so that XKit Rewritten can subscribe to soft navigation events.

More elegant form of #453.

Potentially useful for simplifying logic where something must happen when a navigation event removes an element from the page (as one can't detect removals with `pageModifications`), but when using both one can't be sure of the order they're fired in and thus must be careful, presumably.

### DIFF
--- a/src/scripts/show_originals.js
+++ b/src/scripts/show_originals.js
@@ -5,7 +5,6 @@ import { onNewPosts } from '../util/mutations.js';
 import { keyToCss } from '../util/css_map.js';
 import { translate } from '../util/language_data.js';
 import { userBlogs } from '../util/user.js';
-import { onNavigation } from '../util/on_navigation.js';
 
 const hiddenAttribute = 'data-show-originals-hidden';
 const lengthenedClass = 'xkit-show-originals-lengthened';
@@ -127,8 +126,6 @@ export const main = async function () {
   disabledBlogs = [...whitelist, ...showOwnReblogs ? nonGroupUserBlogs : []];
 
   onNewPosts.addListener(processPosts);
-
-  onNavigation.register(() => console.log('navigated to', location.href));
 };
 
 export const clean = async function () {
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Note that this demonstrates something Redpop could reasonably just do. My guess is that no one uses `window.tumblr.on('navigation')`, but some might reasonably use `window.addEventListener('navigation')`.

AprilSylph (dismissed):
me likey

marcustyphoon:
API surface musings include:

- As in my previous PR, should the callbacks be called with the current and previous locations? Would this be reliable (i.e. is there ever a case where the URL meaningfully changes but the navigate hook is not called? Probably yes?) I wouldn't implement that until a script needed it, **but**:
- If that is not desired or not possible, is there any point of this file not just being  just an inject statement that one calls for its side effect? There's really no difference between registering and unregistering a callback and just adding and removing an event listener on `window`. Of course, importing a file for its side effect is easy to forget and generally feels pretty stupid; this way makes it clear why the logic works. If Redpop were to fire that event, though, I would just delete this and make any script that uses it use `addEventListener` directly.

marcustyphoon:
Oh... hm. I wonder what happens when XKit Rewritten gets updated by Firefox with this code running. If I recall correctly, old event listeners in the extension sandbox are killed... but I don't think the shim that we inject here would, so unless it was written very specifically not to do this, there would be two shims, which would fire two custom events, both of which would trigger the handler in the fresh web extension context.

AprilSylph:
> there would be two shims, which would fire two custom events, both of which would trigger the handler in the fresh web extension context.

Add a random string to the event name? I think we still have a good random string generator in `crypto.js`

AprilSylph (approved):
nice

marcustyphoon:
oops I should have renamed that

Issue #1301, opened 2023-10-09
Mutual Checker: Only show notifications from mutuals


enchanted-sword:
### Concept

In a similar vein to the "Only show posts from mutuals on the dashboard" option of Mutual Checker, would it be possible to add a feature to only show notifications from mutuals in the activity feed?

The only downside I can see with this is that if you have a high volume of notifications from people you don't follow, you may only be able to see a few notifications at any time without going to the activity page.
![image](https://github.com/AprilSylph/XKit-Rewritten/assets/81981111/fb4edd96-40d7-4bfa-b391-3ae9c95d14e8)
I don't see a way around this without making the activity preview load more notifications, which it specifically doesn't in favour of redirecting users to the activity page via the "See Everything" button.

PR #1303, opened 2023-10-13
Mutual checker notification filter

zbeok:
### Description

Addresses #1301 by adding a new option to only show notifications from mutuals. Added common functionality (ie: unburying notifications) to utils that may be useful to other activity page features such as Notification Block. 

### Testing steps

- Open Tumblr with Mutual Checker and notification filter enabled.
- Navigate to the activity page or trigger the popup.
- Ensure that there are no errors in console and that only mutual notifications pop up.

I'm new to contributing to open source so any and all kind feedback is welcomed!

### DIFF
--- a/src/scripts/mutual_checker.css
+++ b/src/scripts/mutual_checker.css
@@ -7,7 +7,6 @@ svg.xkit-mutual-icon {
   margin-right: 0.5ch;
 }
 
-[data-notification-feed] [data-mutual-checker-hidden],
 [data-timeline="/v2/timeline/dashboard"] [data-mutual-checker-hidden] article {
   display: none;
 }
--- a/src/scripts/mutual_checker.css
+++ b/src/scripts/mutual_checker.css
@@ -1,12 +1,13 @@
 svg.xkit-mutual-icon {
   vertical-align: text-bottom;
 
-	height: 1.23em;
+  height: 1.23em;
   margin-top: 0;
   margin-left: 0;
-	margin-right: 0.5ch;
+  margin-right: 0.5ch;
 }
 
+[data-notification-feed] [data-mutual-checker-hidden],
 [data-timeline="/v2/timeline/dashboard"] [data-mutual-checker-hidden] article {
   display: none;
 }
--- a/src/scripts/mutual_checker.js
+++ b/src/scripts/mutual_checker.js
@@ -12,8 +12,7 @@ const hiddenAttribute = 'data-mutual-checker-hidden';
 const mutualsClass = 'from-mutual';
 const postAttributionSelector = `header ${keyToCss('attribution')} > span:not(${keyToCss('reblogAttribution')}) a`;
 
-const styleElement = buildStyle();
-const buildCss = () => `${keyToCss('notification')}:not([data-mutuals]) { display: none !important; }`;
+const styleElement = buildStyle(`${keyToCss('notification')}:not([data-mutuals]) { display: none !important; }`);
 
 const regularPath = 'M593 500q0-45-22.5-64.5T500 416t-66.5 19-18.5 65 18.5 64.5T500 583t70.5-19 22.5-64zm-90 167q-44 0-83.5 18.5t-63 51T333 808v25h334v-25q0-39-22-71.5t-59.5-51T503 667zM166 168l14-90h558l12-78H180q-8 0-51 63l-42 63v209q-19 3-52 3t-33-3q-1 1 0 27 3 53 0 53l32-2q35-1 53 2v258H2l-3 40q-2 41 3 41 42 0 64-1 7-1 21 1v246h756q25 0 42-13 14-10 22-27 5-13 8-28l1-13V275q0-47-3-63-5-24-22.5-34T832 168H166zm667 752H167V754q17 0 38.5-6.5T241 730q16-12 16-26 0-21-33-28-19-4-57-4-3 0-1-51 2-37 1-36V421q88 0 90-48 1-20-33-30-24-6-57-6-4 0-2-44l2-43h635q14 0 22.5 11t8.5 26v543q0 5 4 26 5 30 5 42 1 22-9 22z';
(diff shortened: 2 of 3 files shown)


enchanted-sword:
Beat me to it, was about to open a PR for this lol.

This is a decent solution, however it's a lot more complicated than it needs to be. Notification react props have a boolean `mutuals` property that can be used to determine if the notification is generated by a mutual without manually comparing it to the users' list of mutuals.

In addition, it probably isn't helpful to hide _all_ non-mutual notifications as this also includes things such as blaze notifications, asks, post flagging status, etc.

enchanted-sword:
One other problem that this encounters (as does my implementation, and probably the current version of NotificationBlock) is that hiding notifications this way doesn't get rid of the wrapper and dividing line between notifications, so the empty wrappers sort of stack up between notifications 
image

src/util/react_props.js
```diff
@@ -44,6 +45,32 @@ const unburyBlog = () => {
...
+      return notification;
+    } else {
+      fiber = fiber.return;
+    }
+  }
+};
+
+/**
+ * @param {Element} notificationElement - An on-screen notification
+ * @returns {Promise<object>} - The notification's buried notificationObject property
```
AprilSylph, line 65:
```suggestion
 * @returns {Promise<object>} - The notification's buried `notification` property
```

src/util/mutations.js
```diff
@@ -55,6 +55,11 @@ export const onNewPosts = Object.freeze({
   removeListener: callback => pageModifications.unregister(callback)
 });
 
+export const onNewNotifications = Object.freeze({
+  addListener: callback => pageModifications.register(`${notificationSelector}`, callback),
```
AprilSylph, line 59:
```suggestion
  addListener: callback => pageModifications.register(notificationSelector, callback),
```

AprilSylph (changes requested):
not a super comprehensive review on account of the fact that i am tired and ill, sorry.

i think i like the util changes, though honestly it would be nice to split that out into another PR and make use of it in NotificationBlock (since i think that's where `unburyNotification` is from anyway?)

enchanted-sword:
Also, I double-checked, and notifications from mutuals seemingly aren't added to rollups, so it isn't necessary to check them.

nightpool:
@enchanted-sword this was a recently added feature, in the past 2-3 weeks.
Screenshot_20231014-175012

enchanted-sword:
> Notification react props have a boolean mutuals property that can be used to determine if the notification is generated by a mutual without manually comparing it to the users' list of mutuals.

Now that I've tested both methods more, I can definitely say that checking the `mutuals` property is a better solution performance-wise.

zbeok:
Thank you for all the feedback! I'll try to carefully incorporate all the comments, though if it's urgent to get this passed feel free to override this PR (LMK if you do) since I'm still learning the ropes and not the most fast. Regardless I'll be sure to split the util work out as well in a later separate PR as well!

AprilSylph:
This is going to need updating from upstream, now that #1306 is merged :stuck_out_tongue:

Please re-request my review via the sidebar once the conflicts are resolved!

src/scripts/mutual_checker.css
```diff
@@ -1,12 +1,13 @@
...
 
-	height: 1.23em;
+  height: 1.23em;
   margin-top: 0;
   margin-left: 0;
-	margin-right: 0.5ch;
+  margin-right: 0.5ch;
 }
 
+[aria-label="Notification"][data-mutual-checker-hidden],
```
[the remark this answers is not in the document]
zbeok: Agree to the above, though I currently have an unsatisfactory hack to actually target the container (messy chained parentElements); I don't know what CSS selector or otherwise to target with (there's nothing like data-timeline for the notification feed). If you could suggest a different way to select I would love to know.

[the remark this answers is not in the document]
AprilSylph: So, what we want to do here is actually move the CSS... out of the CSS file. That's what `buildStyle` is for - we create a stylesheet from inside the JS file, where all our utils like `keyToCss` are available. Then, we add and remove that stylesheet to the document in `run()` and `clean()`, respectively. There are a ton of other features that have this pattern if you go looking!

Now, since notifications can actually come in two containers (the activity popover, and the activity page), we need to target both of them. `keyToCss` has a neat trick where you can actually put in any number of keys, and get given a selector that targets all of them at the same time. So in this case, it'd be something like...

```js
${keyToCss('notifications', 'activityPopover')} [data-mutual-checker-hidden] {
  display: none;
}
`);
```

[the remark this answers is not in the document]
AprilSylph: I was going to also detail how to use `.closest()` to find the wrappers without chaining `.parentElement`, but it looks like... that's not necessary anymore...? I'm not seeing borders on notifications right now...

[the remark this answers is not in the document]
zbeok: That makes sense- I've reverted back to using `buildStyle` instead of the css page. I currently don't see the wrappers when I delete the notifications as well.

One more note though: since this was previously brought up, unless we iterate on the children of the feed itself rather than using `notificationSelectors`, date labels wouldn't be affected (leaving some dates devoid of notifications which is functionally fine but aesthetically undesirable).

src/scripts/mutual_checker.js
```diff
@@ -12,15 +12,32 @@ const hiddenAttribute = 'data-mutual-checker-hidden';
 const mutualsClass = 'from-mutual';
 const postAttributionSelector = `header ${keyToCss('attribution')} > span:not(${keyToCss('reblogAttribution')}) a`;
 
+const styleElement = buildStyle();
+const buildCss = () => `${keyToCss('notification')}:not([data-mutuals]) { display: none !important; }`;
```
AprilSylph, line 16: You can set the contents of the style element directly if they don't need to change based on preferences:

```suggestion
const styleElement = buildStyle(`${keyToCss('notification')}:not([data-mutuals]) { display: none !important; }`);
```

src/scripts/mutual_checker.js
```diff
@@ -84,10 +101,19 @@ export const main = async function () {
   ]);
 
   onNewPosts.addListener(addIcons);
+  if (showOnlyMutualNotifications) {
+    styleElement.textContent = buildCss();
```
AprilSylph, line 105:
```suggestion
```

src/scripts/mutual_checker.js
```diff
@@ -12,15 +12,32 @@ const hiddenAttribute = 'data-mutual-checker-hidden';
...
 let showOnlyMutuals;
+let showOnlyMutualNotifications;
 let icon;
 
+const processNotifications = (notificationElements) => {
+  notificationElements.forEach(async notificationElement => {
+    // Captures the current date wrapper element before async shenanigans.
+    const notification = await notificationObject(notificationElement);
+    if (notification) {
+      const { mutuals } = notification || undefined;
```
AprilSylph, line 33: - destructuring anything from `undefined` will throw a type error
- this `|| undefined` will never be reached because you're already testing for `notification` being truthy in the line above

if you're trying to make `mutuals` fall back to `undefined` if it doesn't exist on `notification`, that happens by default. accessing nonexistent properties in JS doesn't cause a type error, it just gives you `undefined`.

```suggestion
      const { mutuals } = notification;
```

src/scripts/mutual_checker.js
```diff
@@ -12,15 +12,32 @@ const hiddenAttribute = 'data-mutual-checker-hidden';
...
 const following = {};
 const mutuals = {};
 
 let showOnlyMutuals;
+let showOnlyMutualNotifications;
 let icon;
 
+const processNotifications = (notificationElements) => {
+  notificationElements.forEach(async notificationElement => {
+    // Captures the current date wrapper element before async shenanigans.
```
AprilSylph, line 30: I don't see any code relating to the date wrapper here? We're just grabbing the React props and actioning on them

```suggestion
```

zbeok:
Thanks for being patient with your explanations. It's appreciated! I've amended what you've suggested.

AprilSylph (approved):
A nice addition, thank you!

Issue #1285, opened 2023-09-22
No Recommended: Blogs sidebar only partially hidden


marcustyphoon:
### Platform

MacOS 13.5 arm64

### Browser

Chrome 116.0.5845.187

### Addon version

v0.22.2

### Details

<img width="455" src="https://github.com/AprilSylph/XKit-Rewritten/assets/8336245/ad3a87ec-68e4-4024-9c9f-f62c7829f1e8">

The No Recommended "Hide recommended blogs in the sidebar" option hides the blogs in this element, but not the element itself.

PR #1318, opened 2023-10-30
No Recommended: Fix blog sidebar/radar hiding

### LINKED ISSUE #1316: No Recommended: Recommended blog hiding/radar hiding nonfunctional
### Platform

MacOS 13.6 arm64

### Browser

Chrome 117.0.5938.149

### Addon version

#983 / master

### Details

Recommended blogs and the radar are not hidden from the sidebar when the appropriate No Recommended options are enabled.

This is not correct now:

https://github.com/AprilSylph/XKit-Rewritten/blob/f5b13c19ca5f45840c1346fbba6f5e5b0a3185a5/src/scripts/no_recommended/hide_recommended_blogs.js#L19

as per:

Screenshot 2023-10-25 at 8 22 58 PM

marcustyphoon:
### Description

Fixes the No Recommended options to hide the recommended blogs sidebar item and the radar.

Resolves #1316.

### Testing steps

Enable No Recommended.

- Toggle the "Hide recommended blogs in the sidebar" option and confirm that it functions correctly (including on tagged pages, the code for which was not changed).
- Toggle the "Hide the Tumblr Radar" option and confirm that it functions correctly.

### DIFF
--- a/src/scripts/no_recommended/hide_radar.js
+++ b/src/scripts/no_recommended/hide_radar.js
@@ -9,11 +9,11 @@ const styleElement = buildStyle(`[${hiddenAttribute}] { display: none; }`);
 const checkForRadar = function (sidebarTitles) {
   sidebarTitles
     .filter(h1 => h1.textContent === translate('Radar'))
-    .forEach(h1 => h1.parentNode.setAttribute(hiddenAttribute, ''));
+    .forEach(h1 => h1.closest('aside > *').setAttribute(hiddenAttribute, ''));
 };
 
 export const main = async function () {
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
I would say that blog hiding on Explore-like pages is not working correctly since the container is still visible, but we can fix that separately later.


Explore | Tagged
-|-
Screen Shot 2023-10-30 at 22 40 26 | Screen Shot 2023-10-30 at 22 40 36

marcustyphoon:
Yeah, that should probably be what #1285 is about.

PR #1328, opened 2023-11-11
CleanFeed: Prevent text overlapping on small images

### LINKED ISSUE #1324: Cleanfeed: Image cover label overlapping on small images
### Platform

Windows 10

### Browser

FF Nightly 121

### Addon version

v0.23.1

### Details

Small images hidden by Cleanfeed seem to have text labels that overlap instead of wrapping.
image
The image I noticed it on here had a default width of 68px and only displayed correctly when resized to 115px, but I tested some other hidden images and they all seemed to have different breakpoints based on the initial width.

enchanted-sword:
### Description
Prevents the "Hidden by CleanFeed" text from overlapping on certain small images, as can be seen [here](https://www.tumblr.com/meowlgbt/733113897690775552/kitty-can-be-doggy-as-a-treat).
Resolves #1324.

### Testing steps
Enable CleanFeed, add the blog in the post above to the filter list, and view the image cover

### DIFF
--- a/src/scripts/cleanfeed.css
+++ b/src/scripts/cleanfeed.css
@@ -12,8 +12,10 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  text-align: center;
   height: 100%;
   width: 100%;
+  line-height: 16px;
   overflow: hidden;
 
   background-color: rgb(var(--white));
--- a/src/scripts/cleanfeed.css
+++ b/src/scripts/cleanfeed.css
@@ -15,7 +15,7 @@
   text-align: center;
   height: 100%;
   width: 100%;
-  line-height: 1;
+  line-height: 1.5;
   overflow: hidden;
 
   background-color: rgb(var(--white));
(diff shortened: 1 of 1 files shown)


marcustyphoon:
`line-height: 1;` is equivalent, right? Might be a bit more semantic than a pixel value that happens to correspond to the font size.

enchanted-sword:
> `line-height: 1;` is equivalent, right? Might be a bit more semantic than a pixel value that happens to correspond to the font size.

True, I forgot that's a thing.

src/scripts/cleanfeed.css
```diff
@@ -12,8 +12,10 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  text-align: center;
   height: 100%;
   width: 100%;
+  line-height: 1;
```
AprilSylph, line 18: This is _sort of_ a nitpick but I might as well require this since releases are slow...

`line-height: 1.5` is the standard for readability for "regular" sized text:

```suggestion
  line-height: 1.5;
```

AprilSylph (changes requested):
Thanks for figuring this one out!

AprilSylph (approved):

Issue #1236, opened 2023-08-23
Feature: Search Timelines


AprilSylph:
> **Search Timelines**
> Find posts by text match

I am envisioning a feature that adds a search bar to the top of all known timelines. When this search bar is typed into, each post is checked for a match for the input. If it passes the check (has a match), the post stays visible. If it fails the check (doesn't have a match), the post is hidden.

The search bar should debounce input/change events to avoid re-checking posts more often than needed.

Emptying the search bar should make the feature perform cleanup on the current timeline.

What to check for text matches:
- Every block in the `content` array
- Every block in the `trail`
- Every blogname in the `trail`
- The post author
- The reblog parent
- Tags

What to check within blocks: anything that is, or can become, visible to the user. Text within text blocks is obvious, but also things like inline link URLs, image alt text, link block titles, audio block track/album/artist names, etc.

Matching method: substring. To copy Tumblr's example, a search for "horse" should match a post containing the word "abhorsement".

The following nice-to-haves should _not_ be included in the feature's base implementation, for the sake of reviewer sanity. They may be considered as followup enhancements to the base feature, to be considered and reviewed separately.
- Match highlighting
- Preferences for which timelines to activate on

marcustyphoon:
> the feature should not attempt to hook each timeline up to toggle preferences initially.

Sorry, could you clarify this one?

AprilSylph:
It shouldn't have preferences like Collapsed Queue for controlling which timelines the feature activates on, in the first iteration of the feature.

Just deciding how we would want to split those preferences (i.e. whether we allow enabling it on Drafts but not Posts, or to include every `/blog/:blogName/*` timeline under the same toggle) is its own cognitive task, and should not be included in the feature's base implementation, but should be considered as an enhancement to the feature afterwards.

marcustyphoon:
Ah, gotcha! 👍

enchanted-sword:
I'm interested in working on this feature, if nobody has started yet.

marcustyphoon:
@AprilSylph Is this still intended to be active, or should it be closed due to Tumblr's virtual scroller changes that—sort of—enable control-F?

AprilSylph:
I think <kbd>CTRL</kbd>+<kbd>F</kbd> works well enough that this can be closed, yeah.

PR #1329, opened 2023-11-16
New feature: Timeline Search

enchanted-sword:
### Description
Adds a "Timeline Search" feature that adds a searchbar to all timelines and filters which posts are seen based on the searchbar input query, as discussed in #1236.

### Testing steps
- Enable the Timeline Search feature
- Enter a search query on a page with a timeline

### DIFF
--- a/src/scripts/_index.json
+++ b/src/scripts/_index.json
@@ -28,6 +28,7 @@
   "tag_tracking_plus",
   "themed_posts",
   "timeformat",
+  "timeline_search",
   "timestamps",
   "trim_reblogs",
   "tweaks",
--- a/src/scripts/timeline_search.css
+++ b/src/scripts/timeline_search.css
@@ -1,7 +1,7 @@
 .xkit-timeline-search-container {
   color: rgba(var(--white-on-dark),.65);
   background-color: rgba(var(--white-on-dark),.25);
-  border-radius: 3px;
+  border-radius: 8px;
   justify-content: space-between;
   align-items: center;
   width: 100%;
--- a/src/scripts/timeline_search.css
+++ b/src/scripts/timeline_search.css
@@ -1,7 +1,7 @@
 .xkit-timeline-search-container {
   color: rgba(var(--white-on-dark),.65);
   background-color: rgba(var(--white-on-dark),.25);
-  border-radius: 8px;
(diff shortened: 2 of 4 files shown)


enchanted-sword:
Very much a proof-of concept implementation, I went with the first search method that came to mind, which is converting the NPF representation of the post to a string and filtering it with regex to separate out the desired fields. It currently searches these fields:
```js
const filters = {
  name: /"name":"([\w\d]*)"/g,
  blogName: /"blogName":"([\w\d]*)"/g,
  rebloggedRootName: /"rebloggedRootName":"([\w\d]*)"/g,
  rebloggedFromName: /"blogName":"([\w\d]*)"/g,
  summary: /"summary":"([^"]*)"/g,
  text: /"text":"([^"]*)"/g,
  tags: /"tags":\[((?:"(?:[^"]*)",?)*)\]/g,
  artist: /"artist":"([^"]*)"/g,
  album: /"album":"([^"]*)"/g,
  audioTitle: /url:"[^"]*","title":"([^"]*)"/g,
  sitename: /"sitename":"([^"]*)"/g,
  linkTitle: /sitename:"[^"]*","title":"([^"]*)"/g,
  description: /"description":"([^"]*)"/g
};
```
The string that contains all these fields is stored as a property on the postElement to avoid having to re-fetch the timelineObject for every search operation.

As is to be expected with filtering potentially large amounts of posts, there is a bit of lag when clearing a query after loading a large number of posts, but I'm not sure if there's anything that can be done about that on our end. One small optimization idea I had was to cache the previous query and skip unhiding posts that matched the previous query if the next query contains the previous one as a substring.

One future preference to possibly implement alongside the ones discussed in the original issue is including blog titles and descriptions in the searched fields, it could be handy at times but in most cases you probably wouldn't want every search result to be a seemingly random post with no body text matches until you realize the query shows up a single time in a random blog's description.

That aside, at least from all the testing I've done, it actually seems to work surprisingly well, both for finding specific posts on my timeline as well as functioning as a general filter for the timeline (or like the blog subscriptions tab with no associated push notifications XD). It's also useful when searching `/blog/[user]` pages, because, while to be fair there are probably reasons why, the default Tumblr blog search is pretty much useless.

PR #1330, opened 2023-11-18
Manifest V3-compatible injection demo

marcustyphoon:
### Description

This is a functional MV3 port (without the change to manifest.json) that works in both Chrome and Firefox without per-browser modification, i.e. slightly updated #921. There are a number of methods to do our main world script injection; this one uses bubbled custom events and moves the imported function bodies into their own files, using a cached dynamic import to load them for performance (this should be compared against simply setting a script element's `src` property, for example).

I had to append the target element as the last argument to the injected function unconditionally; this means injected functions can't have optional arguments. We could do other stuff: make it the first argument, change the signature to take an object of named fields, whatever.

### Testing steps

to write

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -77,8 +77,18 @@
     return installedScripts;
   };
 
+  const initPageContext = () => {
+    const { nonce } = [...document.scripts].find(script => script.getAttributeNames().includes('nonce'));
+    const script = document.createElement('script');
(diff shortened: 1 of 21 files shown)


marcustyphoon:
- file names and locations are arbitrary/little thought was put into them
- injected function bodies in their current positions are block-commented (would be removed if merging this; this just makes for easier comparison/lower risk of merge conflicts)

PR #1341, opened 2023-12-02
No Recommended: Change blog and tag carousel selectors

### LINKED ISSUE #1335: No Recommended not hiding recommended tags
### Platform

Windows 10

### Browser

Brave 1.60.125

### Addon version

v0.23.1

### Details

despite having the recommended tags option selected, i am still seeing recommended tags. reloading the page sometimes works but it doesn't all the time. i've also noticed that when the recommended tags are on my dashboard, some posts are blocked from showing up, but are able to be seen if the recommended tags block is gone.

image

enchanted-sword:
### Description
Fixes #1335 by changing the blog carousel and tag card selectors to match the current element classes
```js
const blogCarouselSelector = `${listTimelineObjectSelector} ${keyToCss('blogRecommendation')}`;
```
```js
const tagCardSelector = keyToCss('tagCard');
```

### Testing steps
- Enable the No Recommended feature and the two related preferences
- Scroll through the timeline

### DIFF
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -11,7 +11,7 @@ const styleElement = buildStyle(`
 `);
 
 const listTimelineObjectSelector = keyToCss('listTimelineObject');
-const blogCarouselSelector = `${listTimelineObjectSelector} ${keyToCss('blogCarousel')}`;
+const blogCarouselSelector = `${listTimelineObjectSelector} ${keyToCss('blogRecommendation')}`;
 
 const hideBlogCarousels = blogCarousels => blogCarousels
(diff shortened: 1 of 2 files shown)


marcustyphoon:
I wonder if we could just do like `[data-cell-id^="timelineObject:carousel"]`. Whether that's more or less robust is just complete guesswork, though.

marcustyphoon (approved):

enchanted-sword:
> I wonder if we could just do like `[data-cell-id^="timelineObject:carousel"]`. Whether that's more or less robust is just complete guesswork, though.

I thought about that initially, but the same id scheme is used for both blog and tag carousels, so that doesn't work if they're separate preferences.

AprilSylph (approved):

PR #1344, opened 2023-12-05
Fix no user colors not applying to text in the editor

### LINKED ISSUE #1093: AccessKit: No user colors does not apply to post editor
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 114.0.5735.106

### Addon version

v0.21.9

### Details

As reported by helptheexo on the New XKit Discord, the AccessKit option to turn off user-colored text does not apply to the trail items that are shown in the editor when reblogging a post.

This is probably also true of some other AccessKit features.

enchanted-sword:
### Description
Fixes #1093 by hiding user-set colors for trail items and asks in the editor.

### Testing steps
- Enable AccessKit and the remove user-set colors preference
- Open a post with user-set colors in the post editor
- Open an ask with user-set colors in the post editor

### DIFF
--- a/src/scripts/accesskit.css
+++ b/src/scripts/accesskit.css
@@ -50,7 +50,9 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
   color: rgb(var(--accent));
 }
 
-body.accesskit-no-user-colours article span[style^="color"] {
+body.accesskit-no-user-colours article span[style^="color"],
+body.accesskit-no-user-colours [data-testid="blocks-reblog-trail"] span[style^="color"],
+body.accesskit-no-user-colours [data-testid="blocks-ask"] span[style^="color"] {
   color: inherit !important;
 }
 


src/scripts/accesskit.css
```diff
@@ -50,7 +50,9 @@ body.accesskit-blue-links article a[target="_blank"][href^="https://href.li/"] {
   color: rgb(var(--accent));
 }
 
-body.accesskit-no-user-colours article span[style^="color"] {
+body.accesskit-no-user-colours article span[style^="color"],
+body.accesskit-no-user-colours [data-testid="blocks-reblog-trail"] span[style^="color"],
+body.accesskit-no-user-colours [data-testid="blocks-ask"] span[style^="color"] {
```
AprilSylph, line 55: Targeting elements based on `data-testid` is unfortunately not reliable. We (those of us working with Redpop's source code) actually have a lint warning against setting this attribute outside of mocked components in unit tests, since the ideal way to use Testing Library is to query for elements by accessibility attributes, which `data-*` attributes are not.

AprilSylph (changes requested):

src/scripts/accesskit.css
enchanted-sword: Hm. In that case, the next closest selector I can see that doesn't use `keyToCss()` would be `body.accesskit-no-user-colours #glass-container span[style^="color"]:not(.tumblr-color)`, but I wouldn't know if that could affect anything outside of the text blocks.

AprilSylph: Moving away from static CSS is probably the better thing to do here.

PR #1348, opened 2023-12-08
No Recommended: Reliably select blog carousels via react props

### LINKED ISSUE #1346: No Recommended: Recommended blogs not hidden on tagged pages
### Platform

MacOS 13.6 arm64

### Browser

Chrome 117.0.5938.149

### Addon version

master 8c270824

marcustyphoon:
### Description

Enough classname targeting.

This uses our `timelineObject` utility to read the react internal data of recommendation carousels, reliably identifying those which are blog recommendations (as opposed to tag recommendations). This is, of course, fragile to changes in the underlying API data, but it seems better than trying to target a classname, ~~since there aren't any unique "blog recommendation"y ones in this element.~~

Resolves #1346.

### Testing steps

- Optionally unrevert the test commit for visibility
- Confirm that No Recommended's blog carousel hiding functionality works on the dashboard and on tag pages and doesn't hide tag recommendations (I did not do the latter; I can't find any)
- Confirm that No Recommended's blog carousel hiding functionality works on anywhere else blog recommendation carousels can appear, if any; I don't know the full list.

### DIFF
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -11,13 +11,13 @@ const styleElement = buildStyle(`
   [${hiddenAttribute}] > div :is(img, video, canvas) { display: none }
 `);
 
-const blogCarouselSelector = `${keyToCss('listTimelineObject')} ${keyToCss('carouselWrapper')}`;
+const carouselSelector = `${keyToCss('listTimelineObject')} ${keyToCss('carouselWrapper')}`;
 
-const hideBlogCarousels = blogCarousels =>
-  blogCarousels.forEach(async blogCarousel => {
-    const { elements } = await timelineObject(blogCarousel);
+const hideBlogCarousels = carousels =>
+  carousels.forEach(async carousel => {
+    const { elements } = await timelineObject(carousel);
     if (elements.some(({ objectType }) => objectType === 'blog')) {
-      const timelineItem = getTimelineItemWrapper(blogCarousel);
+      const timelineItem = getTimelineItemWrapper(carousel);
       if (timelineItem.previousElementSibling.querySelector(keyToCss('titleObject'))) {
         timelineItem.setAttribute(hiddenAttribute, '');
         timelineItem.previousElementSibling.setAttribute(hiddenAttribute, '');
@@ -27,7 +27,7 @@ const hideBlogCarousels = blogCarousels =>
 
 export const main = async function () {
   document.documentElement.append(styleElement);
-  pageModifications.register(blogCarouselSelector, hideBlogCarousels);
+  pageModifications.register(carouselSelector, hideBlogCarousels);
 };
 
 export const clean = async function () {
--- a/src/scripts/no_recommended/hide_blog_carousels.js
+++ b/src/scripts/no_recommended/hide_blog_carousels.js
@@ -5,15 +5,10 @@ import { timelineObject } from '../../util/react_props.js';
 
 const hiddenAttribute = 'data-no-recommended-blog-carousels-hidden';
 
-/*
 const styleElement = buildStyle(`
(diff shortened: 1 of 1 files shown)


marcustyphoon:
(Clarification: this isn't the only way to solve the linked issue.)

src/scripts/no_recommended/hide_blog_carousels.js
```diff
@@ -10,17 +11,18 @@ const styleElement = buildStyle(`
   [${hiddenAttribute}] > div :is(img, video, canvas) { display: none }
 `);
 
-const listTimelineObjectSelector = keyToCss('listTimelineObject');
-const blogCarouselSelector = `${listTimelineObjectSelector} ${keyToCss('blogRecommendation')}`;
+const blogCarouselSelector = `${keyToCss('listTimelineObject')} ${keyToCss('carouselWrapper')}`;
```
AprilSylph, line 14: Wouldn't this now be more accurately... `carouselSelector`? (Or `carouselWrapperSelector` to be more verbose, but I don't think it's a useful distinction to make here.)

marcustyphoon: Good point! And the elements selected by them are carousels of either type, so that makes sense to change too.

src/scripts/no_recommended/hide_blog_carousels.js
```diff
@@ -10,22 +11,23 @@ const styleElement = buildStyle(`
...
-const hideBlogCarousels = blogCarousels => blogCarousels
-  .map(getTimelineItemWrapper)
-  .filter(timelineItem =>
-    timelineItem.previousElementSibling.querySelector(keyToCss('titleObject'))
-  )
-  .forEach(timelineItem => {
-    timelineItem.setAttribute(hiddenAttribute, '');
-    timelineItem.previousElementSibling.setAttribute(hiddenAttribute, '');
+const hideBlogCarousels = carousels =>
+  carousels.forEach(async carousel => {
```
AprilSylph, line 17: This isn't really part of the review, just thinking out loud: isn't this pattern kinda silly? Naming an argument just to loop through it? It feels like there should be a way to avoid that without changing how `pageModifications` works, but I just can't quite think of it...

marcustyphoon: Yeah, one could take a callback and run it on every result, for sure. I guess the questions are:
a) Do we ever not do all of them? Yes—sometimes we fire only on the first, though that's iirc only when we assume there's one, in which case it wouldn't matter.
b) Do we ever do anything else? I don't think so but I would have to check?

marcustyphoon: Ah, yeah, we do do some other things, seems like. We could have two registerable options, though.

AprilSylph: Mutual Checker does something weird IIRC.

marcustyphoon: Sidebar util definitely does some... things.

AprilSylph: I'm still hung up on how to do it, _without modifying `pageModifications`_, though. It feels like I should know how to do it.

marcustyphoon: Oh—I mentally misread "without" as "by." Interesting! Nothing immediately comes to mind.

AprilSylph: I think I might have fungus in my brain.

AprilSylph (approved):
- [x] Blog carousels hidden on the dashboard
- [x] Blog carousels hidden on hubs
- [x] Tag carousels _not_ hidden on the dashboard

it's good code

PR #1359, opened 2023-12-28
Experiment: Bundle script files

marcustyphoon:
### Description

I don't know how much more I'll work on this—possibly none—but this is a quick (~3hr) demo using [esbuild](https://esbuild.github.io/) to bundle all of the script .js files into a single content script (and strip type information, allowing the use of typescript files). 

It seems like no one has really put effort into a plugin that parses a webextension's `manifest.json` for a bundler. There's `@parcel/config-webextension`, but Parcel [doesn't support top level await I guess](https://github.com/parcel-bundler/parcel/issues/4028) so, you know, uh. The esbuild config used here is a mess.

_(Edit: I don't think this matters for Firefox, but Chrome automatically updates non-content-script JS files when you save an unpacked extension, but requires a manual update for content script files; one could adjust this to have a no-op content script that just imports the actual bundled file to preserve just-refresh-Tumblr "HMR"-like behavior in dev mode.)_

### Testing steps

- clone the branch
- `npm i`
- `npm run esbuild` or `npm run esbuildwatch`
- load `/dist` as your extension folder instead of `/src`


marcustyphoon:
For additional fun, one can also use the JSX transform in place of `dom`, given a few tweaks to the function signature.

PR #1367, opened 2024-01-17
Show Originals: Allow filtering whitelisted/own blogs

marcustyphoon:
One of those things that feels like an easy functionality win but is immediately hard to explain and reason about every consequence of once you think about it at all. Don't worry about merging this too much; I'm probably one of a fairly small number of people who want it.

### Description

This quality of life tweak slightly adjusts the behavior of Show Originals' whitelist and "always show my own reblogs" settings. Previously, they would cause viewing a blog that you have whitelisted or own in the blog view to unconditionally show every post, so the Show Originals controls were locked to `All Posts` on those blogs. Sometimes one wants to filter original posts from a specific blog to find something, and having to temporarily un-whitelist it or turn off the own reblogs checkbox to do so feels a bit silly.

This replaces the lock with a "default to `All Posts` but enable both buttons" behavior on those blogs, and ignores the own-reblogs and whitelist rules when processing posts in the blog view so that the originals-only modes work when clicked.

Caveat: With the current simple-ish PR code, this means there's no longer a way to filter a group blog you're part of by "original posts and my reblogs." For correctness, this tweak to the post processing (ignoring the own-reblogs and whitelist rules) should apply only to blog view blogs whose posts would unconditionally otherwise be shown, i.e. whitelisted blogs and own _non-group_ blogs, but that's a bit hard to represent in code succinctly; still poking at it.

### DIFF
--- a/src/scripts/show_originals.js
+++ b/src/scripts/show_originals.js
@@ -4,7 +4,7 @@ import { getPreferences } from '../util/preferences.js';
 import { onNewPosts } from '../util/mutations.js';
 import { keyToCss } from '../util/css_map.js';
 import { translate } from '../util/language_data.js';
-import { userBlogs } from '../util/user.js';
+import { userBlogNames } from '../util/user.js';
 
 const hiddenAttribute = 'data-show-originals-hidden';
 const lengthenedClass = 'xkit-show-originals-lengthened';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Oh, and as a bit of background, #677 implements the full-fat version of this kind of behavior, insofar as it allows one to manually view the posts on a blog one has muted, and continues to respect the rest of one's mute list while in that override mode. I did that using multiple attributes that get applied during the post processing step based on the reason that a post is hidden, such that the modal logic could hide some and not others. Of course, I'm trying to avoid that level of complexity here.

marcustyphoon:
I just thought of a less silly way to do this (I think). Also, this method isn't compatible with Patio.

PR #1377, opened 2024-01-20
Show Originals: Function on blogs on Tumblr Patio UI 

marcustyphoon:
### Description

This improves and cleans up the code that Show Originals uses to determine what mode to apply to a timeline, making it treat user blogs viewed in the in the "Tumblr Patio" feature preview the same way it treats the blog view modal.

Specifically:

- Blog timelines are detected by parsing the `timeline` property instead of assuming anything in a blog view container is a blog. The regex is specific enough to exclude single post views, so extra code is not needed to exclude them.
- "Channel" timelines are excluded by checking for the post compose bar above them.
- Disabled blog checking is simplified by ordering the "where am I" code.

### Testing steps

Confirm that Show Originals inserts its control element and functions on:
- https://www.tumblr.com/dashboard
- https://www.tumblr.com/dashboard/blog_subs
- https://www.tumblr.com/timeline/blog_subscriptions (I guess—kind of pointless now)
- the blog view modal
- the permalinked blog view https://www.tumblr.com/wip
- the "More from @[user]" section beneath a single post viewed in the blog view modal, but not on the post itself https://www.tumblr.com/wip/721828490934124544/hey-there
- the following feed in the "Tumblr Patio" feature preview 
- a specific blog viewed in in the "Tumblr Patio" feature preview 

Confirm that Show Originals does not insert its control element on:
- "channel" pages (https://www.tumblr.com/blog/blogname)


marcustyphoon:
(I suppose I could get cute and have each Patio tab remember its Show Originals mode state separately, using the id attributes on the things. Would have to fall back to not doing that in case those ids go away, though, and also is that really worth the extra code? Probably not.)

AprilSylph (approved):

PR #1378, opened 2024-01-20
feat(Show Originals): Allow filtering whitelisted/own blogs

marcustyphoon:
Don't worry about reviewing this too much; I'm probably one of a fairly small number of people who want it.

### Description

This quality of life tweak slightly adjusts the behavior of Show Originals' whitelist and "always show my own reblogs" settings. Previously, they would cause viewing a blog that you have whitelisted or own* in the blog view to unconditionally show every post, so the Show Originals controls were locked to `All Posts` on those blogs. Sometimes one wants to filter original posts from a specific blog to find something, and having to temporarily un-whitelist it or turn off the own reblogs checkbox to do so feels a bit silly.

This replaces the lock with a "default to `All Posts` but enable both buttons" behavior on those blogs, and adds additional CSS to whitelisted reblogs during processing so that they can be hidden (only) in the forced originals-only modes.

_*unless it's a group blog, in which case the posts might not all be your own_

### Testing steps

- Enable Show Originals and navigate to a whitelisted blog. Confirm that the Show Originals toggle is off but can be enabled, and that it works correctly when enabled.
- Enable Show Originals and navigate to your own blog with the show own reblogs preference enabled. Confirm that the Show Originals toggle is off but can be enabled, and that it works correctly when enabled.


marcustyphoon:
- [x] copy #1984 ~~code~~ _comment structure?_ for maintainability

PR #1382, opened 2024-01-24
Improve inject reliability on webkit

marcustyphoon:
### Description

In short, this leaves our injected script elements in the DOM while they're pending and being observed-by-a-mutation-observer, rather than appending-and-immediately-removing them.

Of course, we don't officially support every browser, but there are lots of ways to install web extensions. I've been running a dev branch of XKit Rewritten in Safari for a few days and sometimes things just wouldn't load. This seems? to help? I'll have to do more testing to be sure.

(Frankly I'm even less sure how the old way could lead to inconsistent behavior—surely it should either always work or never work, unless there's some window of n milliseconds in which Webkit still registers event listeners on things removed from the DOM? Yeah, no idea.)

Aside: now I'm extremely curious if #920 works.

### Testing steps

- confirm basic functionality

### DIFF
--- a/src/util/inject.js
+++ b/src/util/inject.js
@@ -38,9 +38,11 @@ export const inject = async (func, args = [], target = document.documentElement)
         if (mutations.some(({ attributeName }) => attributeName === 'data-result')) {
           observer.disconnect();
           resolve(JSON.parse(script.dataset.result));
+          script.remove();
         } else if (mutations.some(({ attributeName }) => attributeName === 'data-exception')) {
           observer.disconnect();
           reject(JSON.parse(script.dataset.exception));
+          script.remove();
         }
       });
 
@@ -49,7 +51,6 @@ export const inject = async (func, args = [], target = document.documentElement)
         attributeFilter: ['data-result', 'data-exception']
       });
       target.append(script);
-      script.remove();
     });
   } else {
     target.append(script);


marcustyphoon:
> I'll have to do more testing to be sure.

Oh, yeah, this absolutely works.

nightpool (approved):
I'm kind of surprised this ever worked on Chrome, considering that MutationObservers are supposed to run async.

marcustyphoon:
It's not even the observer being async that's weird—it's that it tracks mutations on disconnected elements. The script element is gone already when the async callback runs to create the mutation, let alone when the observer processes it:

```diff
  const script = dom('script', { nonce }, null, [`{
+   const script = document.currentScript;
    const { dataset } = document.currentScript;
    const ${name} = ${func.toString()};
    const returnValue = ${name}(...${JSON.stringify(args)});
    ${async
      ? `returnValue
          .then(result => {
+           console.log('setting dataset: ', '${func.name}', document.currentScript, script.isConnected);
            dataset.result = JSON.stringify(result || null);
          })
          .catch(exception => { dataset.exception = JSON.stringify({
            message: exception.message,
            name: exception.name,
            stack: exception.stack,
            ...exception
          })})
        `
      : 'dataset.result = JSON.stringify(returnValue || null);'
    }
  }`]);
```

```
setting dataset: apiFetch null false
```

marcustyphoon:
Interestingly, #920 seems to work reliably without modification so far. So I guess an event listener on a detached element is fine, as opposed to a mutation observer.

AprilSylph (approved):

PR #1384, opened 2024-01-28
Vanilla Video: Fix blank video on pause in Safari

marcustyphoon:
### Description

I don't know why this works. Someone please help.

This fixes an issue where, in Safari, pausing a video with Vanilla Videos on causes it to go white (or, if popped out, black). It does so by adjusting the DOM position of the Vanilla Videos video element one higher (outside the `videoContainer` element). Are there listeners on that element which do something weird? I have no idea.

### Testing steps

- Confirm that Vanilla Video works as expected in Chromium and Firefox. (I did test Chromium; I did not Firefox.)


marcustyphoon:
Seems like even this doesn't completely solve the problem if Cleanfeed is affecting the video in question as well?

marcustyphoon:
Nope, it's definitely something more than just this.

marcustyphoon:
I'm seeing this happen on Tumblr's video elements in my Safari 17.6 with no extensions at all, so it's probably just Safari being screwed up?

PR #1397, opened 2024-02-06
Tweaks: Remove unread post count from tab title

### LINKED ISSUE #1360: Correct notification count in tab title when ignoring unread following posts
### Concept

The `Hide the Home/Following unread badge` tweak currently only hides the actual badge via css. This has the downside of leaving the actual browser tab title unaffected, which is constantly updated to `(n) Tumblr`, where `n` is the inbox + unread count.  
image
It would be nice if the Unread count was ignored here too, although this would of course increase the complexity of the patch from a simple css inject.  
Since realistically, noone would want the Unread badge hidden but the unread count included in the window title, this should likely not be a separate patch, but an update of the exisitng one.  
*The easiest way to accomplish both badge hiding and the window title fix is probably patching the `unread` field in the `/counts` response to `0`. This would also make the badge hiding immune to tumblr css changes.*

*My apologies if this is what is meant by the "Tab titles" entry in the `Feature Progress/unplanned` board, but it's pretty ambiguous*

marcustyphoon:
### Description

As discussed and prototyped in the linked issue, this is April's code that adds functionality to the `Hide the Home/Following unread badge` tweak to also hide the unread post count in the browser tab title.

Resolves #1360.

### DIFF
--- a/src/scripts/tweaks.json
+++ b/src/scripts/tweaks.json
@@ -90,7 +90,7 @@
     },
     "hide_following_notification_badge": {
       "type": "checkbox",
-      "label": "Hide the Home/Following unread badge",
+      "label": "Hide the Home/Following unread count",
       "default": false
     },
     "hide_activity_notification_badge": {
--- a/src/scripts/tweaks/hide_following_notification_badge.js
+++ b/src/scripts/tweaks/hide_following_notification_badge.js
@@ -1,14 +1,31 @@
 import { keyToCss } from '../../util/css_map.js';
+import { dom } from '../../util/dom.js';
 import { buildStyle } from '../../util/interface.js';
 import { translate } from '../../util/language_data.js';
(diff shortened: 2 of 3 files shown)


marcustyphoon:
I think it would probably make sense to adjust the name of the tweak as well. Something like `Hide the unread post count in the`—ugh I don't know how to make it succinct.

AprilSylph:
I think `Hide the Home/Following unread count` would make sense. Previously it was limited to the badge, now it hides the count everywhere... right?

marcustyphoon:
I don't think people who are looking for it because they don't like it being in the tab titles will find it that way, but yeah.

AprilSylph (changes requested):
https://github.com/AprilSylph/XKit-Rewritten/pull/1397#issuecomment-1930807363

i maintain that "unread count" makes way more sense now

AprilSylph (approved):

PR #1408, opened 2024-03-02
Quick Tags: Preserve tags when scrolling

### LINKED ISSUE #1380: Quick Tags: Added tags disappear when scrolling
### Platform

MacOS 14.2.1 arm64

### Browser

Chrome 120.0.6099.216

### Addon version

0.23.3

### Details

When one adds tags to a post via Quick Tags, XKit Rewritten edits the added tags onto the post itself (Tumblr's code isn't aware of this). Due to the "virtual scroller" performance improvement code, Tumblr removes posts from the page when they're scrolled off of the screen and adds them back when they're scrolled back on; this process removes XKit's tag edits and doesn't ask the server if the post has changed, so the post will appear as if Quick Tags' edits weren't done. Refreshing the timeline in question or navigating to a different Tumblr page fixes this, of course.

This would be a lot of work to fix any way that I can think of and is fairly minor, so my guess is that we won't fix it, but I'll report it anyway so that it's documented.

marcustyphoon:
> This would be a lot of work to fix any way that I can think of and is fairly minor, so my guess is that we won't fix it

Eh, it wasn't actually _that_ many lines of code.

### Description

This:

- Uses Trim Reblogs' button-adding code in Quick Tags
- Extracts the "add fake tags" code into a function and re-applies it to posts when an edited post is added back to the page by the "virtual scroller" code.

Resolves #1380.

### Testing steps

- Add tags to a post with Quick Tags
- Scroll away until the post disappears, then scroll back
- Confirm that the post still appears to have the added tags
- Edit the post with the normal Tumblr editor, changing its tags again
- Scroll away until the post disappears, then scroll back
- Confirm that no erroneous changes persist on the post

### DIFF
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -2,6 +2,7 @@ import { cloneControlButton, createControlButtonTemplate } from '../util/control
 import { keyToCss } from '../util/css_map.js';
 import { dom } from '../util/dom.js';
 import { filterPostElements, getTimelineItemWrapper, postSelector } from '../util/interface.js';
+import { translate } from '../util/language_data.js';
 import { megaEdit } from '../util/mega_editor.js';
 import { showErrorModal } from '../util/modals.js';
 import { onNewPosts, pageModifications } from '../util/mutations.js';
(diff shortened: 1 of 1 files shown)


src/scripts/quick_tags.js
```diff
@@ -253,6 +265,8 @@ export const clean = async function () {
   $(`.${buttonClass}`).remove();
   $(`.${excludeClass}`).removeClass(excludeClass);
   $(`.${tagsClass}`).remove();
+
+  editedTagsMap = new WeakMap();
```
AprilSylph, line 269: fucked up that JS has `Map.prototype.clear()` but not `WeakMap.prototype.clear()`

AprilSylph (approved):
Neat!

src/scripts/quick_tags.js
marcustyphoon: I'm pretty sure that would become a regular `Map`, since it would need to keep pointers to all of its entries then!

AprilSylph: 🤔 I suppose so...

PR #1412, opened 2024-03-15
AccessKit Disable Gifs: Minor code refactoring

marcustyphoon:
### Description

Should result in no functional changes. Just some cleanup and refactoring I did before experimenting with the Disable Gifs code.

In a script with this many classnames I find it easier to use css-in-js so you don't lose track of them all.

### Testing steps

https://www.tumblr.com/search/gif
Test Disable Gifs in a post with a gif photoset and in the tag search sidebar on the community header and suggested tag previews.


src/scripts/accesskit/disable_gifs.js
```diff
@@ -14,24 +77,19 @@ const pauseGif = function (gifElement) {
...
     canvas.getContext('2d').drawImage(image, 0, 0);
-
-    const gifLabel = document.createElement('p');
-    gifLabel.className = gifElement.clientWidth && gifElement.clientWidth < 150
-      ? 'xkit-paused-gif-label mini'
-      : 'xkit-paused-gif-label';
-
-    gifElement.parentNode.append(canvas, gifLabel);
+    gifElement.parentNode.append(canvas);
+    addLabel(gifElement);
```
AprilSylph, line 83: I'm pretty sure these two lines of code just need to be the other way around. We're using element insertion order to determine the rendering layer to avoid the need to set `z-index` and potentially fighting with Tumblr's Z indices.
```suggestion
    addLabel(gifElement);
    gifElement.parentNode.append(canvas);
```

AprilSylph (changes requested):
Paused GIFs in posts are currently missing GIF labels with this change.

marcustyphoon:
Ooh. Never like seeing that.

src/scripts/accesskit/disable_gifs.js
```diff
@@ -1,9 +1,72 @@
...
+  background-color: rgb(var(--secondary-accent));
+}
+
+.${backgroundGifClass}:not(:hover) > div {
+  color: rgb(var(--black));
+}
+`);
+
+const addLabel = (element, inside = false) => {
+  if (element.parentNode.querySelector(`.${labelClass}`) === null) {
```
marcustyphoon, line 61: Ah, it's a race condition in the original code that this "fix" only half-mitigates.

Master branch:



Looks like what I assume is hydration stuff causes `processGifs` to run twice, and we only deduplicate if the elements have been added. If `pauseGif` is queued and waiting for `image.onload`, we don't deduplicate.

AprilSylph (approved):

PR #1418, opened 2024-03-23
Refactor text formatting functions

marcustyphoon:
### Description

Should result in no functional changes. This extracts some utility functions that are used in the modals of various scripts into a utility file to reduce repetition.

Open to bikeshedding (different file name, interface.js, modals.js, whatever).

### DIFF
--- a/src/scripts/mass_deleter.js
+++ b/src/scripts/mass_deleter.js
@@ -2,6 +2,7 @@ import { dom } from '../util/dom.js';
 import { megaEdit } from '../util/mega_editor.js';
 import { modalCancelButton, modalCompleteButton, showErrorModal, showModal } from '../util/modals.js';
 import { addSidebarItem, removeSidebarItem } from '../util/sidebar.js';
+import { dateTimeFormat } from '../util/text_format.js';
 import { apiFetch } from '../util/tumblr_helpers.js';
 
 const timezoneOffsetMs = new Date().getTimezoneOffset() * 60000;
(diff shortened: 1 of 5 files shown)


AprilSylph (commented):
if you want bikeshedding... here's some extreme bikeshedding for you...!

---

I feel like trying to follow XKit 7's pattern of grouping utils just for the sake of grouping them was kinda silly, and I regret doing it. Rather than think up names for collections of utils which don't even share any code, I think the easier and clearer thing is to have one export per util file, e.g.

- `src/util/construct_relative_time_string.js`
  - Defines `thresholds`
  - Defines `relativeTimeFormat`
  - Exports `constructRelativeTimeString`
- `src/util/date_time_format.js`
  - Exports `dateTimeFormat`
- `src/util/elements_as_list.js`
  - Exports `elementsAsList`

Of course, this pattern doesn't work for util files which export several interdependent values (like `user_blogs.js`), but at least that way we're only doing the group-naming work when it's actually necessary.

marcustyphoon:
Spoken like a true developer-whose-project-at-work-uses-a-bundler, I kid, I kid, don't sue me, I'm innocent :P 

I certainly don't think I would go unequivocally to the opposite extreme, where one winds up with a folder full of function names in alphabetical order. If a set of functions will usually be used together or would need all need to be refactored at the same time, I think it makes sense to be able to find one from the other via context. But yeah, I think there's certainly places for it. `interface.js` is pretty much a grab bag, iirc. I feel like it might make sense to put the functions that take an NPF object together (might have to add one to normalize community posts?)

marcustyphoon:
Reopening this discussion: are you still an advocate of splitting these up? I am not, but I am open to going into it more.

AprilSylph:
Nah, I think grouping is fine now actually. I do think `interface.js` is stupid because its name is too generic to be meaningful, but `text_format.js` is definitely clear enough.

PR #1420, opened 2024-03-24
Minor CSS optimizations

marcustyphoon:
### Description

I don't know how to measure CSS rendering performance, but there's no way this matters at all.

This tweaks some CSS selectors where the final selector (which, if I understand correctly, is evaluated first) results in ~10 classes due to ambiguous CSS map entry naming.


marcustyphoon:
Another one we probably want to remove: https://github.com/AprilSylph/XKit-Rewritten/blob/master/src/features/tweaks/hide_mini_follow.js#L9

edit: https://github.com/AprilSylph/XKit-Rewritten/pull/2097

AprilSylph (approved):

PR #1422, opened 2024-03-25
Show Originals: Add option to hide only reblogs you follow

### LINKED ISSUE #1386: Option to only show reblogs of accounts I'm not following
### Concept

I see a lot of reblogs and duplicated content on my feed. The majority of reblogs I see are for accounts I already follow, so I'm bound to see the original on my feed too, and I much prefer to like the original. However, I don't want to hide reblogs entirely because I do want to discover new content and people. So, I'd much prefer it if there was an option to hide reblogs unless I'm not following the account that is being reblogged, and ideally also the option to choose whether its all other reblogs that are hidden or just those without any contributed content. This would probably make the most sense as a sub-option under the Show Originals option. Hope that makes sense! Thanks :)

marcustyphoon:
### Description

As per the linked issue, this adds an option to Show Originals that shows reblogs if you don't follow the account that's being reblogged. This makes it into a "my friends all reblog each other and I don't really need to see the same content a bunch of times" filter.

Tried it out myself and I can definitely see the appeal, though my dash is mostly people who don't know each other and so most of the content is new to me.

Resolves #1386.

### Testing steps

If desired, unrevert the test commit, enable Show Originals and both "always show reblogs" options, go to the "all posts" tab, and observe that reblogs that you wouldn't expect to see on your dash twice are highlighted green, while self reblogs and reblogs of other users you follow are highlighted red.

### DIFF
--- a/src/scripts/show_originals.json
+++ b/src/scripts/show_originals.json
@@ -20,7 +20,7 @@
     },
     "showReblogsOfNotFollowing": {
       "type": "checkbox",
-      "label": "Always show reblogs of accounts you don't follow",
+      "label": "Always show reblogs of blogs I don't follow",
       "default": false
     },
     "whitelistedUsernames": {
--- a/src/scripts/show_originals.json
+++ b/src/scripts/show_originals.json
@@ -18,6 +18,11 @@
       "label": "Always show reblogs with contributed content",
       "default": false
     },
+    "showReblogsOfNotFollowing": {
+      "type": "checkbox",
+      "label": "Always show reblogs of accounts you don't follow",
+      "default": false
+    },
     "whitelistedUsernames": {
       "type": "textarea",
       "label": "Always show reblogs from these blogs (comma-separated)",
--- a/src/scripts/show_originals.js
+++ b/src/scripts/show_originals.js
@@ -15,6 +15,7 @@ const includeFiltered = true;
 
 let showOwnReblogs;
 let showReblogsWithContributedContent;
+let showReblogsOfNotFollowing;
 let whitelist;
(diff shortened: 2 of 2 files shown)


src/scripts/show_originals.json
```diff
@@ -18,6 +18,11 @@
       "label": "Always show reblogs with contributed content",
       "default": false
     },
+    "showReblogsOfNotFollowing": {
+      "type": "checkbox",
+      "label": "Always show reblogs of accounts you don't follow",
```
AprilSylph, line 23: label nits:
- We typically refer to the user in first-person in preferences, e.g. "Hide my own posts on the dashboard", "Always show my own reblogs"
- Blogs ≠ Accounts

```suggestion
      "label": "Always show reblogs of blogs I don't follow",
```

AprilSylph (approved):
otherwise LGTM

AprilSylph (changes requested):
oops, wrong radio button ^_^;

src/scripts/show_originals.json
marcustyphoon: Yeah, I waffled about that. "Reblogs of blogs" sounds funny. But, indeed, more correct!

AprilSylph: "posts reblogged from blogs I don't follow" would be the more verbose and less funny-sounding option if we care about that...

AprilSylph (approved):
yay

(up to you whether to follow the alternative wording suggestion in https://github.com/AprilSylph/XKit-Rewritten/pull/1422#discussion_r1582707573)

PR #1428, opened 2024-03-29
Mutual Checker: Fix icon insertion on mobile devices

### LINKED ISSUE #1405: Mutual Checker does not function on mobile layout
### Platform

MacOS 13.2.1 arm64, iOS 17.2.1

### Browser

Chrome 122.0.6261.57, iOS Safari 17.2.1

### Addon version

0.23.3

### Details

On viewport widths of less than 540px, the `span` referenced here does not exist and Mutual Checker therefore does not work.

https://github.com/AprilSylph/XKit-Rewritten/blob/0f982348212eab7f2ad2ec09fa5b3fe36b333fa3/src/scripts/mutual_checker.js#L13

marcustyphoon:
### Description

As noted in the linked issue, Mutual Checker does not insert icons when the page viewport width is ≤540px. (Icons will persist when the page is narrowed through this boundary, though.) This is because the code added in #1239 assumes that a popover wrapper exists.

This fixes this issue by:
 - Improving `postAttributionSelector` using a complex selector list in `:not()`, which correctly semantically represents "element that does not have a certain parent" without relying on a certain element existing instead. I have no idea why this pattern is not documented on MDN, by the way.
 - Defaulting to the element itself in `getPopoverWrapper` if the target element isn't surrounded in a popover wrapper, which the mobile-mode attribution isn't.

Resolves #1405.

### Testing steps

- Open your dash with a wide viewport width. Find a post from a mutual with a mutual icon.
- Resize your browser viewport to <540px wide (zooming in makes this easier on a desktop browser). Confirm that the mutual icon is still there. (This was fine before this PR.)

- With your browser viewport <540px wide, open your dash and find a post from a mutual. Confirm that it has a mutual icon.
- Resize your browser viewport to a wide viewport width. Confirm that the mutual icon is still there.

### DIFF
--- a/src/util/interface.js
+++ b/src/util/interface.js
@@ -22,14 +22,12 @@ export const getTimelineItemWrapper = element =>
   element.closest(cellSelector) || element.closest(listTimelineObjectSelector);
 
 /**
- * @param {Element} element Element within a popover wrapper
- * @returns {Element | null} The outermost popover wrapper
+ * @param {Element} element - Element that may be within a popover wrapper
+ * @returns {Element} The outermost popover wrapper or the element
  */
 export const getPopoverWrapper = element => {
   const closestWrapper = element.closest(targetWrapperSelector);
-  return closestWrapper?.parentElement?.matches(targetWrapperSelector)
-    ? closestWrapper.parentElement
-    : closestWrapper;
+  return closestWrapper?.parentElement?.closest(targetWrapperSelector) ?? closestWrapper ?? element;
 };
 
 /**
--- a/src/util/interface.js
+++ b/src/util/interface.js
@@ -26,11 +26,8 @@ export const getTimelineItemWrapper = element =>
  * @returns {Element} The outermost popover wrapper or the element
  */
 export const getPopoverWrapper = element => {
-  let result = element;
-  while (result.parentElement.closest(targetWrapperSelector)) {
-    result = result.parentElement.closest(targetWrapperSelector);
-  }
-  return result;
+  const closestWrapper = element.closest(targetWrapperSelector);
+  return closestWrapper?.parentElement?.closest(targetWrapperSelector) ?? closestWrapper ?? element;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/util/interface.js
```diff
@@ -22,14 +22,12 @@ export const getTimelineItemWrapper = element =>
...
- * @returns {Element | null} The outermost popover wrapper
+ * @param {Element} element - Element that may be within a popover wrapper
+ * @returns {Element} The outermost popover wrapper or the element
  */
 export const getPopoverWrapper = element => {
   const closestWrapper = element.closest(targetWrapperSelector);
-  return closestWrapper?.parentElement?.matches(targetWrapperSelector)
-    ? closestWrapper.parentElement
-    : closestWrapper;
+  return closestWrapper?.parentElement?.closest(targetWrapperSelector) ?? closestWrapper ?? element;
```
AprilSylph, line 30: I feel like, strictly, a function named `getPopoverWrapper` should not return anything other than a popover wrapper. The usages should do `?? element` instead.

(Annoying? Kinda. Clearer? Definitely.)

AprilSylph (changes requested):
Otherwise LGTM, assuming this is tested and working in Chrome latest.

I wonder what the browser support for this kind of usage of `:not()` really is? Is it every browser that supports selector lists in `:not()`, was descendant-testing included from the start...?

src/util/interface.js
marcustyphoon: ...I have no idea why I didn't think of that :D

marcustyphoon:
> I wonder what the browser support for this kind of usage of `:not()` really is? Is it every browser that supports selector lists in `:not()`, was descendant-testing included from the start...?

I tested it myself for https://github.com/mdn/content/issues/32880, and yes. I legitimately believe that MDN not having an example (and having, imo, misleading example text) has caused the lack of awareness and usage.

AprilSylph (approved):

PR #1429, opened 2024-03-29
Cleanfeed: Hide link card images

marcustyphoon:
### Description

Refactors Cleanfeed a bit and hides the images in link cards... which, let's be real, probably comprise more than half of the images on this site that you wouldn't want someone looking over your shoulder at work and seeing.

### Testing steps

Tempted as I am to link to a search that will almost certainly have porn links :P I mostly tested this on https://www.tumblr.com/april-codes/filter-by/link.


src/scripts/cleanfeed.js
```diff
@@ -44,9 +44,38 @@ export const main = async function () {
...
+
+    background-color: rgb(var(--white));
+    background-image: linear-gradient(rgba(var(--black), 0.07), rgba(var(--black), 0.07));
+    content: "Hidden by CleanFeed";
+    color: rgba(var(--black), 0.65);
+    object-fit: cover;
+  }
+  ` + localFlaggedBlogs
+      .map(username => `[title="${username}"] img[alt="${translate('Avatar')}"] { filter: blur(20px); }`)
+      .join('');
```
AprilSylph, line 77: - We should definitely extract the result of this `.map()` to its own constant, to then be joined into one big selector, to avoid repeating the style declaration, i.e.
  ```js
  const localFlaggedBlogsAvatarSelector = localFlaggedBlogs
     .map(username => `[title="${username}"] img[alt="${translate('Avatar')}"]`)
     .join(', ');

  styleElement.textContent = `
    ${localFlaggedBlogsAvatarSelector} {
     filter: blur(20px);
    }

    [...]
  `;
  ```
- We could reduce the repetition even further using `:is()` for `[title="${username}"]` selectors to avoid repeating the `img` selector part, i.e.
  ```js
  const localFlaggedBlogsTitleSelector = `:is(${
    localFlaggedBlogs.map(username => `[title="${username}"]`).join(', ')
  })`;

  styleElement.textContent = `
    ${localFlaggedBlogsTitleSelector} img[alt="${translate('Avatar')}"] {
      filter: blur(20px);
    }
    
    [...]
  `;
  ```

AprilSylph (commented):
Nice! Thoughts on further improvements to the style element construction below.

src/scripts/cleanfeed.js
marcustyphoon: Oh yeah! Hm, my "prioritize minimum diff over just making code good" brain is doing the thing again :D

AprilSylph: My philosophy is that extraneous changes are fine, so long as they're both
1. actually good changes to make, and
2. are siloed into their own commits so that they can be reviewed individually instead of just making one big messy diff that's doing multiple things.

(I know the everything-mess is the default diff view in GitHub, but if we were satisfied with defaults, we wouldn't be here working on this project, now would we?)

marcustyphoon:
Tempted to add an `a` to that `:is()`, as in theory the CSS engine should be able to walk up searching for `a` elements before doing a whole bunch of `title` checks on every `div`, but I'm not positive that's how CSS engines work in this particular case and also I don't know offhand if that element is 100% for sure always an `a`.

src/scripts/cleanfeed.js
marcustyphoon: I'd always wondered how much you look at the individual commits!

AprilSylph: I usually look at the combined diff for my first review, and then per-commit or "changes since last review" thereafter... and then, if I feel the need, one last glance-over of the final combined diff before I leave an approval.

AprilSylph (approved):

PR #1430, opened 2024-03-29
Cleanfeed: Add mature post text blurring

marcustyphoon:


AprilSylph:
> To my knowledge this has never been requested, I don't think, but we could do it anyway.

We can revisit this if there's ever demand for it.

PR #1439, opened 2024-04-18
Fix Communities compatibility via props override

marcustyphoon:
### Description

See discussion #1438 and issue #1441.

One of the ways to fix the discrepancy in behavior between posts in communities (which have an `authorBlog` field in their react prop data containing what would normally be in `blog`) is to simply replace the `blog` field's contents when we query for said data.

This is probably a bad idea for a number of reasons, such as future developer confusion and inconsistency between API calls (would need a helper there too, I guess) and react prop data queries. I only bring this up because, that I can tell, the Tumblr frontend seems to do this in some cases.

(This kind of brings into question why the fields actually are this way, but I can imagine a fair number of good reasons it has to be that way ¯\\\_(ツ)_/¯) 

### Testing steps

- Enable Themed Posts on a communities page and confirm that posting users' blog colors appear.
- You know, other stuff, if we go forward with this method.

### DIFF
--- a/src/util/react_props.js
+++ b/src/util/react_props.js
@@ -25,7 +25,14 @@ const unburyTimelineObject = () => {
  */
 export const timelineObject = async function (postElement) {
   if (!timelineObjectCache.has(postElement)) {
-    timelineObjectCache.set(postElement, inject(unburyTimelineObject, [], postElement));
+    timelineObjectCache.set(
+      postElement,
+      inject(unburyTimelineObject, [], postElement).then(data =>
+        data && data.community && data.authorBlog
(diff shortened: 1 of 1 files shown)


AprilSylph:
> This is probably a bad idea for a number of reasons

1. Editing posts in Communities (the post won't exist on the author blog)
2. Reblogging posts from Communities? (same as above) (not sure if you can do this actually)

marcustyphoon:
Oh, yeah, editing would definitely explode. (I mean, technically we try to always grab live data from the API before editing—stale data from race conditions bad—but of course even having to entertain the thought means you're going down the wrong path.)

PR #1440, opened 2024-04-18
Cleanfeed, Quick Reblog: Include submitting blogs/group blog authors

marcustyphoon:
### Description

As noted in #1441, posts can have not only the single blog that "owns" them, but can also have a secondary blog associated with them: either the submitting blog if the post is a submission, the author blog if the post is on a group blog with the "show author portraits" setting enabled, or the author blog if the post is part of Tumblr Communities.

This expands Cleanfeed's blog-based processing and Quick Reblog's blog-based suggested tag generation to include these secondary blogs.


marcustyphoon:
Notes:
- Mutual checker works with no changes on Communities pages, but it's firing API requests unnecessarily rather than parsing `authorBlog` when `blog` doesn't match. The code change to improve this is somewhat awkward, though.

AprilSylph (approved):

PR #1444, opened 2024-04-30
Combine option pane CSS

marcustyphoon:


marcustyphoon:
So these are not the same.

Might look at this more later to understand why, might not.

PR #1446, opened 2024-04-30
Quick Tags: Minor refactors

marcustyphoon:
### Description

Refactors a few bits of code in Quick Tags that I found confusing on first read. They might originate from before `dom()`; I don't recall. Should result in no functional changes.

### Testing steps

- Confirm that the quick tags post popup appears correctly and works correctly, including fake tag generation.
- Confirm that the quick tags editor popup appears correctly and works correctly.

### DIFF
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -25,7 +25,6 @@ let autoTagAsker;
 let controlButtonTemplate;
 
 const popupElement = dom('div', { id: 'quick-tags' });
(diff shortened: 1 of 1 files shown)


src/scripts/quick_tags.js
```diff
@@ -53,29 +52,25 @@ const checkLength = ({ currentTarget }) => {
   }
 };
 popupInput.addEventListener('input', checkLength);
```
AprilSylph, line 54: we could also move `popupInput`'s definition here and set the `input` event listener during creation

PR #1450, opened 2024-05-03
Mega-edit util: Remove unnecessary inject()

marcustyphoon:
### Description

Uhhhh... silly question: do we actually need this to be injected? That I can tell, Tumblr doesn't modify/instrument window.fetch(), and if this fails, having it potentially pop up in Sentry logs seems like all downside.

### Testing steps

Either:
 - Unrevert the test commit
 - Confirm quick tags still works

Or:
 - Confirm tag replacer still works

### DIFF
--- a/src/util/mega_editor.js
+++ b/src/util/mega_editor.js
@@ -1,5 +1,3 @@
-import { inject } from './inject.js';
-
 let formKey;
 
 const pathnames = {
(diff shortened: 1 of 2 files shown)


src/util/mega_editor.js
```diff
@@ -44,15 +42,11 @@ export const megaEdit = async function (postIds, options) {
...
-    async (resource, body) =>
-      fetch(resource, {
-        method: 'POST',
-        headers: { 'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8' },
-        body
-      }).then(async response =>
-        response.ok ? response.json() : Promise.reject(await response.json())
-      ),
-    [`https://www.tumblr.com/${pathname}`, $.param(requestBody)]
+  await fetch(`https://www.tumblr.com/${pathname}`, {
```
AprilSylph, line 45: this should be a `return` for if we ever want to access the response or, more importantly, any errors

```suggestion
  return fetch(`https://www.tumblr.com/${pathname}`, {
```

marcustyphoon: You get errors with await! But yeah, should be able to get the response too, huh.

AprilSylph:
I feel like this is probably not safe. I think the fact that a content script `fetch` works for these endpoints is likely an oversight on Tumblr's part.

If you're familiar enough with New XKit's code, you'll know that we use the most bogstandard method possible to call any API endpoints, `XMLHttpRequest`, which AFAIK neither the old nor the new Tumblr web app modifies in its page context. But, we still inject it. Why? At some point, someone at Tumblr realised "hey, these `/svc/*` endpoints aren't public, so only we should be able to use them", and slapped on a requirement that the `Referer` header of the request have a `www.tumblr.com` origin. Setting the referer header manually can't be done; browsers don't allow that. We have to run the request in the page context.

Since the mass post editor endpoints aren't under `/svc/*`, I guess they weren't affected by this, but... if a bunch of requests to them start coming in with no `Referer` header, that might look like non-legitimate behaviour. If someone notices that, what are they going to do? Probably block them in the same way.

PR #1451, opened 2024-05-06
Modals: Make small buttons consistent width

marcustyphoon:
### Description

Sets a minimum width on modal action buttons so that there isn't a slight asymmetry when exactly two small buttons are beside each other.

See: https://github.com/AprilSylph/XKit-Rewritten/pull/1279#issuecomment-2096317910


marcustyphoon:
Not a fan.

| before | after |
| - | - |
|  |  |

PR #1452, opened 2024-05-12
Move wiki content source into repository

marcustyphoon:
### Description

You can't PR a github wiki even though they're repositories. I wondered if you could put the source of the wiki in the repository itself to fix this, and apparently you [kind of can](https://github.com/marcustyphoon/wiki-sync-test).

Requires a third-party github action. There are a bunch on the actions marketplace, none of which really stood out to me. Syncing changes back from the wiki to the repository seemed annoying, so this is a one-way sync (single source of truth). Drawback: you can't use the wiki editor anymore, which may provide a better editing experience than whatever one uses to edit .md files.

Is this a good idea? ¯\\\_(ツ)\_/¯ 

### Testing steps

n/a, unless you want to fork the whole repo to an organization, merge this, then edit a file in the wiki folder.

Make sure this PR is up to date with the current wiki content if you merge it, though, naturally, otherwise it'll overwrite any changes between its creation/last update and the merge.

### DIFF
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -2,7 +2,7 @@ name: Sync wiki content
 on:
   push:
     branches:
-      - main
+      - master
     paths:
       - wiki/**
       - .github/workflows/sync-wiki.yml
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -1,0 +1,25 @@
+name: Sync wiki content
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - wiki/**
+      - .github/workflows/sync-wiki.yml
+  workflow_dispatch:
+
+concurrency:
+  group: sync-wiki
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  sync-wiki:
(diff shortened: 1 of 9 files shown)


AprilSylph:
I kinda like the wiki not existing in forks.

Also, I trust all repository contributors to make any wiki changes you deem necessary, without my approval. This is **our** project, not just mine! It just happens to live on my user instead of a specially-created organisation because, to be honest, I didn't plan very far ahead, and now it would be too annoying to move.

If potential or already-made changes need to be discussed, we have the discussions tab.
The "Tech Office" category is fine to use to discuss documentation too.

marcustyphoon:
Makes sense!

To be honest, I wish I could PR changes in this case less for approval and more for organization; it would be nice to be able to queue up a release's wiki changes in advance and cross-reference the wiki diff to the commit list.

(It would, naturally, be ideal to be able to make wiki changes directly in the PR that necessitates them, though the desire to have the wiki's current contents synced with the released version of the extension rather than the state of the master branch kind of precludes just doing that.

Although now that I think about it, one could have a `wiki` branch that's the source for the wiki and update that branch from `master` only at release time, which would do exactly that; that's sort of neat... or, wait, no, just only run the sync action on release, I'm smart. Anyway, eh, still feels messy)

PR #1453, opened 2024-05-13
bump version to 0.23.5

marcustyphoon:
### Description

Drafts a release.

Feel completely free to ignore this! No idea if you have a plan in mind, here. (I know you mentioned 1.0 in passing somewhere).

Depending on the potential timeline, I think it _could_ make sense to start the timers early on getting #1276 distributed and on https://github.com/new-xkit/XKit/pull/2135 eventually being mergeable, separate from any new functionality if that's likely to happen soon. On the other hand, it very easily could not make sense!

### Testing steps

n/a

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "XKit Rewritten",
-  "version": "0.23.4",
+  "version": "0.23.5",
 
   "short_name": "XKit",
   "author": "April Sylph",


AprilSylph (approved):
I actually wanted to do a release yesterday, but didn't quite have the energy.

And as I type that, I remember I can do everything about a release that I usually do locally (version bump + release notes) via the web UI... shame on me...

AprilSylph:
1.0.0 is going to be the Manifest V3 release, I think. XKit can't ever really be "finished" in the same way my simpler addons (e.g. Palettes for Tumblr) can be, so waiting for that for v1.0.0 would be silly. But I think the MV3 conversion deserves a major version bump for sure.

marcustyphoon:
Hm, trying to think just for fun of what a user could experience that would feel like a major... userscript support, obviously :3 Stick it on the roadmap for 2.0! /joke

PR #1460, opened 2024-05-21
 Quick Tags/Trim Reblogs: Fix buttons in Russian and Hindi

### LINKED ISSUE #1459: Quick Tags/Trim Reblogs: Broken in Russian and Hindi
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 123.0.6312.87

### Addon version

master

### Details

Same cause as #1424: Tumblr's normalized language data given in `window.tumblr` only gives one result back if an English-language string corresponds to multiple localized strings, so our entire `translate()` helper is unreliable when used as part of a CSS selector.

In this case, in Russian, both Редактировать and Изменить correspond to to `Edit`.

Not sure what would be the best way to correct this properly (from both sides). Changing the format of the `languageData` to have multiple values corresponding to one key would be cool, though there would obviously have to be backwards compatibility. In a sense I wish ambiguous values weren't included, but that's silly when you're using it to localize extension UI. We could ignore the helper and parse the non-normalized language data, but that's fragile.

Anyway, in the short term we need to target the edit button in a different way. `#managed-icon__edit` should be fine, I imagine.

marcustyphoon:
### Description

As per the linked issue, we can't rely on the translate utility to target DOM elements with CSS/query selectors, as there may be more than one localized string and some may not match the utility's output.

I think this should work? The edit button is always an icon, i.e. not "Edit," right?

Resolves #1459, but the pattern applies more broadly across the extension and another issue should track ensuring they're all made consistent.

### Testing steps

- Ensure that Quick Tags and Trim Reblogs buttons are consistently added to editable posts (including in Russian and Hindi; language should not matter now).

### DIFF
--- a/src/scripts/quick_tags.js
+++ b/src/scripts/quick_tags.js
@@ -2,7 +2,6 @@ import { cloneControlButton, createControlButtonTemplate } from '../util/control
 import { keyToCss } from '../util/css_map.js';
 import { dom } from '../util/dom.js';
 import { filterPostElements, getTimelineItemWrapper, postSelector } from '../util/interface.js';
-import { translate } from '../util/language_data.js';
 import { megaEdit } from '../util/mega_editor.js';
 import { modalCancelButton, modalCompleteButton, showErrorModal, showModal } from '../util/modals.js';
 import { onNewPosts, pageModifications } from '../util/mutations.js';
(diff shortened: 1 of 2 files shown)


src/scripts/quick_tags.js
```diff
@@ -218,8 +217,9 @@ const processPosts = postElements => filterPostElements(postElements).forEach(po
   const existingButton = postElement.querySelector(`.${buttonClass}`);
   if (existingButton !== null) { return; }
 
-  const editButton = postElement.querySelector(`footer ${controlIconSelector} a[href*="/edit/"][aria-label=${translate('Edit')}]`);
-  if (!editButton) { return; }
+  const editIcon = postElement.querySelector(`footer ${controlIconSelector} a[href*="/edit/"] use[href="#managed-icon__edit"]`);
+  if (!editIcon) { return; }
+  const editButton = editIcon.closest('a');
```
AprilSylph, line 222: I think we should just bump the browser versions high enough to use `:has()`. We'll need to bump them pretty high for MV3 soon anyway.

marcustyphoon: Fair. I admit to feeling like there's some sort of purity in having the final version available for a block of browser versions not have a regression bug, though I imagine the whole "I run outdated firefox on purpose" thing probably doesn't apply to this particular set of them.

AprilSylph (approved):

PR #1463, opened 2024-05-22
AccessKit Disable GIFs: Fix recommended blog card hover

marcustyphoon:
### Description

At present, AccessKit does pause GIFs in recommended blog carousels, but they don't unpause on hover. This makes the recommended blog cards into hover-to-unpause targets.

<details><summary>outdated; see comment below</summary>

At present, AccessKit does pause GIFs in recommended blog carousels, but a) they don't unpause on hover, and b) there are sometimes two overlapping gif labels visible due to the way Tumblr makes a blurred background copy of non-square images*. This ~hides~ _removes_ the duplicate labels and makes the recommended blog cards into hover-to-unpause targets.

*~~I can't get a screenshot of this because by the time I reverted this PR, I no longer had any of these. Developing modifications for a nondeterministic feature is annoying.~~ **edit:** https://www.tumblr.com/tagged/gif reliably has some:

</details>

### Testing steps

- Get a recommended blog carousel card with a non-square GIF, ~~by the grace of god or whatever~~ by scrolling down https://www.tumblr.com/tagged/gif.
- Confirm that it is paused correctly, unpauses correctly when the card is hovered, and has a correct-looking label when paused.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -58,10 +58,7 @@ export const styleElement = buildStyle(`
 }
 `);
 
-const blogCarouselBackgroundSelector = `${keyToCss('background')} > *`;
-
 const addLabel = (element, inside = false) => {
-  if (element.matches(blogCarouselBackgroundSelector)) return;
   if (element.parentNode.querySelector(`.${labelClass}`) === null) {
     const gifLabel = document.createElement('p');
     gifLabel.className = element.clientWidth && element.clientWidth < 150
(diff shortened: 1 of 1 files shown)


marcustyphoon:
ooh, this nukes parts of the dom during clean(). I may not be very smart

marcustyphoon:
...Huh. Interesting. I can't currently reproduce the behavior in the screenshot; the canvas element we create is covering the entire square even when the element it's supposed to cover is not. I guess this must be a redpop layout change?

AprilSylph (approved):

PR #1464, opened 2024-05-24
Automatic `styleElement` addition/removal

marcustyphoon:
### Description

Random idea.

This factors the common "append style elements in `main()`; remove style elements during `clean()`" pattern into scripts just exporting `styleElement`, and makes `main()` and `clean()` optional since they may not be needed at all. Very simple scripts look slightly more like user-styles-with-special-sauce now.

Very possibly not worth the review work to make sure I didn't typo anything; I just thought it was sort of a cute idea.


marcustyphoon:
Oops—missed an if around `clean().then(main)`, didn't I. Should probably:

- [ ] refactor to be smarter about this

AprilSylph (approved):
Love it!

PR #1465, opened 2024-05-25
feat(Tag Tracking+): Fetch a consistent number of posts

marcustyphoon:
### Description

This makes Tag Tracking+ fetch an arbitrary round number of posts, to match XKit 7's functionality (per user request).

I mean, no, obviously it doesn't; there's no API endpoint for that, and if there was, we would already be using it. This actually makes Tag Tracking+ fetch posts from a tag until it either a) finds a post the user has seen already, b) runs out of posts to fetch, or c) has fetched at least said arbitrary round number of posts, so that it can tell the user there are more than that number.

The "simplest" way I could think to implement this very-not-simple control flow was to early return out of a function when we definitively know the unread post count. There are definitely other ways to do it, possibly better ones; it's honestly quite a fun challenge to see what clear form you can find, IMO. Try it!

There's no minimum API delay implemented at present; with `maxUnreadCount = 10` this at maximum doubles our normal API requests per second (as I observe 7 or 8 posts being fetched per request). I think XKit 7's count was 10... possibly maybe?

Split diff view is recommended.

### Testing steps

- Confirm that Tag Tracking+ displays `10+` for tags with more than 10 unread posts and the correct number otherwise.


marcustyphoon:
One can, of course, obfuscate the real counts less:

```diff
    if (!resource || posts.every(isUnread) === false) {
      return posts.filter(isUnread).length;
    }
  }
  return Infinity;
};

const refreshCount = async function (tag) {
  if (!trackedTags.includes(tag)) return;

  let unreadCountString = '⚠️';

  try {
    const unreadCount = await getUnreadCount(tag);
    unreadCountString = unreadCount > maxUnreadCount ? `${maxUnreadCount}+` : unreadCount;
  } catch (exception) {
    console.error(exception);
  }
```

AprilSylph:
I don't think this detail is worth the extra code or the extra API requests. One request per tag is nice and simple.

Two years later, and I haven't seen any actual demand for this, so I don't feel compelled to pursue it in spite of the above.

marcustyphoon:
Yeah, I think I've only had it requested the once. Also, #1813; we don't really want to be increasing the API load of this thing, at the moment, imo.

PR #1466, opened 2024-05-27
upgrade to Manifest V3

AprilSylph:
### Description

This PR makes the following changes:

- Performs some directory renames
  - `src/browser_action` &rarr; `src/action` to match the [manifest V3 rename](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action)
  - `src/scripts` &rarr; `src/features` to match the user-facing nomenclature we use
  - `src/util` &rarr; `src/utils` for my own personal satisfaction
- Re-tools `inject()` to work with static files to satisfy MV3 CSP
  - All injectable scripts are now found in their own files
  - The first argument to `inject()` is now a path to be resolved by `browser.runtime.getURL()`
  - Each injectable script is now individually responsible for interacting with `document.currentScript.dataset` to grab arguments and transmit the result of its operation
- Updates the manifest format to V3
- Adds a new banner to the addon popup prompting for host permission for `*://www.tumblr.com/*`; this is necessary because [some browsers treat `host_permissions` as optional permissions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions#requested_permissions_and_user_prompts) :roll_eyes:
- Bumps the minimum browser versions to [those which support `:has()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility) for our own later convenience; also to squash invalid `web-ext lint` warnings about Firefox for Android compatibility based on the minimum Gecko version instead of the minimum Gecko Android version
- Bumps the addon version to `1.0.0`, and removes the "Beta" label from the version displayed in the popup

#### Feature testing: Installation

To make sure the new permissions prompt is working:
1. Install the branch version of the addon to a browser which treats `host_permissions` as optional, e.g. Firefox
2. Open the XKit Rewritten control panel
    - **Expected result**: The warning banner appears at the top, with the `Always allow on www.tumblr.com` button
3. Click `Always allow on www.tumblr.com`
    - **Expected result**: You are prompted to grant the host permission
4. Grant the host permission
    - **Expected result**: The warning banner disappears

#### Smoke testing

Test each feature to make sure it's still working as it was before.
Panorama is excluded from this list, since it is already broken.

- [x] AccessKit
- [x] Anti-Capitalism
- [x] Classic Search
- [x] CleanFeed
- [x] Collapsed Queue
- [x] Hide Avatars
- [x] Limit Checker
- [x] Mass Deleter
- [x] Mass Privater
- [x] Mass Unliker
- [x] Mirror Posts
- [x] Mutual Checker
- [x] No Recommended
- [x] NotificationBlock
- [x] Open in Tabs
- [x] Painter
- [x] PostBlock
- [x] Quick Reblog
- [x] Quick Tags
- [x] Quote Replies
- [x] Scroll to Bottom
- [x] Seen Posts
- [x] Shorten Posts
- [x] Show Originals
- [x] Tag Replacer
- [x] Tag Tracking+
- [x] Themed Posts
- [x] TimeFormat
- [x] Trim Reblogs
- [x] Tweaks
- [x] Vanilla Audio
- [x] Vanilla Video

#### Usage testing

> [!IMPORTANT]
> If you are using regular Firefox (i.e. not Nightly or Developer Edition), you _must_ install the branch version of the addon via `about:debugging`, which means the addon **and its config** will be lost as soon as you close the browser!

1. On your main browser where you use XKit Rewritten, download a config backup
2. Disable the release version of XKit Rewritten
3. Load the branch version of XKit Rewritten:
    - On Chromium: unzip the addon, then load as unpacked via `chrome://extensions`
    - On Firefox Developer Edition or Firefox Nightly:
      - Disable the addon signature requirement (`about:config` &rarr; `xpinstall.signatures.required` &rarr; **`false`**)
      - Install the addon ZIP via `about:addons` &rarr; :gear: &rarr; Install Add-On From File
    - On Firefox: Install the addon ZIP as a temporary extension via `about:debugging`
4. Restore your config on the branch version of XKit Rewritten
5. Use Tumblr as normal for the rest of the day
    - **Expected result**: Other than some modifications now being applied a few frames slower, your Tumblr/XKit experience should be the same as ever


src/utils/inject.js
```diff
@@ -0,0 +1,39 @@
...
+
+  return new Promise((resolve, reject) => {
+    const attributeObserver = new MutationObserver((mutations, observer) => {
+      if (mutations.some(({ attributeName }) => attributeName === 'data-result')) {
+        observer.disconnect();
+        resolve(JSON.parse(script.dataset.result));
+        script.remove();
+      } else if (mutations.some(({ attributeName }) => attributeName === 'data-exception')) {
+        observer.disconnect();
+        reject(JSON.parse(script.dataset.exception));
```
marcustyphoon, line 28: Not the world's greatest error printout in Firefox :D Perhaps we should construct an `Error` here and throw that. Not the fault of this diff, but.

src/utils/inject/unbury_timeline_object.js
marcustyphoon, line 1: Asynchronous inject can result in race conditions where there were previously not any. (Well, maybe this isn't what you would think of of as a race condition? Anyway, by the time the code runs, the target may not be attached to the DOM any more, e.g. if you've scrolled past it.)

Presuming we don't want to pursue instant injection, at least at the moment, this should be fairly easy to handle. I would probably just decline to run `unburyTimelineObject` and never resolve the inject promise, at first thought? That would stall the function awaiting the inject forever rather than making it have to put a conditional on the result. Not sure how garbage collection works in that case, though; we don't want to leave DOM elements un-GC'd because they have event listeners pending.

src/utils/react_props.js
```diff
@@ -4,47 +4,20 @@ import { primaryBlogName, userBlogNames, adminBlogNames } from './user.js';
...
-      return blog ?? blogSettings;
-    } else {
-      fiber = fiber.return;
-    }
-  }
-};
-
 const unburyNotification = async () => {
   const notificationElement = document.currentScript.parentElement;
   const reactKey = Object.keys(notificationElement).find(key => key.startsWith('__reactFiber'));
```
marcustyphoon, line 23: A forgotten migration!

marcustyphoon (changes requested):

marcustyphoon (changes requested):
https://github.com/AprilSylph/XKit-Rewritten/blob/381f0235df8a827673cedf1a71b90eb89ad859c9/src/utils/react_props.js#L102

also, this one!

marcustyphoon:
What do you think about bumping to Chrome 121 for https://caniuse.com/mdn-css_properties_scrollbar-width (so we don't have a Firefox-only feature) or perhaps Chrome 117 for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy (which I found a use for in #1482)?

AprilSylph:
All features smoke-tested and working in Firefox 128.0b3.

AccessKit, PostBlock, and Tweaks tested and working in Chromium 126.0.6478.55.

Will merge with one approving review. If you'd like to test extensively on your browser of choice, please feel free to edit the PR description to reset the checklist.

marcustyphoon:
- [x] most features smoke tested in safari
- [ ] smoke test in Firefox for Android

src/manifest.json
```diff
@@ -65,13 +67,14 @@
...
+  "minimum_chrome_version": "105",
   "browser_specific_settings": {
     "gecko": {
-      "strict_min_version": "89.0a1"
+      "id": "{6e710c58-36cc-49d6-b772-bfc3030fa56e}",
+      "strict_min_version": "121.0"
     },
     "gecko_android": {
-      "strict_min_version": "113.0"
+      "strict_min_version": "121.0"
```
marcustyphoon, line 77: Currently unable to test on Android; probably will be able to tomorrow. But will this need to be 128? Has anyone tested on 127?

marcustyphoon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1889402
https://github.com/mdn/content/pull/34094

Uh. Hm. Are we supposed to request host permissions in manifest.json?

Obviously in Chromium land, no we aren't, since we only have static content scripts. But if in Firefox land we do need host permissions, and if in the future Firefox will ask the user during install if they want to grant the host permissions we request in the manifest, and if in the present Firefox will apparently unconditionally the host permissions we request in the manifest...

marcustyphoon (dismissed):
Smoke tested in Firefox for Android 126.0.1, and by "smoke tested" I mean "it took me more than ten minutes just to get web-ext to load the damn thing on a Nexus 5X and I cannot for the life of me get it to work a second time, but I think Vanilla Videos was working that one time, probably."

(Aside: if it were me, I might consider rebasing, squashing to a "rename everything without changes" commit and an "implement MV3" commit, then merge committing that, so they're separate in the repo history. Kind of a pain, but might make it easier to merge into PR branches.)

marcustyphoon:
Also tested in Firefox for Android 121.1.0.

Also tested: my patience.

AprilSylph:
> (Aside: if it were me, I might consider rebasing, squashing to a "rename everything without changes" commit and an "implement MV3" commit, then merge committing that, so they're separate in the repo history. Kind of a pain, but might make it easier to merge into PR branches.)

I could cherrypick out the renames into a separate PR, merge that, then rebase this PR with those changes already in the main branch?

marcustyphoon:
That would work! Definitely not a big deal if I had to guess.

marcustyphoon:
We could also consider setting the minimum Firefox version to at least 127, in which users [will (it seems)](https://bugzilla.mozilla.org/show_bug.cgi?id=1889402) continue to have the extension work in the background without manually granting origin permissions. I did test this using web-ext and it does seem to work in that case, while 126 requires the click.

There may be other MV3 improvements in 128, as well, but that releases 4 weeks from now and I'm not keen on delaying bugfix/feature updates for Firefox users until then-plus-however-long-it-takes-people-to-update.

AprilSylph:
hm. c9d5924 should've been included in #1507, technically. oh well.

AprilSylph:
also, to be clear, my plan for this weekend is:
1. today, get Manifest V3 merged, but not released
2. tomorrow, fix the TimelineV2 shenanigans, and _then_ do a release

marcustyphoon (approved):
Squash merged the old branch, reverted the result, squash merged this branch, merged the old revert and the new squash, got an empty diff. Conclusion: the resulting file state of this PR has not changed, and thus tests should continue to be valid.

AprilSylph:
Next up: time to rebase #1501.

PR #1467, opened 2024-05-27
Manifest V3 instant injection demo

marcustyphoon:
### Description

There are a number of methods to do instant main world script injection; this one uses a cached dynamic import  to load script body files, bubbled custom events to request their execution and communicate their result, and the event target of said bubbled event and the `this` property on the script function to communicate the target element when needed.

Yes, I actually found a use for `Function.prototype.bind`/`Function.prototype.apply`; only took about a million years. No, `page_context.js` probably isn't in the right place in the repo.

Previously: #1330; preceding work is linked in each PR.


AprilSylph:
I like this. I feel like we should get the MV3 branch merged first, though.

PR #1472, opened 2024-05-31
Create optional eslint pre-commit hook

marcustyphoon:
### Description

Something I was playing around with in another repository; I have no experience using pre-commit lint hooks personally yet so I don't know if I find them helpful or clunky.

This allows developers to—optionally—install a pre-commit hook that runs eslint locally on staged files before they're committed, preventing the commit if they are not in the semistandard style. This is presumably nice for avoiding getting yelled at by the CI action.

(Traditional `husky` installs include a `"prepare": "husky"` package script, ensuring that all developers have the hook enabled. This does not.)

### Testing steps

- As per the readme, run `npm run enable-hooks`.
- Make a change with a lint error (`let a;`), attempt to commit it, and confirm that the developer is informed about the error and must fix it before continuing.
- Make a change with no lint error and observe the small added delay when committing it.
- Make a change with a lint error (`let a;`) on a different branch and confirm that no hooks run (I chose `husky` primarily because it does not break in this scenario).
- Run `npm run disable-hooks`, make a change with a lint error (`let a;`), attempt to commit it, and confirm that it is no longer prevented.


AprilSylph:
Questions:
1. What's the rationale behind making it optional? I feel like it would be a smoother experience to install the hook automatically.
2. Why not do the full lint with `npm test`? `web-ext lint` errors are errors on CI, so...
3. Could this be a precursor to us switching to the actual `semistandard` package for true autoformatting? :D

marcustyphoon:
1. I know some developers don't like commit hooks! For example, the MV3 branch would have been impossible to work on in advance of web-ext being updated without bypassing hooks on each commit, and my setup doesn't have an easy way to bypass. I've seen arguments about whether they discourage frequent committing, as well, given that you still have tests run on push via CI. Didn't know what your opinion was on 'em.
2. Fair; definitely an option! (I do wish it was faster. On another repo I use `concurrently` for multiple tasks; wouldn't be hard to set up here, though I'm not sure how well modified cli output in hooks works with some git setups.)
3. I don't know of a difference between using the `semistandard` package and just running eslint!

marcustyphoon:
thaaaaaat was not an intentional review request. codeowners is... interesting. but of course, if you are interested in revisiting this, april, feel free!

package.json
```diff
@@ -4,15 +4,18 @@
...
   "scripts": {
     "test": "web-ext lint",
-    "posttest": "eslint src/",
+    "posttest": "eslint src/ --cache",
     "autofix": "eslint src/ --fix",
     "start": "web-ext run",
-    "build": "web-ext build"
+    "build": "web-ext build",
+    "enable-hooks": "husky",
+    "disable-hooks": "git config --unset core.hooksPath"
```
AprilSylph: Couldn't we avoid the extra dev dependency and just set `core.hooksPath` ourselves here?

(Proposed folder rename: `.husky` &rarr; `.githooks`)

.gitignore
AprilSylph, line 1: Alphabetised, please :friendysmile:

package.json
marcustyphoon: Yeah, I don't know why I thought these were so complicated back in the day. (Husky allows per branch hooks, which we don't need, and does nothing if you don't autoinstall.)

marcustyphoon:
Eh, that's really my fault; I didn't press re-review!

(I wasn't sure if, because my initial review request was by accident, it should have been considered—well, see, to end this sentence I would have to talk about the concept of things being officially on or not on your review queue if they're marked as ready-for-review? Which I recall you mentioning doesn't really make sense? But we've also acknowledged that it doesn't make sense to send you 45 concurrent review requests? So I don't know how that's officially supposed to work.)

PR #1475, opened 2024-06-01
Quick Reblog: Restore changed popup font size/weight

### LINKED ISSUE #1474: Quick Reblog menu Scaled too Big
### Platform

macos 14.5

### Browser

Firefox 126.0.1

### Addon version

v0.23.5

### Details

The quick reblog popup seems to be drawn twice as large as previously.

Screenshot 2024-05-31 at 15-53-10 Tumblr

marcustyphoon:
### Description

Resolves #1474.

As noted in the linked issue, a change to Tumblr's CSS resulted in the Quick Reblog menu's fonts changing size and weight, since we did not previously specify them.

(Not sure exactly who I'm writing this note for, as the main person who reads these is the one who pointed the information out to me—thanks April—but the change appears to be that `button`, `input`, `textarea`, and `select` elements have `font: inherit`; presumably some or all did not previously and used browser defaults.)

I'm not actually at all certain that this actually does restore the appearance the menu had prior to both changes, with and without Palettes for Tumblr overrides/XKit 7 Old Blue, in both Firefox and Chromium. I could find one screenshot I had incidentally (Chromium, Palettes-with-font-override) and side-by-side it looks fine?

No idea if this results in Favorit being used where it previously wasn't, or something like that.

### Testing steps

Unless one has screenshots to compare, this might just have to be the eye test?

### DIFF
--- a/src/scripts/quick_reblog.css
+++ b/src/scripts/quick_reblog.css
@@ -6,6 +6,8 @@
   border-radius: 3px;
   overflow: hidden;
   color: rgb(var(--black));
+  font-size: .875rem;
+  font-weight: normal;
   background-color: rgb(var(--white));
   background-image: linear-gradient(rgba(var(--black), 0.13), rgba(var(--black), 0.13));
   display: flex;


AprilSylph (approved):
Yup, this is normal-looking.

PR #1481, opened 2024-06-02
Refactor: Use `:has()` in edit button detection

marcustyphoon:
### Description

#1460, but the right way around once we require Firefox 121.

Also includes a small refactor that was originally in #1501.


AprilSylph (approved):

marcustyphoon:
Forgot to document that this change is included in #1573.

PR #1482, opened 2024-06-02
Parse `data-timeline-id` on Tumblr's new timeline component

### LINKED ISSUE #1480: Orignal post tab no longer showing on blogs
### Platform

Windows 11

### Browser

Firefox 12.0.1

### Addon version

v0.23.5 (beta)

### Details

The original posts tab that used to appear on blogs is no longer showing

firefox_13vhfcN27G
firefox_Dc7aZQYh9F
firefox_hXR0nC95lB

### LINKED ISSUE #1488: Breakage in V2 timelines with endless scrolling off
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 125.0.6422.77

### Addon version

0.23.6

### Details

Sorry for the overly-technical title; this problem includes some upcoming Tumblr code tweaks and it's too complicated to specify which of its problems occur now and which will occur soon.

In V2 timelines with `data-timeline-id`, specifically with endless scrolling disabled, each `listTimelineObject` element is surrounded in a classless `<div>` element (as opposed to one with `data-cell-id` when endless scrolling is enabled, or in a V1 timeline regardless of whether endless scrolling is enabled).

This causes our `getTimelineItemWrapper` utility to fall back to selecting the `listTimelineObject` element itself, which is fine in a lot of cases, but because it is no longer selecting the outermost element of a post, this breaks, at least:

- Recommended carousel hiding and the carousel-becomes-a-line tweak, as stepping up the DOM to find the title doesn't work
- Show Originals (like I said in the PR where I changed this, I knew this one was fragile):

https://github.com/AprilSylph/XKit-Rewritten/blob/a92be15496450118418166676fb3c978868cb20f/src/scripts/show_originals.css#L1

Presumably, the correct way to do this is to update `getTimelineItemWrapper` to handle this case correctly. This isn't exactly easy, though, when both the `div` we want to target and the parent `div` that we do not want to target have no HTML attributes at all.

Ideas?

marcustyphoon:
### Description

Tumblr has a new internal `TimelineV2` component that changes some attributes we use to detect timelines and which posts a timeline is on. There is currently an incomplete rollout of this migration, partially behind some feature flags.

This:

- Moves all of our code and CSS selectors that get run on a timeline element to determine its identity to a utility file (is "consolidation of concerns" a thing?), mostly by changing the `filterPostElements` function's `timeline` option field to take (an) arbitrary function(s) rather than a regular expression. This allows us to update the selector logic without grepping scripts, and using a function as the API lets us use complex behavior like `matches()` if necessary to deal with edge cases during migration or that result from redpop bugs. Bikeshed away.
- Implements parsing of the new`data-timeline-id` attribute for all of the timelines we currently use which at time of updating this description either use the new code or do so if forced with flags... **except** Tumblr Patio columns, for which I'm waiting on tweaks to the ids of since they're too ambiguous at the moment.
- Fixes `getTimelineItemWrapper` on V2 timelines with endless scrolling off, on which the `cell` element has no classname; see the linked issue.

I experimented with a) generating timeline filter functions and timeline css selectors automatically from the same source data and b) making that source data exhaustive in #1485. It was interesting but kind of bad. It does make this diff look extremely boilerplate-y by comparison, though.

Resolves #1488. Resolves #1480.

### Testing steps

If I've left this line in the PR description I haven't exhaustively tested this yet; keep that in mind.

Tests should include having endless scrolling on/off, and having feature flags on/off. I will not post my method for toggling feature flags in production here but it's very easy.

I have also been running a user style to make things quick and easy:

```css
[data-timeline-id] {
  border-top: 2px dashed red;
}

[data-timeline] {
  border-top: 2px dashed green;
}

[data-timeline][data-timeline-id] {
  border-top: 2px dashed blue;
}

[data-timeline-id]::before {
  display: block;
  content: attr(data-timeline-id);
  color: red;
  margin: 1ch;
}

[data-timeline]::before {
  display: block;
  content: attr(data-timeline);
  color: green;
  margin: 1ch;
}
```


marcustyphoon:
Aside: here's a fun use of the fairly-new `groupBy` global, for... well, "performance", in theory. Why process each timeline element for every post that gets added instead of processing each one once?

Well, because performing a simple regex, even from a string, is so fast it doesn't show up in the chromium profiler, and this code is way too long (`Map`s are so clunky) and requires us to bump minimum browser versions, is why. Not worth it.

```js
  if (timeline) {
    const timelineFilters = [timeline].flat().filter(Boolean);

    const postsByTimeline = Map.groupBy(
      postElements,
      postElement => postElement.closest(timelineSelector)
    );

    const filtered = [...postsByTimeline.entries()]
      .map(([timelineElement, postElements]) =>
        timelineElement &&
        timelineFilters.some(timelineFilter => timelineFilter(timelineElement))
          ? postElements
          : []
      )
      .flat();

    postElements = filtered;
  }
```

marcustyphoon:
- [ ] fix `getTimelineItemWrapper` when used in carousel hiding code.

src/util/interface.js
```diff
@@ -51,10 +54,15 @@ export const filterPostElements = function (postElements, { excludeClass, timeli
...
-    postElements = postElements.filter(postElement => timeline.test(postElement.closest('[data-timeline]')?.dataset.timeline));
-  } else if (timeline) {
-    postElements = postElements.filter(postElement => timeline === postElement.closest('[data-timeline]')?.dataset.timeline);
+  if (timeline) {
+    const timelineFilters = [timeline].flat().filter(Boolean);
+    postElements = postElements.filter(postElement => {
+      const timelineElement = postElement.closest(timelineSelector);
+      return (
+        timelineElement &&
+        timelineFilters.some(timelineFilter => timelineFilter(timelineElement))
```
marcustyphoon, line 63: javascript brain: "we have a cute way to shorten "do any of these array items fulfill this singular predicate function'; SURELY there must be an equivalently cute way to shorten "does this singular value fulfill any of this array of predicate functions!"

narrator: there isn't.

AprilSylph (approved):
Tested all modified features with and without endless scrolling disabled; I see no immediate issues.

Have you submitted a Support ticket for the better IDs thing (Patio / Peepr)? I haven't been finding the time to sneak in the work, so it feels like it should be ticketed... somewhere...

marcustyphoon:
Haven't! Will do that.

AprilSylph:
Also, is there anything else you want me to look at before putting together the v1.0.0 release?

marcustyphoon:
I do have a category of small diffs that fix regressions or potential pitfalls that I use the `bugfix` label for (that could also include #1486, I guess). Could take a glance at those!

PR #1484, opened 2024-06-05
Quick Reblog: Don't clear popup state on error

marcustyphoon:
### Description

Currently, Quick Reblog erases the popup state (actually sets `lastPostID = null`, the flag to erase the popup state when the popup next appears, technically) unconditionally during the reblog process. This means that API errors not only cause a failure of the user action, but result in them losing their entered tags.

This moves the clear action to after the await, so that it only occurs on success. Users can thus try again, and/or copy their tags elsewhere in order to do the reblog through the normal post form without losing their work.

I had to go git blame the conditional here because I couldn't figure out exactly what it was for.

### Testing steps

- Force an API fetch failure (such as by unreverting the test commit).
- Write tags into the quick reblog popup form.
- Attempt the reblog. Dismiss the failure modal.
- Hover over the reblog button on the same post. Confirm that the entered tags are still in the popup.
- Hover over the reblog button on the a different post. Confirm that the popup is reset.
- Perform a successful reblog and confirm that Quick Reblog works as normal.

(I didn't actually bother to test the delay-the-reblogPost-function-long-enough-to-move-to-another-post-while-it's-happening thing.)


marcustyphoon:
Alternate documentation:

```js
      const popupNotMoved = lastPostID === postID;
      if (popupNotMoved) {
        popupElement.remove();
        lastPostID = null;
      }
  ```

But, I dunno, is that clear enough?

AprilSylph (approved):
I like the current version. Big fan of code blocks that only have comments. It's readable code and documentation at the same time, it's perfect.

PR #1492, opened 2024-06-11
Implement Scroll to Bottom on Timeline V2 (modal branch)

### LINKED ISSUE #1491: Scroll to Bottom breaks on some timelines; blog lists; activity
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 125.0.6422.77

### Addon version

dev branch

### Details

https://github.com/AprilSylph/XKit-Rewritten/blob/a92be15496450118418166676fb3c978868cb20f/src/scripts/scroll_to_bottom.js#L10-L13

Scroll to Bottom stops after the first jump on all V2 timelines. I believe it is due to the `:last-child` here not matching the div containing the loader element any more, as the last child of the V2 timeline is an empty div (a react-intersection-observer inView, if the google results for `_unobserveCb` are to be believed).

marcustyphoon:
likely to be rebased.

also I can't PR a branch against an external PR branch; that's sad

### Description

This implements Scroll to Bottom on V2 timelines.

As noted in the linked issue, on V2 timelines, the loader element we need to detect so that we keep scrolling conditionally:

- Isn't the last child of its parent anymore (and doesn't have a classname iirc?)
- Doesn't exist unless the timeline is actively loading content

Thus, this:

- Fixes the selector
- Implements new logic: scrolling will stop after a short delay unless a loader is added during the delay window. (This is basically equivalent to "debounce the stopscrolling function" in some meaningful sense, though I believe literally doing that has some bad edge cases. Didn't think this through much; if that works we should do it so the code is easier to understand.)

Resolves #1491


AprilSylph:
Not gonna review this one, sorry.

Issue #1441, opened 2024-04-20
Incorrect blog detection in Tumblr Communities


marcustyphoon:
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 123.0.6312.87

### Addon version

v0.23.4

### Details

The Tumblr Communities feature, which is currently in a closed beta, contains posts that work differently than other posts in the Tumblr ecosystem, and also include changes to the way the API data underlying posts are used. This currently causes/probably will cause some XKit Rewritten functionality not to work properly within communities.

Particularly, community posts are displayed (generally speaking) as if they belong to the blog of the user who composed them, but the API data is more similar to a submission from that user to a generic blog.

In the case of Themed Posts, this just results in erroneous behavior: we want to match the theme we render a post with the way it otherwise appears, so we'll need a method to get a post's "blog" that isn't actually just the `blog` property of its API data. A number of other scripts look like they'll fall into this category (I can't tell right now; I'm not in the beta).

In some other cases, it would make sense for scripts to treat the submitting blog of a submission equally with the owning blog. For example, Quick Reblog automatically suggests a tag for the blog name; it may as well suggest the submitting blog's name too. In these cases there doesn't need to be special behavior for community posts.

marcustyphoon:
Needs tweaking:
- Themed Posts
- "Restore attribution" tweak

Maybe:
- Mutual checker works with no changes on Communities pages, but it's firing API requests unnecessarily rather than parsing `authorBlog` when `blog` doesn't match. The code change to improve this is somewhat awkward, though.
- I can't test Quick Tags on Communities.
- I can't test NotificationBlock on Communities (blockPostFilter may be wrong, and the apiFetch to check if the post is muted probably crashes if the button actually is clickable? most likely fine though)

PR #1496, opened 2024-06-12
Themed Posts: Function in Communities

### LINKED ISSUE #1494: Restore Attribution tweak breaks Communities blog links
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 125.0.6422.77

### Addon version

master

### Details

As per: https://www.tumblr.com/communities/communities-feedback/post/752845459562856448/bug-report-clicking-on-a-posters-username-in-a

The "restore attribution links" tweak causes a broken author link in Tumblr Communities.

marcustyphoon:
### Description

Related: #1441

As noted in the first linked issue, posts can have not only the single blog that "owns" them, but can also have a secondary blog associated with them: either the submitting blog if the post is a submission, the author blog if the post is on a group blog with the "show author portraits" setting enabled, or the author blog if the post is part of Tumblr Communities.

Notably, in the third case, the Tumblr UI reverses how these two fields affect how a post is rendered (the community owns the post, but the post looks like the "authorBlog" owns it).

Themed Posts' rendering should be consistent with how a post looks, so this overrides the field it queries to determine post color only in the case of Communities posts.

(I had changes to other scripts in this PR but none of them wound up being entirely correct.)

### Testing steps

- Confirm that Themed Posts functions normally outside of Communities.
- Confirm that Themed Posts colors posts in Communities with the author's blog color.


marcustyphoon:
Notes:
- Mutual checker works with no changes on Communities pages, but it's firing API requests unnecessarily rather than parsing `authorBlog` when `blog` doesn't match. The code change to improve this is somewhat awkward, though.
- I can't test Quick Tags on Communities.
- I can't test NotificationBlock on Communities (blockPostFilter may be wrong, and the apiFetch to check if the post is muted probably crashes if the button actually is clickable? most likely fine though)

src/scripts/themed_posts.js
```diff
@@ -30,7 +30,8 @@ const processPosts = async function (postElements) {
   filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
     if (postElement.matches(blogViewSelector) && !enableOnPeepr) return;
 
-    const { blog, trail = [], content } = await timelineObject(postElement);
+    const { blog: owningBlog, authorBlog, trail = [], content, community } = await timelineObject(postElement);
+    const blog = community ? authorBlog : owningBlog;
```
AprilSylph, line 34: not much of a fan of renaming `blog` in the destructure here. how about...

```suggestion
    const { blog, authorBlog, trail = [], content, community } = await timelineObject(postElement);
    const visibleBlog = community ? authorBlog : blog;
```

src/scripts/themed_posts.js
```diff
@@ -30,7 +30,8 @@ const processPosts = async function (postElements) {
...
 
     const blogData = [
       blog,
```
AprilSylph, line 37:
```suggestion
      visibleBlog,
```

AprilSylph (approved):

PR #1507, opened 2024-06-18
upgrade directory structure

AprilSylph:
### Description

- See: https://github.com/AprilSylph/XKit-Rewritten/pull/1466#pullrequestreview-2125997125

Renames `src/scripts` to `src/features` to match the terminology we use on the Wiki; renames `src/util` to `src/utils` because I think having a non-plural name for this directory is silly; and of course renames `src/browser_action` to `src/action` for the impending Manifest V3 upgrade.

All these renames were done with `git mv`, which... is better than a regular `mv` somehow I think...?

### Testing steps

- [ ] Ensure that the control panel still renders all features and feature preferences correctly
- [ ] Ensure that enabled features (particularly ones that have stylesheets) still activate on Tumblr


marcustyphoon (approved):
Seems to work fine, and presuming that we wind up with the same diff in the end as the MV3 branch had when it included these commits, we've already tested the result.

If this were being done completely separately from the MV3 commit, it would make sense to include the docs changes here, but this is only a merge optimization anyway so it makes no difference.

Issue #1508, opened 2024-06-21
Mutations util: Feature errors can break other features


marcustyphoon:
### Platform

MacOS 14.4.1 arm64

### Browser

Chrome 125.0.6422.77

### Addon version

v0.23.6 (+ dev commits)

### Details

https://github.com/AprilSylph/XKit-Rewritten/blob/a7bad978ff9978eecb8dc437dae7ef9af3a5571f/src/util/mutations.js#L77-L82

If a callback registered with `pageModifications` throws, and it is not an async function, onBeforeRepaint will throw and the entire set of callbacks remaining in the queue will not be executed. This means that a bug in a feature could become visible to the user as an entirely different feature not functioning, resulting in their bug report being very difficult to find the root cause of.

This, as they say in the business, Bad.

Only vaguely related: #1001 (in a vague sense the opposite of that PR in a number of ways, actually)

marcustyphoon:
This should probably be done for other parts of the extension too.

PR #1512, opened 2024-06-21
Mutations util: Isolate callback errors

marcustyphoon:
### Description

Wraps callbacks passed to the mutations util in try-catch so they don't impact each other when they throw.

Resolves https://github.com/AprilSylph/XKit-Rewritten/issues/1508.

Function could, essentially equivalently, be `const isolateErrors = async func => func()`, basically as a more-self-documenting #1509—if anything I think in some browsers and circumstances that actually makes the error log _better_, compared to `console.error`, if I recall correctly.

### Testing steps

You know... I'm not sure of a great way to reliably trigger the scenario to test this? I can maybe hack something together to ensure that modifiers get called in a certain order to demonstrate the problem. The place I ran into it was `createCaughtUpLine` throwing with a null input, per #1488, and breaking the boring tag chiclets tweak.


src/utils/mutations.js
```diff
@@ -7,6 +7,14 @@ const addedNodesPool = [];
 let repaintQueued = false;
 let timerId;
 
+const isolateErrors = func => {
+  try {
+    func();
```
AprilSylph: i will allow myself to bikeshed because i am approving the PR

**nit**: prefer unabbreviated names where possible; prefer names that don't need to be abbreviated to avoid confusion


```suggestion
const isolateErrors = callback => {
  try {
    callback();
```

AprilSylph (approved):

PR #1514, opened 2024-06-22
Main world injection: Execute functions without delay; refactors

marcustyphoon:
### Description

This adjusts the method used to inject code into the unsandboxed "main world." The goal is to begin the execution of an injected function within the same event loop invocation that it is called in when possible; this both helps avoid UI jank and helps reduce the need to handle race conditions resulting from other page mutations occurring between the call and execution (crashes due to processing detached elements).

Also included are some additional refactors of #1466.

There are a number of methods to do "instant" main world script injection; this one uses a cached dynamic import  to load script body files, bubbled custom events to request their execution and communicate their result, and the event target of said bubbled event and the `this` property on the script function to communicate the target element when needed.

Yes, I actually found a use for `Function.prototype.bind`/`Function.prototype.apply`; only took about a million years. No, `page_context.js` probably isn't in the right place in the repo.

Previously: #1467; preceding work is linked in each PR.

Resolves #1513 (you can't do that anymore, can you)


.eslintrc.json
```diff
@@ -15,7 +15,8 @@
     "sourceType": "module"
   },
   "rules": {
-    "import/no-cycle": "error"
+    "import/no-cycle": "error",
+    "jsdoc/require-jsdoc": "off"
```
AprilSylph, line 19: what's up with this addition?

marcustyphoon: Right now we don't have jsdoc for injected function bodies, and with this PR they can't be arrow functions (because we're making use of `this`). We could potentially document them, though it won't show up in intellisense because `inject` of course isn't a real import.

AprilSylph: don't we have tons of functions missing JSdoc already, though?

marcustyphoon: I haven't looked into it extensively, but the `jsdoc/require-jsdoc` defaults are a bit odd and they don't treat function declarations, function expressions, and arrow functions the same way.

Is required:

```js
export function controlTagsInput ({ add, remove }) { /* ... */ };
```

```js
export default function controlTagsInput ({ add, remove }) { /* ... */ };
```

Not required:

```js
export const controlTagsInput = function ({ add, remove }) { /* ... */ };
```

```js
const controlTagsInput = function ({ add, remove }) { /* ... */ };
export default controlTagsInput;
```

```js
export const controlTagsInput = ({ add, remove }) => { /* ... */ };
```

```js
const controlTagsInput = ({ add, remove }) => { /* ... */ };
export default controlTagsInput;
```

AprilSylph: ah, fun. okay.

well, can we disable per-file instead, then?
or (slightly less messy) add an override just for `src/utils/inject/**`?

marcustyphoon: Maybe `get_notification_props` should be in `src/features/quote_replies/inject`, hm.

marcustyphoon: Maybe a better way to do this, actually, is to put an eslint config file in each relevant directory.

src/page_context.js
AprilSylph, line 1: what's the official term that the Chrome scripting API uses? would be nice to match that re: this filename

AprilSylph: ah, right, "main world".

marcustyphoon: Oh, yeah, that makes sense. "This is the content script that gets run in the main world." Unfortunately the content script that's run in the isolated world is called `main.js`, so that's... a thing.

I considered putting the script in question in `src/content_scripts`, too—kind of feel like that makes the most sense? It's sort of interesting because I could see a world (sigh) in which someday they're both literally just referenced in manifest.json and one of them just has `"world": "main"`—definitely should be side-by-side in that case.

AprilSylph: anything that gets run in the main world is by definition not a content script, though - they don't have access to the content script APIs.

what do you think about a totally new directory, `src/main_world`, to house this script (i.e. `src/main_world/index.js`) and all the scripts that it imports? i don't really like the `src/utils/inject/**` or `src/features/{featureName}/{injectable}.js` patterns, seems better to keep them all in one place, but i was kinda rushed at the time.

marcustyphoon: > anything that gets run in the main world is by definition not a content script, though

Sounds like there will have to be some term redefinition at some point :D https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/registerContentScripts can target either world. But, yeah.

> what do you think about a totally new directory

Sure, seems reasonable. I can't think of a strong opinion either way—one separates the calling function from the called function (confusing) but groups the things that need to follow special rules (reasonable), the other does the opposite.

marcustyphoon: https://github.com/marcustyphoon/XKit-Rewritten/compare/mv3-instant-bind...marcustyphoon:XKit-Rewritten:mv3-move-main-world Ah, right. I cant PR my own PR. Like this?

(Doing it this way with no folder hierarchy, one could additionally change it to `inject('unbury_timeline_object', [], postElement)` or whatever, but I think it's probably clearer the way it is, so a new developer could unambiguously find the executed code.)

AprilSylph: yep, beautiful.

marcustyphoon:
Smoke tested in Safari again, and by "smoke tested" I mean "I spent \*checks time tracker* half an hour freaking out over Safari not loading anything from the new directory until I realized just rebuilding in xcode is insufficient and you apparently need to run `xcrun safari-web-extension-converter [sourcedir] --rebuild-project [destination].xcodeproj` or something I guess. anyway show originals seems to work."

AprilSylph (approved):

AprilSylph:
Will also need a docs update for `src/main_world/`. Let's do that separately though...

AprilSylph:
Will get to TimelineV2 reviews after supper (~2hrs' time?)

marcustyphoon:
> Will also need a docs update for `src/main_world/`. Let's do that separately though...

Oh, right!

marcustyphoon:
I feel like I should be bikeshedding the merge commit title more but that's kind of a stupid thought huh.

marcustyphoon:
...Huh. I guess you actually can close a PR with another PR. Okay then.

PR #1516, opened 2024-06-22
PostBlock: Block posts instantly

### LINKED ISSUE #894: PostBlock: Block posts instantly (minor)

marcustyphoon:
### Description

This applies similar adjustments to Postblock's storage code as implemented for Seen Posts in #523, making the asynchronous storage query only occur on enable and when blocking a post rather than delaying post processing. This means blocked posts won't be briefly visible when navigating to another timeline within Tumblr.

Resolves #894.

Closes #898; the method used is different, but these are pretty much equivalent. (I did this method mostly because it was fewer changed lines of code.)

~~Weirdly, I currently don't see the appearing-then-disappearing effect this is supposed to solve; still investigating why, but it seems as though it's related to the debounce on part of the virtual scroller rendering code.~~


src/features/postblock.js
```diff
@@ -10,9 +10,9 @@ const meatballButtonLabel = 'Block this post';
 const hiddenAttribute = 'data-postblock-hidden';
 const storageKey = 'postblock.blockedPostRootIDs';
 
-const processPosts = async function (postElements) {
-  const { [storageKey]: blockedPostRootIDs = [] } = await browser.storage.local.get(storageKey);
+let blockedPostRootIDs = [];
 
+const processPosts = async function (postElements) {
```
AprilSylph, line 15: looks like this no longer needs to be an async function?

looks like `onButtonClicked` doesn't need to be async either...

marcustyphoon: me: "aha I will just add `"require-await": "warn"`
eslint: `✖ 163 problems (0 errors, 163 warnings)`
me: "...never mind"

marcustyphoon:
> Weirdly, I currently don't see the appearing-then-disappearing effect this is supposed to solve; still investigating why, but it seems as though it's related to the debounce on part of the virtual scroller rendering code.

Ah, now I am seeing it, at least on Patio. Interesting.

Aside: by "it" I of course mean FOUP ("flash of unblocked post"). It's like soup, but worse in every way, and also it's not in any way like soup.

AprilSylph (approved):

PR #1518, opened 2024-06-22
Inject util: Stop processing disconnected elements

marcustyphoon:
### Description

As noted in https://github.com/AprilSylph/XKit-Rewritten/pull/1466#discussion_r1616413489 and previously addressed by #1501, it is possible for exceptions to be logged to the javascript console when we call certain injected functions on a target element that is no longer in the page by the time the injected function is executed.

As of #1514, it is usually fairly difficult for this to happen.

Nevertheless, it should be possible; the first time an injected function is executed it has to be imported asynchronously, for example.

This therefore uses a similar "just don't trigger the listener and resolve the promise" method to prevent any processing code from being run in this now-extremely-rare case. Fortunately, the new structure makes this a oneliner.

### Testing steps

- Confirm basic functionality.

If one wants to observe the behavior this fixes:
- Load the extension without this PR. Enable at least one feature that calls `timeline()`
- Open the javascript console on Tumblr Patio (with at least one column that the feature processes).
- Click the Tumblr Patio button, triggering a refresh (redpop does not use soft navigate in this instance, so this will perform a full page load and full extension initialization).
- During the resulting post load, click "home" to navigate to the dashboard. This will—not always, but fairly reliably with the right click timing—log a bunch of errors to the console as `unbury_timeline_object` throws on every Patio post for every feature that uses it.
- Perform this same sequence of events with this PR and confirm there are no longer exceptions logged.

### DIFF
--- a/src/main_world/index.js
+++ b/src/main_world/index.js
@@ -8,7 +8,7 @@ document.documentElement.addEventListener('xkitinjectionrequest', async event =>
     moduleCache[path] ??= await import(path);
     const func = moduleCache[path].default;
 
-    if (target instanceof Node && target.isConnected === false) return;
+    if (target.isConnected === false) return;
 
     const result = await func.apply(target, args);
     target.dispatchEvent(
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Oh, wait... I guess the target of an event has to be a Node, doesn't it. I forgot it defaulted to `document.documentElement`.

AprilSylph (approved):

PR #1520, opened 2024-06-23
Main world injection: Prevent race condition

marcustyphoon:
### Description

#1514 relies on the fact that fetching a json file and performing an extension storage query will take longer than the browser fetching the main_world script file, as we need the latter to be performed before importing modules with side effects or executing code that uses injection.

This performs a back-and-forth handshake to confirm that the main_world script file has been executed before proceeding with main install.

Random aside: to improve extension load speed slightly, one can do:

```css
    const [
      installedScripts,
      { enabledScripts = [] }
    ] = await Promise.all([
      getInstalledScripts(),
      browser.storage.local.get('enabledScripts'),
      mainWorldReady
    ]);
 ```

but I can't figure out how to make the code not look stupid.

### Testing steps

- Confirm basic extension functionality.
- If desired, unrevert the test commit and confirm that basic functionality appears with a significant delay on page load.

I have no tests for the bad case this prevents, as I can't for the life of me get it to happen.

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -76,24 +76,27 @@
     return installedScripts;
   };
 
-  const initMainWorld = () => {
+  const initMainWorld = () => new Promise(resolve => {
+    document.documentElement.addEventListener('xkitinjectionready', resolve, { once: true });
+
     const { nonce } = [...document.scripts].find(script => script.getAttributeNames().includes('nonce'));
     const script = document.createElement('script');
     script.type = 'module';
     script.nonce = nonce;
     script.src = browser.runtime.getURL('/main_world/index.js');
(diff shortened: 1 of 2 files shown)


AprilSylph:
> but I can't figure out how to make the code not look stupid.

I don't think it looks _that_ stupid, and it serves a clear purpose of "do these things in parallel instead of sequentially".
I've also seen the exact same pattern in the Redpop codebase.

marcustyphoon:
Yeah, I mean, it's definitely the way that you do that. I just wish the inputs and outputs lined up... and that there wasn't a mismatch in the inputs and outputs in this case.

AprilSylph (approved):

PR #1525, opened 2024-06-24
Timeline ID util: Process current Tumblr Patio IDs

marcustyphoon:
### Description

Adds the current Tumblr Patio timeline IDs to the timeline ID utility, so that we don't need to add them later if they are released in their current state.

The uuid v4 regex is fairly specific, so I can't imagine this will have any side effects.


AprilSylph (approved):
Don't have Patio on either of my non-work accounts, but I remember enough from working on the Patio code directly that these seem correct to me :D

PR #1528, opened 2024-06-27
quick tags search mvp

marcustyphoon:
### Description

Search box for Quick Tags.

I had the idea that there could be a floating/hovering search icon in, say, the bottom right of the area with the quick tags buttons, and clicking on it would hide it and replace it with a search box. Haven't bothered implementing that yet.

Aside The Quick Tags code could use some refactoring IMHO.


marcustyphoon:
Cleanup: Moving incomplete PRs without linked issues to my fork.

PR #1534, opened 2024-07-02
Main world injection: Don't stringify results/errors

marcustyphoon:
### Description

As noted in https://github.com/AprilSylph/XKit-Rewritten/pull/1466#discussion_r1616412097, Firefox—for whatever reason—gives a useless "uncaught exception: Object" message if you throw an object in an async function, and we currently throw objects out of injected functions due to the need to serialize them to JSON.

Fun fact, however: **we don't need to serialize them to JSON**. `CustomEvent`s support sending any `StructuredClone`-able value in Chromium, and Firefox... \*mumble mumble xray vision mumble\*.

This PR therefore removes the serialization step entirely.

Oddly, I'm not observing the `cloneInto()` call I needed to make for Firefox to allow this in [`1052f23` (#1284)](https://github.com/AprilSylph/XKit-Rewritten/pull/1284/commits/1052f23b2a5d583e7f54240d025d12274ac9afbb) being necessary, when testing this current code in Firefox 126. I don't know whether this is due to MV3, due to a change in Firefox, or both. Therefore, **big caveat:** merging this PR without would require us to confirm that `cloneInto()` is not required in any Firefox version we support.

Not well-tested at time of writing. I would like to see that this works on `Response` objects, ideally.

### DIFF
--- a/src/main_world/index.js
+++ b/src/main_world/index.js
@@ -10,21 +10,11 @@ document.documentElement.addEventListener('xkitinjectionrequest', async event =>
 
     const result = await func.apply(target, args);
     target.dispatchEvent(
-      new CustomEvent('xkitinjectionresponse', { detail: JSON.stringify({ id, result }) })
+      new CustomEvent('xkitinjectionresponse', { detail: { id, result } })
     );
   } catch (exception) {
     target.dispatchEvent(
-      new CustomEvent('xkitinjectionresponse', {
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Cleanup: Moving this PR to my fork as I don't trust it enough. I really would like to solve this issue (#1954) though.

PR #1535, opened 2024-07-06
Refactor: Use Array.prototype.at

marcustyphoon:
### Description

We may now use Array.prototype.at due to the browser version requirement bumps in #1466.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

### Testing steps

I confirmed that both avatars and sidebar item insertion in the mobile layout drawer still work (Chrome 126).

### DIFF
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -75,7 +75,7 @@ ${postSelector} footer button[aria-label="${translate('Reblog')}"]:not([role])
 const renderBlogAvatar = async () => {
   const { value: selectedUuid } = blogSelector;
   const { avatar } = userBlogs.find(({ uuid }) => uuid === selectedUuid);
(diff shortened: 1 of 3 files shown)


AprilSylph (approved):
nice

(aside: I'm somehow surprised that we weren't able to use this before; the presence of `Array.prototype.at` is how Redpop gates whether to use its "modern" or "fallback" build. I wouldn't have guessed that just a few weeks ago, there could have been people using the fallback version of Redpop but the latest version of XKit Rewritten. I thought we were more bleeding-edge than that. Good to know that now, we are!)

marcustyphoon:
Indeed, and I have already seen someone whose OS can't run the extension on either Chrome or Firefox :D Such is life without transpilation. And... also without bothering to update MacOS.

PR #1536, opened 2024-07-06
Quick Tags, Quick Reblog: Fix popup horizontal cutoff

### LINKED ISSUE #1369: Quick Reblog/Quick Tags: Popups can be hidden on Tumblr Patio
### Platform

MacOS 14.2.1 arm64

### Browser

Chrome 120.0.6099.216

### Addon version

0.23.3

### Details

Quick Reblog and Quick Tags can have their popups not fully be visible in the "Tumblr Patio" feature preview.

Using the alternate position for Quick Reblog's popup that's currently used in narrow-ish-viewport blog view modal timelines could help here. Quick Tags is a bit weirder, since the Quick Tags button can have various horizontal positions (e.g. on the queue/drafts pages), but it also tends to have less problems so maybe it can be ignored.

Alternatively, the popup could be attached higher in the DOM tree so that it would be "on top of" the edges of the post column, which would fix this entirely, sort of like how XKit 7 did it. (Not as high up as XKit 7 did do it, though, please.)

marcustyphoon:
### Description

Right now, it's possible for Quick Tags/Quick Reblog popups to extend past the left or right of their post elements (fundamentally fine) and be clipped off by the edge of the viewport or post column (not fine) in various circumstances. This fixes this by:
 - Removing the special "popup is on the left of the button" position entirely.
 - Adding logic that finds the closest ancestor element to the popup which would horizontally clip content, whatever it is, and shifts the popup's horizontal position to the left if necessary to keep the popup within that element's bounding box. _(not implemented: shifting it right if it's too far to the left)_

A simpler option, of course, would be to keep the popup horizontally within the post (`const preventOverflowTarget = target.closest(postSelector)`). When the viewport is wide enough not to need this I think it feels a bit more natural to leave it centered, but the difference is very small.

Another option, essentially the opposite, would be to target the viewport (adjusting position only on phone screens) and properly float the popups above e.g. the Patio column edges. I'm not sure if there's an elegant way to do this (`HTMLDialogElement`'s `showModal()`, maybe?)

Incidental changes:
 - Unifies the popup placement code between Quick Tags and Quick Reblog.
 - Adjusts the Quick Reblog element DOM position back to where it used to be; `closest('div')` apparently now matches a "root" `div class="v2JIl"` element in <990px viewports that I don't think existed when we wrote the 12px override thingy.

Resolves #1369. Fixes mobile-device issue reported in the New XKit Discord that hasn't been made into an issue yet.

### Testing steps

todo: write full list of tests performed. (patio narrow column, peepr wide, peepr narrow, embedded blog view, tablet/mobile layout, narrow phone layout via browser zoom /// top/bottom positions)

### DIFF
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -130,13 +130,6 @@ export const appendWithoutOverflow = (element, target, defaultPosition = 'below'
   target.appendChild(element);
 
   const preventOverflowTarget = closestWithOverflow(target);
-
-  // debug
-  preventOverflowTarget.dataset.tempHighlight = '';
-  setTimeout(() => {
-    delete preventOverflowTarget.dataset.tempHighlight;
-  }, 1000);
-
   const preventOverflowTargetRect = preventOverflowTarget.getBoundingClientRect();
   const elementRect = element.getBoundingClientRect();
 
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -118,9 +118,13 @@ export const postType = ({ trail = [], content = [], layout = [] }) => {
 
 const getClosestWithOverflow = element => {
   const parent = element.parentElement;
-  if (!parent) return element;
-  if (getComputedStyle(parent).overflowX !== 'visible') return parent;
-  return getClosestWithOverflow(parent);
+  if (!parent) {
+    return element;
+  } else if (getComputedStyle(parent).overflowX !== 'visible') {
+    return parent;
+  } else {
+    return getClosestWithOverflow(parent);
+  }
 };
 
 export const appendWithoutOverflow = (element, target, defaultPosition = 'below') => {
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -115,3 +115,35 @@ export const postType = ({ trail = [], content = [], layout = [] }) => {
   else if (content.some(({ type }) => type === 'link')) return 'link';
   else return 'text';
 };
+
+const closestWithOverflow = element => {
+  const parent = element.parentElement;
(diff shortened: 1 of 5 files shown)


marcustyphoon:
- Should these positioning styles become classname and data attribute based, and live in one of our static stylesheets? Probably. I ran into a whole mess in #1528 trying to refactor the popups because the current logic overwrites the classname property and thus doesn't let you use classes for anything else

src/utils/interface.js
```diff
@@ -115,3 +115,28 @@ export const postType = ({ trail = [], content = [], layout = [] }) => {
   else if (content.some(({ type }) => type === 'link')) return 'link';
   else return 'text';
 };
+
+const closestWithOverflow = element => {
+  const parent = element.parentElement;
+  if (!parent) return element;
+  if (getComputedStyle(parent).overflowX !== 'visible') return parent;
+  return closestWithOverflow(parent);
+};
```
AprilSylph, line 124: two nitpicks:
1. I think this function should be called `getClosestWithOverflow`; just `closestWithOverflow` makes me disappointed that this isn't a method on `Element.prototype` (which we should _not_ pollute, by the way)
2. The two single-line `if` statements are cute, but since this is basically an `if`/`else if`/`else`, I think it might be more readable if it actually was one

AprilSylph (approved):
Neat. Tested on 320px-wide viewport, edge of screen is avoided; tested on narrow (1000px-wide) desktop Peepr, edge of container is avoided! :tada: I'll just trust your screenshot that the same logic works on Patio too...

PR #1538, opened 2024-07-10
specify static main world script in manifest.json

marcustyphoon:
### Description

Yep: this works.

Rather than creating and appending a script element to run the main world script in the correct world, this specifies it as a static content script with the `"world": "MAIN"` property.

My assumption is that content scripts get injected in order and are executed synchronously...? So there, uh, **should** be no race conditions with this? Probably?

Requires Firefox 128, which was released today, so we obviously don't want to do this now.

**edit:** Requiring Firefox 128 is no longer impactful (it's the earliest ESR we support anyway), but requiring Chrome 111 drops our Windows 7 support. If doing so, we may as well go to Chrome 116, as it has the same operating system support.

### Testing steps

todo


marcustyphoon:
- [ ] make main_world script classic

marcustyphoon:
I wonder if this is incompatible with #1851 or if Firefox magically handles this case correctly.

Be sure to test Firefox extension update/reload heavily if we do this.

AprilSylph:
Aw, I liked the look of this one.

marcustyphoon:
I mean, we can definitely do it if we want to standardize on "we don't support windows 7" eventually.

PR #1542, opened 2024-07-16
Quote Replies: Option to reply as a reblog

### LINKED ISSUE #1414: Quote Replies: Option to reply as a reblog
### Concept

Putting this here so I don't forget to look into it at some point.

Sometimes you want to quote reply in a reblog of the post replied to rather than as a new post. This currently isn't supported by quote replies. It's _probably_ possible? If there's a long reblog thread that gets a reply there's some ambiguity around where a reply-as-reblog should start from in the chain.

Sure, one could argue that you can just reply in the replies in this case, but that's true of every use of quote replies too.

marcustyphoon:
### Description

Working proof of concept.

This implements the ability for Quote Replies to—rather than creating a new draft post with the quoted reply—create a draft reblog of the target post with the quoted reply.

UI-wise, this currently adds a second button next to the quote replies button on notifications, this isn't great but I didn't love the alternatives I thought of either: a) open a modal with both options every time, which is clunky, or b) always do one or the other based on a user preference, which isn't how I would expect anyone to actually want to use this. Also, no idea what icon to use (two quote icons with subscript plus/reblog icons would be kind of neat maybe). 

Currently not sure what happens when someone clicks on a big reblog chain and replies to you under it. Does this target the reblog or the root post? I had nothing to test this on and haven't gotten around to making a test case.

Resolves #1414.

### DIFF
--- a/src/features/quote_replies.js
+++ b/src/features/quote_replies.js
@@ -1,7 +1,7 @@
 import { keyToCss } from '../utils/css_map.js';
 import { dom } from '../utils/dom.js';
 import { inject } from '../utils/inject.js';
-import { showErrorModal } from '../utils/modals.js';
+import { hideModal, modalCancelButton, showErrorModal, showModal } from '../utils/modals.js';
 import { notificationSelector } from '../utils/interface.js';
 import { pageModifications } from '../utils/mutations.js';
 import { notify } from '../utils/notifications.js';
@@ -23,6 +23,7 @@ const dropdownSelector = '[role="tabpanel"] *';
 let originalPostTag;
 let tagReplyingBlog;
 let newTab;
+let enableReblogReplies;
 
 const processNotifications = notifications => notifications.forEach(async notification => {
   const { notification: notificationProps, tumblelogName } = await inject(
@@ -36,7 +37,7 @@ const processNotifications = notifications => notifications.forEach(async notifi
   const activityElement = notification.querySelector(activitySelector);
   if (!activityElement) return;
 
-  activityElement.after(dom(
+  const button = dom(
     'button',
     {
       class: `${buttonClass} ${notification.matches(dropdownSelector) ? dropdownButtonClass : ''}`,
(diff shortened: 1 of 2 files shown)


AprilSylph:
Personally I think a default-on preference would be good. It makes more sense to keep the full context by default—discarding it should be considered a fringe use-case.

marcustyphoon:
> Personally I think a default-on preference would be good. It makes more sense to keep the full context by default—discarding it should be considered a fringe use-case.

Hmm! Interesting—I'll have to think about whether I agree with that. I could see it.

marcustyphoon:
> Currently not sure what happens when someone clicks on a big reblog chain and replies to you under it. Does this target the reblog or the root post? I had nothing to test this on and haven't gotten around to making a test case.

As expected, the root post is reblogged (replies aren't tied to the place on the reblog chain where the replying user was writing the reply). Makes the most sense as the default, of course... but in many cases one would want their reblog to include a bunch of the reblog chain.

How would one implement this? Presumably by adding a Quote Replies button to the reply itself, when viewed below a reblog chain. But maybe that's a different PR.

AprilSylph:
If someone replies to your reblog of a post, the target post is your reblog for you. If someone replies to a reblog of your root post, the target post is the root post for you. Ergo, reblogging the target post with the reply will give as much context as is possible to obtain, which I think is fine.

---

Also, I haven't looked at the code yet, so maybe you already thought of this - but if the target post is deleted or otherwise cannot be reblogged, we should fall back to creating a new post.

AprilSylph:
Ugh. Nevermind, I thought all replies on reblogs got sent to the reblogger's activity feed as activity on that reblog, but no, it's only when you don't also own the root post. Ownership of the root post takes precedence. That's... really annoying.

marcustyphoon:
Hm. I hadn't thought to separately test being/not being the original post author.

That complexity and hoping that complexity can be reduced from Tumblr's side aside, I'm thinking more and more that the ability to use this on replies in the post notes is important.

> Also, I haven't looked at the code yet, so maybe you already thought of this - but if the target post is deleted or otherwise cannot be reblogged, we should fall back to creating a new post.

I did handle "cannot be reblogged" in that way; hadn't handled network errors or deleted posts elegantly yet.

marcustyphoon:
Cleanup: Closed in favor of #1863.

PR #1544, opened 2024-07-16
Exclude hidden posts from page search

### LINKED ISSUE #1543: Hidden posts behave poorly with control-F
### Platform

macOS 14.4.1 arm64

### Browser

Firefox 129.0b1, Chrome 126.0.6478.127

### Addon version

v1.0.1

### Details

When a post is hidden by any of our post hiding features, searching the page for a term that is in that post will result in hits if and only if the post in question is far off the screen, due to the way tumblr's  virtual scroller component works. When the browser attempts to scroll the post into view, the viewport will be left scrolled into a seemingly random location, and the search result in question will actually disappear, causing thrashing (because we do prevent control-F from finding the **rendered content**, but cannot do the same to the **placeholder searchable text**).

This would be easy to fix from the redpop side (simply wrap `cachedTextContent` in a div instead of directly setting the cell's content to the string, so we can hide it without hiding the cell—this would also make it easier to style in a non-silly way, I imagine), but is as far as I am aware impossible to fix from the xkit rewritten side.

I can submit this to the tumblr feedback form, though! Besides that it should probably be wontfix labelled unless someone knows of a way to cheat here (can you replace text contents with css?)

marcustyphoon:
### Description

When endless scrolling is enabled and a post is hidden by any of our post hiding features, searching the page for a term that is in that post will result in hits if and only if the post in question is far off the screen, because the searchable text rendered by tumblr's virtual scroller component is not affected by our hiding css. 

This adjusts the CSS that we use to create hidden posts so that post elements will have both their element children and text node children removed from the DOM, preventing them from being searched. It uses the css `content` property to replace post element contents (whatever they are) with a no-op "image", which is [supported as of Firefox 113](https://developer.mozilla.org/en-US/docs/Web/CSS/content#browser_compatibility). We cannot simply `display: none` the target post elements themselves, as we need them to continue to be rendered and have bounding boxes so that j/k scrolling code continues to find them.

Notably, this also does the same thing as #593, i.e. presumably making Tumblr's observer elements (correctly) think the user is not viewing the post in question. As mentioned in that PR, I don't know if there are actual consequences of this.

(Height is set to 0 because apparently in Chromium unset-height gradients are half as tall as they are wide. No idea why this is; a linear gradient should have no intrinsic height. This isn't necessary in Firefox. But sure, I guess, why not.)

Resolves #1543.

### Testing steps

- Confirm that each feature still hides posts correctly and shows posts again when it's toggled off, including when scrolled far down the dashboard, with endless scrolling enabled.
- Confirm that each feature still hides posts correctly and shows posts again when it's toggled off, including when scrolled far down the dashboard, with endless scrolling disabled.
- Copy some text from a post that will be hidden with Show Originals. Hide the post with Show Originals. Scroll far away from it and search the page for the post text. Confirm that there are no results.
- Enable Show Originals and Seen Posts. Scroll slowly down the dashboard to mark only original posts as seen. Navigate away from and back to the dashboard, dimming seen posts. Disable Show Originals and confirm that the non-original posts that are now visible have not been dimmed, because Seen Posts correctly determined that the user had not seen them.
- Confirm that hiding posts on a masonry view works (e.g. by using postblock on a masonry view search page).

Tested in Firefox 121.
Show Originals smoke tested in Chrome 126, desktop Safari 17.4.1, and iOS Safari 18.1.


AprilSylph (approved):
Neat.

You'll need to update this for #1753 (and maybe some others I've forgotten about) now though! :smiley:

marcustyphoon:
Indeed, I'll have to update some PRs—oh, we just. merged one of them. I see :D

marcustyphoon:
- [x] searched codebase for "article"
- [x] tested "Hide posts from non-joined communities in For You" feature

PR #1545, opened 2024-07-16
Seen Posts: Fix soft refresh

### LINKED ISSUE #1539: Seen Posts may not dim posts on soft refresh
### Platform

macOS 14.5

### Browser

Firefox 128.0

### Addon version

v1.0.0

### Details

As of a few days ago, clicking the 'Home' button on the Tumblr dashboard no longer dims posts that were just seen. Only a full refresh of the page correctly dims newly seen posts.

https://github.com/user-attachments/assets/08b5770c-cee9-4e6b-81ba-2d5003ca73ac

marcustyphoon:
### Description

Resolves #1539. Method:

- Detect, via a hopefully-very-specific CSS selector, the addition of an element that only gets added when the user manually soft refreshes a V2 timeline (by using pull to refresh, clicking the tumblr logo, home button, or top of a patio column—are there other ways to trigger this?)
- For each post in a refreshed timeline, if we've already seen it, dim it.

This should be identical in effect to the code in `dimPosts` that is specifically run when `isFirstRender === true`, since that's what we're trying to simulate in this special case. Perhaps we could refactor both to make this slightly more obvious:

```js
    // dimPosts
    if (seenPosts.includes(id) === false) {
      observer.observe(postElement.querySelector('article header + *'));
    }
    if (isFirstRender && seenPosts.includes(id) === true) {
      timelineItem.setAttribute(dimAttribute, '');
    }
    
    // onSoftRefresh
    if (seenPosts.includes(id) === true) {
      timelineItem.setAttribute(dimAttribute, '');
    }
```

(I also tried having `onSoftRefresh` just call `dimPosts`, which works and happens to have no side effects to my knowledge, but the reasoning to conclude that is complicated.)

Is there a better way to do this?

### Testing steps

Feature flag modification may be required to test this on Patio.

Note that pull-to-refresh is a valid way of soft refreshing; one can test that using the responsive design mode in a desktop browser's dev tools.

### DIFF
--- a/src/features/seen_posts.js
+++ b/src/features/seen_posts.js
@@ -1,8 +1,8 @@
-import { filterPostElements, getTimelineItemWrapper, postSelector } from '../utils/interface.js';
+import { filterPostElements, getPostElements, getTimelineItemWrapper, postSelector } from '../utils/interface.js';
 import { getPreferences } from '../utils/preferences.js';
-import { onNewPosts } from '../utils/mutations.js';
+import { onNewPosts, pageModifications } from '../utils/mutations.js';
 import { keyToCss } from '../utils/css_map.js';
-import { followingTimelineFilter, followingTimelineSelector } from '../utils/timeline_id.js';
+import { followingTimelineFilter, followingTimelineSelector, timelineSelector } from '../utils/timeline_id.js';
 
 const excludeAttribute = 'data-seen-posts-done';
 const timeline = followingTimelineFilter;
@@ -12,6 +12,8 @@ const dimAttribute = 'data-seen-posts-seen';
 const onlyDimAvatarsClass = 'xkit-seen-posts-only-dim-avatar';
 const hideClass = 'xkit-seen-posts-hide';
 const lengthenedClass = 'xkit-seen-posts-lengthened';
+
+const softRefreshLoaderSelector = `${followingTimelineSelector} ${keyToCss('container')}:has(+ ${keyToCss('scrollContainer')}) > ${keyToCss('knightRiderLoader')}`;
 
 const storageKey = 'seen_posts.seenPosts';
 let seenPosts = [];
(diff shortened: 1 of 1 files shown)


marcustyphoon:
- [ ] update selector to work when endless scrolling is off, in which case there is no `scrollContainer` and the loader is followed immediately by the classless would-be `cell` divs containing post elements
 - [ ] instead of doing something so fragile, figure out a better way to solve this problem
 - [ ] give up and go live in the woods

marcustyphoon:
Note: if you're on a "channel" page and you post a new post, it looks like the timeline soft refreshes. Seen posts doesn't work there, though, and this doesn't happen on the main dashboard timeline, so this should be fine.

src/features/seen_posts.js
```diff
@@ -70,6 +72,20 @@ const dimPosts = function (postElements) {
...
+  const refreshedPostElements = getPostElements({ timeline: element => refreshedTimelineElements.includes(element), includeFiltered });
+
+  for (const postElement of refreshedPostElements) {
+    const { id } = postElement.dataset;
+    const timelineItem = getTimelineItemWrapper(postElement);
+
+    if (seenPosts.includes(id)) {
+      timelineItem.setAttribute(dimAttribute, '');
+    }
+  }
```
AprilSylph, line 86: instead of duplicating any code from `dimPosts`, couldn't we just do

```js
$(refreshedPostElements).removeAttr(excludeAttribute);
dimPosts(refreshedPostElements);
```
?

marcustyphoon: Yep! Could also add a `reprocessPosts = false` second argument, also, and do `const isFirstRender =reprocessPosts || /* */;`.

This relies on `MutationObserver`'s `observe` method being idempotent, which... I assume it is; can't say I've tested that (or some timer logic I can't be bothered to reason about). And not adding anything problematic to call twice to `dimPosts`, of course, which seems like an unlikely thing for us to do.

Also lets one skip `getPostElements` and query the timeline elements for `postSelector`.

src/features/seen_posts.js
```diff
@@ -52,14 +54,14 @@ const lengthenTimelines = () =>
...
-const dimPosts = function (postElements) {
+const dimPosts = function (postElements, reprocessPosts = false) {
   lengthenTimelines();
 
   for (const postElement of filterPostElements(postElements, { timeline, includeFiltered })) {
     const { id } = postElement.dataset;
     const timelineItem = getTimelineItemWrapper(postElement);
 
-    const isFirstRender = timelineItem.getAttribute(excludeAttribute) === null;
+    const isFirstRender = reprocessPosts || timelineItem.getAttribute(excludeAttribute) === null;
```
AprilSylph, line 64: I think instead of modifying this constant's condition, I would do `else if (isFirstRender || reprocessPosts)` on L69. Otherwise we need a new name for this constant, and I can't think of one that's as clear as `isFirstRender` which describes this OR.

AprilSylph (approved):

PR #1549, opened 2024-07-17
chore: Remove webextension polyfill

marcustyphoon:
### Description

Included in Chrome's MV3 changes are, according to them, adding promise support to webextension APIs. So long as the major browser vendors do their standardization correctly, this means that https://github.com/mozilla/webextension-polyfill will serve only to fix the name mismatch between the `chrome` and `browser` namespaces. One could thus only do that.

Let's see if it works! Sure seems to.

### Testing steps

Confirm extension functionality in Chromium and Firefox.


marcustyphoon:
Notably, I'm not one hundred percent sure if
```js
globalThis.browser ??= globalThis.chrome;
```
is precisely equivalent to
```js
if (typeof browser === 'undefined') {
  globalThis.browser = chrome;
}
```
but I see no reason it wouldn't be and is a nicer oneliner if you need to use it in multiple places (https://github.com/AprilSylph/Outbox-for-Tumblr/pull/96). I was conservative and [copied MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background) here.

There are, of course, a million ways to write this. One can just as easily use `window.browser ??= window.chrome;`, for example; maybe that makes the most sense here with the justification that it'd the most recognizable to any web extension developer. `globalThis` is good to know about, though, since it works in web workers.

AprilSylph:
> Ugh, I guess putting this in the content scripts folder doesn't make sense, because it's referred to in action. But it probably shouldn't go in lib since it's not vendored, and it probably shouldn't go in util because it exists on a higher hierarchy level than everything in there. Uhh... I dunno, thoughts?

Straight into `src/`.

src/manifest.json
```diff
@@ -52,7 +52,6 @@
       ],
       "run_at": "document_end",
       "js": [
```
AprilSylph, line 54:
```suggestion
      "js": [
        "browser_polyfill.js",
```

AprilSylph (changes requested):
Broken in Chrome as of latest commit.

src/manifest.json
marcustyphoon: Whoops.

However: one second, so is... the master branch?

AprilSylph: `Uncaught (in promise) TypeError: Cannot convert undefined or null to object`

Ooops.

marcustyphoon: Hmm, yeah, in retrospect it probably should have occurred to me to test toggling a feature, not just toggling a feature preference :D

AprilSylph:
Would we be losing anything by just using `chrome` directly, instead of `browser`?

marcustyphoon:
The sense of not bowing to the chromium team's bullying, but other than that, I don't think so!

AprilSylph (approved):

marcustyphoon:
https://developer.chrome.com/docs/extensions/develop/concepts/browser-namespace#polyfill

Oh, hey, fun note: we do have to use the old syntax for async responses to `browser.runtime.onMessage` because of this. Now, we don't currently use `browser.runtime.onMessage`, because we have no background script for browser support reasons (combined with "not wanting to make the chrome and firefox manifest files different" reasons), so this is irrelevant. However, if we did do a browser support version bump in the future, and picked chrome 128 or 138 (latest supported in macos 10.15/11), we would be able to use background scripts without adjusting manifests, but would have to use the legacy syntax. That's kind of silly!

Issue #1550, opened 2024-07-18
Investigate and improve on-update behavior in Firefox (again)


marcustyphoon:
### Concept

With the release of v1.0.1 to Firefox, I'm seeing various reports that the features silently stop working until you refresh/open a new tab. Reports include the radar hiding tweak, hide my own posts on the dash tweak, possibly No Recommended sidebar hide...

Unfortunately, to my knowledge, all of these work fine when you simply toggle the extension off and on in Firefox; hopefully this isn't impossible to reproduce in development. If it is possible to reproduce in development, we can probably(?) fix it. For example, maybe the script element used to initialize the main world handlers needs to not have `type: module` or something, since module scripts don't get executed for side effects multiple times. Ah, damn, that's probably it, isn't it.

(Wait... I guess one can just reproduce this by installling the 1.0.0 xpi, opening Tumblr, then firing the update in a new tab. Nice. How one can test a fix, I'm not so sure.)

marcustyphoon:
Aside: Oh, hey, toggling the extension duplicates the Quote Replies action button. Only the new one works, since we attach the listener directly to the element, so it's not like a check that a button already exists would help in this case; we need to remove it.

Perhaps we should add `element.dataset.xkitRewritten = '';` to the `dom` util and just get rid of everything, not just `buildStyle` style elements, on boot. Clutters the dev tools, though.

PR #1551, opened 2024-07-18
Fix extension restarts in Firefox

marcustyphoon:
### Description

Ah, fuck, yeah, it was my fault. I essentially already knew this, too.

If the script element created in initMainWorld is a module script, disabling and re-enabling the extension adds another identical module script to the page. This won't execute anything, because modules are only executed once, and so the whole extension will break on that page.

This changes the main world script (back) to a classic script, fixing this.

See #1550; probably fixes the main issue there, but I haven't fully tested that yet.

### Testing steps

In Firefox, turn the extension off and then back on. Without this PR, observe that everything breaks. With this PR, smoke test the extension and confirm it works.

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -81,7 +81,6 @@
 
     const { nonce } = [...document.scripts].find(script => script.getAttributeNames().includes('nonce'));
     const script = document.createElement('script');
-    script.type = 'module';
     script.nonce = nonce;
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
I don't think I even need to manually test this one, given the simplicity of the diff (excluding whitespace).

Yay side-effects, yay dynamic `import` working outside of module scripts.

PR #1553, opened 2024-07-18
On-update behavior: Remove all extension DOM elements

marcustyphoon:
### Description

see #1550

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -18,6 +18,7 @@
         rel: 'stylesheet',
         href: browser.runtime.getURL(`/features/${name}.css`)
       });
+      link.dataset.xkitRewritten = '';
       document.documentElement.appendChild(link);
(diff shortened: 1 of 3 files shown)


marcustyphoon:
On reflection, places where this would be a problem unless we specifically declined to use `dom()` or added an argument to it or alternative util export whose result would not be removed on extension reboot:
- Subtle Activity Mutual tweak, where we empty the contents of the real text node
- AccessKit Disable GIFs, where we wrap tumblr elements
- Collapsed Queue, where we wrap tumblr elements
- Maybe Classic Search? I don't remember.

This would also do inconsistent things with, for example, Shorten Posts, which needs a class removal to undo part of its effect.

Going back to https://github.com/AprilSylph/XKit-Rewritten/issues/451#issuecomment-1056679543 as the best-so-far way to handle this.

PR #1559, opened 2024-08-08
New Tweak: Subtle reply original poster labels

marcustyphoon:
### Description

This creates equivalents for the subtle/no activity mutuals indicator tweaks for the "original poster" indicators in threaded replies.

I kind of feel like we should only do the first one?

### Testing steps

Be sure to test with timestamps both enabled and disabled; it changes the DOM structure a bit and the selector for these labels had to be kind of weird because `isOriginalPoster` is used in multiple places.


src/features/tweaks.json
```diff
@@ -63,6 +63,16 @@
...
+    "subtle_reply_original_poster": {
+      "type": "checkbox",
+      "label": "Make original poster indicators on replies more subtle",
+      "default": false
+    },
+    "hide_reply_original_poster": {
+      "type": "checkbox",
+      "label": "Hide original poster indicators on replies",
+      "default": false
+    },
```
AprilSylph, line 75: yeah, cut this one.

AprilSylph (commented):
Why does `transitionStyleElement` exist separately from `styleElement`, and why is there a delay in adding it to the DOM?

marcustyphoon:
This is a trick I also used in the subtle activity mutuals tweak. It just means that when you toggle the feature on with any elements it affects visible on the page, they jump to the collapsed position rather than animating.

marcustyphoon:
Tested and working with #1754.

PR #1560, opened 2024-08-08
Tweaks: Create empty reply button tweak

marcustyphoon:
### Description

Creates a tweak that hides the large reply button in post replies when it doesn't have a reply count. One can still reply to a reply from the meatballs menu, so the button serves no purpose in this case so long as users are aware of this.

This both saves vertical space and, in my opinion, dramatically increases readability of the replies section in some cases by adding more visual distinction between posts at the root of a thread and the thread itself; indentation isn't that much of a distinguishing factor.

Would, of course, need to be removed from the extension in a timely fashion if the reply option gets removed from the meatballs menu (which I would be annoyed about, but I could see happening, as without this tweak it doesn't really make sense). I would probably hack the button into a new location in that case (float? lmao)

### DIFF
--- a/src/features/tweaks.json
+++ b/src/features/tweaks.json
@@ -63,6 +63,11 @@
       "label": "Hide following/mutuals indicators on notifications",
       "default": false
     },
+    "hide_empty_reply_button": {
+      "type": "checkbox",
+      "label": "Hide the large reply button on replies without a thread",
+      "default": false
+    },
     "hide_footer_tooltips": {
       "type": "checkbox",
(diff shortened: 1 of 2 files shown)


AprilSylph:
hmm, it's a no from me on this one. i don't think the reply button is really that annoying, and yeah, i'd be scared about the option in the meatball menu getting removed; it _is_ technically redundant now, after all.

marcustyphoon:
Fair! I'll make it a user style.

PR #1564, opened 2024-08-10
Tweaks: Organize tweaks

marcustyphoon:
### Description

| before | after |
| - | - |
|  |  |

We have a bunch of tweaks that naturally go together due to being variations on each other, affecting similar parts of the UI, etc. It's not that easy to parse visually, and I think a lot of people miss some of the features due to this.

I was thinking about ways to make this a little more clear in a way without noticeably changing how the preference UI looks; this is (the most subtle version of) that. Preferences are grouped a little bit, and some of the names are tweaked to read more consistently within the groups. Thoughts?

(If we want to do something like this, this feels very much like one of those "this PR is yours now; poke it with a stick until you like the resulting UI" ones.)

### DIFF
--- a/docs/Chapter 4.1 - Feature manifests.md
+++ b/docs/Chapter 4.1 - Feature manifests.md
@@ -67,11 +67,11 @@ It is recommended to use camelCase for each preference name, so that the script 
 - Type: String
 - Required: Yes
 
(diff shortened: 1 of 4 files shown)


marcustyphoon:
Cleanup: Moving incomplete PRs without linked issues to my fork.

PR #1567, opened 2024-08-15
TimeFormat: Improve overflow on reply timestamps

### LINKED ISSUE #1562: TimeFormat: Improve overflow on reply timestamps
### Concept

Reply timestamps are now inline with usernames (and the "original poster" label), differing from how post headers/reblog headers/reblog notes are displayed. This can cause layout issues when things become very long:

I thought about making timeformat have a toggle to not apply to replies or an optional different format for them, or something. Perhaps just adjusting the whitespace css on replies is ideal. Perhaps timeformat should automatically check bounding box stuff and automatically not apply if there's not enough space. Lots of options.

The feature added in #1559 helps here.

marcustyphoon:
### Description

This restricts formatted timestamps that are inline with blog links to a single line, simply clipping any overflow. Resolves #1562.

Compatible with #1559:

image


src/features/timeformat.js
```diff
@@ -1,11 +1,42 @@
...
+
+[data-formatted-time][title]::before,
+[data-formatted-time][title]::after {
+  cursor: help;
+}
+
+${keyToCss('blogLinkWrapper')}:has(+ [data-formatted-time]) {
+  flex: none;
+}
+${keyToCss('blogLinkWrapper')} + [data-formatted-time] {
```
AprilSylph, line 34: the stylelint in my brain was activated here

```suggestion

${keyToCss('blogLinkWrapper')} + [data-formatted-time] {
```

AprilSylph (commented):
one little nit, otherwise :+1:

AprilSylph (approved):

PR #1574, opened 2024-08-17
Quick Reblog: Enable reblogging posts into communities

marcustyphoon:
### Description

Adds communities the user has joined to the Quick Reblog blog selector dropdown.

Note that posts drafted or queued into a community (presumably) disappear into the ether, so this puts up an error modal if the user tries to do so. (Ideally it should probably grey out the buttons in question; I was hoping one could css select based on a `<select>`'s current value but I'm not sure you can, so this would probably have to be a bunch of js.)

Should alreadyReblogged exclude reblogging into a community? ¯\\\_(ツ)\_/¯ 

### Testing steps

At time of writing, I have a spare community with invite link [tumblr.com/join/oki7cQsr](https://www.tumblr.com/join/oki7cQsr).

todo: list performed tests

### DIFF
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -49,6 +49,7 @@ let lastPostID;
 let timeoutID;
 let suggestableTags;
 let accountKey;
+let joinedCommunities;
 
 let popupPosition;
 let showBlogSelector;
@@ -66,17 +67,15 @@ const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
 const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
 const blogHashes = {};
+const avatarUrls = {};
 
 const reblogButtonSelector = `
 ${postSelector} footer a[href*="/reblog/"],
 ${postSelector} footer button[aria-label="${translate('Reblog')}"]:not([role])
 `;
 
-const renderBlogAvatar = async () => {
-  const { value: selectedUuid } = blogSelector;
-  const { avatar } = userBlogs.find(({ uuid }) => uuid === selectedUuid);
-  const { url } = avatar.at(-1);
-  blogAvatar.style.backgroundImage = `url(${url})`;
+const renderBlogAvatar = () => {
+  blogAvatar.style.backgroundImage = `url(${avatarUrls[blogSelector.value]})`;
 };
 blogSelector.addEventListener('change', renderBlogAvatar);
 
@@ -200,6 +199,10 @@ const reblogPost = async function ({ currentTarget }) {
   };
 
   try {
+    if (state !== 'published' && joinedCommunities.some(({ uuid }) => uuid === blog)) {
+      throw new Error('Posts cannot be queued/drafted to communities!');
+    }
+
     const { meta, response } = await apiFetch(requestPath, { method: 'POST', body: requestBody });
     if (meta.status === 201) {
       makeButtonReblogged({ buttonDiv: currentReblogButton, state });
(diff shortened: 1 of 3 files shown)


src/features/quick_reblog.js
```diff
@@ -331,12 +334,24 @@ export const main = async function () {
     alreadyRebloggedLimit
   } = await getPreferences('quick_reblog'));
 
+  joinedCommunities = await apiFetch('/v2/communities')
+    .then(({ response }) => response)
+    .catch(() => []);
```
AprilSylph, line 339: This should probably be exported by `src/utils/user.js`, no? It doesn't come from `/v2/user/info` but it _is_ basic user-specific data that we might need for other features in the future.

---

feature future
future feature
future feature future
feature future feature

marcustyphoon: Oh yeah, probably!

src/utils/user.js
```diff
@@ -39,3 +44,8 @@ export const adminBlogs = userInfo?.blogs?.filter(blog => blog.admin) ?? [];
  * {string[]} adminBlogNames - An array of blog names the current user is admin of
  */
 export const adminBlogNames = adminBlogs.map(blog => blog.name);
+
+/**
+ * {object?} userInfo - The contents of the /v2/user/info API endpoint
+ */
+export const joinedCommunities = fetchedCommunitiesInfo.response;
```
AprilSylph, line 51: :coppy:

marcustyphoon: er.

marcustyphoon: not sure how I feel about AI suggestions but I wouldn't mind a commit hook with a little language model that just piped up and said "hey this is probably wrong"

src/features/quick_reblog.js
```diff
@@ -67,17 +67,15 @@ const alreadyRebloggedStorageKey = 'quick_reblog.alreadyRebloggedList';
 const rememberedBlogStorageKey = 'quick_reblog.rememberedBlogs';
 const quickTagsStorageKey = 'quick_tags.preferences.tagBundles';
 const blogHashes = {};
+const avatarUrls = {};
```
AprilSylph, line 70: When we're not setting our own key-values as part of the definition, I think maps should be actual `Map`s. Empty `Record`s are spooky, unnecessary square bracket notation is spooky.

(Sin committed: criticise someone for following an established pattern in a file!)

Bonus points if you convert `blogHashes` to a `Map` as well. <3

marcustyphoon:
```diff
- const savedBlogUuid = Object.keys(blogHashes).find(uuid => blogHashes[uuid] === savedBlogHash);
+ const savedBlogUuid = [...blogHashes].find(([uuid, hash]) => hash === savedBlogHash)?.[0];
```

```diff
- const savedBlogUuid = Object.keys(blogHashes).find(uuid => blogHashes[uuid] === savedBlogHash);
+ const savedBlogUuid = [...blogHashes.keys()].find(uuid => blogHashes.get(uuid) === savedBlogHash);
```

hm

marcustyphoon: That's interesting! I'm curious if you had that stance before (I assume) doing a lot of typescript. That was kind of my experience.

AprilSylph (changes requested):
I think it's worth doing the JS to hide the Queue and Draft buttons in the popup entirely when a community is selected.

The good news is that the CSS isn't complicated because I'm awesome, it just needs to be something like `.actionButtons.communitySelected button:not([data-state="published"]) { display: none; }`. The bad news is that you'll need another `change` event listener on `blogSelector` that looks up whether the new value is from `joinedCommunities` or not to add or remove the necessary class from `actionButtons`.

...which I guess isn't too bad because you could probably make a `Set` of just the user's joined community UUIDs for easier lookup?
Maybe even define and export that in `utils/user.js`.
Maybe even make an equivalent for the user's blogs too, I'm sure we could make use of that.

It is 10pm here. I'm logging off before I develop crippling insomnia again.

marcustyphoon:
> ...which I guess isn't too bad because you could probably make a Set of just the user's joined community UUIDs for easier lookup?

I couldn't think of an advantage or disadvantage of Array.includes vs Set.has, really. Just put in what was consistent with the rest of the util, for now, but I could see it making sense to make any export of a bunch of unique strings a set. If we're never going to be doing union/intersection/etc with these it's just a matter of taste, I assume.

AprilSylph (approved):
:heart:

PR #1582, opened 2024-08-19
feat(NotificationBlock): UI to show blocked posts

marcustyphoon:
### Description

Simplified version of #1071, this time without saving the user's blog UUIDs into storage (per precedent set in #450).

This implements a UI for NotificationBlock's preference pane similar to PostBlock's, including additions in #719. That is, it displays a list of posts that one has blocked notifications for, including the ability to click to navigate to the posts in question (so that, for example, one can migrate to the native "notification muting" feature instead).

This requires opening a post purely based on its post id, something which the Tumblr API isn't able to do. Therefore, this assumes the post is probably on one of the currently logged in user's blogs, and simply checks each blog the user has access to in descending order of post count for the post in question.

If the user uses container tabs, they'll have to right click the link and open it in the correct, logged-in container tab for this to work. The only way I could think of to do this was to make the link real and include a query parameter in it, ~~which is a bit clunky since it persists once you close the blog view modal. This is therefore of last resort.~~

### Testing steps

todo: document tests


marcustyphoon:
- [x] make link color reasonable in firefox dark mode

marcustyphoon:
- [x] check/sync #1755, #1445 #2115, #2117, other potential changes

marcustyphoon:
> I think we shouldn't potentially open the following feed during this process; it unexpectedly moves the "You're caught up" carousel (even if it's not called that anymore).
> 
> Could we change this to use the blog view as a base instead? That way we even load the right route chunks! e.g. `https://www.tumblr.com/@${uuids[blockedPostID]}?postblock-id=${blockedPostID}`
> 
> _Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/pull/719#discussion_r3116596822_

We don't want to move the "You're caught up" carousel in this feature either... but we don't know what blog we're trying to navigate to in advance. It would be great if there was a valid redpop route that just didn't have anything on it, but like, why would there be?

PR #1584, opened 2024-08-19
feat(Tweaks): Apply badge hiding to PWA

marcustyphoon:
### Description

The "Hide the Home/Following unread count" tweak removes the count in question not only from the sidebar badge and mobile menu badge, but from the page title. This count shows up in another place, though: the badge on the icon when Tumblr is installed as a progressive web app (in certain setups). This removes that badge, too.

For background information about PWA badges, see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Display_badge_on_app_icon.

### Testing steps

Following a second account which is logged in in another browser or profile makes generating following-feed activity on demand easier if one wants to test this.

- Install XKit Rewritten on Chrome/Edge on macOS or Windows (per MDN; I used Chrome on macOS). Check out this PR; refresh any open Tumblr tabs in the chosen browser (active tabs without this code will break it).
- Follow the prompts to install Tumblr as a progressive web app.

<p />

- Open the PWA and wait (on a non-dash page) until there's a visible new posts badge.
- Enable the "Hide the Home/Following unread count" tweak.
- Confirm that the PWA badge disappears along with the sidebar badge and count in the page title.

<p />

- Optionally, unrevert the test commit.
- Open the PWA and enable the "Hide the Home/Following unread count" tweak.
- Wait (on a non-dash page) until the new post count increments; this is easier to see with the test commit, but you can also observe the title or badge in inspect element.
- Confirm that the PWA badge does not reappear.
- Disable the "Hide the Home/Following unread count" tweak.
- Confirm that the PWA badge reappears eventually (the next time the count would update).


src/features/tweaks/hide_following_notification_badge.js
```diff
@@ -22,10 +22,17 @@ const onTitleChanged = () => {
   const newTitle = rawTitle.replace(/^\(\d{1,2}\) /, '');
```
AprilSylph: ...does `(99+)` never appear in the page title?

marcustyphoon: ...ah. shit.

this PR or a separate one?

AprilSylph: Separate, please.

PR #1586, opened 2024-08-20
feat(NotificationBlock): Add optional placeholders for hidden notifications

marcustyphoon:
### Description

Adds placeholder text to the activity page in places where at least one notification is blocked. This is done a bit intelligently; a group of sequential blocked notifications on the same day are replaced with a single placeholder.

Not sure if I would make this on-by-default or off-by-default. 

### Testing steps

- Confirm that blocking/unblocking notifications on a post causes relevant notifications to appear/disappear without having to refresh the page.

Enable the new preference.

- Block notifications for a post with multiple notifications in a row. Confirm that a placeholder element replaces the blocked notifications.
- Block notifications for a post with multiple notifications in a row, spanning multiple days. Confirm that a placeholder element replaces the blocked notifications for each day, resulting in the day labels still making sense.

Feel free to express an opinion about the default status of this new preference.


src/features/notificationblock/index.js
```diff
@@ -15,10 +17,27 @@ const meatballButtonUnblockLabel = 'Unblock notifications';
...
+  display: none !important;
+}
+
+body.${placeholdersClass} :is(${firstHidden}, ${firstHiddenOfDay})::after {
+  display: block;
+  padding-bottom: 1ch;
+
+  content: "(hidden notifications)";
+  text-align: center;
+  color: rgba(var(--black), .65);
```
AprilSylph: Should update this for the new design system:

```suggestion
  color: var(--content-fg-secondary);
```

src/features/notificationblock/index.js
```diff
@@ -29,7 +48,13 @@ const processNotifications = (notificationElements) => {
...
       const blockablePostId = notification.actions?.longTap?.meta?.postId;
 
-      notificationElement.dataset.targetRootPostId = targetRootPostId || targetPostId || blockablePostId || '';
+      const rootId = targetRootPostId || targetPostId || blockablePostId;
+
+      if (blockedPostTargetIDs.includes(rootId)) {
+        notificationElement.parentElement.setAttribute(hiddenAttribute, '');
+      } else {
+        notificationElement.parentElement.removeAttribute(hiddenAttribute);
+      }
```
AprilSylph: Tiny refactor opportunity:

```suggestion
      notificationElement.parentElement.toggleAttribute(
        hiddenAttribute,
        blockedPostTargetIDs.includes(rootId),
      );
```

src/features/notificationblock/index.js
```diff
@@ -101,14 +126,16 @@ const unblockPostFilter = async ({ id, rebloggedRootId }) => {
...
-    blockedPostTargetIDs = changes[storageKey].newValue;
-    styleElement.textContent = buildCss();
+    blockedPostTargetIDs = changes[storageKey].newValue ?? [];
+    pageModifications.trigger(processNotifications);
   }
 };
 
 export const main = async function () {
+  const { placeholders } = await getPreferences('notificationblock');
+  document.body.classList[placeholders ? 'add' : 'remove'](placeholdersClass);
```
AprilSylph: Another one:

```suggestion
  document.body.classList.toggle(placeholdersClass, !!placeholders);
```

AprilSylph (commented):
**Idea**: Could we use CSS [`counter()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/counter) to efficiently add a number to the placeholder text? I've never used it before, but it sounds like we could?

marcustyphoon:
**me:** I mean surely not, browser compat will be an issue. Like, just to be sure, I'll check the supporting Firefox version—I'm sorry, does that say version **one**?

marcustyphoon:
Well Kind Of Maybe: https://github.com/marcustyphoon/XKit-Rewritten/commit/07d1b5b6347394aa980f57aae7f324ccf30d757a

**edit:** primarily-js version for comparison: https://github.com/marcustyphoon/XKit-Rewritten/commit/80356d1b272c7fca167daa01dd5c81dd3e3db462 / https://github.com/marcustyphoon/XKit-Rewritten/commit/d22fc8c206e7d6f749dcd0bddff74ce1d0f58322. Have to think more about lifecycle bugs this way (i.e. what happens if you scroll up into a virtualized notification list?), which is a definite drawback, though I think it's mostly more clear?

marcustyphoon:
- [ ] Fix preference not instantly taking effect
- [ ] `dateSeparatorWrapper` no longer exists

PR #1588, opened 2024-08-21
Quick Reblog: List communities above blogs

marcustyphoon:
### Description

This is a UI tweak that makes a ton of sense if the user has a lot of sideblogs and is not in a lot of communities. If the user is in enough communities that the top (presumably most-used) one now scrolls off the screen when it would have otherwise been, like, list item 3, that's rather silly.

As a general rule, placing the default-selected menu item in the middle makes accessing the whole list easier. So there is some upside. But, eh, this could definitely be argued to be more confusing for some users, while the current way is at least clear even when it's kind of inconvenient.

Not worth a preference, certainly.

### DIFF
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -2,7 +2,7 @@ import { sha256 } from '../utils/crypto.js';
 import { timelineObject } from '../utils/react_props.js';
 import { apiFetch } from '../utils/tumblr_helpers.js';
 import { postSelector, filterPostElements, postType, appendWithoutOverflow } from '../utils/interface.js';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
One could even invert the communities order, such that the default position is closest to both the first blog and the first community, but that's even more confusing even as it's even more ergonomic.

One could count the user blogs and user communities and put the one with fewer entries on top; this would imho be an amazing default until the moment it flipped on a user, in which case it would be a significant problem. If this **were** a preference, leaving it unset until the first time the content script ran and having the content script set the preference based on this only if it's unset would be pretty cute.

PR #1589, opened 2024-08-22
No Recommended: Fix fallback behavior when hiding the entire dash

marcustyphoon:
### Description

Back in #1008 I added "Too many recommended posts to hide!" fallback code when No Recommended hid way too many posts in a row. In #1213, I apparently made a typo and completely broke it, which we (quite reasonably, imho) didn't notice.

This fixes it... mostly. There are some edge cases, but it's good enough, I think.

Notably, the virtual scroller means that some cell elements are empty and thus won't match `postSelector` even though they are actually posts.

### Testing steps

This one might be best to treat as a "just look at the diff," honestly? Or, more accurately, "hey, can't be worse than it was before." But I committed the code I was using with a bunch of random logging and a simulation of having a lot more recommended posts.

### DIFF
--- a/src/features/no_recommended/hide_recommended_posts.js
+++ b/src/features/no_recommended/hide_recommended_posts.js
@@ -34,10 +34,10 @@ export const styleElement = buildStyle(`
 const precedingHiddenPosts = ({ previousElementSibling: previousElement }, count = 0) => {
   // If there is no previous sibling, stop counting
   if (!previousElement) return count;
-  // If the previous sibling is not a post, skip over it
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
yep i will just trust you on this one :+1: 

i'll release v1.0.2 tomorrow unless there's anything else you think \*needs\* to be included in it

marcustyphoon:
nope, looks good to me! I keep a `user-review-requested:aprilsylph` search saved in my clipboard manager, hopefully my management of that reflects your review queue status

Issue #1461, opened 2024-05-21
Tweaks: Activity badge tweak broken in Chinese (Taiwan)


marcustyphoon:
### Platform

~untested~ MacOS 14.4.1 arm64

### Browser

~untested~ Chrome 125.0.6422.77

### Addon version

~untested~ 0.23.6 + dev commits

### Details

~~I didn't directly test this (no activity atm), but~~ as in #1459, with the Tumblr language set to 繁體中文 (台灣), `translate('Activity')` yields `動態`, but the activity button label is `活動心電圖`, so the "Hide the Activity notification badge" tweak is ~almost certainly~ broken.

PR #1591, opened 2024-08-24
Create advancedTranslate util

marcustyphoon:
### Description

Creates a utility function that, rather than returning only a single translated value for each root string as Tumblr's built-in helper does, returns all translated values associated with the root string including plurals and contexts with the same string. The utility falls back to `tumblr.languageData` if this extraction method breaks.


marcustyphoon:
See also: #1461

AprilSylph:
1. huh, i didn't see this one
2. oh hey you found the (output of) the `*.po` files, neat
3. i wonder how much lift it would be for me to add a new property to `tumblr.languageData` that just... has this... (i'm not working until next week, though)

marcustyphoon:
Closing this version of this PR, as evidently we would want to write plural handling logic into it. Reopenable.

PR #1593, opened 2024-08-24
feat(Tag Replacer): Improve mode logic; UI messages 

### LINKED ISSUE #913: Tag Replacer: Clarify dialogue message in add-tags mode
### Concept

In append-only mode, the Tag Replacer modal dialogue message may as well say that that's what going to happen rather than saying the tag will be replaced.

marcustyphoon:
### Description

Tag Replacer's tag input boxes can be parsed into four essential modes: add tag(s), remove a tag, replace a tag, or invalid input that shouldn't be allowed. This PR writes new, clearer logic to determine which is true, allowing the UI to be more customized depending on mode and more clear about which of the modes is active.

Resolves #913.

Features:
- Modal UI messages reflect add-tags mode separately from remove-tags and replace-tags mode
- Modal UI continue button changes text/color in real time based on add/remove/replace mode resulting from current text input
- Modal UI continue button is disabled when text input is invalid or would result in no changes to post tags

Bug fixes:
- Better parsing of text like `,,,` in tag inputs

Functional changes:
- Putting multiple comma-separated tags in the old tag input box is treated as an error state instead of silently removing the commas and treating the result as one tag

Easier to add in future if desired: replace-tag-with-different-capitalization, allow multiple tag inputs.

Makes sense to add (in future?): note in modal about add mode.

### Testing steps

- Confirm that combinations of inputs in the remove tag and add tag input text boxes correctly result in add, remove, or replace modes being shown as the submit button text.
- Confirm that the submit button becomes disabled when there is nothing, is only white space or commas, or are multiple tags in the remove tag box, or when replacing a tag only with itself or a differently-capitalized version of itself.
- Confirm that add (single/multiple), remove, and replace (with single/multiple) modes have appropriately phrased confirmation messages.
- Confirm that add, remove, and replace modes work as expected.


src/features/tag_replacer/index.js
```diff
@@ -21,7 +21,7 @@ const showInitialPrompt = async () => {
       dom('input', { type: 'text', name: 'oldTag', required: true, placeholder: 'Required', autocomplete: 'off' }),
     ]),
     dom('label', null, null, [
-      'Add this new tag:',
+      'Add new tag/tags:',
```
AprilSylph: Prefer condensing "singular/plural" language where possible:

```suggestion
      'Add these new tag(s):',
```

Preceding verbiage subject to debate, suggestion contains my subjective opinion.

src/features/tag_replacer/index.js
```diff
@@ -41,13 +63,35 @@ const showInitialPrompt = async () => {
...
+  const oldTags = oldTagInput.value
+    .replace(/"|#/g, '')
+    .split(',')
+    .map(tag => tag.trim())
+    .filter(Boolean);
+  const newTags = newTagInput.value
+    .replace(/"|#/g, '')
+    .split(',')
+    .map(tag => tag.trim())
+    .filter(Boolean);
```
AprilSylph: This could be DRYer, e.g.:

```js
const getTagsFromInput = (inputElement) => inputElement.value
  .replace(/"|#/g, '')
  .split(',')
  .map(tag => tag.trim())
  .filter(Boolean);

const processTagInputs = (oldTagInput, newTagInput) => {
  const oldTags = getTagsFromInput(oldTagInput);
  const newTags = getTagsFromInput(newTagInput);
```

AprilSylph (approved):
Definitely an improvement!

PR #1594, opened 2024-08-25
User util: Improve extension load speed

marcustyphoon:
### Description

This takes around 100ms off the extension's initial load speed by performing both fetches in the user util at once.

### Testing steps

- Confirm basic extension functionality.
- Modify the communities API URL so that it's incorrect and its fetch will fail. Confirm that Quick Reblog does not have communities as reblog targets, but that the extension continues to function otherwise.


AprilSylph (approved):
Tested via request blocking rather than path modification. Works!

Issue #1595, opened 2024-08-27
Mass Privater: use `before` parameter for initial fetch


AprilSylph:
Surprised that this isn't already included.

When fetching posts to make private, we should include the `before` parameter in the request path, set to the timestamp entered in the datetime picker in the UI.

We're currently fetching all posts on the blog, optionally filtered by tag(s)...
https://github.com/AprilSylph/XKit-Rewritten/blob/b2c80ad16ea6d96b6c7d6337ceadbee67089905e/src/features/mass_privater.js#L221-L227

...and then we're doing our own filtering...
https://github.com/AprilSylph/XKit-Rewritten/blob/b2c80ad16ea6d96b6c7d6337ceadbee67089905e/src/features/mass_privater.js#L208-L210

...when really we just shouldn't be fetching posts from before the selected datetime in the first place.

Additionally: we're not setting `limit: 50` on the non-tag request, when we should be (assuming it makes a difference).

---

https://github.com/tumblr/docs/blob/master/api.md#posts--retrieve-published-posts

> | Parameter       | Type            | Description                                                                                                                                                                                                         | Default                 | Required? |
> |-----------------|-----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------| --------- |
> | **before**      | Number          | Returns posts published before a specified Unix timestamp, in seconds.                                                                                                                                              | False                   | No |

AprilSylph:
@marcustyphoon I'm assigning this to myself because I kinda need this (I have way too many posts on `@april` between now and when I want to private from), but I wouldn't mind if you picked this up either.

AprilSylph:
Oh shit, checking over the API docs more thoroughly, we could also filter by multiple tags instead of doing one round of fetches per entered tag:

https://github.com/tumblr/docs/blob/master/api.md#posts--retrieve-published-posts

> Note that `tag` can either be a string or an array of strings:
> ```
> Fetch posts using this one tag:
> /v2/blog/{blog-identifier}/posts?tag=something
>  or fetch posts using all of these tags:
> /v2/blog/{blog-identifier}/posts?tag[0]=something&tag[1]=example
> ```

marcustyphoon:
...Huh. Assuming you're right about this, I have no idea how I missed it. 

Edit after the multiple tags comment: wait **what**

marcustyphoon:
[...]

PR #1603, opened 2024-08-28
 Mass Deleter: Improve drafts fetching

### LINKED ISSUE #1596: Mass Deleter: Include recent posts in fetch progress
### Concept

As noted in https://github.com/AprilSylph/XKit-Rewritten/issues/1595#issuecomment-2312548185, when copying Mass Deleter's code into Mass Privater, I added a bit in order to update the progress counter of the post fetching based on the total number of fetched posts, ignoring the timestamp filter. Without this, someone with a huge post count who selected a long-ago timestamp would see the UI appear to freeze for a while.

Mass Deleter can, specifically in the case of deleting drafts where we allow filter-by-timestamp, have this added.

marcustyphoon:
### Description

This improves the fetch speed of Mass Deleter's delete-drafts mode and displays an incrementing counter based on the total number of posts fetched as well as the number that meet the date range criteria.

See #1595.
Resolves #1596.

Do you think we should also include the `before` parameter on every api fetch that could plausibly be updated to support it, even when the endpoint doesn't?


marcustyphoon:
- [ ] address #1610

AprilSylph:
> Do you think we should also include the `before` parameter on every api fetch that could plausibly be updated to support it, even when the endpoint doesn't?

nah, that seems silly

AprilSylph (approved):

marcustyphoon:
> Do you think we should also include the `before` parameter on every api fetch that could plausibly be updated to support it, even when the endpoint doesn't?

Fun fact: I don't remember what this refers to or means even a little bit.

AprilSylph:
> > Do you think we should also include the `before` parameter on every api fetch that could plausibly be updated to support it, even when the endpoint doesn't?
> 
> Fun fact: I don't remember what this refers to or means even a little bit.

at the time, we had recently updated Mass Privater to use the `before` query parameter in its API URL to avoid fetching all posts on the target blog and just filtering to before the input timestamp. now we start fetching posts on the target blog starting at the input timestamp instead. but, this parameter only exists for published posts on a blog, not drafts or queue--you were proposing we add it here under the assumption it would be a no-op, just in case tumblr supports it in the future.

PR #1610, opened 2024-09-09
Mass Privater: Display total fetched count without filtering

### LINKED ISSUE #1609: Mass Privater: "(checked {x})" count can be incorrect
https://github.com/AprilSylph/XKit-Rewritten/blob/18298b6ad8127ced464852b4d5be539ae9b163d4/src/features/mass_privater.js#L202-L214

If Mass Privater fetches posts that are already privated, it doesn't count them in the "(checked {x})" text.

This can lead to a state where the feature is fetching pages upon pages of already-privated posts (since there is no way to request "only public posts" from the API, or pretend that we're logged-out when fetching...) but the UI is not updated for minutes or potentially even hours at a time.

marcustyphoon:
### Description

Resolves #1609.

This changes the "checked x" count that Mass Privater shows while fetching posts to include all posts without any filtering, ensuring that it continues to increment in real time even if the posts currently being fetched aren't relevant.

This includes counting the same post multiple times if it has multiple searched tags and gets fetched multiple times; do you think this makes sense? (Maybe it makes "checked x" slightly inaccurate; that could be "fetched x" or "downloaded x" something, I dunno.) There's a tradeoff here.

### Testing steps

I did not really test this, as I don't have enough private posts to notice the difference.

### DIFF
--- a/src/features/mass_privater.js
+++ b/src/features/mass_privater.js
@@ -192,26 +192,24 @@ const privatePosts = async ({ uuid, name, tags, before }) => {
     ]
   });
 
-  const allPostIdsSet = new Set();
+  let fetchedPosts = 0;
   const filteredPostIdsSet = new Set();
 
   const collect = async resource => {
     while (resource) {
       await Promise.all([
         apiFetch(resource).then(({ response }) => {
-          const posts = response.posts
+          response.posts
             .filter(({ canEdit }) => canEdit === true)
-            .filter(({ state }) => state === 'published');
-
-          posts.forEach(({ id }) => allPostIdsSet.add(id));
-
-          posts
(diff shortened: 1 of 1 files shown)


AprilSylph:
> This includes counting the same post multiple times if it has multiple searched tags and gets fetched multiple times; do you think this makes sense? (Maybe it makes "checked x" slightly inaccurate; that could be "fetched x" or "downloaded x" something, I dunno.)

After much thought, I'm concluding that I don't care. It's probably a rare enough scenario that it's not worth the mental effort to accommodate writing 100% accurate copy for.

AprilSylph (approved):
Works. Easily tested on a blog with only 22 posts, 20 of which were private.

PR #1612, opened 2024-09-10
Quick Reblog: Reimplement tag suggestion UI

### LINKED ISSUE #1102: Quick Reblog: Inconsistent suggestion behavior on tag field click
### Platform

MacOS 13.2.1 arm64

### Browser

Firefox 115.0b9 / Chrome 114.0.5735.106

### Addon version

v0.21.9

### Details

With Quick Reblog's "suggest tags from the post being reblogged" option enabled, clicking in the empty tag field in Chrome will immediately open the suggested tags `datalist` element, while in Firefox, one must type something into the tag field and press backspace or comma to make the element appear.

This is an inconsistency due to differences in browser implementation, and very well might not be fixable. If we do prefer the instant-appear Chromium behavior, however, I wouldn't be surprised if there's a way to trick Firefox into showing the element on focus if the field is empty.

(Written based on a user report that the tag suggestion feature wasn't working which turned out to be due to the Firefox behavior; this would be a reason to prefer the Chromium behavior.)

### LINKED ISSUE #1611: Suggest tags from post reblogged not working on firefox mobile
### Platform

Android 14

### Browser

Firefox Beta 131.0b3 (Build #2016042863),

### Addon version

1.03

### Details

have no problems using this feature on desktop chrome but mobile firefox beta (i believe beta is necessary for the extension?) does not display any menu for tags from previous post.
Screenshot_20240909-150953

marcustyphoon:
### Description

This manually implements a component similar to the HTML `datalist` element that we're currently using to show the user tag suggestions. This should result in:

- Support for Firefox for Android. Resolves #1611.
- Consistent UI between browsers that we can customize; see the previously linked issue for a buggy example in Kiwi Browser, and note that desktop Firefox and Chrome have noticeably different UI designs (neither of which, imo, look great).
- Suggestions containing only the suggested tag instead of everything in the tags input box. Usable suggestions when there is more than a small amount of content in the tags input box.
- A suggestion box that opens consistently, with a single click, in all browsers. Resolves #1102.

At time of writing, the DOM structure and CSS I created are... uh, let's just say "probably improvable." This feels like one of those PRs that could use the "April fixes everything" special.

### Testing steps

Needs further testing for modal logic edge cases. Needs further testing for overflow (e.g. with long suggested tags/many suggested tags)

### DIFF
--- a/src/features/quick_reblog.css
+++ b/src/features/quick_reblog.css
@@ -4,7 +4,6 @@
   box-shadow: 0 0 15px 0 rgba(0,0,0,.5);
   padding: 2px;
   border-radius: 3px;
-  overflow: hidden;
   color: rgb(var(--black));
   font-size: .875rem;
   font-weight: normal;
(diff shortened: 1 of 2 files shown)


marcustyphoon:
I suppose this could be generic and in a util, come to think of it.

marcustyphoon:
Hiding the element when the parent doesn't match`:focus-within` doesn't work for two reasons:

- in Safari, when clicking the new tag suggestions element, focus is lost before the click action is processed, and everything breaks
- clicking the new tag suggestions element outside of one of the actual action buttons drops focus and hides the tag suggestions element

Thus:

- [ ] remove that and hide the element on click if the click is outside of the element

PR #1619, opened 2024-09-16
Fix static styles failing to update

### LINKED ISSUE #1354: Improve Firefox automatic update CSS inconsistency
### Concept

As noted in discussion #1353, when Firefox pulls a new extension update, changes to scripts' static stylesheets aren't applied until one opens a new tab (while changes to script code and dynamic stylesheets are, potentially leading to temporary inconsistent behavior).

Not sure we actually have a method to improve this (obviously besides using only style elements, which does have some significant advantages), but even if that's the case it seems worth having an issue filed.

marcustyphoon:
### Description

> You wanna know what's really scaring me? The link-element-pointing-to-a-css-file-in-the-extension-package method, when loaded via web-ext run, doesn't even show changes when I press refresh on the page. I have to command-shift-R. Surely that can't reflect the real update experience... right?
> 
> _Originally posted by @marcustyphoon in https://github.com/AprilSylph/XKit-Rewritten/issues/1354#issuecomment-2352212011_

This is what I had to do to get the automatic reload from `web-ext run` to apply changes to static stylesheets included with features: have the load process remove the link element, then add a new link element with a different (cache busted) URL.

Presumably, this resolves #1354. Maybe only one of these two changes would be sufficient to fix the issue in practice; I'm not sure. Both were required when I used web-ext. At some point maybe I'll test just loading the extension as unpacked?

(Not sure if this is technically a performance regression.)

### Testing steps

- Load the extension into Firefox and open Tumblr. Enable a feature with a static stylesheet (say, AccessKit).
- Add something random like `body { border: 10px solid red; }` to the feature stylesheet.
- Reload/update the extension, if you aren't using `web-ext`to doing that for you.
- Confirm that the style is applied.

<br />

- Disable the feature.
- Confirm that the link element is removed and the style is no longer applied.

There may be a more accurate way to simulate an extension update, like loading an old version's xpi. I haven't tried that yet (gotta go find a change to a static stylesheet in the release history that would be really easy to test, especially given other bugs in the Firefox autoupdate process we've recently fixed).

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -6,6 +6,8 @@
 
   const restartListeners = {};
 
+  const randomId = Math.random();
+
   const runScript = async function (name) {
     const scriptPath = browser.runtime.getURL(`/features/${name}.js`);
     const { main, clean, stylesheet, styleElement, onStorageChanged } = await import(scriptPath);
@@ -16,8 +18,9 @@
     if (stylesheet) {
       const link = Object.assign(document.createElement('link'), {
         rel: 'stylesheet',
-        href: browser.runtime.getURL(`/features/${name}.css`)
+        href: browser.runtime.getURL(`/features/${name}.css?v=${randomId}`)
       });
+      link.className = 'xkit';
       document.documentElement.appendChild(link);
     }
     if (styleElement) {
@@ -50,7 +53,7 @@
       clean().catch(console.error);
     }
     if (stylesheet) {
-      document.querySelector(`link[href="${browser.runtime.getURL(`/features/${name}.css`)}"]`)?.remove();
+      document.querySelector(`link[href^="${browser.runtime.getURL(`/features/${name}.css`)}"]`)?.remove();
     }
     if (styleElement) {
       styleElement.remove();
@@ -97,7 +100,7 @@
   });
 
   const init = async function () {
-    $('style.xkit').remove();
+    $('style.xkit, link.xkit').remove();
 
     browser.storage.onChanged.addListener(onStorageChanged);
 


src/content_scripts/main.js
```diff
@@ -6,6 +6,8 @@
 
   const restartListeners = {};
 
+  const randomId = Math.random();
```
AprilSylph, line 9: shouldn't we use `getRandomHexString()` instead here? for even more randomness.

AprilSylph (approved):
leaving a pre-approval; code comment withstanding. disabled the "Dismiss stale pull request approvals when new commits are pushed" protection rule so you no longer need a re-review after addressing nit feedback.

src/content_scripts/main.js
marcustyphoon: Hm. We'd have to inline it or dynamic import it, but a hex string definitely feels more semantic than a number between 0 and 1. I don't think the entropy matters; we realistically could have used `Math.random().toString(16).replace('0.', '')` as a oneliner, come to think of it.

But that just makes me think: why is this random at all? It may as well be `Date.now()`.

AprilSylph: +1 to `Date.now()`, vanishingly unlikely that XKit boots and the addon is updated in the _same millisecond_

marcustyphoon: Ironically, if I did the math correctly, the chance of that happening is around 3.5 million times higher than the chance of `String(Math.random())` giving the same result twice. Of course, using Date.now() guarantees that if this does happen it's when refreshing during an update, rather than... well, hm, wait, hold on, no...

With the Math.random() call:
- As far as I'm aware, we probably have enough bits of seed entropy and the PRNG is good enough to assume that each possible 64 bit float value that comes out of `Math.random()` is equally likely and doesn't depend on prior values without making the math too inaccurate (53 bit mantissa).
- There are 1e16 possible outputs of `String(Math.random())`. I would have to think about whether the base conversion makes them unequally likely enough to matter, but probably not significantly.
- Each time we update the extension, we can assume the browser downloads the update at an arbitrary random time. For each tumblr tab the user has open, this process occurs, and if we hit the `[average number of open tabs this user has] / 2^53 ≈ [average number of open tabs this user has] / 1e16` chance, one of their tabs will have broken styles until it's hard refreshed.

With Date.now():
- Each time we update the extension, we can assume the browser downloads the update at an arbitrary random time. Bad results occur if the user clicks to open a new tab, XKit loads, Date.now() is called, the browser processes the update, XKit is loaded again in the same tab, Date.now() is called again within the same millisecond, etc.
- If we pretend these actions occur in independent processes and have independent timing windows, the chance that this happens depends on the user's new tab open frequency; there are 3e10 milliseconds in a year, and we drop (say) 10 releases in a year, so each time we drop a release, if we hit a `[average number of tumblr tabs this user opens in a year] / 3e9` chance, one of their tabs will have broken styles until it's hard refreshed. This is many orders of magnitude more likely than the above case in the best case scenario.
- These tasks actually run in the javascript main process, though, which isn't multithreaded. Therefore, a) if the event that the Math.random() call is in exceeds 1ms of duration following the call, which in the profiler it looks like it would not necessarily, b) if the event that the web extension update task is in exceeds 1ms of duration before the Math.random() call, c) the sum of those times exceeds 1ms, or d) if the scheduler does not and can not run the web extension update task immediately at this point in the middle of a bunch of React bootstrapping, then the chance of having broken styles due to this particular issue is actually zero.

The chance that this analysis is entirely incorrect without this final note is, of course, probably quite a bit higher than one in a million, making the rest of it vanish within the error bars, though, so like, there is that.

PR #1621, opened 2024-09-16
Popup: Clear storage optionally on backup restore

### LINKED ISSUE #1191: Importing backup keeps old storage keys
### Platform

MacOS 13.5 arm64

### Browser

Chrome 115.0.5790.170

### Addon version

master

### Details

https://github.com/AprilSylph/XKit-Rewritten/blob/248912f7c22e5c129a52563acde76a7b4e65b620/src/browser_action/render_backup.js#L57-L58

When one imports a backup of their preferences, any keys in the backup get set, but any keys not in the backup do not get removed, as one might expect. (adding `await browser.storage.local.clear();` would implement this behavior.)

Of course, there are probably situations where one might want the current, merge-my-import-into-my-current-preferences behavior, honestly; one could argue that this is only sort of a bug. It would probably be way too confusing to implement two different restore buttons that allow the user to pick either behavior and that only appear in the case where the pasted backup does not contain some of the keys which are currently set, though it's not like it's hard.

marcustyphoon:
### Description

This adds a checkbox to the backup restore interface in the specific scenario where the user's pasted backup is missing some of the values that the extension storage currently contains. As noted in the linked issue, this scenario currently results in those old values persisting. With this checkbox selected, the extension will remove those values as part of the restore process.

Most people should not see this, as they're presumably loading a backup with the same amount of information or more information than their storage currently contains

Resolves #1191.

### Testing steps

- Export your data, if you're testing this with your real data. Toggle some preferences to generate some data if you aren't.

<br />

- Paste your current data into the restore box. Confirm that no checkbox appears and that restoring works as expected.

<br />

- Paste `{}` into the restore box. Uncheck the resulting checkbox. Confirm that restoring works as expected but results in no change.
- Paste a subset of your current data into the restore box with a value change. Uncheck the resulting checkbox. Confirm that restoring works as expected but results in no change.
- Paste `{}` into the restore box. Check the resulting checkbox. Confirm that restoring removes all of your data.


marcustyphoon:
Aside: I had no idea you could do this!

```css
[data-stale-keys]::after {
  content: "Clear " attr(data-stale-keys) " Existing Value(s)";
}
```

AprilSylph (changes requested):
I really think we should just clear the storage by default and have the import work on a whole-replacement basis. The usefulness of configuration merges is nil, in my opinion, when it's only a merge of top-level values and not any of the values that would genuinely be useful to merge, e.g. AlreadyReblogged entries.

A "This will overwrite your existing configuration!" disclaimer that appears if the current config has any keys at all would be more easily understood.

marcustyphoon:
Yeah, makes sense 👍

PR #1623, opened 2024-09-17
Tweaks: Improve contributed content tweak compatibility

### LINKED ISSUE #873: readability issue with Themed Posts + Highlight contributed content
### Platform

Windows 10

### Browser

Chrome

### Addon version

v0.21.0

### Details

when `Themed Posts`->`Theme every reblog trail item individually` and  `Tweaks`->`Highlight contributed content on reblogs` are both enabled, the contributed content highlight's background color overrides the blog theme background color, but the text color stays as the one from the blog theme. when the blog theme's text color is similar to the contributed content highlight this can make the text difficult to read

[example post](https://www.tumblr.com/im-mage/702140803842244608/example-post) (`Themed Posts`->`Enable theming on the blog view` also needs to be enabled to see the issue here since it's a blog link)

image

marcustyphoon:
### Description

As noted in the linked issue, the "highlight contributed content" tweak looks like it applies a subtle color highlight to the relevant trail item, but it's actually overriding the background with the `follow` palette color, which generally has that effect on most palettes but isn't compatible with customization.

This applies a low-opacity accent color instead. 7% matches more closely with the current color on the True Blue palette, but I like ~10% better in general.

I also tried `blue` instead of `accent`.

> I prefer blue on:
> cement, strongly
> 
> I prefer deprecated-accent on:
> vampire, moderately
> goth rave, slightly

Most notably, though we override `accent` in Themed Posts, but we don't override `blue`, so `accent` gives more stable results.

In addition, this disables the tweak's effect entirely when Themed Posts' "Theme every reblog trail item individually" setting is enabled; I guess it could make sense not to do this since it'll still work when people have default blog themes? It seems like a lot of the time it would be pointless, though.

Resolves #873.

### Testing steps

The "alternative color demo" commit demos a way to add a second color option and compare side-by-side; one can find a post with a decent reblog chain and shift-P through different palettes.

- Confirm that contributed content trail items look reasonable with Themed Posts disabled in each palette.
- Confirm that contributed content trail items look reasonable with Themed Posts enabled. (Changing palettes should have no effect.)

### DIFF
--- a/src/features/tweaks/highlight_contributed_content.js
+++ b/src/features/tweaks/highlight_contributed_content.js
@@ -2,6 +2,6 @@ import { buildStyle } from '../../utils/interface.js';
 
 export const styleElement = buildStyle(`
 [data-is-contributed-content="true"]:not([data-xkit-themed]) {
-  display: flow-root; background-color: rgb(var(--deprecated-accent), 0.1);
+  display: flow-root; background-color: rgb(var(--deprecated-accent), 0.13);
 }
 `);
--- a/src/features/tweaks/highlight_contributed_content.js
+++ b/src/features/tweaks/highlight_contributed_content.js
@@ -4,4 +4,7 @@ export const styleElement = buildStyle(`
 [data-is-contributed-content="true"] {
   display: flow-root; background-color: rgb(var(--deprecated-accent), 0.07);
 }
+:has(+ [data-is-contributed-content="true"]) {
+  display: flow-root; background-color: rgb(var(--blue), 0.07);
+})
 `);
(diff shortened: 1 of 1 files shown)


src/features/tweaks/highlight_contributed_content.js
```diff
@@ -1,7 +1,7 @@
 import { buildStyle } from '../../utils/interface.js';
 
 export const styleElement = buildStyle(`
-[data-is-contributed-content="true"] {
-  display: flow-root; background-color: rgb(var(--follow));
+[data-is-contributed-content="true"]:not([data-xkit-themed]) {
+  display: flow-root; background-color: rgb(var(--deprecated-accent), 0.1);
```
AprilSylph, line 5: when I last looked at this (in July 2022, ouch), I calculated that a _13_%-opacity accent matched the luminance of `--follow`
```suggestion
  display: flow-root; background-color: rgb(var(--deprecated-accent), 0.13);
```

AprilSylph (approved):

PR #1625, opened 2024-09-18
Implement remove buttons on own post tags

### LINKED ISSUE #549: Feature: Tag remover (...tag editor?)
### Concept

This isn't a fully formed idea, but I often find myself editing posts right after posting just to remove the automatically added `#Youtube` tag and wishing that Quick Tags could be used not only to add tags without a context switch, but to remove them as well. It would be even cooler to extend this to fixing typos in tags as well.

I could imagine the removal-only variant of this feature manifesting itself as delete icons that appear on tags on your own posts when hovered, or the typo-fixing variant manifesting itself as a Quick Tags-like popup containing a text box populated with the full comma separated tags string, or both, or an integration with Quick Tags itself. None of these ideas stand out as comprehensive and native-feeling yet, though, so I would love it if someone had a better idea.

marcustyphoon:
### Description

Core implementation of a feature I've wanted for a long time, but that is somewhat niche: the ability to remove tags from a post with one click. This is very nice when tags are automatically added to posts based on their source (i.e. `#Youtube`) but is otherwise not useful very often.

Resolves #549 (which I intentionally let go stale).

Things one should debate:
- If merging this functionality, should it be part of Quick Tags, be its own feature, be... a tweak (which _probably_ makes the most sense scope-wise but completely violates the norms for how complex a tweak is), etc?
- Should every tag on an editable post be removable or just the auto-added ones? What is the list of auto-added ones?

Ultimately my guess is that no one can come up with a justification for implementing this, but hey, I made it.

### Testing steps

- Create a post (or multiple posts) with a youtube embed, an instagram embed, a spotify embed, a bandcamp embed, and a soundcloud embed, and some random tags. Save the post; Tumblr will give it extra tags.
- Enable the new Quick Tags option. Confirm that the auto-added tags gain x buttons that will remove the tag in question, but other tags do not.
- Click a remove button. Confirm that the tag is removed.
- Use Quick Tags to add back the same tag (with the same capitalization). Confirm that the tag is added and appears with a remove button.
- Confirm that posts with a relevant tag which the user cannot edit don't have remove buttons.


marcustyphoon:
- [x] #1980

PR #1627, opened 2024-09-26
Timeline ID util: Add blog uuids as possible identifiers

marcustyphoon:
### Description

Fun fact: you can navigate to a "channel" page using a blog uuid instead of a blog name. The page will work, and its timeline id or timeline attribute will contain the blog uuid. For example, https://www.tumblr.com/blog/t:Fft_7iTGnBlHdl-UtlllwQ or https://www.tumblr.com/blog/t:Fft_7iTGnBlHdl-UtlllwQ/queue should work if you're logged in to your main account.

This adds those cases to the values matched by the regexes in the timeline id util.


marcustyphoon:
Just clicked those links, and that didn't work, so... I guess you can't anymore.

PR #1639, opened 2024-10-28
feat(Mass Privater): Add mass unprivate functionality

### LINKED ISSUE #1626: Mass Unprivater as an option to Mass Privater
### Concept

In the same way how you can mass private post based on tags and time frame, I would like to unprivate post based on the same things instead of going through anything one by one.

marcustyphoon:
### Description

Per a number of user requests, this adds the ability to unprivate posts to Mass Privater, including by tag. There's currently no realistic way to do this without an extension, since you can't find all of your private posts (that contain a certain tag or otherwise) through the Tumblr UI. No, it's not fast, but it's faster than manually scrolling through your archive, that's for sure.

Resolves #1626.

~~Currently contains the diff of #1610; must be rebased after it is merged (or close it if merged first).~~

### Testing steps

Oh boy. Um.

- [ ] Confirm that Mass Privater's "private posts" mode still works as it used to.
- [ ] Confirm that Mass Privater's "unprivate posts" mode functions (slowly, at one post per second)...
  - [ ] in general, with no errored post list if there are no errors
  - [ ] with a before date selected
  - [ ] with tags selected
- [ ] Confirm that Mass Privater's "unprivate posts" mode displays a list of errored posts if some of its posts are not npf-editable (populated up-front) or fail (e.g. disconnecting network)


src/features/mass_privater.js
```diff
@@ -178,86 +178,116 @@ const showPostsNotFound = ({ name }) =>
...
-      }),
-      sleep(1000)
-    ]);
+      await Promise.all([
+        apiFetch(`/v2/blog/${uuid}/posts/${postData.id}`, {
+          method: 'PUT',
+          body: {
+            ...createEditRequestBody(postData),
+            state: 'published'
+          }
```
AprilSylph, line 272: hmmm. the fact that this will only work on NPF/text/answer posts is... concerning.

i feel like we should have a separate "incompatible" count to show to the user, rather than trying to action on them here, knowing that the operation will always fail.

do we have a util for NPF compatibility yet? i feel like we need one.
if a post is stored in blocks format, or the original type is text, or the original type is note, it should be editable via NPF endpoints.

marcustyphoon: Oh, shoot.

marcustyphoon: (And yes we do; it currently returns `isBlocksPostFormat || shouldOpenInLegacy === false;`).

marcustyphoon: Made something. A bit rudimentary, maybe, but I'm not sure how to improve it.

AprilSylph: Maybe it should be a `<table>`...?

marcustyphoon:
Today I learned: I have absolutely no idea how to control overflow on a table cell.

AprilSylph:
(Don't know why I put this in my review queue when it's still marked as "help wanted".)

marcustyphoon:
Oh, this one.

I might just give up on table and switch to ~~flexbox~~ _grid_; I don't think I've ever gotten a table to do anything close to what I wanted ever.

marcustyphoon:
¯\\\_(ツ)\_/¯ 

(aside: I guess this could also just be a list of links, with some placeholder text added for the posts without a summary. or... yeah, idk. hm, or should the links be formatted timestamps instead of post ids?)

marcustyphoon:
maybe? idk

marcustyphoon:
Current look. Suggestions welcome.

AprilSylph:
Hmm. You know what, what's wrong with just a scrollable list of clickable post URLs? If the post has a summary, it should have a slug too...

PR #1648, opened 2024-11-10
Mass Deleter: Show blog name in confirm modal

### LINKED ISSUE #1647: Mass Deleter: Show blog name in confirm modal
### Concept

Yeah, yeah, it says "this blog" and the button only appears on blog pages. But it's still kind of freaky to press confirm on "delete all of my stuff" without the UI definitely showing you which blog you mean. Should be a simple little refactor to query and display the target blog name one modal earlier.

marcustyphoon:
### Description

Does a refactor I probably should have done a while ago (making `createTagSpan`/`createBlogSpan` part of the modals util rather than duplicating them) and then uses the latter to clarify the target blog in Mass Deleter's confirm modal, as per the linked issue.

Resolves #1647.


AprilSylph (approved):
Works as described. Also, since we're modifying at least one created `<button>` here, I wouldn't mind a `button()`/etc. shorthand refactor being performed prior to merging.

marcustyphoon:
Sure. Aside: would you like:
- new-DOM-util PRs for everything
- one new-DOM-util PR for everything
- to do it yourself
- to do it yourself; you can start with my diff checked out
- neither, we do it as we touch files
- something else

AprilSylph:
Do it as we touch usages of `dom()`.

Oh right, and I should mark that thing as deprecated via JSdoc, shouldn't I...

marcustyphoon:
Oh—do you mean "convert the entire file" or just the part the PR touched?

AprilSylph:
Needing to touching a `dom()` usage should be the trigger to convert the entire file, ideally in its own commit.
If it's a big lift (like in Quick Reblog), the refactor should be PR'd by itself first for in-depth regression testing.

AprilSylph:
...or afterwards works too, actually.

PR #1653, opened 2024-11-17
feat(Quick Tags): Add suggestion-from-reblog-source feature

marcustyphoon:
### Description

Quick Reblog has a feature to quickly copy relevant tags from a post when reblogging the post, but if one drafts or queues a reblog without XKit (e.g. on mobile) and wants to do this same thing after the fact, there's no equivalent feature.

This adds pretty much the same feature to Quick Tags.

Notably, one could use the datalist feature on the Quick Tags text input box (making equivalent functionality to Quick Reblog) or add more buttons with the suggestions. The former is kind of annoying because the text input box is at the top of the modal, so the datalist covers the tag buttons, but UI consistency is important. I dunno.


src/features/quick_tags/index.js
```diff
@@ -124,7 +121,26 @@ const togglePopupDisplay = async function ({ target, currentTarget: controlButto
...
+    if (showReblogSuggestions) {
+      const { rebloggedFromUuid, rebloggedFromId } = await timelineObject(controlButton.closest(postSelector));
+      if (rebloggedFromUuid && rebloggedFromId) {
+        const { response: { tags, blogName, postAuthor, rebloggedRootName } } = await apiFetch(`/v2/blog/${rebloggedFromUuid}/posts/${rebloggedFromId}`);
+
+        const suggestableTags = tags;
+        if (blogName) suggestableTags.push(blogName);
+        if (postAuthor) suggestableTags.push(postAuthor);
+        if (rebloggedRootName) suggestableTags.push(rebloggedRootName);
+        const tagsToSuggest = suggestableTags.filter((tag, index, array) => array.indexOf(tag) === index);
```
AprilSylph: Spotted: code copied from something I wrote that used an array when it should really have used a `Set`.

marcustyphoon:
Well... I guess this should use the tab interface from Quick Reblog.

I feel like it might look kind of silly without the context of other elements being around it, but eh, what can you do.

PR #1654, opened 2024-11-28
Tweaks: Fix subtle following/mutuals indicators option

### LINKED ISSUE #1649: Tweaks: "Make following/mutuals indicators on notifications more subtle" option is broken
### Platform

Steam Deck

### Browser

Firefox 132.0.1 (64-bit)

### Addon version

v1.0.4

### Details

Activity items are using a new label component that the "Make following/mutuals indicators on notifications more subtle" tweak does not affect, seemingly in any capacity.

marcustyphoon:
### Description

The "make following/mutuals indicators on notifications more subtle" tweak's purpose is to hide the text of the indicator in question, leaving only the smaller icon unless the user hovers over it. Staff removed the icons in question, breaking the feature; this adds the icons back and restores the functionality.

It also makes some tweaks required to work in the new activity notification types and to exclude the similar elements on the communities members page (as per the name of the feature; we could also make it work there but that seems like a separate thing).

Resolves #1649.

~~Improvement on #1568.~~

- [x] handle new activity item format, addressing note in #1692

### Testing steps

- Confirm that the feature functions correctly on a new "replied to your post"/"replied to you in a post" notification from a mutual/followed user on both the activity page and in the activity modal.
- Confirm that the feature functions correctly on the classic style of notification (most other notification types) from a mutual/followed user on both the activity page and in the activity modal.
- Confirm that the correct icons are added to both following and mutual labels.
- Confirm that the mutual/following indicators on the communities members page are unaffected.


marcustyphoon:
Well, this is a bit inconvenient. This stops happening if I set the display property on the random-div-containing-a-zero-width-space-for-wrapping-reasons to `inline-block`, I think, but t's not the most elegant thing to target with a CSS selector.

src/features/tweaks/subtle_activity_mutuals.js
```diff
@@ -3,9 +3,10 @@ import { keyToCss } from '../../utils/css_map.js';
 import { buildStyle } from '../../utils/interface.js';
 import { dom } from '../../utils/dom.js';
 
-const labelSelector = `${keyToCss('followingBadgeContainer', 'mutualsBadgeContainer')}:has(> svg)`;
+const labelSelector = `${keyToCss('followingBadgeContainer', 'mutualsBadgeContainer')}:is(${keyToCss('activity', 'activityItem')} *)`;
```
AprilSylph, line 6: i think this can be simplified?

i was wondering why i never thought of using `:is()` to filter to parent selectors, when i realised, wait, CSS just lets you do that without any tricks

```suggestion
const labelSelector = `${keyToCss('activity', 'activityItem')} ${keyToCss('followingBadgeContainer', 'mutualsBadgeContainer')}`;
```

src/features/tweaks/subtle_activity_mutuals.js
```diff
@@ -15,13 +16,25 @@ export const styleElement = buildStyle(`
...
   width: 0;
 }
 
-a:not(:hover) ${labelSelector} > svg {
+${keyToCss('tumblelogName', 'activityHeader')}:not(:hover) ${labelSelector} > svg {
   margin-left: 0;
 }
+
+/* fixes hover detection when covered by the "activityItemLink" <a> element */
+${keyToCss('activityHeader')} div:has(> .${spanClass}) {
```
AprilSylph, line 28: **nit**: generally i find it easier to parse CSS comments that are about "what the rule/declaration does" if they're inside the rule. otherwise it kinda breaks the flow of "what does this target" &rarr; "what does this do"

```suggestion
${keyToCss('activityHeader')} div:has(> .${spanClass}) {
  /* fixes hover detection when covered by the "activityItemLink" <a> element */
```

AprilSylph (approved):
works a treat! couple code concerns to be addressed before merge, no need for re-review but please do re-test after modifying. :purple_heart:

src/features/tweaks/subtle_activity_mutuals.js
marcustyphoon: Oh yeah. I've absolutely done it on purpose for readability reasons when composing selectors, but I have no idea why I did it here :D

marcustyphoon:
Oh. Hah—that :is() was on purpose.

marcustyphoon:
Tested in Firefox 121 and Chrome 105.

However, something that wasn't in my list of tests is "what if Tumblr removes `#managed-icon__profile-double`; the conditional I wrote to handle that caused broken behavior instead of causing the feature to do nothing. Fixed by making the check an early return and moving the other DOM mutations after it; fix tested by editing a typo into one of the hrefs.

AprilSylph (approved):
i don't think this needed re-review to be honest

Issue #1658, opened 2024-12-05
Mutual Checker, NotificationBlock, Quote Replies: Some notifications are not processed


marcustyphoon:
There appears to be a tweaked Tumblr notification ~~element~~ **type*** which is only used sometimes and which does not match our `notificationSelector`. This breaks Quote Replies, as per the quoted comment, as well (I think) as Mutual Checker and NotificationBlock on those notifications.

***Edit:** Ah, I see this is not just a frontend change; the information provided by the backend has changed as well. This may require some nontrivial code rewriting, particularly for Quote Replies.

<img width=501 src="https://github.com/user-attachments/assets/e538fb39-c7c9-417e-9f7d-8631b4cd568e">

---

> hello! I'm also finding that though I've refreshed, toggled off/on, that I'm unable to even see the option to click in order to use quote replies. Now it's been awhile since I've used the feature, so I don't fully remember how it works, but I'm really not getting anything showing in order to do it, I'm wondering if it's something that could have been broken when tumblr implemented their new way of replying to replies within the replies? 
![image](https://github.com/user-attachments/assets/c10d6652-69e9-4e51-bf13-3e190cd94e11)
![image](https://github.com/user-attachments/assets/7ff87954-4535-419a-9cbd-8a8008d41506)
![image](https://github.com/user-attachments/assets/5e693ec6-52ac-4125-a593-75de074088e8)

_Originally posted by @lauraxjane in https://github.com/AprilSylph/XKit-Rewritten/issues/1655#issuecomment-2518848059_

marcustyphoon:
Notably, this currently causes some activity elements to have `[role="listitem"]` and others not (`[role="listitem"] { outline: 5px double red; outline-offset: -5px }` looks pretty silly), which is probably an oversight on Tumblr's part.

lauraxjane:
I'm not gonna lie, I know only the basics of general html coding, like for tumblr themes, is this something I should be able to edit myself or is it something that is requiring a fix on the extension's end? so sorry to be kinda dumb here! but appreciate any and all help, thank you!

marcustyphoon:
No worries! Yes, this will require (a lot of) work on the extension side to fix. Looking into it!

marcustyphoon:
This issue should be fixed in v1.0.10, which is now available on all supported browsers!

PR #1659, opened 2024-12-05
Mutual Checker, NotificationBlock, Quote Replies: Process new reply format

marcustyphoon:
Well, #1658 is fairly problematic.

### Description

This contains (somewhat rudimentary) fixes for the linked issue, in which the relevant features are broken by certain kinds of activity item that use a new internal data format, containing only enough information to render the notification rather than all of the internal data we use to add functionality.

- Mutual Checker can still easily determine if the replying user is a mutual, since Tumblr renders that.
- NotificationBlock wants to know the root post of the target post, which just so happens to work in this case because the activity items in question are on original posts, but will be a problem in the future if all activity items switch to this API.
- As per below discussion, with some slightly hairy URL and NPF parsing, Quote Replies can look up and fetch the correct reply and get the summary text from the notification data.

<details>
<summary>outdated Quote Replies info</summary>

- Quote Replies... is a somewhat complicated scenario because we **might** technically have enough information to look up the relevant post and post activity using an API fetch, but only via very convoluted methods (the activity item links to the post; we could parse the url to get the relevant information to API fetch the post but we have no timestamp with which to easily find the reply). For now, just to make it work on all posts rather than keep getting confused users, I've processed the activity item itself into a rudimentary post body:

This truncates the reply text and generally looks rather stupid, but it's better than nothing. Note also that it does not currently exclude community posts (I didn't see a way to do so). I did test the code on a reply on a community post.

</details>

### Testing steps

To easily distinguish affected activity items, run `document.documentElement.append(Object.assign(document.createElement('style'), { textContent: '[aria-label="Notification"]:not([role="listitem"]) { outline: 1px solid red; outline-offset: -3px }' }))`.

Mutual Checker:
- Enable the "only show notifications from mutuals" option.
- Confirm that from-mutuals notifications of both the old and new types are visible.
- Confirm that other notifications of both the old and new types are hidden.

NotificationBlock:
- Find or generate a new-type notification about one of your posts; this will begin with "Replied to your post."
- Activate NotificationBlock on the linked post and confirm that the notification is hidden.

Quote Replies:
- Confirm that Quote Replies adds action buttons to "replied to your post", "replied to you in a post", and "mentioned you on a post" notifications. (I do this by enabling responsive design mode and emulating a mobile device, which will show the buttons without hovering them.)
- Confirm that Quote Replies does not add action buttons to "replied to you in a [community name] post" or "mentioned you in a comment in [community name]" notifications.
- Confirm that Quote Replies generates draft posts with the correct contents from all three supported notification types, from both the activity page and activity modal.
- Break the Quote Replies API fetch. Confirm that Quote Replies generates draft posts from "replied to your post" / "replied to you in a post" notifications, with degraded formatting. Confirm that "mentioned you on a post" notifications will instead show an error modal (this code path is unchanged and we didn't write a fallback into it).
- Confirm that the links in the generated draft posts all point to the correct accounts/urls.


src/features/quote_replies.js
```diff
@@ -14,9 +14,53 @@ const storageKey = 'quote_replies.draftLocation';
...
+  margin: 10px 0 0;
+}
+
+@media (hover: hover) {
+  button.xkit-quote-replies svg {
+    opacity: 0;
+    transform: scale(0);
+  }
+
+  ${notificationSelector}:is(:hover, :focus-within) button.xkit-quote-replies svg {
```
marcustyphoon, line 54: Probably should have done the switch to styleElement in two commits, now that I think about it. For reference, this should be the only changed line in the CSS.

marcustyphoon:
Hm... maybe this partial implementation of Quote Replies should include an informational modal each time it's used?

AprilSylph:
> This truncates the reply text and generally looks rather stupid, but it's better than nothing. Note also that it does not currently exclude community posts (I didn't see a way to do so). I did test the code on a reply on a community post.

brain dump of my current thinking about how to fetch specific replies:

1. blog identifier and post ID can be reliably parsed `notification.actions.tap.href`. if `new URL(temp1.__reactFiber$i3lhlxz35zo.return.return.memoizedProps.notification.actions.tap.href).pathname.split('/')[1] === 'communities'` then we know it's a communities URL, otherwise it's a blog URL. blog URLs need any leading `@` stripped to be used as an API blog identifier, communities handles need `@@` prepended to be used as an API blog identifier.
2. fetch replies on the target post using the above and the `notification.timestamp`
3. filter fetched replies to those with a matching timestamp
4. pick the first fetched reply that has a partial match for the `notification.body.content[1].text` (with the trailing ellipsis omitted)
5. if there is no such match (how can this happen? newlines? can we massage the fetched reply text in the same way as the tumblr API?), fall back to the notification indented text

also `notification.avatar.badgeImage` might be useful for splitting on different activity types at some point idk

marcustyphoon:
Wait, there is a timestamp... hm, why did I think there wasn't a timestamp? I have screenshots of the API data from when I was working on this and there absolutely is a timestamp. (Unless that's the post timestamp or something absurd, but why would that be the case).

Er, stay tuned.

src/features/quote_replies.js
```diff
@@ -55,12 +99,58 @@ const processNotifications = notifications => notifications.forEach(async notifi
...
+  } = notificationProps;
+  const summaryFormatting = bodyDescriptionContent.formatting.find(({ type }) => type === 'semantic_color');
+
+  try {
+    const [, targetTumblelogName, targetPostId] =
+      /^\/@?([a-z0-9-]{1,32})\/([0-9]{1,20})\//.exec(new URL(actions.tap.href).pathname);
+
+    const targetPostSummary = bodyDescriptionContent.text.slice(summaryFormatting.start + 1, summaryFormatting.end - 1);
+
+    return await processReply({ type, timestamp, targetPostId, targetTumblelogName, targetPostSummary });
```
marcustyphoon, line 118: Should this be two lines (`const { content, tags } = await processReply(/* */); return { content, tags };`)? I know some people instinctively recoil at "return await" but it's necessary here to trigger the synchronous catch block.

AprilSylph:
I'm gonna sit on this one this week too, but only because there's something cooking which should make reply fetching a _lot_ more reliable...

AprilSylph:
Important note for future reference: Mutual Checker and NotificationBlock could've been fixed months ago if their changes had been split into their own PRs.

Granted, all these updates rely on updated utils, but if you include the same util changes across multiple PRs, you open the possibility to first 1) get the util changes merged with the least effort from the reviewer (i.e. with the lowest-reviewer-effort feature update) and then 2) update all the PRs so that the util changes aren't included in feature updates, which further reduces reviewer effort.

AprilSylph (approved):
- [x] Mutual Checker
  - [x] Confirm that from-mutuals notifications of both the old and new types are visible.
  - [x] Confirm that other notifications of both the old and new types are hidden.
- [x] NotificationBlock
  - [x] Activate NotificationBlock on the linked post and confirm that the notification is hidden.
- [ ] Quote Replies
  - [x] Confirm that Quote Replies adds action buttons to "replied to your post", "replied to you in a post", and "mentioned you on a post" notifications.
  - [x] Confirm that Quote Replies does not add action buttons to "replied to you in a [community name] post" or "mentioned you in a comment in [community name]" notifications
  - [x] Confirm that Quote Replies generates draft posts with the correct contents from all three supported notification types, from both the activity page and activity modal
  - [x] Break the Quote Replies API fetch. Confirm that Quote Replies generates draft posts from "replied to your post" / "replied to you in a post" notifications, with degraded formatting.
    - [x] Confirm that "mentioned you on a post" notifications will instead show an error modal (this code path is unchanged and we didn't write a fallback into it).
  - [x] Confirm that the links in the generated draft posts all point to the correct accounts/urls.

marcustyphoon:
> Important note for future reference: Mutual Checker and NotificationBlock could've been fixed months ago if their changes had been split into their own PRs.
> 
> Granted, all these updates rely on updated utils, but if you include the same util changes across multiple PRs, you open the possibility to first 1) get the util changes merged with the least effort from the reviewer (i.e. with the lowest-reviewer-effort feature update) and then 2) update all the PRs so that the util changes aren't included in feature updates, which further reduces reviewer effort.

Oh, interesting—I've of course never been on the reviewer side in a meaningful capacity, but I had guessed the preference would be for the opposite! I'll keep that in mind for the future.

Issue #1655, opened 2024-11-29
Quote Replies error


symphonicdemise:
### Platform

Windows 10

### Browser

Google Chrome 131.0.6778.86 (Official Build) (64-bit)

### Addon version

1.0.5

### Details

When attempting to quote reply someone, it pulls up an error code "APIError [400.0] at https://www.tumblr.com/api/v2/blog/t:9gVdF5k4vDDVbASF1VSvkA/posts ()" Has anyone else come across this issue or is there a fix?

I have already attempted to turn off the "Quote Replies" option, restarted Google chrome and turned it back on and still got the error.

marcustyphoon:
Hmm. Does it work on a different reply on a different post? (You might have to reply to yourself to test this, but I believe that works.) Is there something unusual about the reply you're quoting, such as the user who replied being a deleted blog? Do you have an original post tag set in the Quick Tags feature?

lauraxjane:
hello! I'm also finding that though I've refreshed, toggled off/on, that I'm unable to even see the option to click in order to use quote replies. Now it's been awhile since I've used the feature, so I don't fully remember how it works, but I'm really not getting anything showing in order to do it, I'm wondering if it's something that could have been broken when tumblr implemented their new way of replying to replies within the replies? 
![image](https://github.com/user-attachments/assets/c10d6652-69e9-4e51-bf13-3e190cd94e11)
![image](https://github.com/user-attachments/assets/7ff87954-4535-419a-9cbd-8a8008d41506)
![image](https://github.com/user-attachments/assets/5e693ec6-52ac-4125-a593-75de074088e8)

marcustyphoon:
@lauraxjane This looks like a separate problem! I've made an issue to track that here: #1658

PR #1662, opened 2024-12-10
Quote Replies: Add action buttons to replies in post notes

marcustyphoon:
As a bit of an aside to the actual issue in question, https://github.com/AprilSylph/XKit-Rewritten/issues/1655#issuecomment-2518848059 also brought up that it's reasonable to expect Quote Replies to be usable from the replies themselves as well as the activity pane. It doesn't... but we can probably do that!

WIP.

### DIFF
--- a/src/features/quote_replies.js
+++ b/src/features/quote_replies.js
@@ -8,10 +8,12 @@ import { notify } from '../utils/notifications.js';
 import { getPreferences } from '../utils/preferences.js';
 import { buildSvg } from '../utils/remixicon.js';
 import { apiFetch, navigate } from '../utils/tumblr_helpers.js';
-import { userBlogs } from '../utils/user.js';
+import { userBlogNames, userBlogs } from '../utils/user.js';
+import { registerReplyMeatballItem, unregisterReplyMeatballItem } from '../utils/meatballs.js';
 
 const storageKey = 'quote_replies.draftLocation';
 const buttonClass = 'xkit-quote-replies';
+const meatballButtonId = 'quote_replies';
 const dropdownButtonClass = 'xkit-quote-replies-dropdown';
 
 const originalPostTagStorageKey = 'quick_tags.preferences.originalPostTag';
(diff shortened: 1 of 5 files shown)


marcustyphoon:
Man, I wish I had typescript and something similar to Zod. I mean, jesus.

marcustyphoon:
- [ ] #1710
- [ ] #1712

marcustyphoon:
q: why did you make a PR that would immediately completely explode when it's merged with another active PR?
a: I am very smart

marcustyphoon:
> q: why did you make a PR that would immediately completely explode when it's merged with another active PR?
> a: I am very smart

Yeah, doing that "merge" took me slightly over an hour. Fun times.

(The actual issue was that I—at the time, cleverly—reused the `verbiage`/`formatting`/`content` code that's now part of the `processReply` function when quoting from the post notes. It would be possible to factor things now so that's still the case, but the `processReply`/`processGenericReply` flow is now complicated enough to understand that I think it's best to leave that on its own and live with the `verbiage`/`formatting`/`content` code effectively getting copy-pasted. DRY is a nice principle but it's not the be-all and end-all.)

marcustyphoon:
Hm, this might be easier with the `/v2/blog/[blog]/post/[post id]/replies/[reply id]/permalink?mode=replies` endpoint used for replies permalinks.

marcustyphoon:
> Hm, this might be easier with the `/v2/blog/[blog]/post/[post id]/replies/[reply id]/permalink?mode=replies` endpoint used for replies permalinks.

yeah, I dunno, maybe... not.

src/features/quote_replies/index.js
```diff
@@ -163,36 +166,93 @@ const processReply = async ({ type, timestamp, targetPostId, targetTumblelogName
...
-    ? await processGenericReply(notificationProps)
-    : await processReply(notificationProps);
+    ? await createGenericActivityReplyPost(notificationProps)
+    : await createActivityReplyPost(notificationProps);
+
+  openQuoteReplyDraft(tumblelogName, content, tags);
+};
+
+const determineNoteReplyType = ({ noteProps, parentNoteProps }) => {
+  if (userBlogNames.includes(noteProps.note.blogName)) return false;
```
marcustyphoon, line 212: Interesting note from testing: this line excludes quote replying yourself, which probably makes sense in general... but a) maybe it doesn't (something something group blogs), and b) that's inconsistent with the behavior on activity items, which totally does let you quote reply yourself!

marcustyphoon:
- [ ] fix this on narrow posts (`notesPopover`), in which the meatballs menu is apparently... completely different???

marcustyphoon:
You know... putting this in meatballs doesn't necessarily make a ton of sense, does it. No one's going to find it there. And now there's a juicy space right next to the like button...

marcustyphoon:
- [ ] Handle trail footer items correctly

PR #1665, opened 2024-12-13
Migrate to neostandard and eslint@9

marcustyphoon:
### Description

On the logic that it's better to prepare in advance for any problems that could stem from one of your dependencies being... [in some state besides "definitely has active community support,"](https://github.com/standard/standard/issues/1948) this tries out the https://github.com/neostandard/neostandard project prerelease. The rather painful-to-read eslint 9 config file format aside—nothing one can do about that, really—it seems like a solid direction.

Branch may be rebased.


marcustyphoon:
See #1815.

PR #1672, opened 2024-12-25
Anti-Capitalism: Make blazed post hiding optional

marcustyphoon:
### Description

Blazed posts are somewhat different than third-party algorithmic advertising, and some users of this feature might want to continue to see them. ~~I'm not actually sure if they still show up with e.g. ublock origin lite enabled (it was hard enough just finding a blazed post to test this on at all), but it's not like one has to do that.~~ **edit:** _ah, managed to get one in testing and they do! nice._

Note that this doesn't apply to the kind of advertisement that appears sort of like a post with a large CTA button at the bottom (not sure if this is what Blaze Pro is). I get these from Tor Books; they do have post ids but do not have the orange blaze icon anywhere.

### Testing steps

- Find a blazed post on the dashboard. Enable Anti-Capitalism and confirm that the post is hidden only when the checkbox is disabled.
- Find a blazed post on the dashboard with the new post header flag enabled. Enable Anti-Capitalism and confirm that the post is hidden only when the checkbox is disabled.

### DIFF
--- a/src/features/anti_capitalism.js
+++ b/src/features/anti_capitalism.js
@@ -16,7 +16,9 @@ const processVideoCTAs = videoCTAs => videoCTAs
 
 export const main = async () => {
   const { includeBlazed } = await getPreferences('anti_capitalism');
-  const blazeFilter = includeBlazed ? '' : ':not(:has(header use[href="#managed-icon__badge-blaze"], header use[href="#managed-icon__ds-blaze-filled-16"]))';
+  const blazeFilter = includeBlazed
+    ? ''
+    : ':not(:has(header use[href="#managed-icon__ds-blaze-filled-16"]))';
 
   styleElement.textContent = `
     [${hiddenAttribute}] > div,
--- a/src/features/anti_capitalism.js
+++ b/src/features/anti_capitalism.js
@@ -6,12 +6,7 @@ const hiddenAttribute = 'data-anti-capitalism-hidden';
 
 const listTimelineObjectInnerSelector = keyToCss('listTimelineObjectInner');
 
-export const styleElement = buildStyle(`
-[${hiddenAttribute}] > div,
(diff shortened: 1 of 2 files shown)


src/features/anti_capitalism.js
```diff
@@ -1,28 +1,34 @@
...
+export const styleElement = buildStyle();
 
 const processVideoCTAs = videoCTAs => videoCTAs
   .map(getTimelineItemWrapper)
   .filter(Boolean)
   .forEach(timelineItem => timelineItem.setAttribute(hiddenAttribute, ''));
 
 export const main = async () => {
+  const { includeBlazed } = await getPreferences('anti_capitalism');
+  const blazeFilter = includeBlazed ? '' : ':not(:has(header use[href="#managed-icon__badge-blaze"], header use[href="#managed-icon__ds-blaze-filled-16"]))';
```
AprilSylph: We should be able to drop the old selector now, right?

```suggestion
  const blazeFilter = includeBlazed
    ? ''
    : ':not(:has(header use[href="#managed-icon__ds-blaze-filled-16"]))';
```

src/features/anti_capitalism.json
```diff
@@ -7,5 +7,12 @@
     "background_color": "#cd0000"
   },
   "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#anti-capitalism",
-  "relatedTerms": [ "sponsored" ]
+  "relatedTerms": [ "sponsored" ],
+  "preferences": {
+    "includeBlazed": {
+      "type": "checkbox",
+      "label": "Hide blazed posts",
```
AprilSylph: **Nit**: "Blaze" is a brand name

```suggestion
      "label": "Hide Blazed posts",
```

marcustyphoon:
- [x] still works

PR #1683, opened 2025-01-04
AccessKit Disable GIFs: Refactor async callbacks

marcustyphoon:
### Description

Disable GIFs currently sets some `onload` callbacks to wait for image element sources to be loaded. This replaces them with the [decode() HTMLImageElement instance method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode) that I found while looking through MDN for something different-but-related, removing some nesting. Conveniently, this doesn't require a conditional for whether the image is already loaded (try `$0.decode().then(() => console.log('done'))` on an already-loaded image).

"Hide whitespace" is recommended.

### Testing steps

- Confirm that Disable GIFs continues to work as expected.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -68,21 +68,23 @@ const addLabel = (element, inside = false) => {
   }
 };
 
-const pauseGif = function (gifElement) {
+const pauseGif = async function (gifElement) {
+  await gifElement.decode();
+
   const image = new Image();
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Annoyingly, this seems to just sometimes fail in Firefox in my testing.

marcustyphoon:
Ah, figured out why and noted it in an MDN issue: https://github.com/mdn/content/issues/38034

PR #1686, opened 2025-01-05
chore(deps): remove package-lock.json name property

marcustyphoon:
### Description

Due to https://github.com/npm/cli/issues/4608, a "name" property was added to the package-lock.json file by `npm audit fix` in 01fe042bac00fb5c6d1a3f0a4abd9c1ab45c6fe3 even though `npm install` removes it again.

This commits the diff caused by running `npm install`. Of course, if the linked issue gets resolved in the other direction (always creating a name property in package-lock.json based on the folder name if there is no name in package.json), this would be incorrect, but this is how it is at the moment. (Also, that wouldn't make much sense as a resolution, imho).

### Testing steps

n/a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "XKit-Rewritten",
       "license": "GPL-3.0",
       "dependencies": {
         "jquery": "^3.7.1",


AprilSylph (approved):
lol i thought it would just stay there. thanks npm

PR #1689, opened 2025-01-06
Refactor: Standardize capitalization of "onclick"

marcustyphoon:
### Description

This standardizes the capitalization of "onclick" to all-lowercase.

For some reason—I think a misguided notion that it was correct and the opposite was a mistake—I used "onClick" in #910, even though it's "onclick" in the utilities exported from the same file. This isn't used in any current feature, but it is used in #677, so that PR must be updated after merging this one.

### DIFF
--- a/src/features/scroll_to_bottom.js
+++ b/src/features/scroll_to_bottom.js
@@ -56,7 +56,7 @@ const stopScrolling = () => {
   scrollToBottomButton?.classList.remove(activeClass);
 };
 
-const onClick = () => active ? stopScrolling() : startScrolling();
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
> [!WARNING]
> Approving based on code review only, please make sure this is tested before merging!

yummy yum delicious inoculation against this ever becoming a React codebase

marcustyphoon:
- [ ] confirm Scroll to Bottom functionality (npm test)
- [ ] grep registerBlogMeatballItem; confirm no uses

PR #1693, opened 2025-01-09
Timeformat: Apply to new communities post header

marcustyphoon:
### Description

Applies timeformat to the timestamps in the new unique post header used in tumblr communities (`communitiesRedpopPostChrome`).

~~Not sure if this will break anything else, like mutual checker~~ yes it does.

Is this actually... desirable? I mean, I assume so, though the aesthetics aren't great.

### Testing steps

- Confirm that Timeformat works on community posts with the new header (feature flag enabled).

### DIFF
--- a/src/features/timeformat.js
+++ b/src/features/timeformat.js
@@ -28,7 +28,8 @@ export const styleElement = buildStyle(`
   cursor: help;
 }
 
-${keyToCss('blogLinkWrapper')}:has(+ [data-formatted-time]) {
+${keyToCss('blogLinkWrapper')}:has(+ [data-formatted-time]),
+${keyToCss('info')}:has(+ ${keyToCss('timestamp')} > [data-formatted-time]) {
   flex: none;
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Ah, this might become more relevant than I realized.

AprilSylph:
This is introducing some unnecessary height to the post header as-is... gonna see if I can fix it.

AprilSylph:
Flex items being their own display type is so weird, always throws me off.

I'm gonna merge this now, but just to let you know, I am aware of this fuckass behaviour and will fix it with a Tumblr-side change in the coming week:

Screen Shot 2025-03-02 at 20 10 15

(The separator needs to stop being on the `.timestamp::before` and instead be its own element inside the `.info` so that it disappears when `.info` is collapsed to a width of zero, but that also means it will stay at the end of the line when we allow wrapping via the CSS I just committed here.)

Issue #1069, opened 2023-05-12
Mutual Checker: Possible UI inconsistencies with third party blazed posts


marcustyphoon:
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 112.0.5615.137

### Addon version

v0.21.7

### Details

A post blazed by someone who is not the original poster's header is displayed as if it is a post created by the blazing user, not the author. The post data has `blazerBlog` and `blog` fields.

In this scenario, there is a mismatch between the `blog` field and the top-of-post UI; it may be worth investigating if this breaks mutual checker and if there are any other scripts with assumptions that this violates.

PR #1695, opened 2025-01-10
Mutual Checker: Apply to new communities post header

### LINKED ISSUE #1694: Mutual Checker: Icons are not added to community posts
### Platform

macOS 14.7.1 arm64

### Browser

Chrome 131.0.6778.205

### Addon version

v1.0.5

### Details

Mutual Checker icons are not added to user links in the new unique post header used in tumblr communities (`communitiesRedpopPostChrome`).

Technical notes: I don't really want to use `keyToCss('title')` to target these elements, since that results in 79 classes. (`header keyToCss('userBlock') keyToCss('headline')` is all well and good, but what does one do then?) `[rel="author"]` looks promising.

marcustyphoon:
### Description

It appears that it doesn't need to be any more complex than this!

Applies Mutual Checker to the timestamps in the new unique post header used in tumblr communities (`communitiesRedpopPostChrome`) by taking advantage of the `rel="author"` property. I see no reason why additional elements would be added with this property and I assume(?) it's fine to rely on the property not being forgotten/lost in the future?

Resolves #1694.

### Testing steps

- Confirm that Mutual Checker works on community posts with the new header (feature flag enabled).
- Confirm that Mutual Checker works on non-community posts.

I'd like to find a blazed-by-someone-else post to see what happens here (semi-related: #1069) but that's quite unlikely so I may have to just read the code. But, like... eh, who cares, also.

### DIFF
--- a/src/features/mutual_checker.js
+++ b/src/features/mutual_checker.js
@@ -12,7 +12,7 @@ import { followingTimelineSelector } from '../utils/timeline_id.js';
 const mutualIconClass = 'xkit-mutual-icon';
 const hiddenAttribute = 'data-mutual-checker-hidden';
 const mutualsClass = 'from-mutual';
-const postAttributionSelector = `header ${keyToCss('attribution')} a:not(${keyToCss('reblogAttribution', 'rebloggedFromName')} *)`;
+const postAttributionSelector = 'header a[rel="author"]';
 
 const onlyMutualsStyleElement = buildStyle(`${keyToCss('notification')}:not([data-mutuals]) { display: none !important; }`);
 


marcustyphoon:
> I'd like to find a blazed-by-someone-else post to see what happens here

Found one. They have two `rel="author"` elements, only the first of which is in the header element (the one corresponding to the blog that blazed the post, not the post author). I believe we currently don't look at any react data, so that's the one we'll process (i.e. if you enable "Only show posts from mutuals on the dashboard" and your mutual blazes someone else's post, you'll see the post; if someone else blazes your mutual's post you won't see it, would be my assumption).

Probably not the ideal behavior but the chance this happens is tiny and the consequence really doesn't matter, so I don't think it's worth further consideration.

AprilSylph (approved):
### :relieved:

PR #1703, opened 2025-01-15
Quick Reblog: Exit cleanly without user info

### LINKED ISSUE #1702: User util: Intentionally fail loading when users endpoint fails
### Concept

Due to Tumblr server issues over the past few days, some users have been encountering a scenario where Quick Reblog doesn't have their own blog, but does have communities, causing accidental reblogs into communities and general confusion. We could improve this somewhat by having the utility return nothing if the users endpoint dies, as there isn't likely to ever be a scenario where a feature depends on the util but would be happy with only getting communities.

marcustyphoon:
### Description

Quick Reblog currently:
- Logs a mysterious-looking error to the console if and only if the user is logged out or the user info api fetch fails and the "Remember the last selected blog in the popup" option is enabled.
- Works, but only shows communities as reblog targets, if and only if the user info api fetch fails, the communities info api fetch doesn't fail, **and** the "Remember the last selected blog in the popup" option is disabled.

None of this really makes any sense as error-handling behavior. This PR makes the extension exit cleanly (though we could add a log here) if there is no user info, thus making it so that Quick Reblog:
- Does nothing if the user is logged out.
- Does nothing if the user info api fetch fails.
- Never only shows communities as reblog targets.
- Works, but doesn't show communities as reblog targets, if and only if the user info api fetch works but the communities info api fetch fails (say, if Staff changes that endpoint as part of communities development).

Resolves #1702, though not in the way proposed by the issue text.

### Testing steps

- Confirm that Quick Reblog appears and shows both user blogs and communities as reblog targets.
- Break the user info fetch. Confirm that Quick Reblog does not appear and does not log an error.
- Break the communities info fetch. Confirm that Quick Reblog appears and shows only user blogs reblog targets.
- Log out. Confirm that Quick Reblog does not appear and does not log an error.

### DIFF
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -316,6 +316,7 @@ const preventLongPressMenu = ({ originalEvent: event }) => {
 };
 
 export const main = async function () {
+  if (!userBlogs.find(({ primary }) => primary === true)) return;
   ({
     popupPosition,
     showBlogSelector,
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -2,7 +2,7 @@ import { sha256 } from '../utils/crypto.js';
 import { timelineObject } from '../utils/react_props.js';
 import { apiFetch } from '../utils/tumblr_helpers.js';
 import { postSelector, filterPostElements, postType, appendWithoutOverflow } from '../utils/interface.js';
-import { joinedCommunities, joinedCommunityUuids, userBlogs } from '../utils/user.js';
+import { joinedCommunities, joinedCommunityUuids, primaryBlog, userBlogs } from '../utils/user.js';
 import { getPreferences } from '../utils/preferences.js';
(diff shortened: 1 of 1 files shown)


src/features/quick_reblog.js
```diff
@@ -316,6 +316,7 @@ const preventLongPressMenu = ({ originalEvent: event }) => {
 };
 
 export const main = async function () {
+  if (!userBlogs.find(({ primary }) => primary === true)) return;
```
AprilSylph, line 319: Let's DRY this

https://github.com/AprilSylph/XKit-Rewritten/blob/317a3de992b53c323808fb8296daadccd9a253b2/src/utils/user.js#L32-L35

```suggestion
  if (!primaryBlog) return;

```

marcustyphoon: Oh yeah! Totally forgot we had that.

AprilSylph (approved):

PR #1707, opened 2025-01-22
AccessKit Disable GIFs: Hide animated images instead of covering them

marcustyphoon:
### Description

Right now, Disable GIFs hides most animated images when they aren't hovered by covering them with a non-animated one, and handles transparency by setting an opaque "white" background on the cover so the real, animated image can't be seen through transparent parts. This works great as long as the background is actually supposed to be "white" (which it is, in posts, even with Themed Posts). It doesn't work if one wants to apply that part of the Disable GIFs to other parts of the Tumblr UI which may have different background colors (in this case, the "recommended" element at the top of the Tumblr TV interface), and also might use a bit more CPU than the alternative.

What's the alternative? We simply hide the animated image when it's not hovered, just like we hide the cover when it is hovered. Thus, only one image is ever displayed at a time, and there's no need to implement anything special to make transparent images work.

~~This also means that we no longer care about the layer order at all, and in fact it's slightly more elegant due to the way CSS selectors work to put our canvas element before (i.e. what would have been "behind") the native animated image, similar to how Vanilla Videos inserts its replacement element before the native element and hides the native element with a `~` selector.~~ _removed for consistency with possible future development_

(Aside: #1458, which I've done a reasonable amount of testing on, already uses the main technique used here. It will merge conflict with this, but it needs some refactoring anyway, hence its draft status.)

### Testing steps

- Confirm that Disable GIFs functions correctly on transparent GIFs, e.g. https://www.tumblr.com/animatedtext.
- Confirm that Disable GIFs functions in Firefox and Chromium.
- Confirm that there's no unusual delay/animation when moving the cursor onto or off of a paused GIF; there should be a seamless transition between the paused and animated images.
- Open https://www.tumblr.com/search/gif in masonry view and confirm that small GIFs, such as the related tags in the sidebar and side-by-side GIFs in photosets, have small GIF labels.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -81,7 +81,7 @@ const pauseGif = async function (gifElement) {
       canvas.classList.add(canvasClass);
       canvas.getContext('2d').drawImage(image, 0, 0);
       gifElement.parentNode.append(canvas);
-      addLabel(gifElement);
+      addLabel(canvas);
     }
   };
 };
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -80,8 +80,8 @@ const pauseGif = async function (gifElement) {
       canvas.className = gifElement.className;
       canvas.classList.add(canvasClass);
       canvas.getContext('2d').drawImage(image, 0, 0);
-      addLabel(gifElement);
       gifElement.before(canvas);
+      addLabel(gifElement);
     }
   };
 };
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -36,8 +36,6 @@ export const styleElement = buildStyle(`
 .${canvasClass} {
   position: absolute;
   visibility: visible;
-
-  background-color: rgb(var(--white));
 }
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Ooh, fascinating!

In Firefox, this causes a white flash when one hovers over an image for the first time in some circumstances, which I had always wondered about when developing #1458. Turns out there's an explanation: Firefox uses a default value of "async" for images' [decoding property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding#async), which means that if we hide the real image element and do so quickly enough, the browser will download it but will not decode it until it becomes visible again.

This can be fixed completely cleanly by `gifElement.decoding = 'sync'`... or, to avoid making a DOM mutation, by just calling `gifElement.decode()` to instruct the browser to begin decoding!

Note that:
- As I noted in https://github.com/mdn/content/issues/38034, `gifElement.decode()` (currently?) fails in Firefox on `loading="lazy"` images that haven't loaded yet, so we need to fire it after waiting for `gifElement.complete`/`gifElement.onload`
- `gifElement.decode()` returns a promise, but in this case we have no reason to actually wait for it; waiting for it is a regression here.

PR #1708, opened 2025-01-22
AccessKit Disable GIFs: Use content replacement

marcustyphoon:
### Description

As a proof of concept, this modifies Disable GIFs by entirely replacing the technique of inserting canvas elements into the DOM with a new technique: overriding the content of non-hovered image elements using the [css content property](https://developer.mozilla.org/en-US/docs/Web/CSS/content). This also contains various commits from other Disable GIFs PRs.

Because this version of Disable GIFs is implemented entirely in CSS—besides the labels; maybe they could be pseudos, but you can't pseudo an `` so probably not—this removes basically all restrictions on what GIF elements we can hit. Animated blog headers are now fair game, as are the thumbnail images in the search dropdown. DOM structure and layout don't matter.

The drawbacks of doing this are mainly that its performance can be pretty bad, e.g. when navigating to the tagged page for "gif" (it's a lot of blob URLs to create at once) and that I've observed some weird flickering when moving the cursor off of an affected GIF in some cases. I would not actually merge this as-is, though; we have a much better method for pausing GIFs in posts, which is to use Tumblr's native poster element. That has far better performance than either our current method or the one used here. The technique in this PR is best for niche GIFs throughout the UI that don't have posters (setting `content` if they are true image elements and overriding `background-image` when they are not), and these are generally few in number and/or can't be paused by any other method anyway.


marcustyphoon:
Included in the mentioning PR.

PR #1563, opened 2024-08-09
AccessKit: Apply Disable GIFs to all WebP images

marcustyphoon:
### Description

Still not sure if this is a good idea (in this state). There are, apparently, a fair number of non-animated WebP images.

This
- Applies Disable GIFs to WebP images (including non-animated ones; it is [not particularly easy](https://github.com/AprilSylph/XKit-Rewritten/pull/1402) to [distinguish the two](https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon:XKit-Rewritten:webp-riffchunks?expand=1)).
- Makes the label accurately indicate the source image type, so at least when it false positives on a non-animated image the user knows why. 

Resolves #1401.

### Testing steps

Animated WebP: https://www.tumblr.com/changes/758172275271991296/threaded-replies-are-here
Animated GIF: https://www.tumblr.com/@radar/757906400962920448, https://www.tumblr.com/april-codes/700206746833240064
Non-animated GIF: https://www.tumblr.com/@radar/753920129042153472
Non-animated WebP: https://www.tumblr.com/minmos/744793127262601216/front-facing-egrets-making-me-lose-control-of, https://www.tumblr.com/@radar/755279084134744066, https://www.tumblr.com/@radar/750568043810340864

PR #1709, opened 2025-01-23
AccessKit Disable GIFs: Parse; disable animated WebP images

### LINKED ISSUE #1401: AccessKit: Pause wepb
### Concept

Same as the Pause GIFs option, but applicable to animated webp.

marcustyphoon:
### Description

This applies Disable GIFs to animated WebP images—but not, if possible, to non-animated WebP images—by downloading them and parsing them with the new and not-amazingly-documented [WebCodecs API](https://developer.mozilla.org/en-US/docs/Web/API/WebCodecs_API).

Supersedes #1402 (which vendored a library to do this) and #1563 (which put labels on webp images regardless of whether they were animated or not).

Resolves #1401.

### Technical notes

Unlike the "create an image element, set its source, and read its data once it loads" method of obtaining image data used by Disable GIFs currently (and used in #1681), the fetch call used here doesn't seem to reuse the browser image cache, and I didn't manage to change this using cache headers or anything. So I think that checking whether an image is animated using this method has a server load cost as well as a time cost. This therefore intentionally and potentially incorrectly assumes that any gif (or gifv) image is animated and skips the check, only running it on webp images, which are currently pretty rare.

`ImageDecoder` is feature detected; when tested in Firefox 128, this falls back to trying to de-animate all webp images (or  to ignoring webp images if `isAnimatedDefault = false` is set; this is kind of arbitrary).

I also thought about processing images as if they're animated while simultaneously checking whether they're animated, waiting for both to finish, then removing the processing if the image was actually stationary. This would reduce processing latency (the amount of time the user sees the animated image) compared to the current sequential async code paths. That's a bit complicated, though.

### Testing steps

Animated WebP: https://www.tumblr.com/changes/758172275271991296/threaded-replies-are-here
Animated GIF: https://www.tumblr.com/@radar/757906400962920448, https://www.tumblr.com/april-codes/700206746833240064
Non-animated GIF: https://www.tumblr.com/@radar/753920129042153472
Non-animated WebP: https://www.tumblr.com/minmos/744793127262601216/front-facing-egrets-making-me-lose-control-of, https://www.tumblr.com/@radar/755279084134744066, https://www.tumblr.com/@radar/750568043810340864

Confirm that:
- animated WebP images in posts are paused
- non-animated WebP images in posts are not paused
- animated GIF images in posts are paused
- non-animated GIF images in posts... are still incorrectly paused, as it would be too expensive to check every GIF

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -89,9 +89,13 @@ const isAnimated = memoize(async (sourceUrl) => {
     data: response.body,
     preferAnimation: true
   });
-  await decoder.decode();
-  controller.abort();
-  return decoder.tracks.selectedTrack.animated;
+  try {
+    await decoder.decode();
+    return decoder.tracks.selectedTrack.animated;
+  } finally {
+    decoder.close();
+    controller.abort();
+  }
 });
 
 const pauseGif = async function (gifElement) {
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -93,10 +93,7 @@ const isAnimated = memoize(async (sourceUrl) => {
 });
 
 const pauseGif = async function (gifElement) {
-  if (!await isAnimated(gifElement.currentSrc)) {
-    console.log("this webp image isn't animated!", gifElement);
-    return;
-  }
+  if (!await isAnimated(gifElement.currentSrc)) return;
   const image = new Image();
   image.src = gifElement.currentSrc;
   image.onload = () => {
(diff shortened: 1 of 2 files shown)


marcustyphoon:
As a fun optimization, this streams the fetch response into the decoder and aborts the fetch as soon as the browser has enough information to know if it's animated or not. (This only downloads like 10% of the file, according to devtools on whatever random image I tested!)

Unfortunately, in Firefox this results in a console error (`Uncaught (in promise) DOMException: Closed decoder`) being logged that as far as I can tell can't be caught. Not sure if this is supposed to happen—yes, I checked [the spec](https://w3c.github.io/webcodecs/#imagedecoder-interface)—but it's pretty annoying.

marcustyphoon:
Included in the mentioning PR.

PR #1711, opened 2025-01-23
Vanilla Video: Add toggle to enable/disable in Tumblr TV

marcustyphoon:
### Description

Vanilla Video prevents video autoplay, which kind of goes against the idea of the Tumblr TV thingy. This adds a toggle to its preferences so that the user can disable it specifically there if they want to. I set it to be enabled by default, because that's probably what most XKit users would expect?

### Testing steps

- Confirm that the toggle turns Vanilla Video on and off inside the Tumblr TV interface.

### DIFF
--- a/src/features/vanilla_video.js
+++ b/src/features/vanilla_video.js
@@ -5,6 +5,7 @@ import { keyToCss } from '../utils/css_map.js';
 const vanillaVideoClass = 'xkit-vanilla-video-player';
 
 let defaultVolume;
+let tumblrTvEnable;
 
 const cloneVideoElements = videoElements => videoElements.forEach(videoElement => {
   if (videoElement.previousElementSibling?.classList.contains(vanillaVideoClass)) return;
@@ -32,8 +33,14 @@ const cloneVideoElements = videoElements => videoElements.forEach(videoElement =
 
 export const onStorageChanged = async function (changes, areaName) {
   const {
+    'vanilla_video.preferences.tumblrTvEnable': tumblrTvEnableChanges,
     'vanilla_video.preferences.defaultVolume': defaultVolumeChanges
   } = changes;
+
+  if (tumblrTvEnableChanges && tumblrTvEnableChanges.oldValue !== undefined) {
+    clean().then(main);
+    return;
+  }
 
   if (defaultVolumeChanges && defaultVolumeChanges.oldValue !== undefined) {
     ({ newValue: defaultVolume } = defaultVolumeChanges);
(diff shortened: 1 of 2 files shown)


src/features/vanilla_video/index.js
```diff
@@ -33,17 +34,25 @@ const cloneVideoElements = videoElements => videoElements.forEach(videoElement =
 
 export const onStorageChanged = async function (changes) {
   const {
+    'vanilla_video.preferences.tumblrTvEnable': tumblrTvEnableChanges,
     'vanilla_video.preferences.defaultVolume': defaultVolumeChanges
```
AprilSylph: **nit**: prefer destructuring preferences in the same order that they're defined

```suggestion
    'vanilla_video.preferences.defaultVolume': defaultVolumeChanges,
    'vanilla_video.preferences.tumblrTvEnable': tumblrTvEnableChanges,
```

src/features/vanilla_video/feature.json
```diff
@@ -20,6 +20,11 @@
         { "value": "100", "label": "100%" }
       ],
       "default": "100"
+    },
+    "tumblrTvEnable": {
+      "type": "checkbox",
+      "label": "Enable on Tumblr TV",
+      "default": true
```
AprilSylph: From my very brief time using Tumblr TV just now, videos seem to be rare enough that I think I as a user would prefer the illusion of them just being high-quality moving images.

Tumblr TV is basically an attempt at cloning TikTok or Instagram Reels<sup>[?]</sup> anyway, so IMO the autoplay is more expected there than in regular posts.

Also, given that preferences are shown when you enable a feature, I feel that it's more important to consider if existing users of Vanilla Video would want this option to enable itself without their input. I haven't seen anyone else note that it currently doesn't work in Tumblr TV, which really does make this more of a feature than a bugfix, and features generally should not enable themselves.

(This is in contrast to the reblog-buttons-are-links option to Classic Footer, where the increased number of clicks to open the reblog editor was one of the primary complaints I saw about the new post footer, and where all communication regarding Classic Footer's creation explicitly deemed the feature incomplete without that option.)

```suggestion
      "default": false
```

marcustyphoon: > I haven't seen anyone else note that it currently doesn't work in Tumblr TV

Are you saying that... hold on.

Are you saying that, though it currently does work on Tumblr TV, no one has reported this, so it's as if it didn't, and thus—

Er, sorry, could you clarify this statement/paragraph? :D 

(This inquiry aside, I'm down for changing the default.)

AprilSylph: ...I might be stupid. In my defense, I have less blood in me today.

The rest of what I wrote stands. I will attribute the lack of anyone noting any interactions between Vanilla Video and Tumblr TV to abysmal adoption of Tumblr TV.

marcustyphoon: Gotcha. So, essentially, we can ignore the prior behavior, the new toggle is a feature, it makes sense for the new feature to be off by default. Yep, agreed!

PR #1719, opened 2025-02-06
AccessKit Disable GIFs: Speed up pausing using "poster" source

marcustyphoon:
### Description

Without significant architectural changes, this improves the speed at which AccessKit Disable GIFs pauses images, typically making it faster than GIF loading itself and thus making the pausing effect appear "instant," by fetching the smaller already-paused poster source instead of the larger animated source.

This gracefully falls back to the current behavior when Tumblr doesn't provide a poster (ex: animated webp images).

This technique is generally compatible with my other Disable GIFs PRs; most (including all non-draft ones) should merge cleanly, one requires implementation of this technique by hand, and one would entirely remove all of the code that this modifies, making it irrelevant.

### Testing steps

- Confirm that Disable GIFs functions correctly.
- Confirm that, after the initial page load, Disable GIFs generally pauses GIFs as they appear on the page without a period during which the GIF is animated.
- Confirm that hovering a GIF that Disable GIFs has paused, but which still has the loader element in the top right, hides the GIF label and doesn't cause any erroneous visuals (image disappearing, flickering). Network throttling can help test this.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -68,9 +68,9 @@ const addLabel = (element, inside = false) => {
   }
 };
 
-const pauseGif = function (gifElement) {
+const pauseGif = function (gifElement, src) {
   const image = new Image();
-  image.src = gifElement.currentSrc;
+  image.src = src;
   image.onload = () => {
     if (gifElement.parentNode && gifElement.parentNode.querySelector(`.${canvasClass}`) === null) {
       const canvas = document.createElement('canvas');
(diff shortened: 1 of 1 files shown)


marcustyphoon:
...Huh. Apparently the code without this PR generally doesn't have the brief bit of animation slipping in either; I don't remember this being the case and I'm not sure how that would work (there should have to be a delay while we wait for the `Image` we create to load in which the animated image should be visible, unless Tumblr is delaying hiding the posters). Maybe it's related to having a fast device, I dunno.

Anyway, this is still quicker and more efficient, but that puts it more in "refactor" territory than "feature" territory. Speeding up pausing a GIF is only a feature when it wouldn't have been paused during that time period anyway, which apparently it sometimes is. (It's not on initial page loads, though.)

marcustyphoon:
I have planned PRs that will remove the code this affects, so closing.

PR #1721, opened 2025-02-08
AccessKit Disable GIFs: Blur animated images while processing

marcustyphoon:
### Description

Fun bit of CSS experimentation. This applies a temporary blur effect to images with GIF backgrounds that will be animated only for a moment while we're processing them. Like #1128 and #960, care is taken not to apply the effect when the extension is first loaded or the feature is switched on to avoid an unblurred -> blurred -> unblurred chain. Not sure this is really any better than not doing it, but it's cute.

Have you heard of the [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) CSS property? I sure hadn't.

Branch based off of #1729.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "webextension-polyfill": "^0.12.0"
       },
       "devDependencies": {
+        "@types/dom-webcodecs": "^0.1.13",
         "chrome-webstore-upload-cli": "^3.3.1",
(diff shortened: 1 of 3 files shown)


marcustyphoon:
This was supposed to be a just for fun thing, but, like... I **really** like it actually? :D But I think CSS blurs can perform really badly on older hardware, so my guess is it's too risky.

PR #1725, opened 2025-02-12
No Recommended: Fix blog/community hiding with Tumblr Savior enabled

### LINKED ISSUE #1724: No Recommended: Hiding blogs/communities fails with Tumblr Savior enabled
### Platform

MacOS 14.7.1

### Browser

Chrome 132.0.6834.160

### Addon version

v1.0.6

### Details

No Recommended's "Hide recommended communities between posts" and "Hide recommended blogs between posts" features can fail if another extension also tries to hide part of the element when endless scrolling is enabled. Tumblr Savior has a setting that's on by default that causes this, which is likely the most common cause.

This is, to be clear, an XKit Rewritten bug, **not** a bug with any extension that can trigger this!

The sequence of events (I assume) is:

- Another extension hides the title element of the carousel (using CSS, the correct, should-be-harmless-and-compatible-with-other-extensions way!)
- Tumblr's "virtual scroller" code realizes that the title element is not on the screen and removes it from the page
- XKit Rewritten's code runs, doesn't find the element it's expecting, and does nothing

https://github.com/AprilSylph/XKit-Rewritten/blob/529a998ba27df4cc9b4aeb3a7c467545118deda6/src/features/no_recommended/hide_blog_carousels.js#L21-L24

Shoutout to the users in the New XKit discord server who kept testing and providing screenshots until I tracked this down.

marcustyphoon:
### Description

This fixes the "Hide recommended communities between posts" and "Hide recommended blogs between posts" features of No Recommended when certain other extension features are enabled, such as Tumblr Savior's "Hide timeline objects" setting, and endless scrolling is enabled.

It may improve the reliability of both features with endless scrolling in general.

Resolves #1724.

### Technical details

As noted in the linked issue, Tumblr's virtual scroller may remove the "titleObject" element from the DOM before our code processes the carousel element. We thus need to not rely on its presence.

### Testing steps

These features are a pain to test because the elements they hide don't always show up and will show up less if you repeatedly view a timeline they're allowed to appear on. The For You tab seems to have some carousels a bit more reliably.

- Confirm that blog and community carousels are hidden by the relevant features with endless scrolling on.
- Confirm that blog and community carousels are hidden by the relevant features with endless scrolling off.
- Enable endless scrolling, apply the CSS `.rZlUD[data-cell-id*="timelineObject"] {display:none!important;}` or install Tumblr Savior and enable its rule which does so, and confirm that blog and community carousels are hidden by the relevant features.

### DIFF
--- a/src/features/no_recommended/hide_blog_carousels.js
+++ b/src/features/no_recommended/hide_blog_carousels.js
@@ -18,8 +18,11 @@ const hideBlogCarousels = carousels =>
     const { elements } = await timelineObject(carousel);
     if (elements.some(({ objectType }) => objectType === 'blog')) {
       const timelineItem = getTimelineItemWrapper(carousel);
-      if (timelineItem.previousElementSibling.querySelector(keyToCss('titleObject'))) {
-        timelineItem.setAttribute(hiddenAttribute, '');
+      timelineItem.setAttribute(hiddenAttribute, '');
+      if (
+        timelineItem.previousElementSibling.querySelector(keyToCss('titleObject')) ||
+        timelineItem.previousElementSibling.dataset.cellId?.startsWith('timelineObject:title')
+      ) {
         timelineItem.previousElementSibling.setAttribute(hiddenAttribute, '');
       }
     }
--- a/src/features/no_recommended/hide_blog_carousels.js
+++ b/src/features/no_recommended/hide_blog_carousels.js
@@ -18,11 +18,11 @@ const hideBlogCarousels = carousels =>
     const { elements } = await timelineObject(carousel);
     if (elements.some(({ objectType }) => objectType === 'blog')) {
       const timelineItem = getTimelineItemWrapper(carousel);
-      timelineItem.setAttribute(hiddenAttribute, '');
       if (
         timelineItem.previousElementSibling.querySelector(keyToCss('titleObject')) ||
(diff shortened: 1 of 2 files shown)


src/features/no_recommended/hide_blog_carousels.js
```diff
@@ -18,8 +18,11 @@ const hideBlogCarousels = carousels =>
     const { elements } = await timelineObject(carousel);
     if (elements.some(({ objectType }) => objectType === 'blog')) {
       const timelineItem = getTimelineItemWrapper(carousel);
-      if (timelineItem.previousElementSibling.querySelector(keyToCss('titleObject'))) {
-        timelineItem.setAttribute(hiddenAttribute, '');
```
AprilSylph, line 22: didn't we have the timeline-hiding code under this preceded-by-title gate so that the management carousel at the top of the blog subs page/tab wouldn't get hidden?

marcustyphoon: Oh, hm. Yeah, probably. That element no longer exists, but once reminded of its existence I'm more inclined to make the title element required, yeah.

AprilSylph (approved):

src/features/no_recommended/hide_blog_carousels.js
marcustyphoon: > That element no longer exists

For posterity: it doesn't there, but does on https://www.tumblr.com/dashboard/crushes.

PR #1706, opened 2025-01-22
AccessKit Disable GIFs: Apply to Tumblr TV landing page

marcustyphoon:
### Description

I was going to say this is kind of pointless—if you're going to Tumblr TV you want to see animations, right?—but actually, no, there's a big difference between seeing one piece of video content you clicked on and having many things constantly animating as you navigate.

Based on #1681.

**Edit:** Currently causes annoying white backgrounds on transparent gifs in the "recommended" element; I guess I'll finally bother to make [a PR](https://github.com/AprilSylph/XKit-Rewritten/pull/1707) to fix this. A bit difficult to test, as the vast majority of recommendations aren't transparent.

PR #1682, opened 2025-01-04
AccessKit Disable GIFs: Apply to communities browse page/logged-out page

marcustyphoon:
### Description

This applies AccessKit Disable Gifs to https://www.tumblr.com/communities/browse... which is currently the page that's shown at https://www.tumblr.com/ before you log in to Tumblr.

Relies on/contains the commit from #1681.

PR #1729, opened 2025-02-17
AccessKit Disable GIFs: Pause ALL the GIFs!

marcustyphoon:
### Description

Branch based off of #1727.

Supersedes #1682, #1706, and #1708.


marcustyphoon:
https://www.tumblr.com/dashboard/tumblr_tv cards apparently change their background image each time you navigate within them; React changes only the `background-image` style property, so our paused images go out of sync with the on-hover images/sometimes we don't pause an image because it wasn't originally animated.

- [ ] fix this (or just don't process the Tumblr TV page)

PR #1733, opened 2025-02-19
 Tweaks: Apply "hide mini-follow buttons" to new post header

marcustyphoon:
### Description

This updates the "hide mini-follow buttons" tweak so the feature will apply to the new post header that's used in community posts if it's ever used for regular ones. (Okay, they're not really "mini" buttons anymore.)

There's no distinguishing CSS for follow buttons, so I made really sure we don't hit the wrong ones by checking the post data.

At time of writing, this code is for a UI element that doesn't exist.

### DIFF
--- a/src/features/tweaks/hide_mini_follow.js
+++ b/src/features/tweaks/hide_mini_follow.js
@@ -1,6 +1,24 @@
 import { keyToCss } from '../../utils/css_map.js';
-import { buildStyle } from '../../utils/interface.js';
+import { buildStyle, postSelector } from '../../utils/interface.js';
+import { pageModifications } from '../../utils/mutations.js';
+import { timelineObject } from '../../utils/react_props.js';
+
+const hiddenAttribute = 'data-tweaks-hide-mini-follow-hidden';
 
 export const styleElement = buildStyle(`
-article ${keyToCss('followButton')}:not(${keyToCss('postMeatballsContainer')} *) { display: none; }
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Probably won't be able to test this this weekend, but if I recall correctly it's done and works?

AprilSylph:
To be honest I think we should retire this tweak when the new UI launches. The main reason they're so egregious in the current UI (in my eyes) is their repetition, being present for every possible entity... with the move to a maximum of one per post, that problem is (FINALLY) solved in Tumblr proper.

marcustyphoon:
I think that's a fair argument, and also one I... well, okay, I'm not going to say I disagree necessarily :D I haven't been using the new UI regularly, so all I have is first impressions, but those new buttons sure are big.

AprilSylph:
We can keep this PR open and revisit it after launch. Whether or not we get bug reports / feature requests to please hide the big follow buttons will be interesting to see...

marcustyphoon:
- [ ] handle trail items

marcustyphoon:
Uuuuuugh fuck it I'm probably just going to replace the timelineObject processing in this with the translate utility. It's too complicated to handle all of the possible edge cases with how trail items could plausibly use internal data when all we want is "hide a button that says 'follow.'"

I would feel more comfortable doing this if we had a more exhaustive version of the translation dictionary; French has multiple entries for "follow" (google translate translates one of them to "subscribe").

marcustyphoon:
For the sake of making it official, my opinion on the follow buttons if the current code is the final code is "but why though."

PR #1734, opened 2025-02-21
AccessKit Disable GIFs: Use pseudo element for labels

marcustyphoon:
### Description

Okay, now it's **really** all CSS.

This replaces the `<p>` GIF label element with a data attribute and CSS that sets `::after`. 

Branch based off of #1729 for ease of comprehensive testing. I'll probably debate this for a bit and then potentially merge it into other PRs.


marcustyphoon:
As it turns out, yes: there are, that I can tell, a single digit number of ::after elements in redpop... and yet one of them, a post with a poll and an image that's in the blog view sidebar recommended posts section ([e.g.](https://www.tumblr.com/excitementshewrote/757829278953062400/best-marge-simpson-quote-i-thought-googling)) is indeed on an element with a GIF background and thus breaks with this.

Unbelievable.

marcustyphoon:
On the other hand, the `signpostCta` elements currently atop communities has css adjusting `<p>` elements within them, which messes up the non-pseudo label we're currently using if you enable Disable GIFs on it. Ups and downs, then.

PR #1735, opened 2025-02-23
Themed Posts: Adjust identical background/foreground colors.

marcustyphoon:
### Description

I am immeasurably pissed that this doesn't actually work. You should have heard me when I figured out that you could do it.

On supporting browser versions, this makes the text/accent colors used by Themed Posts 10% lighter or darker if they exactly match the background depending on whether they're light or dark, with the "is it light or dark; go the opposite way" logic implemented in pure CSS. Without [sign()](https://developer.mozilla.org/en-US/docs/Web/CSS/sign).

This doesn't work because Tumblr uses the legacy `rgba(var(--someVariable), 0.5)` syntax instead of `rgba(var(--someVariable) / 0.5)` and you can't mix and match modern and legacy syntax.

### DIFF
--- a/src/features/themed_posts.js
+++ b/src/features/themed_posts.js
@@ -51,9 +51,24 @@ const processPosts = async function (postElements) {
           linkColor
         } = theme;
 
+        const hexToRGBAdjusted = color => {
+          if (color === backgroundColor) {
+            const cssSign = val => `(sqrt(pow(${val}, 2)) / ${val})`;
+
+            const isDarkThreshold = 0.5;
+            const negativeIfDark = cssSign(`(l - ${isDarkThreshold})`);
+
+            const adjustmentAmount = 0.1;
+            const newColor = `oklch(from ${color} calc(l - ${adjustmentAmount} * ${negativeIfDark}) c h)`;
+            const adjusted = `from ${newColor} r g b`;
+            if (CSS.supports('color', `rgb(${adjusted})`)) return adjusted;
+          }
+          return hexToRGB(color);
+        };
+
         const backgroundColorRGB = hexToRGB(backgroundColor);
-        const titleColorRGB = hexToRGB(titleColor);
-        const linkColorRGB = hexToRGB(linkColor);
+        const titleColorRGB = hexToRGBAdjusted(titleColor);
+        const linkColorRGB = hexToRGBAdjusted(linkColor);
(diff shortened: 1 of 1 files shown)


marcustyphoon:
- It looks like Tumblr's adjustments are probably in HSL. This results in quite a bit more saturation (at least on the color used in the blog referenced in the linked issue), and actually makes a really nice accent color in that case. I avoided it both for "oklch is technically superior" reasons and as, per https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl, there's a bug in certain chromium versions that would mess up relative `hsl()` colors.
- I previewed the effect using this code, which lets you use the dev tools color picker on `--background-color` to quickly see different backgrounds and adjusted foregrounds on them.

```js
const hexToRGBAdjusted = color => {
  const cssSign = val => `clamp(-1, ${val} * 100000, 1)`;

  const isDarkThreshold = 0.4;
  const direction = cssSign(`(${isDarkThreshold} - l)`);

  const adjustmentAmount = 0.18;
  const newColor = `oklch(from ${color} calc(l + ${adjustmentAmount} * ${direction}) c h)`;
  const adjusted = `from ${newColor} r g b`;
  return adjusted;
};

const backgroundColorRGB = hexToRGB(backgroundColor);
const titleColorRGB = hexToRGB(titleColor);
const linkColorRGB = hexToRGB(linkColor);

styleElement.textContent += `
  [data-xkit-themed="${name}"] {
    --background-color: ${backgroundColor};
    --white: from var(--background-color) r g b;
    --black: ${hexToRGBAdjusted('rgb(var(--white))')};
    --deprecated-accent: ${linkColorRGB};
    --color-primary-link: rgb(var(--deprecated-accent));
  }
`;
```

Also, mdn seems to imply that `CSSStyleDeclaration.getPropertyValue()` may be able to read the computed values of relative color variables, in which case maybe we can round-trip back to rgb numbers? That would be kind of absurd ("the browser doesn't ship native js color manipulation constructs, but you can make a utility function that does it by constructing and reading back values off a DOM element"), but, sure. Noting this because I might come back to this later and not have remembered that I want to try this.

marcustyphoon:
Hey look, a wild undocumented browser bug: https://github.com/mdn/browser-compat-data/issues/26041. Whee!

PR #1740, opened 2025-03-09
 AccessKit Disable GIFs: Optionally delay GIF downloading until hover; refactor

marcustyphoon:
### Description

This adds an option (off-by-default) to AccessKit Disable GIFs to delay downloading of the animated portion of most GIFs until the user actually hovers to load them.

This is implemented in CSS by using—when there is one—Tumblr's native "poster" element as the paused image rather than making our own, hiding either the poster or the animated image depending on the hover state (#1707). Because this is synchronous, choosing to delay the animated image load is as simple as using `display:none` rather than `visibility:hidden` on the animated image.

To avoid having to handle edge cases where the "GIF" label would be underneath another element in certain cases, this also makes the label into an `::after` pseudo element.

As a side effect of using the poster, this fixes the layout of non-square images in recommended blog cards, which (as mentioned in #1463) can expose a second "GIF" label on the background gif; a fix is included to hide this extra label.

 

Supersedes #1458.


AprilSylph:
<sub>Branch has conflicts, please resolve!</sub>

marcustyphoon:
Recreated as #1853 using a branch in this repository instead of a fork.

PR #1749, opened 2025-03-19
Quick Tags: Enable sorting bundles in options pane

### LINKED ISSUE #1748: Quick tags drag and drop organization
### Concept

I have a lot of quick tag bundles for tagging the multiple names of super-heroes in a comic I'm liveblogging. I am organizing them alphabetically by superhero name to make it easier to find them. I just added Andromeda and it took me so many clicks to get her from the very bottom to the very top. Being able to drag and drop the bundle instead would be a great quality of life feature.

marcustyphoon:
### Description

Well, that was shockingly easy.

This adds drag-and-drop reordering to the Quick Tags tag bundles in the options pane.

Resolves #1748.

### Testing steps

- Confirm that tag bundles can be dragged to rearrange them in the options UI and that the drag action rotates the bundle order in the tag popup.
- Confirm that scrolling is not affected negatively on mobile devices.


src/features/quick_tags/options/index.js
```diff
@@ -1,8 +1,23 @@
+/* globals Sortable */
```
AprilSylph, line 1: aw, no ES module version?

marcustyphoon: There is one, iirc! Non minified, but I don't think we care particularly.

marcustyphoon:
Oh—hm. This may cause problems on mobile devices, where a screen full of tag bundles will make scrolling impossible. We probably need to add handles.

marcustyphoon:
Thoughts?

| | |
| - | - |
|  |  |

AprilSylph:
I like the first variant with the drag handles on the left. I can probably figure out how to test it on my Fairphone once the drag handles are committed.

AprilSylph (commented):
The drag handle should have `cursor: grab` instead of the current `cursor: pointer`. Also `cursor: grabbing` when the handle is pressed but not being moved. (Looks like SortableJS is taking care of changing the cursor to `move` for us... neat.)

AprilSylph:
```css
.bundle .drag-handle {
  cursor: grab;
}

.bundle .drag-handle:active {
  cursor: grabbing;
}
```

Are you not seeing the cursor change when you're dragging the "ghost" around?

marcustyphoon:
> Are you not seeing the cursor change when you're dragging the "ghost" around?

Indeed I'm not, in either Chrome or Firefox! Both on MacOS.

marcustyphoon:
Do you think this needs tutorializing? "Drag to rearrange" beneath the header? Probably not, right?

PR #1752, opened 2025-03-22
feat(Scroll to Bottom): Add preference to stop at Changes carousel

marcustyphoon:
this is. so annoying to test :P "yeah you can only test once each time you don't use your tumblr dashboard for a long enough time that the caught up element isn't on the first page of your dash." fun!

### Description

> I suppose that, since Scroll to Bottom doesn't otherwise have an exit condition on a well-populated dashboard, we could add an on-by-default preference for "Stop scrolling the Following feed when I reach the Changes/Trending/etc. carousel"?
> 
> _Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/discussions/1745#discussioncomment-12582154_

This adds the specified preference. Specifically, when the carousel in question is *below the bottom of the viewport when Scroll to Bottom is activated or is* added to the page when Scroll to Bottom is currently active, Scroll to Bottom is deactivated and the page is scrolled to the carousel.

~~This means that if the carousel is already on the page (but not visible to the user) when the Scroll to Bottom button is pressed, scrolling will never stop. We could also add code that the Scroll to Bottom button, when pressed, checks for a carousel that is below the bottom of the viewport and jumps to it instead of activating if so. I think this would be the correct behavior, but is there a more elegant way to do this?~~ Implemented.

Note that I'm not sure if the carousel in question actually always appears, so this might be of limited utility.

### Testing steps

Note some useful modifications for testing:

```js
if (item.elements?.some(({ objectType }) => objectType /* === caughtUpCarouselObjectType */)) {
```

Stops at recommended blog/tag carousels too... which, admittedly, also vanish after a few refreshes, so this isn't always very useful.

```js
if ((Math.random() < 0.05) || item.elements?.some(({ objectType }) => objectType === caughtUpCarouselObjectType)) {
```

Stops at random timeline elements.

---

With endless scrolling enabled:

- Confirm that Scroll to Bottom stops scrolling at the Changes carousel when it is on-screen, but below the bottom of the viewport.
- Confirm that Scroll to Bottom stops scrolling at the Changes carousel when it is loaded from being offscreen.
- Confirm that Scroll to Bottom  continues scrolling if it's enabled when the Changes carousel is on or above the screen.

With endless scrolling disabled:

- This currently does nothing. We definitely could make it stop at the Changes carousel in this case too, but the code is nontrivial and, like... do you think anyone would care?


marcustyphoon:
Not sure this works; the container for the carousel might be able to be added offscreen without the contents being rendered (have not confirmed this, but it's a possible reason why this isn't working).

Considering changing this to "every time the scroll to bottom callback fires, before you scroll down, synchronously-ish* check if the container for the carousel is on-screen and below the current scroll position; scroll to it if it is and scroll to the bottom of the page if it's not." This would necessitate running timelineObject on every cell and (for non-endless scrolling compatibility) every timelineObject element.

Anyway, uh, yeah, unless I'm missing something this is quite a bit more difficult than it probably ought to be.

*within the same tick, before layout

src/features/scroll_to_bottom/index.js
```diff
@@ -88,15 +94,65 @@ const addButtonToPage = async function ([scrollToTopButton]) {
...
+    debouncedCancel();
+  };
+  const observer = new ResizeObserver(callback);
+  const debouncedCancel = debounce(() => {
+    console.log('disconnected reliablyScrollToTarget observer');
+    observer.disconnect();
+  }, 500);
+  observer.observe(document.documentElement);
+  callback();
+};
```
marcustyphoon, line 109: do you ever look at a piece of code and go, "okay, that very obviously needs to be refactored to be shorter" but then see no way to actually do that?

src/features/scroll_to_bottom/index.js
```diff
@@ -88,15 +94,56 @@ const addButtonToPage = async function ([scrollToTopButton]) {
...
+const reliablyScrollToTarget = target => {
+  const callback = () => {
+    window.scrollBy({ top: target?.getBoundingClientRect?.()?.top });
+    debouncedCancel();
+  };
+  const observer = new ResizeObserver(callback);
+  const debouncedCancel = debounce(() => observer.disconnect(), 500);
+  observer.observe(document.documentElement);
+  callback();
+};
```
marcustyphoon, line 106: do you ever look at a piece of code and go, "okay, that very obviously needs to be refactored to be shorter" but then see no way to actually do that?

src/features/scroll_to_bottom/index.js
```diff
@@ -88,15 +94,56 @@ const addButtonToPage = async function ([scrollToTopButton]) {
...
+const reliablyScrollToTarget = target => {
+  const callback = () => {
+    window.scrollBy({ top: target?.getBoundingClientRect?.()?.top });
+    debouncedDisconnect();
+  };
+  const observer = new ResizeObserver(callback);
+  const debouncedDisconnect = debounce(() => observer.disconnect(), 500);
+  observer.observe(document.documentElement);
+  callback();
+};
```
marcustyphoon, line 107: do you ever look at a piece of code and go, "okay, that very obviously needs to be refactored to be simpler and shorter" but then see no way to actually do that?

PR #1753, opened 2025-03-22
No Recommended: Hide posts from non-joined communities in For You

marcustyphoon:
### Description

This adds a No Recommended option that hides posts from non-followed communities in the For You feed, something which Tumblr doesn't have a config option for. Many users find that this type of recommendation, at least at the moment, often isn't relevant to their interests. (Maybe that'll change with increased use of communities, but at least in my experience there simply doesn't seem to be enough communities content for pretty much any of it to be relevant to me.)

Note that Tumblr _does_ have an option not to add posts from communities you **do** follow in the **Following** feed; it's near the bottom of https://www.tumblr.com/settings/dashboard. We naturally won't add an XKit option to do this, as that would be redundant.

### Testing steps

- Confirm that the new No Recommended option works as specified on https://www.tumblr.com/dashboard/stuff_for_you.
- Confirm that the new No Recommended option works as specified on a For You feed in Tumblr Patio.

### DIFF
--- a/src/features/no_recommended.json
+++ b/src/features/no_recommended.json
@@ -13,6 +13,11 @@
       "label": "Hide experimental types of recommended posts",
       "default": true
     },
+    "hide_recommended_community_posts": {
+      "type": "checkbox",
+      "label": "Hide posts from non-followed communities in For You",
+      "default": false
+    },
     "hide_answertime": {
       "type": "checkbox",
       "label": "Hide the answertime banner",
--- a/src/features/no_recommended.json
+++ b/src/features/no_recommended.json
@@ -15,7 +15,7 @@
     },
     "hide_recommended_community_posts": {
       "type": "checkbox",
-      "label": "Hide posts from non-followed communities in For You",
+      "label": "Hide posts from non-joined communities in For You",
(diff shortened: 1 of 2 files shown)


src/features/no_recommended.json
```diff
@@ -13,6 +13,11 @@
       "label": "Hide experimental types of recommended posts",
       "default": true
     },
+    "hide_recommended_community_posts": {
+      "type": "checkbox",
+      "label": "Hide posts from non-followed communities in For You",
```
AprilSylph, line 18: The official terminology decrees that a community cannot be followed, only joined.

```suggestion
      "label": "Hide posts from non-joined communities in For You",
```

AprilSylph (approved):
Don't have Patio on my test account.

New option works on `/dashboard/stuff_for_you`. :white_check_mark:

src/features/no_recommended.json
marcustyphoon: `un-joined`, `unjoined`, no... `communities you haven't joined` is long enough to wrap in an ugly way... yeah, sure, `non-joined`, good enough

PR #1754, opened 2025-03-23
TimeFormat: Fix reply timestamp formatting

marcustyphoon:
### Description

This fixes the layout improvements added in #1110 and an erroneous lack of underline beneath the relative part of the formatted time when hovered in reply timestamps that are links (currently unreleased/in-testing Tumblr code).

### Testing steps

- Enable the "append relative time" option in TimeFormat and find a reply with a relatively long blog name (such as <https://www.tumblr.com/once-more-with-gravitas/778774885212225536/ah-dangit>).
- Confirm that reply timestamps stay as a single line with and without the relevant feature flag.
- Confirm that reply timestamps are underlined consistently when hovered with the relevant feature flag.

### DIFF
--- a/src/features/timeformat.js
+++ b/src/features/timeformat.js
@@ -94,7 +94,7 @@ const formatTimeElements = function (timeElements) {
     const momentDate = moment(timeElement.dateTime, moment.ISO_8601);
     timeElement.dataset.formattedTime = momentDate.format(format);
     if (displayRelative) {
-      timeElement.dataset.formattedTime += '\u2002\u00B7\u2002';
+      timeElement.dataset.formattedTime += '\u00A0\u00B7\u00A0';
       timeElement.dataset.formattedRelativeTime = constructRelativeTimeString(momentDate.unix());
     }
   });
--- a/src/features/timeformat.js
+++ b/src/features/timeformat.js
@@ -61,6 +61,11 @@ ${keyToCss('blogLinkWrapper')} + a:has(> [data-formatted-time]) {
 a > [data-formatted-relative-time]::after {
   display: inline;
 }
+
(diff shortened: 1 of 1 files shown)


AprilSylph:
The on-hover underline offset is weird for me in Firefox; it gets cut off by capital letters:

image

I think we should be applying our own underline, with an offset. Also, let's override back to `cursor: pointer;` for these...

I'll self-assign this and play with it sometime.

marcustyphoon:
That is... very strange-looking. Huh.

AprilSylph:
All fixed up, I think:

Screenshot 2025-06-10 at 16-55-52 Tumblr

AprilSylph (approved):
Would appreciate you double-checking my changes before merging!

marcustyphoon:
Firefox:

- [x] middot spacing looks fine in post headers, replies, and likes
- [x] cursor is pointer in replies
- [x] cursor is help in post headers (dom modified to test; currently there's an `<a>` covering them) and likes
- [x] reply hover underlines look correct

Chrome:

- [x] middot spacing looks fine in post headers, replies, and likes
- [x] cursor is pointer in replies
- [x] cursor is help in post headers (dom modified to test; currently there's an `<a>` covering them) and likes
- [x] reply hover underlines look correct

Safari 17.6:

- [x] middot spacing looks fine in post headers, replies, and likes
- [x] cursor is pointer in replies
- [ ] cursor is help in post headers (dom modified to test; currently there's an `<a>` covering them) and likes
- [x] reply hover underlines look correct

Safari help cursor is not a regression, as this was apparently the case before. Presumably a Safari bug with setting cursor on a pseudo; I didn't look into it.

PR #1755, opened 2025-03-24
Refactor: Use browser.storage.local.onChanged.addListener

marcustyphoon:
### Description

As of Firefox 101, as per https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged, we should be able to use `browser.storage.local.onChanged.addListener` rather than using `browser.storage.onChanged.addListener` and checking if the second callback parameter is `'local'`.

In addition to being a nice refactor, this appears—weirdly enough—to be a bugfix on Safari 18.1.1 on iOS in my testing, because Safari is giving `undefined` as the second callback parameter when I use `browser.storage.onChanged.addListener`, which breaks all of our storage listeners.

### Testing steps

- [] Confirm that changing extension preferences works in Chromium.
- [] Confirm that changing extension preferences works in Firefox.
- [] Confirm that changing extension preferences works in Safari.

### DIFF
--- a/src/action/render_backup.js
+++ b/src/action/render_backup.js
@@ -78,11 +78,7 @@ const localRestore = async function () {
 
 const renderLocalBackup = async function () {
   updateLocalExportDisplay();
-  browser.storage.onChanged.addListener((changes, areaName) => {
-    if (areaName === 'local') {
-      updateLocalExportDisplay();
-    }
-  });
+  browser.storage.local.onChanged.addListener(updateLocalExportDisplay);
 
   localCopyButton.addEventListener('click', localCopy);
   localDownloadButton.addEventListener('click', localExport);
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -33,21 +33,19 @@
       document.documentElement.append(styleElement);
     }
 
-    restartListeners[name] = async (changes, areaName) => {
-      if (areaName !== 'local') return;
-
+    restartListeners[name] = async (changes) => {
       const { enabledScripts } = changes;
       if (enabledScripts && !enabledScripts.newValue.includes(name)) return;
 
       if (onStorageChanged instanceof Function) {
-        onStorageChanged(changes, areaName);
+        onStorageChanged(changes);
       } else if (Object.keys(changes).some(key => key.startsWith(`${name}.preferences`) && changes[key].oldValue !== undefined)) {
         await clean?.();
         await main?.();
       }
     };
 
-    browser.storage.onChanged.addListener(restartListeners[name]);
+    browser.storage.local.onChanged.addListener(restartListeners[name]);
   };
 
   const destroyScript = async function (name) {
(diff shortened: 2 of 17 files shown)


AprilSylph (commented):
> <sup>https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged</sup>
> Note: In Firefox, the information returned includes all keys within the storage area. Also, the callback may be invoked when there is no change to the underlying data. Details of the changed items are found by examining each returned key's [storage.StorageChange](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageChange) object.

Doesn't this mean all our `Object.keys(changes).some(key => key.startsWith` checks need to be updated too?

```js
export const hasStorageKeyChanged = (changes, keyStartsWith) =>
  Object.keys(changes).some(key => (
    key.startsWith(keyStartsWith) &&
    changes[key].oldValue !== changes[key].newValue
  ));
```

(code not tested)

marcustyphoon:
Hm.

- This note seems to appear (with slightly different, but not clearly different, copy... should probably put an mdn issue about that) in the equivalent `storage.onChanged` docs. If there are any problematic behaviors resulting from this, we should already be encountering them in Firefox.
- Some quick testing (`browser.storage.local.set({ enabledScripts, specialAccess, hello: 'world' });` + `export const onStorageChanged = changes => console.log(changes);`) indicates that this is indeed the case. Chrome logs only the enabledScripts value when toggling a script; Firefox logs all of them.
- We only seem to set multiple values in one `storage.local.set` call in one place and seem to set primitive values in very few places; the number of updates we could save by equality checking appears to be zero and if we needed to do so we would need a deep equality check.

So I think—if I haven't missed anything, which is definitely possible—it should be fine as-is!

AprilSylph:
The difference between MDN's descriptions seems to be implying that while  `storage.onChanged`'s first argument includes all the keys present in the triggering `StorageArea.set()` call, `storage.StorageArea.onChanged`'s argument includes all keys present in the entire storage area, not just those from the triggering `StorageArea.set()` call.

> <sup>https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/onChanged</sup>
> Note: In Firefox, the information returned includes all keys within the storage area [storageArea.set](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set) ran against whether they changed or not.

> <sup>https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged</sup>
> Note: In Firefox, the information returned includes all keys within the storage area.

Is the MDN note for `storage.StorageArea.onChanged` just unclear to the point of being incorrect? It does look copy-pasted.

marcustyphoon:
Yeah, I think something bad must have happened during review of this PR: https://github.com/mdn/content/pull/31227.

PR #1758, opened 2025-03-26
Options UI: Fix disabled label wrap

marcustyphoon:
### Description

This fixes an issue that only crops up on narrow options UI widths where the space between feature labels and the (disabled) text wasn't an option for text wrapping, making features with very long names and no space able to have slightly wonky UI alignment.

### Testing steps

- Confirm that disabled UI spacing looks good.
- Either open the options UI in its own window (https://github.com/AprilSylph/XKit-Rewritten/pull/1193#issuecomment-1676173160) and use browser zoom to simulate a very small options window, or load XKit Rewritten into the iPadOS simulator or onto an iPad. Confirm that UI wraps well.

### DIFF
--- a/src/action/configuration.css
+++ b/src/action/configuration.css
@@ -139,14 +139,12 @@ label[for="filter"] {
 }
 
 .script.disabled .title::after {
-  content: "(disabled)";
-  margin-left: 0.5ch;
+  content: " (disabled)";
   font-weight: normal;
 }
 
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):

marcustyphoon:
With this pulled into #1758, I assume there's not much point in actually pressing merge, as it'll just be a conflict that needs to get resolved.

PR #1760, opened 2025-03-28
fix: Modal UI broken in mobile/tablet browser widths

### LINKED ISSUE #1759: Text boxes in modals don't function in mobile/tablet width
### Platform

MacOS 14.7.1 1rm64

### Browser

Firefox 135.0, Chrome 134.0.6998.118

### Addon version

v1.0.8

### Details

When opened from the mobile/tablet sidebar, the text input boxes in e.g. Tag Replacer and Mass Privater can't be interacted with.

I assume something must have changed to make this the case, because surely we would have noticed this before now if not (????)

Probably the same fix as https://github.com/new-xkit/XKit/pull/2101.

marcustyphoon:
### Description

Just as in https://github.com/new-xkit/XKit/pull/2101, this fixes an issue where Tumblr's code would steal focus from and break form input elements in our modal overlay UI because it was trying to prioritize its own. In this case, it was when something like Tag Replacer or Mass Deleter/Privater/Unliker was opened from the sidebar menu where they are in mobile device/tablet viewport widths.

Resolves #1759.

Thanks to Gideon on the New XKit discord for reporting this!

### Testing steps

- Confirm that Tag Replacer/Mass [whatever] features work in <990px and <540px viewport widths.

### DIFF
--- a/src/utils/modals.js
+++ b/src/utils/modals.js
@@ -17,7 +17,10 @@ export const showModal = ({ title, message = [], buttons = [] }) => {
     id: 'xkit-modal',
     tabindex: '-1',
     role: 'dialog',
-    'aria-modal': 'true'
+    'aria-modal': 'true',
+
+    // prevents Tumblr's trapFocusInsideGlass function from stealing focus when opened from mobile drawer
+    'data-skip-glass-focus-trap': ''
   }, null, [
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):

AprilSylph:
I should be able to do some more reviews and get a release out this weekend.

PR #1573, opened 2024-08-17
Quick Tags; Trim Reblogs: Enable in communities (old)

marcustyphoon:
### Description

This enables Quick Tags and Trim Reblogs on posts you authored/reblogged in communities.

See #1441.

Closes #1481.

### Testing steps

At time of writing, I have a spare community with invite link [tumblr.com/join/oki7cQsr](https://www.tumblr.com/join/oki7cQsr).
- Confirm that Quick Tags/Trim Reblogs buttons appear, in the correct order, on posts you can edit in communities and on https://www.tumblr.com/communities.
- Confirm that Quick Tags/Trim Reblogs buttons do not appear, on posts you cannot edit in communities and on https://www.tumblr.com/communities.
- Confirm that the Quick Tags button functions.
- Confirm that the Trim Reblogs button functions and that the preview items appear correctly.
--
- Confirm that the Quick Tags button appears and functions on non-community posts.
- Confirm that the Trim Reblogs button appears and functions on non-community posts.

PR #1775, opened 2025-04-11
update code for upcoming post footer tweaks

marcustyphoon:
### Description

This adjusts Quick Reblog, Quick Tags, and the tweak to hide control button tooltips for the upcoming tweaks to the post footer.

Based on #1573, which means this will have to be rebased (presuming that it's squash merged).


marcustyphoon:
Pretty sure the vertical alignment of the buttons isn't currently correct in this.

marcustyphoon:
Also, needs to exclude asks in the inbox. I forget the right post data field to do this... `state`?

marcustyphoon:
changes required to handle split notes, another possible upcomign post footer tweak:
- add `${keyToCss('engagementAction')}` to `buttonDivSelector` in quick_reblog.js
- set `position: relative` on the `engagementAction` element
- change hide footer tooltips final selector line to `article footer ${keyToCss('controls', 'engagementAction')} ${keyToCss('tooltip')} {`
- ~~question why on earth staff is using `display: grid` instead of `display: flex`~~
- ~~question why the blaze button has `position: absolute; right: 52px;~~
- ~~whatever anyway~~
- honestly just wait and see if the layout/css gets changed to not be inexplicable and hope this works:

```js
      const newControls = postElement.querySelector(keyToCss('engagementControls'));
      if (newControls) {
        newControls.after(clonedControlButton);
        return;
      }
 ```

PR #1800, opened 2025-05-28
Tweaks: Fix "hide control button tooltips" tweak in split-notes footer

marcustyphoon:
### Description

This fixes the "hide footer tooltips" tweak on accounts with the new split-notes post footer.

### Testing steps

- Confirm that the tweak hides tooltips on all footer variations.


AprilSylph (approved):
- [x] Works with old footer
- [x] Works with new footer, combined notes variant
- [x] Works with new footer, split notes variant

Issue #742, opened 2022-09-02
Quick Reblog not working


AprilSylph:
Tumblr is testing new reblog controls, which breaks Quick Reblog's reblog button detection.

My main account was hit with this test, so hopefully I'll be able to craft an update tonight.

AprilSylph:
New selector for the reblog button, while avoiding the reblog button in the menu:
```js
`${postSelector} footer button[aria-label="${translate('Reblog')}"]:not([role])`
```

AprilSylph:
I didn't have time to implement and test this before the test was rolled back to less users, and now neither my main nor testing account are in the bucket.

@marcustyphoon Are any of your accounts still in the test?

marcustyphoon:
Nope. I didn't even get to see it.

Skeithh:
I currently have the new reblog test features on my account, if I can be of any help to test the new implementation

AprilSylph:
@Skeithh if you clone the repository as it is right now, then load `src/` as an unpacked/temporary extension in your browser, you'll be able to test the fix. Be sure to either disable the release version of XKit Rewritten before you do (to avoid confusion; they'll have the same version number) or do so in a separate browser profile.
- [Loading a temporary extension in Firefox](https://developer.mozilla.org/en-US/docs/Tools/about:debugging#Extensions)
- [Loading an unpacked extension in Chrome](https://developer.chrome.com/docs/extensions/mv2/getstarted/#manifest)

partrah:
My account has been affected by this too so I did the load unpacked thing and it's working for me! Although it shows an error if I'm looking at the chrome extensions page, it's still working perfectly so far

AprilSylph:
@partrah Thanks for testing the fix! I've now pushed v0.20.9, which should hit you in a day or two automatically. It's already approved for Firefox, I'll let you know when it's approved for Chrome.

AprilSylph:
Now approved for Chrome!

Hananasaur:
Just a comment on the reblog button fix... it used to also remember the already reblogged/queued posts from the /tagged/ pages, but now it doesn't seem to anymore. (It does remember it for the posts in the dashboard...)

Also thank you for applying such a quick fix!!!

marcustyphoon:
[...]

PR #1801, opened 2025-05-28
Quick Reblog: Fix feature on accounts with new footer variants

marcustyphoon:
### Description

This fixes Quick Reblog on accounts with the new split-notes post footer.

Edit: There are actually two footer variations in testing, so this should address both!

### Testing steps

On all three footer variations:

- Confirm that the Quick Reblog popup appears, in the correct location, when hovering over the reblog button.
- Confirm that the Quick Reblog popup is not covered by Tumblr's control button tooltip, as the tooltip is now hidden by this feature.
- Confirm that alreadyreblogged posts have different-colored reblog icons.

Note: I haven't done a full round of testing on this after my last few changes, unless I've crossed out this line in the description.

### DIFF
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -11,7 +11,7 @@ import { dom } from '../utils/dom.js';
 import { showErrorModal } from '../utils/modals.js';
 import { keyToCss } from '../utils/css_map.js';
 
-const popupElement = dom('div', { id: 'quick-reblog' });
+const popupElement = dom('div', { id: 'quick-reblog' }, { click: event => event.stopPropagation() });
 const blogSelector = dom('select');
 const blogAvatar = dom('div', { class: 'avatar' });
 const blogSelectorContainer = dom('div', { class: 'select-container' }, null, [blogAvatar, blogSelector]);
--- a/src/features/quick_reblog.js
+++ b/src/features/quick_reblog.js
@@ -6,7 +6,6 @@ import { joinedCommunities, joinedCommunityUuids, primaryBlog, userBlogs } from 
 import { getPreferences } from '../utils/preferences.js';
 import { onNewPosts } from '../utils/mutations.js';
 import { notify } from '../utils/notifications.js';
-import { translate } from '../utils/language_data.js';
 import { dom } from '../utils/dom.js';
 import { showErrorModal } from '../utils/modals.js';
 import { keyToCss } from '../utils/css_map.js';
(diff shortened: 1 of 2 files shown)


src/features/quick_reblog.js
```diff
@@ -71,8 +71,19 @@ const avatarUrls = new Map();
 
 const reblogButtonSelector = `
 ${postSelector} footer a[href*="/reblog/"],
-${postSelector} footer button[aria-label="${translate('Reblog')}"]:not([role])
+${postSelector} footer button[aria-label="${translate('Reblog')}"]:not([role]),
+${postSelector} footer a[aria-label="${translate('Reblog')}"][href*="/reblog/"]
```
AprilSylph, line 75: This is the same as the first selector, but with extra specificity. What does that achieve?

src/features/quick_reblog.js
```diff
@@ -71,8 +71,19 @@ const avatarUrls = new Map();
...
 `;
+const buttonDivSelector = `:is(${keyToCss('controls')} > *, ${keyToCss('engagementAction')})`;
```
AprilSylph, line 77: Does this need to be an `:is()`? I only see this used in the calls to `.closest()`, which accepts a selector list. If it's a performance thing, that's fine, but it's unclear.

AprilSylph (commented):
- [x] Confirm that the Quick Reblog popup appears, in the correct location, when hovering over the reblog button.
  - [x] No features
  - [x] New footer, combined notes button
  - [x] New footer, split notes button
- [x] Confirm that the Quick Reblog popup is not covered by Tumblr's control button tooltip, as the tooltip is now hidden by this feature.
- [x] Confirm that alreadyreblogged posts have different-colored reblog icons.
  - [x] No features
  - [x] New footer, combined notes button
  - [x] New footer, split notes button :round_pushpin: 

---

**Bug**: Using Quick Reblog in the split notes variant triggers a click on the notes toggle button. Even clicking the tags field toggles the notes view visibility. Should be easily fixed by preventing click event propagation?

marcustyphoon:
Oof, apparently I didn't "do a full round of testing" since switching this code from the other footer. Not sure why I raced to MVP. (eh, I guess I do know why; the correct assertion is I don't know why I pressed review request when I knew I hadn't done my own job).

src/features/quick_reblog.js
marcustyphoon: I have absolutely no clue.

AprilSylph:
Another bug on the split-notes footer, wasn't sure what the repro steps were before but now I seem to have figured it out:

1. Reblog/Queue/Draft a post using Quick Reblog. The reblog button turns green/purple/red as expected.
2. Mouse over the reblog button. The reblog button stays green/purple/red from Quick Reblog as expected.
3. Mouse over the reblog count. The reblog button turns green from Redpop's own code.
4. Mouse out of the reblog count. The reblog button turns grey again. :x: 

It seems that the entire reblog pill is being re-rendered when moused out of, so we need to put the Quick Reblog action-taken classname a lot higher in the tree. Maybe on the `footer` element itself?

AprilSylph (approved):
Fixed the last bug that I noticed. All seems working otherwise. I'll let you decide whether to conduct your own retests prior to merging.

marcustyphoon:
> ...is it just me, or does the non-link version of the reblog button not exist anymore?
It seems like footer a[href*="/reblog/"] is 100% reliable now.

Seems like it. That change was #742, which I don't specifically remember but is certainly in the rearview mirror, and I don't see any obvious references to a button element with that area label anywhere.

PR #1803, opened 2025-05-28
Quick Tags; Trim Reblogs: Enable in communities

marcustyphoon:
### Description

This enables Quick Tags and Trim Reblogs on posts you authored/reblogged in communities.

Supercedes #1573.

Based on #1802 and will need to be rebased if it is squash merged.

See #1441.

### Testing steps

At time of writing, I have a spare community with invite link [tumblr.com/join/oki7cQsr](https://www.tumblr.com/join/oki7cQsr).
- Confirm that Quick Tags/Trim Reblogs buttons appear, in the correct order, on posts you can edit in communities and on https://www.tumblr.com/communities.
- Confirm that Quick Tags/Trim Reblogs buttons do not appear, on posts you cannot edit in communities and on https://www.tumblr.com/communities.
- Confirm that the Quick Tags button functions.
- Confirm that the Trim Reblogs button functions and that the preview items appear correctly.
--
- Confirm that the Quick Tags button appears and functions on non-community posts.
- Confirm that the Trim Reblogs button appears and functions on non-community posts.


src/utils/control_buttons.js
```diff
@@ -71,7 +71,8 @@ export const insertControlButton = async (postElement, clonedControlButton, butt
   const newEditControlIcon = postElement.querySelector('a[href*="/edit/"]:has(use[href="#managed-icon__ds-pencil-outline-24"])');
 
   if (community) {
-    // not yet implemented
+    clonedControlButton.classList.add('in-community');
+    postElement.querySelector(`${keyToCss('controls')}`).append(clonedControlButton);
```
AprilSylph: **nit**: Unnecessary string creation

```suggestion
    postElement.querySelector(keyToCss('controls')).append(clonedControlButton);
```

marcustyphoon:
Oh, hm.

- [x] fix Trim Reblogs preview

marcustyphoon:
Hm, hadn't thought about copying the button styles.

marcustyphoon:
Do you think the control buttons should be before the share button in community post footers? Obviously if they're styled the old way they should be last, but when they match I feel like maybe share should always be last. Then again, share is always there and these are not, so maybe not?

AprilSylph:
Yeah, I think the order is fine as-is; for blog posts, putting our control buttons next to Tumblr's Edit button specifically makes sense because our control buttons are also ways of editing posts—grouping edit actions together takes precedence over not moving the Delete button.

For community posts, there is no Edit button in the footer, so in the absence of that, we should prioritise not modifying the placement of Tumblr's buttons.

marcustyphoon:
Still kind of mixed feelings about the button style but I don't feel I have anything conclusive or interesting to say on the matter.

This would be an approval if I were not PR owner.

PR #1804, opened 2025-05-28
AccessKit Disable GIFs: Exclude blog sub-avatars

marcustyphoon:
### Description

As an addition to #1746, this prevents AccessKit Disable GIFs from processing blog sub-avatar elements in posts on group blogs.

### Testing steps

I don't think I currently have a test post for this one; the place where I noticed it was in a queue. Might get around to making one.


AprilSylph (approved):
Validated by just... changing a sub-avatar image to a `.gif` via Inspect Element before activating the option in question. No more GIF label! ✅ 

Before | After
-|-
Image | Image

PR #1810, opened 2025-06-03
Quick Reblog: Fix popup hiding when holding E/W on accounts with new footer variants

### LINKED ISSUE #1809: Quick Reblog: Fix popup hiding when holding E/W on accounts with new footer variants
### Concept

I'm not actually sure why, but when holding E or W to use Tumblr's native quick reblog/quick queue, our Quick Reblog popup disappears. This is good!

This doesn't, however, happen in either new footer variant.

marcustyphoon:
### Description

This ensures that the behavior of hiding the Quick Reblog popup when a user uses the E/W keyboard shortcuts to show Tumblr's built-in quick reblog bubbles works in all three footer variants.

Admittedly this CSS looks kind of lazy; I generally try to use `:has()` with a specific selector path (like `:has(> something)`) unless it's in a queryselector/closest/matches call. To my knowledge, though, because selectors are generally evaluated right-to-left, this rule only performs search/invalidation on the parent element of the quick reblog popup, which at any given time is either nothing or a very small number of elements, so I don't expect a noticeable performance impact on even the slowest of systems.

Resolves #1809.

### Testing steps

On all footer variations, confirm that the Quick Reblog popup disappears when holding W or E when hovering over a post's reblog button.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,10 +2,10 @@ root = true
 
 [*]
 end_of_line = lf
+indent_size = 2
+indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
-indent_style = space
-indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,12 +1,12 @@
-*               text=auto
+.gitattributes  export-ignore
+.gitignore      export-ignore
 
-*.png           binary
+*               text=auto
 
 *.css           text  diff=css
 *.html          text  diff=html
 *.md            text  diff=markdown
 
-.gitattributes  export-ignore
-.gitignore      export-ignore
+*.png           binary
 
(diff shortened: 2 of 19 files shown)


AprilSylph (approved):
- [x] No features
  - [x] <kbd>E</kbd>
  - [x] <kbd>W</kbd>
- [x] `postFooterSplitNotesCount`
  - [x] <kbd>E</kbd>
  - [x] <kbd>W</kbd>
- [ ] ~~`postFooterSplitNotesCountSingleAction`~~ Popup does not appear at all with this build
  - [ ] <kbd>E</kbd>
  - [ ] <kbd>W</kbd>

---

Willing to approve anyway because the code looks robust enough. 🙌

marcustyphoon:
After merge of PR enabling Quick Reblog here:

- [x] `postFooterSplitNotesCountSingleAction`
  - [x] <kbd>E</kbd>
  - [x] <kbd>W</kbd>

PR #1815, opened 2025-06-16
chore(deps): upgrade to eslint v9; migrate to neostandard

AprilSylph:
### Description

- closes #1478
- closes #1678
- closes #1679

This is a redo of our linting setup for ESLint v9. It now seems official that [standard](https://github.com/standard/standard) and derivatives are no longer maintained; the updated equivalent is [neostandard](https://github.com/neostandard/neostandard).

I don't like ESLint v9's configuration format, but it only took me the morning to wrap my head around it, and there's something kinda neat about using ES imports for config compiling. Whatever, it works. RIP eslintrc dot json.

Something very nice about neostandard is no more peer dependencies. `eslint-plugin-promise`, `eslint-plugin-n`, and `eslint-plugin-import` can be dropped from our package.json; everything comes included as transitive dependencies (FINALLY), and we can still use the newest fork of `eslint-plugin-import` _via neostandard's exports_ to enable the `no-cycle` rule ([which we still need to do ourselves](https://github.com/neostandard/neostandard/blob/f994154e46528e87428325e78ae70133a7b96927/lib/configs/base.js#L152-L157)).

For some reason, the "flat" version of the `eslint-plugin-jsdoc` recommended config gives us new warnings which were not present before. I'm not attempting to remedy them here, because how to resolve them is a question whose answer has no bearing on the viability of this PR.

<details><summary>Console output</summary>

```
/home/april/Desktop/XKit-Rewritten/src/main_world/api_fetch.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/control_tags_input.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/get_notification_props.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/get_tumblelogname_prop.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/post_request.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/test_header_element.js
  1:16  warning  Missing JSDoc comment  jsdoc/require-jsdoc

/home/april/Desktop/XKit-Rewritten/src/main_world/unbury_blog.js
[...]

PR #1818, opened 2025-06-24
chore: (re-)enable `import-x` recommended lint rules

AprilSylph:
### Description

Re-enables `import/recommended` (now actually `import-x/recommended`) (actually not that either because the format for reusable eslint configs is completely different now) lint rules that were previously removed in #1815.

I was a little too trusting of `neostandard`'s rules. Turns out, it doesn't even enable the `no-unresolved` lint rule, which I found to be kind of important for implementing the new directory structure for `src/features/`, as discussed in #1816.

### Testing steps

Code review and green CI is enough.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,5 +1,5 @@
 import jsdoc from 'eslint-plugin-jsdoc';
-import neostandard from 'neostandard';
+import neostandard, { plugins } from 'neostandard';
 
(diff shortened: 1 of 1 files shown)


marcustyphoon (approved):
Things I should have caught and commented on in my own review, part [probably a big number at this point].

PR #1823, opened 2025-07-03
Quick Reblog: Fix popup on accounts with new footer variant

### LINKED ISSUE #1822: Quick Reblog: Feature nonfunctional on accounts with a new footer variant
### Platform

MacOS 14.7.6 arm64

### Browser

Chrome 137.0.7151.104

### Addon version

master branch

### Details

I was working on fixing an issue yesterday in which the quick reblog popup was somewhat buggy on the new alternative split-notes post footer (clicks in it would also click the button; fixable with `preventDefault()`), but doing it correctly (fixing the insertion location in the DOM) wound up being more complicated than I thought, and today the code changed again and now the feature in its current state is nonfunctional. Still looking into it; I'll file the bug in case you want to as well/since it might take long enough to justify that.

marcustyphoon:
### Description

This fixes an issue on a new slight variant of one of the post footers that's floating around where clicks within the quick reblog popup would trigger the reblog button and open the reblog interface.

Resolves #1822.

### Testing steps

On all 3 footer variations:

- Confirm that the Quick Reblog modal appears when hovering over the reblog button.
- Confirm that the Quick Reblog modal **doesn't** appear when hovering over the reblog button of a post with reblogs disabled.
- Confirm that clicking the blog selector, text boxes, quick tags buttons, and submit buttons in the Quick Reblog modal only do what they're supposed to.
- Confirm that the reblog button changes color when Quick Reblog is used to reblog a post.

Note that the reblog button will become highlighted when mousing over the Quick Reblog popup in one split notes variant but not the other.

### DIFF
--- a/src/features/quick_reblog/index.css
+++ b/src/features/quick_reblog/index.css
@@ -190,9 +190,9 @@ div:first-child + span + #quick-reblog,
 
 #quick-reblog .action-buttons.community-selected button:not([data-state="published"]) { display: none; }
 
-footer.published a[href*="/reblog/"] svg use { --icon-color-primary: rgb(var(--green)); }
-footer.queue a[href*="/reblog/"] svg use { --icon-color-primary: rgb(var(--purple)); }
-footer.draft a[href*="/reblog/"] svg use { --icon-color-primary: rgb(var(--red)); }
+footer.published svg use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]) { --icon-color-primary: rgb(var(--green)); }
+footer.queue svg use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]) { --icon-color-primary: rgb(var(--purple)); }
+footer.draft svg use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]) { --icon-color-primary: rgb(var(--red)); }
 
 :is(.published, .queue, .draft) :is(a[role="button"], button[class=""]) {
   position: relative;
--- a/src/features/quick_reblog/index.js
+++ b/src/features/quick_reblog/index.js
@@ -10,7 +10,7 @@ import { dom } from '../../utils/dom.js';
 import { showErrorModal } from '../../utils/modals.js';
 import { keyToCss } from '../../utils/css_map.js';
 
-const popupElement = dom('div', { id: 'quick-reblog' }, { click: event => event.stopPropagation() });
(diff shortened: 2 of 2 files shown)


marcustyphoon:
Oh, right, I discovered this but forgot that it would have implications for this PR: there's different behavior for "reblogs disabled; has no reblogs" (disabled button) and "reblogs disabled; has reblogs" (button is enabled and only links to the reblogs).

marcustyphoon:
Checking `canReblog` on the post mostly works, but posts on a password-protected blog are not rebloggable but have that value set to true. Checking for `isPasswordProtected` on the blog as well should fix this. If we do this, we should make it an export of `src/utils/react_props.js`.

_Edit: Hah, no, it doesn't fix this, because `isPasswordProtected` doesn't exist on posts in peepr, because it's not queried... which also breaks Tumblr's reblog button disabling functionality. Sending this to support._

Not sure if there's another way yet.

marcustyphoon:
On all 3 footer variations:

- Confirm that the Quick Reblog modal **doesn't** appear when hovering over the reblog button of a post with reblogs disabled that has never been reblogged.
- Confirm that the Quick Reblog modal **doesn't** appear when hovering over the reblog button of a post with reblogs disabled but that has been reblogged before and thus has a reblog/note count.

Note that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post on a password-protected blog on the "channel" page, but does appear when doing so in the blog view modal and will 404 when used. This is a Tumblr bug, as the native reblog button will also be enabled and will also show an error message when clicked in this case.

AprilSylph (approved):
- [x] No features variant
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled that has never been reblogged.
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled but that has been reblogged before and thus has a reblog/note count.
- [x] `postFooterSplitNotesCount`
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled that has never been reblogged.
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled but that has been reblogged before and thus has a reblog/note count.
- [x] `postFooterSplitNotesCountSingleAction`
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled that has never been reblogged.
  - [x] Confirm that the Quick Reblog modal doesn't appear when hovering over the reblog button of a post with reblogs disabled but that has been reblogged before and thus has a reblog/note count.

PR #1820, opened 2025-06-27
chore: restructure `src/features/` directory

AprilSylph:
### Description

- Discussed in #1816

Changelog:
- Moved/renamed the following files:
  - `src/features/<feature_name>.json` &rarr; `src/features/<feature_name>/feature.json`
  - `src/features/<feature_name>.js` &rarr; `src/features/<feature_name>/index.js`
  - `src/features/<feature_name>.css` &rarr; `src/features/<feature_name>/index.css`
  - `src/features/_index.json` &rarr; `src/features/index.json`
- Updated every import in every feature's main script (`utils/` and `lib/` are now one directory higher, matching the traversal previously only necessary in child scripts for e.g. Tweaks)
- Updated every reference to `features/**` in dynamic import statements
- Updated documentation for all of the above
  - Additionally, "feature manifests" are now known as "feature metadata" for greater terminological accuracy

I decided against doing anything with `options/` files right now. To be honest, I'm kind of on the fence about doing anything with them at all; I kind of like the clearer separation between the structures of one-script-per-preference features and the needs-iframe-type-preference-panel features.

### Testing steps

- Load the modified addon in your browser of choice
- Open the configuration panel
  - [ ] **Expected result**: All features are displayed with their correct metadata (icon, name, description, preferences)
- Open a Tumblr tab
- Open your devtools &rarr; Console, clear the console, then filter to errors only
- Enable _every single feature_, one at a time
  - [ ] **Expected result**: There should be no console errors produced by enabling an XKit feature
  - [ ] **Expected result**: Basic functionality of every feature should be able to be confirmed
    - [ ] AccessKit
    - [ ] Anti-Capitalism
    - [ ] Classic Search
    - [ ] CleanFeed
    - [ ] Collapsed Queue
    - [ ] Hide Avatars
    - [ ] Limit Checker
    - [ ] Mass Deleter
    - [ ] Mass Privater
    - [ ] Mass Unliker
    - [ ] Mirror Posts
    - [ ] Mutual Checker
    - [ ] No Recommended
    - [ ] NotificationBlock
    - [ ] Open In Tabs
    - [ ] Painter
    - [ ] Panorama
    - [ ] PostBlock
    - [ ] Quick Reblog
    - [ ] Quick Tags
    - [ ] Quote Replies
    - [ ] Scroll to Bottom
    - [ ] Seen Posts
    - [ ] Shorten Posts
    - [ ] Show Originals
    - [ ] Tag Replacer
    - [ ] Tag Tracking+
[...]

PR #1828, opened 2025-07-07
chore: rename "scripts" to "features" in code

AprilSylph:
### Description

- Relates to https://github.com/AprilSylph/XKit-Rewritten/pull/1507
- Relates to https://github.com/AprilSylph/XKit-Rewritten/pull/1820#discussion_r2172593553

Updates `src/action/**` and `src/utils/preferences.js`' internal-only terminology of "scripts" to "features", to match our new directory structure (as of v1.0.0) and our documentation.

This PR intentionally does not update the `"enabledScripts"` storage key or any of its usages, since this PR is already rather big, and by its nature of doing mass-renaming should be tested thoroughly; adding any migratory code for `"enabledScripts"` &rarr; `"enabledFeatures"` would fall outside the scope of this testing.

I also haven't touched `src/content_scripts/main.js`, because 1) that file really does primarily deal with features' scripts (being the thing that runs them), and 2) it would be confusing to update its terminology while keeping the `enabledScripts` name intact.

### Testing steps

1. Load the modified addon
    - [ ] **Expected result**: All features render their information (icon, title, description) with no obvious visual changes vs the latest release version
    - [ ] **Expected result**: All feature preference types render with no obvious visual changes vs the latest release version
      - [ ] `checkbox` (as used by e.g. AccessKit)
      - [ ] `text` (as used by e.g. Quick Reblog)
      - [ ] `color` (as used by Painter)
      - [ ] `select` (as used by e.g. AccessKit)
      - [ ] `textarea` (as used by e.g. CleanFeed)
      - [ ] `iframe` (as used by e.g. PostBlock)
3. Search for a feature in the control panel
    - [ ] **Expected result**: For a valid search/filter, non-matching feature results are hidden
    - [ ] **Expected result**: For an invalid search/filter, all features are hidden and "No results found" is shown instead
4. Clear your search/filters
    - [ ] **Expected result**: All features are shown again
5. Open the preferences for a feature that has no preferences (e.g. Limit Checker)
    - [ ] **Expected result**: The empty state verbiage says "No preferences available for this feature."
6. Enable or disable a feature
    - [ ] **Expected result**: Your change is reflected on Tumblr, or in the control panel's Backup tab
    (whichever is more convenient to test)
7. Change a feature's preferences
    - [ ] **Expected result**: Your changes are reflected on Tumblr, or in the control panel's Backup tab
    (whichever is more convenient to test)

### DIFF
--- a/src/action/render_features.js
+++ b/src/action/render_features.js
@@ -16,7 +16,7 @@ const getInstalledFeatures = async function () {
 const writeEnabled = async function ({ currentTarget }) {
   const { checked, id } = currentTarget;
   const detailsElement = currentTarget.closest('details');
-  const {
+  let {
     [enabledFeaturesKey]: enabledFeatures = [],
     [specialAccessKey]: specialAccess = []
   } = await browser.storage.local.get();


marcustyphoon:
https://github.com/AprilSylph/XKit-Rewritten/blob/6cb7aefa2edb4d4dd116dd438fe53b667c7d2d4a/docs/Chapter%204.0%20-%20Feature%20framework.md?plain=1#L25

Not exactly relevant, but... kind of relevant?

marcustyphoon:
I think if it were me I would probably do `const enabledFeaturesKey = 'enabledScripts';` and include the usage changes in the same PR without including the storage migration, but it really doesn't matter :D

marcustyphoon (approved):

PR #1831, opened 2025-07-09
refactor: use Web Components to render features in the control panel

AprilSylph:
### Description

- Incorporates and closes #1758

Rewrites how feature elements in the control panel are rendered, now utilising [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components). Since we're already using `<template>` cloning, the MVP of this change was to introduce `<slot>` usage, so that filling out feature information can be done by the browser automagically instead of by us using `querySelector`.

Somewhat unfortunately, using template slots necessitates the usage of the Shadow DOM aspect of Web Components, which means:
1. There is now a hard* CSS barrier between features and the rest of the popup; all the feature-specific styles have been put into their own stylesheet. (*Fun fact: CSS variables actually _do_ get inherited by shadow trees, but nothing else does.)
2. There is now a soft JS barrier between features and the rest of the popup; the search and filter functions have been updated to account for this.
3. There is now an `XKitFeatureElement` class, which does not necessarily need to handle rendering beyond the template cloning... but in the spirit of the technology, I put as much into it as I could.

To be perfectly honest, I was expecting this to be a LoC net removal, what with this being a shiny new tech which I've seen at least one person call React obsolete for, so I'm a little disappointed that it's actually a 100+ line net increase for no behavioural changes. But, it was fun to do, and I think adopting this will (eventually) make it easier to do any visual overhauls of the control panel which, to be clear, [I very much want to do at some point](https://www.tumblr.com/april/784801883183284225/i-dont-know-how-i-didnt-figure-out-how-to-do)).

I have three goals in opening this PR up for review:
1. To see if you've got any reasons not to implement this! Dropping this _just_ because of the net LoC change seems a little silly, but it's certainly a strike against it.
2. To see if you've got any ideas for how to use this tech to refactor this code even better. I might be missing the forest for the trees here, for all I know.
3. Otherwise, to get this approved and merged. And once that's done, I'd like to try turning preferences into their own components too, because again, this was fun!

### Testing steps

1. Load the modified addon
2. Open the control panel
    - [x] **Expected result**: All features render in the correct order
    - [x] **Expected result**: All features render without their preferences open
    - [x] **Expected result**: All features render with icons
    - [x] **Expected result**: All features render with titles
    - [x] **Expected result**: Features that have descriptions render with descriptions
    - [x] **Expected result**: Features that do not have descriptions do not render with descriptions
    (nor do they reserve any space for a description; their titles sit in the middle of the summary box)
    - [x] **Expected result**: All features render with a working help link
    - [x] **Expected result**: Enabled features render with an enabled toggle
    - [x] **Expected result**: Disabled features render with a disabled toggle, and "(disabled)" appended to their titles
3. Enable a disabled feature
    - [x] **Expected result**: The "(disabled)" suffix disappears from the feature title
    - [x] **Expected result**: If the feature has any preferences, its preferences list is opened
    - [x] **Expected result**: If the feature does not have any preferences, its preferences list is not opened
    - [x] **Expected result**: The feature is added to `enabledScripts` in storage (see Backup tab)
4. Disable an enabled feature
    - [x] **Expected result**: The "(disabled)" suffix appears on the feature title
    - [x] **Expected result**: If the feature's preference list was open prior to disabling the feature, the list is closed
    - [x] **Expected result**: The feature is removed from `enabledScripts` in storage (see Backup tab)
5. Click a feature's displayed metadata
    - [x] **Expected result**: If its preferences list was closed, the list is opened
    - [x] **Expected result**: If its preferences list was open, the list is closed
6. Visually compare each preference type with its counterpart in the release version of the addon
    - [x] **Expected result**: `checkbox`-type preferences look identical
    - [x] **Expected result**: `text`-type preferences look identical
    - [x] **Expected result**: `color`-type preferences look identical
    - [x] **Expected result**: `select`-type preferences look identical
    - [x] **Expected result**: `textarea`-type preferences look identical
    - [x] **Expected result**: `iframe`-type preferences look identical
7. Type something into the search box
    - [x] **Expected result**: Features that have matching text in their titles are shown
    - [x] **Expected result**: Features that have matching text in their descriptions are shown
    - [x] **Expected result**: Features that have matching text in their related terms metadata are shown
    - [x] **Expected result**: Features that have matching text in their preferences are shown
    - [x] **Expected result**: When a preference matches your search text, that preference is highlighted
    - [x] **Expected result**: Non-matching features are not shown
8. Change the enabled/disabled filter
    - [x] **Expected result**: When filtering to enabled features only, disabled features are hidden
    - [x] **Expected result**: When filtering to disabled features only, enabled features are hidden
9. Use search+filter to make all features non-matching
    - [x] **Expected result**: All features are hidden
    - [x] **Expected result**: The "No results found" empty state is shown
10. Undo your search/filter to show some features again
    - [x] **Expected result**: The "No results found" empty state disappears


marcustyphoon:
> tech which I've seen at least one person call React obsolete for

A statement which would make me immediately skeptical that said person has sufficient relevant knowledge to be worth listening to, but yeah, I've heard that sentiment at least once as well :D

Random big-picture thoughts:

- There's a hypothetical lit-element version of this PR. Lit is kind of a pain to get working without a build step, but we should be able to vendor it (and potentially use an import map? it looks like import maps are usable with our browser support setup). Lit-element is still, imho, full of boilerplate if you can't use decorators, and native js decorators are Absolutely Not A Thing Yet, so I'm not sure if there's really a point. It does mean component props aren't Kind Of Stupid though, is my understanding.
- There's a hypothetical preact htm version of this PR. My guess is that it's quite clean looking. But that's obviously a whole different story.

Edit: `it looks like import maps are usable with our browser support setup` welp, no, they violate csp when inline and external import maps are not a thing. fun!

marcustyphoon:
Hey, wanna see something silly?

https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon:XKit-Rewritten:decorative-custom-elements I was going to make some of the xkit-inserted DOM elements have custom names in the devtools panel instead of `<div>`, jokingly demonstrating that you can register custom elements in the tumblr page as long as you don't need them to actually do anything. Turns out you don't actually have to register them at all, then, so you can just kind of... do this.

src/action/popup.js
```diff
@@ -1,64 +1,60 @@
...
+  const featureElements = [...document.querySelectorAll('xkit-feature')];
+  const preferenceElements = featureElements.flatMap(({ shadowRoot }) => [...shadowRoot.querySelectorAll('li')]);
+
+  featureElements.forEach(featureElement => {
+    const textContent = featureElement.textContent.toLowerCase();
+    const relatedTerms = featureElement.dataset.relatedTerms.toLowerCase();
+    const shadowContent = featureElement.shadowRoot.textContent.toLowerCase();
+
+    const hasMatch = textContent.includes(query) || relatedTerms.includes(query) || shadowContent.includes(query);
+    featureElement.classList.toggle('search-hidden', !hasMatch);
```
marcustyphoon, line 30: I cannot _believe_. that has been a thing. the whole time.

AprilSylph: Yuuuuup.

src/action/popup.html
```diff
@@ -4,15 +4,20 @@
...
-    <link rel="stylesheet" href="backup.css">
-    <link rel="stylesheet" href="links.css">
+    <link rel="icon" href="../icons/128.png" type="image/png">
+    <link rel="stylesheet" href="../lib/normalize.min.css">
+    <link rel="stylesheet" href="../lib/remixicon/remixicon.css">
+    <link rel="stylesheet" href="../lib/spectrum.css">
+    <link rel="stylesheet" href="./backup.css">
+    <link rel="stylesheet" href="./configuration.css">
+    <link rel="stylesheet" href="./links.css">
+    <link rel="stylesheet" href="./popup.css">
```
AprilSylph, line 14: Decided to use relative notation consistently here so that 1) IDEs can actually locate these files and 2) alphabetising these lines always puts vendored stuff first

src/action/popup.html
```diff
@@ -4,15 +4,20 @@
...
+    <script defer src="../lib/browser-polyfill.min.js"></script>
+    <script defer src="../lib/jquery.min.js"></script>
+    <script defer src="../lib/spectrum.js"></script>
+    <script defer src="./popup.js"></script>
+    <script defer src="./render_backup.js"></script>
+    <script defer src="./render_features.js"></script>
```
AprilSylph: Same here; in the case of JS, putting the vendored stuff first is actually necessary.

src/action/popup.html
```diff
@@ -84,22 +89,26 @@ <h4>Related projects</h4>
       </section>
     </main>
 
-    <template id="feature">
-      <details class="feature">
+    <template id="xkit-feature">
+      <link rel="stylesheet" href="../lib/normalize.min.css">
+      <link rel="stylesheet" href="../lib/remixicon/remixicon.css">
+      <link rel="stylesheet" href="../lib/spectrum.css">
```
AprilSylph: These both have to be duplicated between the main file and the component, sadly.

1. `remixicon.css` is necessary in the document for its `@font-face` definition to take effect (WHY!?) but is also necessary in the component so that the `ri-*` classnames actually apply their glyphs.
    - **musings**: We should probably look at using the SVGs directly instead? Way less of a pain all around.
2. `spectrum.css` is necessary in the document to style the colour picker popups, but also necessary in the component to style the input buttons.
    - **musings**: I wish we had a better library for colour-picking that didn't have this problem.

src/action/render_features.js
```diff
@@ -133,9 +134,148 @@ const renderPreferences = async function ({ featureName, preferences, preference
...
+  #preferences = {};
+
+  set preferences (preferences = {}) {
+    if (Object.keys(preferences).length === 0) return;
+
+    renderPreferences({
+      featureName: this.#featureName,
+      preferences,
+      preferenceList: this.#preferencesList
+    });
```
AprilSylph: `<slot>`-ifying preferences as well is one of the follow-ups I have planned if this gets merged...

src/action/render_features.js
```diff
@@ -133,9 +134,148 @@ const renderPreferences = async function ({ featureName, preferences, preference
...
+
+  get relatedTerms () {
+    return this.#relatedTerms;
+  }
+
+  render ({
+    description = '',
+    icon = {},
+    title = this.#featureName
+  }) {
```
marcustyphoon: Is there a specific reason that this doesn't go in `connectedCallback`? Standardizing on a pattern that makes sense for reactive components, I assume?

Also, is there a reason not to reference XKitFeatureElement properties for these arguments? (Or attributes, of course, but the attribute system on custom elements is... yikes.) Fewer loc?

marcustyphoon:
Not sure how much this is worth it, but some of the custom component setup logic is extractable:

```js
class CustomComponent extends HTMLElement {
  shadowRoot;

  constructor () {
    super();
    const { content } = document.getElementById(this.localName);
    this.shadowRoot = this.attachShadow({ mode: 'open' });
    this.shadowRoot.replaceChildren(content.cloneNode(true));
  }
}

class XKitFeatureElement extends CustomComponent {
  #detailsElement;
  #enabledInput;
  #helpAnchor;
  #preferencesList;

  constructor () {
    super();
    this.#detailsElement = this.shadowRoot.querySelector('details');
    this.#enabledInput = this.shadowRoot.querySelector('input[type="checkbox"]');
    this.#helpAnchor = this.shadowRoot.querySelector('a.help');
    this.#preferencesList = this.shadowRoot.querySelector('ul.preferences');
  }
```

src/action/render_features.js
```diff
@@ -13,29 +13,30 @@ const getInstalledFeatures = async function () {
   return installedFeatures;
 };
 
-const writeEnabled = async function ({ currentTarget }) {
+const handleEnabledInputClick = async function ({ currentTarget }) {
   const { checked, id } = currentTarget;
-  const detailsElement = currentTarget.closest('details');
+  const shadowRoot = currentTarget.getRootNode();
+  const featureElement = shadowRoot.host;
```
marcustyphoon: If this handler was a class method, I assume you could just reference `this`. Whether or not you think that's good organization, though, I don't know. (I don't really have any examples of a real app that uses vanilla js web components to take inspiration from :D)

src/action/render_features.js
AprilSylph: Reasons I had for a `render` function:
- Semi-guaranteeing that `featureName` is set prior to trying to access it (this would also work via public properties + `connectedCallback`)
- Not overriding what the `title` property means (although we don't make use of it, so I guess that's silly)
- I don't know if `connectedCallback` runs before or after painting (I should have just tested it to find out)

I am noticing that the MDN article on custom elements says "The specification recommends that, as far as possible, developers should implement custom element setup in this callback rather than the constructor" so I should at the very least have a go at doing... that.

AprilSylph:
> Not sure how much this is worth it, but some of the custom component setup logic is extractable:

Yesss I had the same idea over the weekend! I think that will be part of the preferences componentisation PR.

src/action/render_features.js
AprilSylph: I did actually try this, and it didn't work for some reason, but I should have another go at it...

src/action/render_features.js
AprilSylph: No visible flash moving this entire block to `connectedCallback`, specification recommendation wins.

src/action/render_features.js
AprilSylph: It was because I didn't make it an arrow function. I am so silly. Now, in my defense, the getters and setters aren't arrow functions and _their_ `this` refers to the class instance...

AprilSylph:
Woohoo, LoC bloat reduced significantly by following the damn spec!

src/action/render_features.js
```diff
@@ -133,88 +104,151 @@ const renderPreferences = async function ({ featureName, preferences, preference
...
+  /** @type {boolean} True if the feature can be enabled. Defaults to `false`. */
+  #disabled = false;
+
+  set disabled (disabled = false) {
+    this.#detailsElement.classList.toggle('disabled', disabled);
+    this.#enabledToggle.checked = !disabled;
+    this.#disabled = disabled;
+  }
+
+  get disabled () { return this.#disabled; }
```
marcustyphoon: I just checked, and you totally can configure eslint to allow (or force!) no newline between specific class fields. That would be quite extreme for a style tweak I could see some/you not even agreeing with though :D

AprilSylph: Yeah, I'd prefer keeping the style rules as they are. Don't want us to bikeshed ourselves.

marcustyphoon: Yep, it's not bad now that there's only one of them.

marcustyphoon:
I assume you've already considered (and either decided not to, or decided to defer) putting the component in a file in the components/ directory. Certainly if you did want to do that it would probably make sense to do it along with changes to the preference rendering.

AprilSylph:
> I assume you've already considered (and either decided not to, or decided to defer) putting the component in a file in the components/ directory. Certainly if you did want to do that it would probably make sense to do it along with changes to the preference rendering.

Yeah, I'm thinking a lot about what I want that directory to be like, really. So far I'm thinking...

```
components/
├── index.js            export class CustomElement { … }
├── <component-name>/   
│   ├── index.css
│   └── index.js        import { CustomElement } from '../index.js';
                        export class NameOfElement extends CustomElement { … }
```

But that makes more sense to put effort into as part of having, um, more than one component. Please consider the current directory structure to be a stupid little baby.

marcustyphoon:
Ooh. Hm. This is... interesting. I observe a significant FOUC... only in Chrome.

Note that this isn't the FOUC that mostly comes up in search results about web components, since we're of course inserting components after defining them; as noted by the [lit docs](https://lit.dev/docs/components/styles/#external-stylesheet) and [this csstricks](https://css-tricks.com/styling-a-web-component/#aa-link-to-external-styles-instead), this is specific to using linked external stylesheets (inlining the styles fixes this; connectedCallback is unrelated as I tested before and after that change; `blocking="render"` is not relevant; a style element with `@import url()` rules does the same thing). 

This does work, though it's a bit... well...

```js
Promise.all(
  this.#styleLinkElements.map(element => new Promise(resolve => element.addEventListener('load', resolve, { once: true })))
).then(() => {
  this.#detailsElement.hidden = false;
});
```

I don't know off the top of my head if you can fetch styles with fetch and put them in via cssom before inserting the element, but that's—no.

marcustyphoon:
> I assume you've already considered (and either decided not to, or decided to defer) putting the component in a file in the components/ directory.

Side thought related to this: I guess it's kind of literally impossible to do this cleanly with pure native web components, because the markup is in a template element in the html file in which one wants to use the component and can't be with the component code and css. No native html partials.

(Also um. Unless there's something I'm not thinking of, I guess one can't make a "totally native" web component that can be... imported... as a javascript module? Without having it use `new DOMParser().parseFromString`, anyway? So that's nice.)

AprilSylph:
What an annoying state of affairs. Chrome continues to be bad for my health.

I had a big comment written up saying "fuck it, let's just put everything in the module file, I keep hearing the term 'CSS-in-JS' everywhere anyway", but then I realised that the vendored CSS makes everything ugly by requiring either:
1. Using `@import` in the inline style element, which will resolve files... relative to the document the style element is inserted into... which would be extremely unsatisfying from a theoretical reusability point-of-view, or
2. Having to, actually, fetch every single stylesheet (as their own files) in the JS via top-level await, and then clone them into every new constructed custom element as inline styles, which feels like a lot of work that I can't think of how to helper-ise?

`src/action/components/xkit-feature/index.css`
```css
details {
  border-bottom: 1px dotted rgb(var(--active-grey));
}

/* etc. */
```

`src/action/components/xkit-feature/index.js`
```js
const componentDocument = new DOMParser().parseFromString(`
  <template id="xkit-feature">
    <details>
      <summary>
        <div class="icon">
          <slot name="icon"></slot>
        </div>
        <div class="meta">
          <h4 class="title"><slot name="title"></slot></h4>
          <p class="description"><slot name="description"></slot></p>
        </div>
        <div class="buttons">
          <a class="help" target="_blank"><i class="ri-fw ri-question-fill" style="color:rgb(var(--black))"></i></a>
          <input type="checkbox" checked class="toggle-button" aria-label="Enable this feature">
        </div>
      </summary>
      <ul class="preferences"></ul>
    </details>
  </template>
`);

const stylesheet = document.createElement('style');
const css = await fetch('./index.css').then(file => file.text());
stylesheet.replaceChildren(css);

export class XKitFeatureElement extends HTMLElement {
  constructor () {
    super();

    const { content } = componentDocument.getElementById(this.localName);
    const shadowRoot = this.attachShadow({ mode: 'open' });
    shadowRoot.replaceChildren(stylesheet.cloneNode(true), content.cloneNode(true));
  }
}
```

See, I gave up with just one stylesheet here. There are four more that we actually need... and then when we create the preference components, the feature component will only need three vendored stylesheets, while the preference components will need just one or two.

I guess what I really want is the as-of-now-fictional `CustomElement` class to somehow programmatically read the template's stylesheets, remove them from the template, fetch their contents, and insert them as inline styles into the template... before any class that extends `CustomElement` gets to run its own `constructor` code.

image

Anyway we should put this on hold because Tumblr just launched the redesigned footer. FUCK

marcustyphoon:
Hmm, lots of stuff up in the air with this one. I'll list some quick notes off the top of my head that may save some time rather than waiting for thoughts to coalesce:

- `Using @import in the inline style element` - doesn't help with this, stylesheets still load async
- `fetch every single stylesheet (as their own files) in the JS via top-level await, and then clone them into every new constructed custom element as inline styles` don't actually make `<style>` dom nodes, push to: shadowroot.adoptedStyleSheets: https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#constructable_stylesheets
- `which feels like a lot of work that I can't think of how to helper-ise?` I feel like it wouldn't be _terrible_?

```js
// component
const componentDocument = /* */;

const externalStyleSheets = await fetchStyleSheets('./some/path/filename.css', './some/other/path/filename.css', 'etc');

class SomeComponent extends XKitComponent {
  constructor () {
    super(componentDocument , externalStyleSheets);
  }

  /* whatever */
}

customElements.define('some-component', SomeComponent);

export default 'some-component';

// util
export class XKitComponent extends HTMLElement {
  shadowRoot;

  constructor (componentDocument, externalStyleSheets) {
    super();

    const { content } = componentDocument.getElementById(this.localName);
    this.shadowRoot = this.attachShadow({ mode: 'open' });
    this.shadowRoot.replaceChildren(content.cloneNode(true));

    this.shadowRoot.adoptedStyleSheets = externalStyleSheets;
  }
}

// util

const cache = {};

export const fetchStyleSheets = (...paths) => {
  paths.forEach(path => {
    cache[path] ??= fetch(path).then(response => response.text()).then(text => new CSSStyleSheet().replaceSync(text));
  })
  return Promise.all(paths.map(path => cache[path]);
}
```

or something idk I wrote this in the github editor. that breaks relative paths probably so I guess some tweaking required. ~~also there should be a cache in that util.~~ anyway.

- there's a world in which one is using these kind of components for all of their markup, in which case they could make all of their "css" files actually js modules with ```export default '/* entire css file goes here */'```. then you can just import css files as part of the regular waterfall. we obviously do not live in that world.
- there is another world where firefox supports `import externalStylesheet from './whatever.css' with { type: 'css' };`. we also do not live in that world and I don't think it's possible to polyfill.
- ~~wait... [it might be possible to polyfill.](https://github.com/guybedford/es-module-shims#css-modules)~~
- ~~holy shit. well, feel free to look into that if you'd like; I have no idea if it's usable here given our restrictions.~~ nope, to no one's surprise, the polyfill violates mv3 csp

marcustyphoon:
probably a possible way to allow relative paths:

```js

// module that wants to use relative paths
const externalStyleSheets = await fetchStyleSheets(['./some/path/filename.css', 'etc'], import.meta.url);

// module that wants to use absolute paths
const externalStyleSheets = await fetchStyleSheets(['/some/path/filename.css', 'etc']);

// util
const cache = {};

export const fetchStyleSheets = (paths, base) => {
  paths.forEach(path => {
    const resolvedPath = new URL(path, base).href;
    cache[resolvedPath] ??= fetch(resolvedPath).then(response => response.text()).then(text => new CSSStyleSheet().replaceSync(text));
  })
  return Promise.all(paths.map(path => cache[path]);
}
```

actually, no, I think either one of these is cleaner; the boilerplate is comprehensible without reading the util jsdoc:

```js

// module that wants to use relative paths
const externalStyleSheets = await fetchStyleSheets(['./some/path/filename.css', 'etc'].map(import.meta.resolve));

// module that wants to use absolute paths
const externalStyleSheets = await fetchStyleSheets(['/some/path/filename.css', 'etc']);

// util
const cache = {};

export const fetchStyleSheets = (paths) => {
  paths.forEach(path => {
    cache[path] ??= fetch(path).then(response => response.text()).then(text => new CSSStyleSheet().replaceSync(text));
  })
  return Promise.all(paths.map(path => cache[path]);
}
```

```js

// module that wants to use relative paths
const externalStyleSheets = await fetchStyleSheets(...['./some/path/filename.css', 'etc'].map(import.meta.resolve));

// module that wants to use absolute paths
const externalStyleSheets = await fetchStyleSheets('/some/path/filename.css', 'etc');

// util
const cache = {};

export const fetchStyleSheets = (...paths) => {
  paths.forEach(path => {
    cache[path] ??= fetch(path).then(response => response.text()).then(text => new CSSStyleSheet().replaceSync(text));
  })
  return Promise.all(paths.map(path => cache[path]);
}
```

but, like, is any of that worth it over just using absolute paths? eh

_edit:_ I wonder if you can pass import.meta.resolve as a function parameter. that would be pretty wild lmao

AprilSylph:
Okay, new plan for this PR:

- I implement the utils we need/want, including the base `CustomElement` class and your stylesheet util
- I implement the idealised directory structure that I previously talked about doing in a follow-up PR
- I don't throw away or redo anything I've done here, since that would be rude when you've already started reviewing/collaborating on this
- The PR gets bigger, inevitably, but I leave it up to you to decide if we should split it up into smaller chunks for easier reviewing, instead of following my instinct to do that unprompted
- ...I don't do any of this today, since it's already 4:10pm somehow (I think I might be addicted to Pokémon GO PvP...)

marcustyphoon:
I don't mind if you redo it! I throw around multiple versions of diffs of what I assume are comparable sizes all the time. If there's a way to organize things that makes sense to you that includes (say) closing this PR and creating more that mention it as prior art, I say go for it. It's not like we're keeping enough code that testing won't be redone.

(Aside: interesting situation comparison with fixing custom palettes in Palettes for Tumblr.)

AprilSylph:
FOUC on Chrome appears to be fixed now...

marcustyphoon:
I can't find definitive documentation about whether the `<!DOCTYPE html><html><body>` is necessary. In practice it doesn't seem to be and `new DOMParser().parseFromString(` examples often call it with bare elements.

Also, would we rather do ```const template = ` ``` and factor the parsing into XKitFeatureElement?

marcustyphoon:
Chromium 105:

1. Load the modified addon
2. Open the control panel
    - [x] **Expected result**: All features render in the correct order
    - [x] **Expected result**: All features render without their preferences open
    - [x] **Expected result**: All features render with icons
    - [x] **Expected result**: All features render with titles
    - [x] **Expected result**: Features that have descriptions render with descriptions
    - [x] **Expected result**: Features that do not have descriptions do not render with descriptions
    (nor do they reserve any space for a description; their titles sit in the middle of the summary box)
    - [x] **Expected result**: All features render with a working help link
    - [x] **Expected result**: Enabled features render with an enabled toggle
    - [x] **Expected result**: Disabled features render with a disabled toggle, and "(disabled)" appended to their titles
3. Enable a disabled feature
    - [x] **Expected result**: The "(disabled)" suffix disappears from the feature title
    - [x] **Expected result**: If the feature has any preferences, its preferences list is opened
    - [x] **Expected result**: If the feature does not have any preferences, its preferences list is not opened
    - [x] **Expected result**: The feature is added to `enabledScripts` in storage (see Backup tab)
4. Disable an enabled feature
    - [x] **Expected result**: The "(disabled)" suffix appears on the feature title
    - [x] **Expected result**: If the feature's preference list was open prior to disabling the feature, the list is closed
    - [x] **Expected result**: The feature is removed from `enabledScripts` in storage (see Backup tab)
5. Click a feature's displayed metadata
    - [x] **Expected result**: If its preferences list was closed, the list is opened
    - [x] **Expected result**: If its preferences list was open, the list is closed
6. Visually compare each preference type with its counterpart in the release version of the addon
    - [x] **Expected result**: `checkbox`-type preferences look identical
    - [x] **Expected result**: `text`-type preferences look identical
    - [x] **Expected result**: `color`-type preferences look identical
    - [x] **Expected result**: `select`-type preferences look identical
    - [x] **Expected result**: `textarea`-type preferences look identical
    - [x] **Expected result**: `iframe`-type preferences look identical
6. Use each preference type
    - [x] **Expected result**: `checkbox`-type preferences function
    - [x] **Expected result**: `text`-type preferences function
    - [x] **Expected result**: `color`-type preferences function
    - [x] **Expected result**: `select`-type preferences function
    - [ ] **Expected result**: `textarea`-type preferences function
    - [x] **Expected result**: `iframe`-type preferences function
7. Type something into the search box
    - [x] **Expected result**: Features that have matching text in their titles are shown
    - [x] **Expected result**: Features that have matching text in their descriptions are shown
    - [x] **Expected result**: Features that have matching text in their related terms metadata are shown
    - [x] **Expected result**: Features that have matching text in their preferences are shown
    - [x] **Expected result**: When a preference matches your search text, that preference is highlighted
    - [x] **Expected result**: Non-matching features are not shown
8. Change the enabled/disabled filter
    - [x] **Expected result**: When filtering to enabled features only, disabled features are hidden
    - [x] **Expected result**: When filtering to disabled features only, enabled features are hidden
9. Use search+filter to make all features non-matching
    - [x] **Expected result**: All features are hidden
    - [x] **Expected result**: The "No results found" empty state is shown
10. Undo your search/filter to show some features again
    - [x] **Expected result**: The "No results found" empty state disappears

- [x] Smoke test in chrome://extensions/?options=[id]

Firefox 121:

1. Load the modified addon
2. Open the control panel
    - [x] **Expected result**: All features render in the correct order
    - [x] **Expected result**: All features render without their preferences open
    - [x] **Expected result**: All features render with icons
    - [x] **Expected result**: All features render with titles
    - [x] **Expected result**: Features that have descriptions render with descriptions
    - [x] **Expected result**: Features that do not have descriptions do not render with descriptions
    (nor do they reserve any space for a description; their titles sit in the middle of the summary box)
    - [x] **Expected result**: All features render with a working help link
    - [x] **Expected result**: Enabled features render with an enabled toggle
    - [x] **Expected result**: Disabled features render with a disabled toggle, and "(disabled)" appended to their titles
3. Enable a disabled feature
    - [x] **Expected result**: The "(disabled)" suffix disappears from the feature title
    - [x] **Expected result**: If the feature has any preferences, its preferences list is opened
    - [x] **Expected result**: If the feature does not have any preferences, its preferences list is not opened
    - [x] **Expected result**: The feature is added to `enabledScripts` in storage (see Backup tab)
4. Disable an enabled feature
    - [x] **Expected result**: The "(disabled)" suffix appears on the feature title
    - [x] **Expected result**: If the feature's preference list was open prior to disabling the feature, the list is closed
    - [x] **Expected result**: The feature is removed from `enabledScripts` in storage (see Backup tab)
5. Click a feature's displayed metadata
    - [x] **Expected result**: If its preferences list was closed, the list is opened
    - [x] **Expected result**: If its preferences list was open, the list is closed
6. Visually compare each preference type with its counterpart in the release version of the addon
    - [x] **Expected result**: `checkbox`-type preferences look identical
    - [x] **Expected result**: `text`-type preferences look identical
    - [x] **Expected result**: `color`-type preferences look identical
    - [x] **Expected result**: `select`-type preferences look identical
    - [x] **Expected result**: `textarea`-type preferences look identical
    - [x] **Expected result**: `iframe`-type preferences look identical
6. Use each preference type
    - [x] **Expected result**: `checkbox`-type preferences function
    - [x] **Expected result**: `text`-type preferences function
    - [x] **Expected result**: `color`-type preferences function
    - [x] **Expected result**: `select`-type preferences function
    - [ ] **Expected result**: `textarea`-type preferences function
    - [x] **Expected result**: `iframe`-type preferences function
7. Type something into the search box
    - [x] **Expected result**: Features that have matching text in their titles are shown
    - [x] **Expected result**: Features that have matching text in their descriptions are shown
    - [x] **Expected result**: Features that have matching text in their related terms metadata are shown
    - [x] **Expected result**: Features that have matching text in their preferences are shown
    - [x] **Expected result**: When a preference matches your search text, that preference is highlighted
    - [x] **Expected result**: Non-matching features are not shown
8. Change the enabled/disabled filter
    - [x] **Expected result**: When filtering to enabled features only, disabled features are hidden
    - [x] **Expected result**: When filtering to disabled features only, enabled features are hidden
9. Use search+filter to make all features non-matching
    - [x] **Expected result**: All features are hidden
    - [x] **Expected result**: The "No results found" empty state is shown
10. Undo your search/filter to show some features again
    - [x] **Expected result**: The "No results found" empty state disappears

- [x] smoke test in about:addons

src/action/components/xkit-feature/index.js
```diff
@@ -0,0 +1,233 @@
...
+        preferenceLabel.setAttribute('for', `${featureName}.${preference.type}.${key}`);
+      } else {
+        preferenceInput.title = preference.label || key;
+      }
+
+      switch (preference.type) {
+        case 'iframe':
+          break;
+        default:
+          preferenceInput.addEventListener('change', this.#writePreference);
```
marcustyphoon, line 111: This fires only when the user deselects the text/textarea. I see this was its own commit so I assume the change is intentional/the debounce method doesn't work without additional changes, but this feels broken to me. Is there a way to make the debounce work? We don't necessarily have to use a closure.

marcustyphoon (changes requested):

marcustyphoon:
Fun off-topic observation: if you open about:addons in Firefox and navigate to an extension's preferences, then drag the browser tab so it becomes its own window, the extension preferences vanish until the tab is returned to the original window... with or without this PR. Sure, Firefox.

AprilSylph:
> I can't find definitive documentation about whether the `<!DOCTYPE html><html><body>` is necessary. In practice it doesn't seem to be and `new DOMParser().parseFromString(` examples often call it with bare elements.

Neat!

> Also, would we rather do `` const template = ` `` and factor the parsing into XKitFeatureElement?

Yes, that would be cleaner...

src/action/components/xkit-feature/index.js
AprilSylph: I wasn't able to find a way to close the popup without firing the `change` event on Firefox desktop, intending for one of us to search for a scenario in which it doesn't... and it sounds like you found one?

Either way, yes I should try harder to fix the debounce. It's such an odd issue—the `input` event's `target` for keypresses inside a web component is... not the `<input>`/`<textarea>` element, but the web component itself?? But `currentTarget` expires if you don't use it immediately (which is why I assume `writePreference` wasn't ever using it) so it'll be a fun one for sure.

AprilSylph:
> Also, would we rather do `` const template = ` `` and factor the parsing into XKitFeatureElement?

Wait, no, because that would do HTML parsing on every new instance of the custom element, instead of HTML parsing once upon module evaluation and then cloning the parsed template contents on every new instance.

src/action/components/xkit-feature/index.js
AprilSylph: Fixed in 7c91594776b1d2b6cf9c775250724a28d1c421fd. `currentTarget` doesn't expire if we immediately extract it from the event object, hooray. The code does look stupid, though.

marcustyphoon:
> Wait, no, because that would do HTML parsing on every new instance of the custom element, instead of HTML parsing once upon module evaluation and then cloning the parsed template contents on every new instance.

Oh... right. I had forgotten that (iirc) htm/preact and lit-html have caching built in, so their parse functions (tagged templates) can be called inside a class body without meaningful performance impact.

I suppose this is another thing we could export (`const domParser = new DOMParser(); export const getTemplateFragment = template => domParser.parseFromString(template, 'text/html').querySelector('template').content`) or use caching for, but on the other hand there's only so far one wants to go down that road before component files become unintelligible without going back and forth with the util file, at which point we've effectively just made an htm/preact or lit-html competitor with less documentation and which when learned confers a skill set that, while valuable, is probably not that likely to be reused :D

marcustyphoon:
> I wasn't able to find a way to close the popup without firing the `change` event on Firefox desktop, intending for one of us to search for a scenario in which it doesn't... and it sounds like you found one?

I did not. But basically all of our textarea preferences are ones which, in the current extension version, cause UI updates immediately when changed, and so that's how I tested them. It just so happens that all uses of `text` preferences (in non-deprecated features) are only queried on a user action, so I didn't notice before I got to `textarea` on the list of preference types.

> the `input` event's `target` for keypresses inside a web component is... not the `<input>`/`<textarea>` element, but the web component itself??

Hm... it doesn't seem to be for me? Oh, I see—it is when called later, because shadow DOM. There does [appear to be](https://stackoverflow.com/a/57963850) a [composedPath() function](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) which lets one query this.

Aside: not sure if it's cleaner to have writePreference not be a class method.

AprilSylph:
> I did not. But basically all of our textarea preferences are ones which, in the current extension version, cause UI updates immediately when changed, and so that's how I tested them. It just so happens that all uses of `text` preferences (in non-deprecated features) are only queried on a user action, so I didn't notice before I got to `textarea` on the list of preference types.

That... makes a lot of sense! :+1: 

> Aside: not sure if it's cleaner to have writePreference not be a class method.

I like having everything on the class so that all the top-level stuff becomes a repeatable consistent pattern, i.e. every component module will have the exact* same set of definitions/exports. (*Name variation notwithstanding...)

src/action/components/xkit-feature/index.js
```diff
@@ -0,0 +1,241 @@
...
+  set disabled (disabled = false) {
+    this.#detailsElement.classList.toggle('disabled', disabled);
+    this.#enabledToggle.checked = !disabled;
+    this.#disabled = disabled;
+  }
+
+  get disabled () { return this.#disabled; }
+}
+
+customElements.define(localName, XKitFeatureElement);
```
marcustyphoon, line 209: XKitFeatureElement could be inlined to a class expression.

src/action/render_features.js
```diff
@@ -149,72 +30,19 @@ const renderFeatures = async function () {
...
-      renderPreferences({ featureName, preferences, preferenceList });
+    const disabled = enabledFeatures.includes(featureName) === false;
+    if (disabled && metadata.deprecated && !specialAccess.includes(featureName)) {
+      continue;
     }
 
-    featureClones.push(featureTemplateClone);
+    const featureElement = XKitFeature();
+    Object.assign(featureElement, { disabled, featureName, ...metadata });
+    featureElements.push(featureElement);
```
marcustyphoon: Could be a oneliner `featureElements.push(XKitFeature({ disabled, featureName, ...metadata }));` with `export const XKitFeature = (properties = {}) => Object.assign(document.createElement(localName), properties);`

marcustyphoon (commented):
Definitely optional, but my standard workflow at the end of a large PR is to look for single-use variables that could be LOC reductions; here are a few.

src/action/popup.js
```diff
@@ -1,64 +1,60 @@
...
-      ) {
-        preferenceElement.classList.add('search-highlighted');
-      } else {
-        preferenceElement.classList.remove('search-highlighted');
-      }
-    });
+document.getElementById('search').addEventListener('input', ({ currentTarget }) => {
+  const query = currentTarget.value.toLowerCase();
+  const featureElements = [...document.querySelectorAll('xkit-feature')];
+  const preferenceElements = featureElements.flatMap(({ shadowRoot }) => [...shadowRoot.querySelectorAll('li')]);
```
marcustyphoon, line 22: The awkwardness of this line hints if anything that _preference elements_ being slotted sort of makes sense.

marcustyphoon (commented):
I guess I should mention that, besides to try out the API, I'm not sure what purpose using slots in this PR serves; having a component that sets its own children is a bit of an odd pattern. They seem primarily designed so external code can interact with the elements in question, which isn't the case for the three slotted elements here. But, also, you know, why not.

src/action/popup.js
AprilSylph: Preferences being slotted is definitely part of the overall plan.

> having a component that sets its own children is a bit of an odd pattern. They seem primarily designed so external code can interact with the elements in question, which isn't the case for the three slotted elements here.

You have a good point here... when building React components, a slot is just a prop that accepts another React element, so by definition a component can't have slots _and_ be in total control of those slots' content.

I will admit that the primary motivation I had for using slots here is just to try out the API, but it's also kind of nice having a declarative way to fill out the content, isn't it...? Although that argument falls apart when we still rely on `querySelector` for setting up all the event listeners, phooey.

The real benefit of slotted elements being outside of the shadow root is real, though. And, if `XKitFeatureElement` needs to be fed the right props in order to render in its own slot... maybe it really would just be best to remove that abstraction and create those slotted elements on the outside instead.

marcustyphoon: Yep, I figure we can revisit that along with preferences, which you said was next on the list, right?

marcustyphoon:
Safari in iOS 18.5:

1. Load the modified addon
2. Open the control panel
    - [x] **Expected result**: All features render in the correct order
    - [x] **Expected result**: All features render without their preferences open
    - [x] **Expected result**: All features render with icons
    - [x] **Expected result**: All features render with titles
    - [x] **Expected result**: Features that have descriptions render with descriptions
    - [x] **Expected result**: Features that do not have descriptions do not render with descriptions
    (nor do they reserve any space for a description; their titles sit in the middle of the summary box)
    - [x] **Expected result**: All features render with a working help link
    - [x] **Expected result**: Enabled features render with an enabled toggle
    - [x] **Expected result**: Disabled features render with a disabled toggle, and "(disabled)" appended to their titles
3. Enable a disabled feature
    - [x] **Expected result**: The "(disabled)" suffix disappears from the feature title
    - [x] **Expected result**: If the feature has any preferences, its preferences list is opened
    - [x] **Expected result**: If the feature does not have any preferences, its preferences list is not opened
    - [x] **Expected result**: The feature is added to `enabledScripts` in storage (see Backup tab)
4. Disable an enabled feature
    - [x] **Expected result**: The "(disabled)" suffix appears on the feature title
    - [x] **Expected result**: If the feature's preference list was open prior to disabling the feature, the list is closed
    - [x] **Expected result**: The feature is removed from `enabledScripts` in storage (see Backup tab)
5. Click a feature's displayed metadata
    - [x] **Expected result**: If its preferences list was closed, the list is opened
    - [x] **Expected result**: If its preferences list was open, the list is closed
6. Visually compare each preference type with its counterpart in the release version of the addon
    - [x] **Expected result**: `checkbox`-type preferences look identical
    - [x] **Expected result**: `text`-type preferences look identical
    - [x] **Expected result**: `color`-type preferences look identical
    - [x] **Expected result**: `select`-type preferences look identical
    - [x] **Expected result**: `textarea`-type preferences look identical
    - [x] **Expected result**: `iframe`-type preferences look identical
6. Use each preference type
    - [x] **Expected result**: `checkbox`-type preferences function
    - [ ] **Expected result**: `text`-type preferences function
    - [x] **Expected result**: `color`-type preferences function
    - [x] **Expected result**: `select`-type preferences function
    - [x] **Expected result**: `textarea`-type preferences function
    - [x] **Expected result**: `iframe`-type preferences function
7. Type something into the search box
    - [x] **Expected result**: Features that have matching text in their titles are shown
    - [x] **Expected result**: Features that have matching text in their descriptions are shown
    - [x] **Expected result**: Features that have matching text in their related terms metadata are shown
    - [x] **Expected result**: Features that have matching text in their preferences are shown
    - [x] **Expected result**: When a preference matches your search text, that preference is highlighted
    - [x] **Expected result**: Non-matching features are not shown
8. Change the enabled/disabled filter
    - [x] **Expected result**: When filtering to enabled features only, disabled features are hidden
    - [x] **Expected result**: When filtering to disabled features only, enabled features are hidden
9. Use search+filter to make all features non-matching
    - [x] **Expected result**: All features are hidden
    - [x] **Expected result**: The "No results found" empty state is shown
10. Undo your search/filter to show some features again
    - [x] **Expected result**: The "No results found" empty state disappears

marcustyphoon (approved):
Seems to work!

I think my review philosophy—for a piece of end-user software, maybe not so much a library—is that if there's only one intended way for the code to function (and affect e.g. storage) and the PR functions in that intended way, I'm comfortable hitting approve; we can always transparently refactor/change/revert/etc the implementation later without an effect to the user. Of course that goes double for a PR with "this will be changed later" on it.

AprilSylph:
> Yep, I figure we can revisit that along with preferences, which you said was next on the list, right?

I committed cad002c87cf720d3b86e3ea2f836503ad7a4abb6 here because it actually ended up being a small LoC saving (+27-34). This way there is no intermediary PR before tackling preferences, and the preferences PR can follow the pattern set here instead of having to itself establish the pattern we want.

...

Hitting merge on this myself immediately after adding two commits feels wrong, but I don't feel like re-review is required, either... maybe I'll just self-review and merge tomorrow?

AprilSylph:
Everything seems to work. How terrifying.

PR #1835, opened 2025-07-12
web components branch: use dom helper

marcustyphoon:
### Description

For your consideration: "Hmm, sure would be nice to have the dom() helper here. Hey... wait a second."

### Testing steps

See target branch PR (#1831) tests.

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -17,7 +17,7 @@
     <script defer src="../lib/spectrum.js"></script>
     <script defer src="./popup.js"></script>
     <script defer src="./render_backup.js"></script>
-    <script defer src="./render_features.js"></script>
+    <script type="module" defer src="./render_features.js"></script>
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Come to think of it, `render` is kind of unnecessary: https://github.com/marcustyphoon/XKit-Rewritten/compare/xkit-feature-web-component-dom...marcustyphoon:XKit-Rewritten:xkit-feature-web-component-dom-inline

AprilSylph:
Eh, I think for this piece of code, I'm less interested in achieving the absolute pinnacle of clean code (shock horror, I know) (...which, anyway, would need a new version of `dom()` that doesn't make attributes and callbacks separate arguments for no good reason) and more interested in seeing how far we can go with just vanilla JS (i.e. none of our framework stuff helping us; not sure if the WebExtension framework itself would be considered non-vanilla...?)

PR #1838, opened 2025-07-13
refactor demo: use htm/preact to render features in the control panel

marcustyphoon:
### Description

I needed an excuse to see if and how preact worked without a bundler or jsx, so, for funsies: here's #1831 using htm/preact. Obviously this doesn't implement any mutable state, so using the preact renderer (vs e.g. hooking htm up to our `dom()` function) is technically pointless, but now one could do so.

It appears that preact neither has officially stable `<Suspense>` (though apparently it's totally usable) nor [React's `use`](https://react.dev/reference/react/use), so converting more async stuff is probably fairly annoying.

### Testing steps

 See #1831 testing steps.


marcustyphoon:
see also: https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon:XKit-Rewritten:xkit-feature-preact-htm-preferences

marcustyphoon:

PR #1840, opened 2025-07-14
refactor demo: use lit-html to render features in the control panel

marcustyphoon:
### Description

Possibly part 2/3 of experimenting with no-bundle component frameworks (#1838 was part 1; part 3 would be lit web components). This uses [lit-html](https://lit.dev/docs/libraries/standalone-templates/)*, the template-string-based templating language used in google's lit framework. Because this only uses the templating language, there's no state hooks (you can call `render` repeatedly if you handle state yourself). We, of course, currently only render the preference pane once outside of backup restores, so lit-html would be sufficient for implementing the code as it currently exists.

Quite a fun thing to do is to look at how this syntax differs from `htm`. Not much, as it turns out; it basically just loses some syntax tricks like attribute spreading and loose html and uses some magic characters for event listeners and boolean attributes: https://github.com/AprilSylph/XKit-Rewritten/commit/09200145e420592b2b9aa598a340ed564df34527

Annoyingly, there's a standalone lit-html bundle but it doesn't include stuff like [`ref`](https://lit.dev/docs/templates/directives/#ref) and [`until`](https://lit.dev/docs/templates/directives/#until) and also isn't published on npm, so I pulled the library in a kind of messy way for the demo.

_*there's better documentation for the standalone library [in the lit v1 docs](https://lit.dev/docs/v1/lit-html/introduction/)_ for whatever reason.

### Testing steps

See https://github.com/AprilSylph/XKit-Rewritten/pull/1831 testing steps.


marcustyphoon:

PR #1841, opened 2025-07-14
chore(Shorten Posts): mark feature as deprecated

AprilSylph:
### Description

Deprecates the Shorten Posts feature in favour of the on-by-default "Shorten long posts" option available in Tumblr's own [dashboard settings](https://www.tumblr.com/settings/dashboard).

I actually meant to do this a lot sooner, as I saw it become a point of confusion even among fellow Tumblr staff to have the dashboard setting disabled but still have post shortening happen.

### Testing steps

1. Load the modified addon with a blank configuration
2. Look for Shorten Posts in the control panel
    - [ ] **Expected result**: The feature is not listed
3. Enable the feature via importing this config in the Backup tab:
    ```json
    { "enabledScripts": ["shorten_posts"] }
    ```
4. Go back to the Configuration tab
    - [ ] **Expected result**: Shorten Posts is listed and enabled, with the "(deprecated)" title suffix
5. Disable the feature
6. Close and reopen the control panel
    - [ ] **Expected result**: Shorten Posts is now listed and can be enabled normally

### DIFF
--- a/src/features/shorten_posts/feature.json
+++ b/src/features/shorten_posts/feature.json
@@ -26,5 +26,6 @@
       ],
       "default": "1"
     }
-  }
+  },
+  "deprecated": true
 }


marcustyphoon (approved):
- [x] it works
- [x] "Sure, I guess. I don't think it would be a valuable use of brainpower to try to come up with an opinion about this, and if we change our minds it's totally reversible."

PR #1848, opened 2025-07-20
Quick Reblog: Fix touchscreen long press context menu prevention

### LINKED ISSUE #1844: Quick Reblog on Mobile Firefox (Android) opening right click dialogue overlay
### Platform

Graphene OS (Android 16)

### Browser

Firefox 140.0.4 (Mobile/Android)

### Addon version

v1.0.12

### Details

Starting today (7/18/25), when using Tumblr on mobile Firefox for Android, if I long press the reblog button to open the XKit quick reblog menu, it will frequently additionally open the right click dialogue used for links over the top of everything else. 

I can't 100% tell what factors are involved in why it only does this some of the time, but it seems to be related to how long I hold the button. Unfortunately the difference in duration is like... milliseconds lol — I've been using the mobile website with XKit for a few months on this phone and prior to this morning it's never done this, with long press standing in for hovering without issue, and has otherwise worked seamlessly. 
 
It's a minor inconvenience but it is still annoying lol so, since I didn't find any existing threads with this issue (unless I missed something), I thought I would report in case anyone else is having a similar problem or knows a workaround.

Thanks for all your hard work!

marcustyphoon:
### Description

In #893, we added code to Quick Reblog that prevents the right-click context menu from appearing when a user with a touchscreen device long-presses a post's reblog button so that the Quick Reblog popup can appear instead. I tried to fix this code on certain accounts in #1823, but my fix was erroneous and breaks it on all accounts. This is the correct fix.

The problem is that we are using a JQuery delegated event listener for `preventLongPressMenu`. JQuery sets the currentTarget property on its synthetic event to the selected element, but the native event is actually on `document.body`.

Resolves #1844 (unless it doesn't; see below)

### Testing steps

You know, I'm not actually clear on whether this section is supposed to indicate that the PR author performed the listed tests or not.

Confirm that long-pressing the reblog button with Quick Tags enabled does not create/open a context menu:

- [x] In Chromium desktop, in responsive design mode simulating a touchscreen
- [x] In Firefox desktop, in responsive design mode simulating a touchscreen
- [ ] In Firefox for android

Confirm that long-pressing the reblog button with Quick Tags enabled opens the Quick Reblog menu:

- [x] In Chromium desktop, in responsive design mode simulating a touchscreen
- ❌ In Firefox desktop, in responsive design mode simulating a touchscreen
- [ ] In Firefox for android(?)

Depending on the results of this testing, #1849 may prove to be necessary?

### DIFF
--- a/src/features/quick_reblog/index.js
+++ b/src/features/quick_reblog/index.js
@@ -324,7 +324,6 @@ const updateRememberedBlog = async ({ currentTarget: { value: selectedBlog } }) 
 const MOZ_SOURCE_TOUCH = 5;
 
 const preventLongPressMenu = ({ currentTarget, originalEvent: event }) => {
-  console.log(currentTarget, currentTarget.matches(reblogButtonSelector));
   if (!currentTarget.matches(reblogButtonSelector)) return;
 
   const isTouchEvent = event.pointerType === 'touch';


AprilSylph:
> You know, I'm not actually clear on whether this section is supposed to indicate that the PR author performed the listed tests or not.

Nope. Testing steps are written by the PR author to document what they believe their changes are doing, and performed by the PR reviewer to validate those beliefs. Like a bug report, confidence in a fix or feature can only be attained by reproducing the same thing in a different environment.

marcustyphoon:
Oh, certainly I don't think that means the reviewer should skip them. But I could see an argument that the author should have performed them before undrafting the PR.

AprilSylph:
I'm assuming all references to Quick Tags in this PR are erroneous...?

marcustyphoon:
Oh, that's weird. I wonder why desktop Firefox doesn't simulate it correctly. I would be suspicious of desktop Firefox not working correctly with a touchscreen, but I did test that in the original implementation.

PR #1850, opened 2025-07-21
fix(Quick Reblog): close menu when Tumblr's reblog button is clicked

AprilSylph:
### Description

Fixes two little things that have been bugging me with Quick Reblog on the new post footer.

#### Problem no.1: Two menus

When the Quick Reblog menu is open, and Tumblr's reblog menu is then opened, the Quick Reblog menu doesn't close. Instead, Tumblr's reblog menu simply goes on top. This looks bad.

Because the Quick Reblog menu prevents `click` events from propagating upwards, this is fixed easily by adding a new event listener on the parent; if the parent receives a click event, the Quick Reblog menu is closed instantly.

This fix also affects the other post footer variants, naturally—it makes sense to me that if you click Tumblr's reblog button, you're signalling that you don't want to interact with Quick Reblog.

For the new footer only, this also involves adding another check before we show the Quick Reblog menu in the first place; if Tumblr's reblog button matches `[aria-expanded="true"]`, its own menu is open and we should refuse to show ours. Without this, it would still be possible to open both menus at once, just not quite as easily.

#### Problem no.2: DOOR STUCK

When opening the post notes via Tumblr's reblog menu, the Quick Reblog menu gets stuck open. This is because Ariakit (correctly) puts keyboard focus back on the reblog button once the reblog menu is closed; meanwhile, our `removePopupOnLeave` function requires the Quick Reblog menu's parent _not_ to match `:focus-within`, with the now-incorrect assumption that the only way that element or its descendants receive focus is through direct interaction or actual keyboard navigation.

We should instead be testing for direct indicators of intent to interact with Quick Reblog for refusing to remove its menu on mouseleave; namely Quick Reblog's activation condition (`:hover` on the reblog button) or Quick Reblog actively being interacted with (`:focus-within` on the popup element).

Before | After
-|-
<video src="https://github.com/user-attachments/assets/7110a7b2-2357-44e2-8abc-e4b309b67730"></video> | <video src="https://github.com/user-attachments/assets/d82caa0d-624a-4408-9d4e-296222db7ecc"></video>

#### All footer variants

1. Load the modified addon
2. Enable Quick Reblog
3. Hover over a post's reblog button
    - [ ] **Expected result**: The Quick Reblog menu appears
4. Click a post's reblog button while the Quick Reblog menu is still open
    - [ ] **Expected result**: The Quick Reblog menu disappears instantly
5. Hover over the reblog button again
    - [ ] **Expected result**: The Quick Reblog menu appears again
6. Repeat steps 3-5 for all three known post footer variants
    - [ ] **Expected result**: No issues arise with opening/closing the Quick Reblog menu

#### With `postFooterSplitNotesCountSingleAction`

1. Hover over a post's reblog button
    - [ ] **Expected result**: The Quick Reblog menu appears
2. Click that same reblog button
    - [ ] **Expected result**: The Quick Reblog menu disappears instantly
3. Click "See reblogs" in Tumblr's reblog menu
    - [ ] **Expected result**: The Tumblr reblog menu is closed
    - [ ] **Expected result**: The post's notes view is opened
4. Hover over the reblog button again
    - [ ] **Expected result**: The Quick Reblog menu appears again
5. Hover somewhere else that is not the reblog button or the Quick Reblog menu
    - [ ] **Expected result**: The Quick Reblog menu disappears on a slight delay

### DIFF
--- a/src/features/quick_reblog/index.js
+++ b/src/features/quick_reblog/index.js
@@ -142,14 +142,17 @@ tagsInput.addEventListener('input', checkLength);
 const showPopupOnHover = async ({ currentTarget }) => {
   if (!currentTarget.matches(reblogButtonSelector)) return;
 
-  const thisPost = currentTarget.closest(postSelector);
+  const buttonDiv = currentTarget.closest(buttonDivSelector) ?? currentTarget.parentElement;
+  if (buttonDiv.matches(':has([aria-expanded="true"])')) return;
 
+  const thisPost = currentTarget.closest(postSelector);
   const { blog, canReblog } = await timelineObject(thisPost);
   if (canReblog === false || blog?.isPasswordProtected) return;
 
   clearTimeout(timeoutID);
 
-  appendWithoutOverflow(popupElement, currentTarget.closest(buttonDivSelector) ?? currentTarget.parentElement, popupPosition);
+  appendWithoutOverflow(popupElement, buttonDiv, popupPosition);
+  popupElement.parentNode.addEventListener('click', removePopupOnClick, { once: true });
   popupElement.parentNode.addEventListener('mouseleave', removePopupOnLeave);
 
   const thisPostID = thisPost.dataset.id;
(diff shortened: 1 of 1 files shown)


marcustyphoon (approved):
- [x] tested in Chrome
- [x] tested in Safari
  - [x] found an unrelated bug when testing in Safari, whee

PR #1847, opened 2025-07-18
Quick Reblog: Fix already-reblogged button coloration in notes view

marcustyphoon:
### Description

On accounts without footer changes, the reblog icon in the mode selector in the notes view was colored green if the user has reblogged the post with quick reblog, etc. This makes sure the coloration is only applied to the correct button.

Not sure if there is a more elegant way to do this.

Resolves #1845.

### Testing steps

On all 3 footer variations:
- Confirm that the reblog button changes color when Quick Reblog is used to reblog a post.
- Confirm that nothing in the notes view changes color when Quick Reblog is used to reblog a post.

PR #1852, opened 2025-07-24
fix(Quick Reblog): restore AlreadyReblogged tick in new post footer variants

### LINKED ISSUE #1846: Quick Reblog: No already-reblogged button checkmark on accounts with new footer variant
### Platform

macOS 14.7.6 arm64

### Browser

chrome 138.0.7204.93

### Addon version

v1.0.12

### Details

In #1823 I omitted bringing the checkmark icon on already-reblogged posts to the new footer variant.

I can't remember if this was an oversight or if I had trouble doing it and intentionally shipped the PR without it, but I think I would have documented it in the case of the latter, so honestly I think I just didn't notice.

AprilSylph:
### Description

- Fixes #1846

Updates Quick Reblog to restore the AlreadyReblogged tick on already-reblogged posts' reblog buttons when the user has one of the two new 2025 post footer variants, via `mask-image` magic that I stole from my past self (the new post header's avatar component uses this to display cut-outs for badges/sub-avatars too).

The magic numbers for how the mask works is... part science, part magic. The first two magic numbers are the ellipsis dimensions, which should match the tick's bounding box dimensions. The positional magic numbers are... kind of just whatever worked? To figure out how to get a perfect overlap, I temporarily added `background-color: red` to the masked SVG element and `background-color: blue` to the tick pseudo-element. Then, I just played with the position until they fit into each other perfectly. For the "single action" reblog button variant, the numbers make sense, the position values for the tick and the mask being identical; for the other new variant, the numbers are backwards for some reason. 🤷‍♀️

This is all done in pure CSS, dropping the CSS map calls introduced in #1847, thanks to the observation that there are now only three reblog button variants:
1. `footer a[href^="/reblog/"]:has(use[href="#managed-icon__reblog"])`—the pre-redesign reblog button
2. `footer a[href^="/reblog/"]:has(use[href="#managed-icon__ds-reblog-24"])`—the "split notes count" reblog button
3. `footer button:has(use[href="#managed-icon__ds-reblog-24"])`—the "single action" reblog button

None of these selectors should target anything in the notes view; only the pre-redesign notes view has a reblog icon, and it's not part of a link. When combining these selectors to target all of them in one go, the `button` selector part must be swapped out for `button:not([role])` so as not to target the `button[role="tab"]` containing the old reblog icon in the pre-redesign notes view.

Oh, and this removes the outdated CSS that changes the tick position on mobile-sized viewports. Now that we're putting the tick on the reblog anchor/button directly, instead of the control button container, the difference in the container size on mobile doesn't affect us anymore. In fact, in attempting to still account for it, the tick just goes in the wrong corner.

Before | After
-|-
Screen Shot 2025-07-24 at 14 07 50 | Screen Shot 2025-07-24 at 14 08 09
Screen Shot 2025-07-24 at 14 10 41 | Screen Shot 2025-07-24 at 14 10 56
Screen Shot 2025-07-24 at 14 08 45 | Screen Shot 2025-07-24 at 14 09 01
Screen Shot 2025-07-24 at 14 09 31 | Screen Shot 2025-07-24 at 14 09 51

### Testing steps

1. Load the modified addon
2. Enable Quick Reblog
3. Reblog a post using Quick Reblog
    - [ ] **Expected result**: The reblog button in the post footer turns green/purple/red
    - [ ] **Expected result**: A tick is added to the reblog button
    - [ ] **Expected result**: The tick does not touch or overlap with the reblog button; the bottom-right corner of the icon disappears to make room for the tick
4. Confirm that the above holds true on all known post footer variants
    - [ ] Pre-redesign
    - [ ] `postFooterSplitNotesCount`
    - [ ] `postFooterSplitNotesCountSingleAction`

### DIFF
--- a/src/features/quick_reblog/index.css
+++ b/src/features/quick_reblog/index.css
@@ -248,13 +248,6 @@ footer:is(.published, .queue, .draft) button:has(use[href="#managed-icon__ds-reb
   left: 21px;
 }
 
-@media (max-width: 989px) {
-  footer:is(.published, .queue, .draft) :is(button:not([role]), a[href^="/reblog/"]):has(use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]))::after {
-    bottom: calc((27px / 2) - 2px);
-    right: calc((27px / 2) - 4px);
-  }
-}
-
 footer.published :is(button:not([role]), a[href^="/reblog/"]):has(use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]))::after { color: rgb(var(--green)); }
 footer.queue :is(button:not([role]), a[href^="/reblog/"]):has(use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]))::after { color: rgb(var(--purple)); }
 footer.draft :is(button:not([role]), a[href^="/reblog/"]):has(use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__reblog"]))::after { color: rgb(var(--red)); }
--- a/src/features/quick_reblog/index.css
+++ b/src/features/quick_reblog/index.css
@@ -195,7 +195,25 @@ footer:is(.published, .queue, .draft) :is(button:not([role]), a[href^="/reblog/"
   position: relative;
 }
 
-/* Common styles for tick pseudo-element */
+/* Cut-out for tick in 2025 post footer, "split notes count" variant */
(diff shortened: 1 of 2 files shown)


AprilSylph:
Tested on Firefox on macOS, Ubuntu, and Windows 10; the tick position changes slightly due to the available fonts on each OS, but the bounding box of the pseudo-element doesn't seem to, at least not significantly enough to break the cut-out effect on the redesign variants.

marcustyphoon:
Ooh, fancy.

Definitely one of those style PRs that's more readable using selector composition in js, but I assume you prefer the use of static styles when possible.

marcustyphoon (approved):
Tested in Chrome 105 and Safari 18.5 on macOS.

PR #1860, opened 2025-08-07
New Tweak: Show tags on filtered posts

### LINKED ISSUE #1664: Feature: Show tags on filtered posts
### Concept

One thing that New XKit's Blacklist has which we don't is the ability to show the tags on filtered posts (not just the one that's filtered). This would probably be quite simple to add.

marcustyphoon:
### Description

Resolves #1664.


marcustyphoon:
This layout is somewhat limited by the CSS-only method; we could certainly insert more DOM elements to have more flexibility. That I can think of, it would require adding wrappers (as we did in #562, for example), which I generally like to avoid.

One thing I thought about was putting `float: right` on the view post button in the collapsed layout; would that enable vertically centering it in this case? Is that better? Eh. I'm not much of a designer.

AprilSylph:
Is this a draft or is it pending my review? It can't be both.

* If the PR is considered non-viable in its current state, it should be a draft
* If a draft is kept open, that indicates that someone is expected to work on it until it is viable
* If the person expected to work on the draft is not the PR author, they should be an assignee on the draft
* If the PR is considered viable in its current state by the author or assignee, it should be marked as ready for review

PR #1865, opened 2025-08-12
Adjust menu backup section layout for chromium compatibility

marcustyphoon:
### Description

Fun fact: setting `display: flex` or setting margin on the child of a `details` element in chromium doesn't seem to work. (I see `slot` in chromium devtools; maybe `details` is implemented as a custom element?) Anyway, this results in slightly incorrect spacing in the backup section of the XKit Rewritten menu, and is kind of a pain if you try to add an element to it.

This PR is... I dunno, one way to solve this. There may be better ones. As mentioned a number of times, I consider myself a CSS layout... intermediate, I guess, at best?

### Testing steps

Confirm that the backup section of the XKit Rewritten menu has correct-looking layout in both Firefox and Chrome when either/neither element is expanded.

### DIFF
--- a/src/action/backup.css
+++ b/src/action/backup.css
@@ -1,7 +1,13 @@
 #backup details {
+  padding: 1ch;
+}
+
+#backup details > .content {
+  margin-top: 0.5em;
+
   display: flex;
   flex-direction: column;
-  padding: 1ch;
+  gap: 0.5em;
 }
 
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
:old-man-yells-at-chromium:
:old-man-yells-at-chrome:
:old-man-yells-at-google:

PR #1868, opened 2025-08-19
feat(AccessKit Disable GIFs): Pause "background" gifs instead of blanking them

marcustyphoon:
### Description

We currently have a bit of workaround behavior where AccessKit "pauses" a small number of GIFs in the Tumblr UI by making them a solid color when they're not hovered. This occurs when Tumblr sets the background-image CSS property to a value that contains an animated image, which our code doesn't know how to freeze.

This correctly pauses those GIFs. The technique used is to parse the background-image property string, extract the image URL, fetch the URL, create a canvas element containing its first frame, then create a blob URL containing the paused image and replace the image URL in the property string with this new paused image URL when the element is not hovered. This preserves the rest of the property string, so things like gradients that are added to the animated version will also be added to the non-animated version.

It's a bit of a gnarly-looking regex, but I didn't find a native way to parse these things or anything. Any ideas to improve that?

Notably, this may be a small memory leak for each unique image that is processed in this way. Blob URLs can be revoked when they are no longer needed, but it is impractical to track when a paused element that references one is garbage collected (I did try; weakrefs are fun). I think for this PR this is probably not an issue; the kinds of UI elements affected by this PR are limited in number and algorithmic/don't change a lot. I have tested versions of this code that make a blob URL for _every image in every post_ and didn't observe crashing or memory use that seemed especially out of the ordinary, but I obviously want to be cautious about this because we know some users use the same Tumblr tabs for extremely large amounts of time. Be aware that this may also be a _complete_ nonissue; it seems to me that Chrome stores (most?) blobs on the file system rather than in-memory, though I couldn't figure out for sure if Firefox also does so.

Anyway, this also adds the communities browse page tiles to the list of paused GIFs; I could have added them earlier but it looked pretty silly to grey them out.

Related prior PR: #1727

Closes #1877.

### Testing steps

- Confirm that all of the `gifBackgroundImage` elements this feature can pause, as listed in the code, are paused correctly (including the gradient overlay on the tagged page sidebar header) and unpause seamlessly on hover.
- Confirm that the `createImageBitmap` code path works by verifying functionality in Firefox 128 ESR or similar or with the code modified to use it. (I download old Firefox versions from https://ftp.mozilla.org/pub/firefox/releases/ and use `npx web-ext run --firefox='/path/to/Firefox 121.app/Contents/MacOS/firefox'`).

Ideally one would confirm that this code works correctly on WebP images and on images with transparency, but finding a relevant element may be nearly impossible. #1862 can be modified to use this code path for all images and tested on the images in https://www.tumblr.com/communities/test-community-46174.


marcustyphoon:
I guess in theory I could make this work on an element with two different images referenced in its background-image css value, but that sounds like a pain, we don't have any of those, and it seems fairly unlikely that Tumblr would do that?

AprilSylph:
> It's a bit of a gnarly-looking regex, but I didn't find a native way to parse these things or anything. Any ideas to improve that?

Nope. What a weird gap in JS access to computed CSS!

src/features/accesskit/disable_gifs.js
```diff
@@ -178,9 +212,29 @@ const processGifs = function (gifElements) {
   });
 };
 
+const sourceUrlRegex = /url\(["'][^)]*?\.(?:gif|gifv|webp)["']\)/g;
 const processBackgroundGifs = function (gifBackgroundElements) {
-  gifBackgroundElements.forEach(gifBackgroundElement => {
-    gifBackgroundElement.classList.add(backgroundGifClass);
+  gifBackgroundElements.forEach(async gifBackgroundElement => {
+    const sourceValue = gifBackgroundElement.style.backgroundImage;
```
AprilSylph: I would use `getComputedStyle` here; no guarantee that the background image isn't defined via the `background` shorthand property.

AprilSylph (approved):
Tested and working. Lovely. Comment above is a concern about future robustness only.

src/features/accesskit/disable_gifs.js
marcustyphoon: Hmm. I thiiiink that should work, yeah. Right? If you specify (with higher specificity) one of the sub-properties of a shorthand value, only that property is overridden? And with multiple background images, the sub-properties are all comma delimited and since we're just replacing a url with a different url the structure should stay exactly the same...

AprilSylph: I'm not saying we should use `background`, I'm saying the existing background image could conceivably be defined via `background`, which would make them invisible to `element.style.backgroundImage` (thus breaking this code), because `element.style` is defined style and not computed style.

marcustyphoon: No, I've got you! To clarify, I'm trying to determine for sure if, were we to switch to `const sourceValue = getComputedStyle(gifBackgroundElement).backgroundImage;` to facilitate the code working with a hypothetical redpop switch from `background-image` to `background`, the code **would** in fact function in that case.

In that case there would be a redpop-set `background` property, potentially representing multiple `background-` values, and then we would extract the `background-image` part (which could thus potentially have multiple comma-separated values), do a string replace on it, and make a `background-image`-setting CSS rule with higher specificity than redpop's `background`.

AprilSylph: Oh. Yes, that's the idea. Us setting `background-image` would indeed override just the background image part of Tumblr's theoretical multi-value `background` declaration. It's like when you set `padding: 8px; padding-left: 0;`.

PR #1874, opened 2025-09-01
fix(Popup): Clear storage optionally on backup restore

### LINKED ISSUE #1191: Importing backup keeps old storage keys

marcustyphoon:
### Description

This ensures that when one imports a backup of their preferences, keys not in the backup get removed, as one might expect.

Resolves #1191.

### Testing steps

- Export your data, if you're testing this with your real data. Toggle some preferences to generate some data if you aren't. Copy your data.
- Paste `{}` into the restore box. Confirm that restoring removes all of your data.
- Paste your current data into the restore box. Confirm that restoring works as expected.
- Confirm that an informative message is shown if data will be overwritten.

### DIFF
--- a/src/action/render_backup.js
+++ b/src/action/render_backup.js
@@ -3,15 +3,17 @@ const localCopyButton = document.getElementById('copy-local');
 const localDownloadButton = document.getElementById('download-local');
 
 const localImportTextarea = document.getElementById('local-storage-import');
+const localOverwriteWarning = document.getElementById('overwrite-warning');
 const localRestoreButton = document.getElementById('restore-local');
 
 const sleep = ms => new Promise(resolve => setTimeout(() => resolve(), ms));
 
-const updateLocalExportDisplay = async function () {
+const onStorageChanged = async function () {
   const storageLocal = await browser.storage.local.get();
   const stringifiedStorage = JSON.stringify(storageLocal, null, 2);
 
   localExportDisplayElement.textContent = stringifiedStorage;
+  localOverwriteWarning.hidden = stringifiedStorage === '{}';
 };
 
 const localCopy = async function () {
@@ -79,8 +81,8 @@ const localRestore = async function () {
 };
 
 const renderLocalBackup = async function () {
-  updateLocalExportDisplay();
-  browser.storage.local.onChanged.addListener(updateLocalExportDisplay);
+  onStorageChanged();
+  browser.storage.local.onChanged.addListener(onStorageChanged);
 
   localCopyButton.addEventListener('click', localCopy);
   localDownloadButton.addEventListener('click', localExport);
--- a/src/action/render_backup.js
+++ b/src/action/render_backup.js
@@ -57,6 +57,8 @@ const localRestore = async function () {
     localRestoreButton.disabled = true;
 
     const parsedStorage = JSON.parse(importText);
+
+    await browser.storage.local.clear();
     await browser.storage.local.set(parsedStorage);
 
     localRestoreButton.classList.add('success');
--- a/src/action/backup.css
+++ b/src/action/backup.css
@@ -19,7 +19,7 @@
   margin: 0;
 }
 
-#backup label {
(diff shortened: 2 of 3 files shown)


src/action/render_backup.js
```diff
@@ -3,15 +3,17 @@ const localCopyButton = document.getElementById('copy-local');
...
 
 const sleep = ms => new Promise(resolve => setTimeout(() => resolve(), ms));
 
-const updateLocalExportDisplay = async function () {
+const onStorageChanged = async function () {
   const storageLocal = await browser.storage.local.get();
   const stringifiedStorage = JSON.stringify(storageLocal, null, 2);
 
   localExportDisplayElement.textContent = stringifiedStorage;
+  localOverwriteWarning.hidden = stringifiedStorage === '{}';
```
AprilSylph: You've heard of "don't use magic numbers", now get ready for "don't use magic strings":

```suggestion
  localOverwriteWarning.hidden = Object.keys(storageLocal).length === 0;
```

I know these are exactly the same from a reliability standpoint, but... still.

Also fun topic of debate: does _zero_ count as a "magic number" here?

marcustyphoon: `= !Object.keys(storageLocal).length` if you really want no magic numbers!

AprilSylph: Heh. Yeah, I don't think zero actually counts as a magic number. Obfuscating the zero by actually checking for falsiness is definitely worse.

marcustyphoon: I'm a big fan of "never check if a value `=== undefined` (e.g. in an if statement) unless you're doing something different with null," but yeah, zero's kind of a different story :D

AprilSylph (commented):
It works, but I'm not happy with the UX of restoring a backup into a fresh install...

When you have no current data, the note/warning is not displayed. But the moment you restore some data, the note appears _in the place the Restore button was_, which destroys the illusion of the Restore button changing into a success message and back again.

But, more specifically, I find the additional warning appearing with the success message at the same time to be... panic-inducing. I feel like I've just done something with consequences I'm only being warned of after the fact. I haven't, of course, but that's how it feels.

I'm not totally certain on how much this will improve things, but I would start with:
1. Moving the warning to be _after_ the Restore button, so that the button/success message always stays in the same place in the layout.
2. To counteract the out-of-order...ness... created by the above action... we might want to make the warning more eye-grabbing by changing "Note:" to "Warning:" and maybe prepending the ⚠️ emoji.
3. Let's move the `hidden` attribute determination from `onStorageChanged()` to `localRestore()`'s `finally` block, so that it only appears or disappears when the next restore becomes possible. Of course, this means also initialising it explicitly in `renderLocalBackup()`, but oh well.

marcustyphoon:
Ah, shoot, now the message doesn't appear if you generate data by toggling something. Hold on...

marcustyphoon:
### Additional Testing steps

- With no data, confirm that the warning message is not shown.
- Toggle some preferences to generate some data and confirm that the warning message is shown.
- Restore `{}` and confirm that, only after the restore button reappears, the warning message disappears.
- Restore some user data and confirm that, only after the restore button reappears, the warning message appears..

PR #1879, opened 2025-09-06
fix(Quick Reblog, Quick Tags): Prevent popovers being covered by next post in older browsers (no container)

### LINKED ISSUE #1876: Quick Reblog: Popup can be covered by next post (info needed)
### Platform

unknown

### Browser

unknown

### Addon version

unknown

### Details

I have seen multiple user posts reporting that the Quick Reblog popup, when in its default "below" position, is sometimes covered by the next post on the timeline. I have not been able to reproduce this (in a supported browser), so I can't fix it without help.

If this happened to you and you can reproduce it, please comment with your operating system, browser and browser version, XKit Rewritten version, whether you have endless scrolling enabled, and which other web extensions you have and whether temporarily disabling all but XKit Rewritten and opening a new Tumblr tab fixes the problem.

Screenshot from a modified version of the extension loaded in Safari, for reference about what I'm talking about:

marcustyphoon:
### Description

As described in the linked issue, the z-indexes on Quick Reblog and Quick Tags' popover elements do not apply on older browsers with Tumblr's new footer because the new footer uses container queries, which in older browsers created a stacking context, making the whole footer have (from the relevant perspective) the same z-index, i,e. not on top of the following post on the timeline.

This is one way of fixing this: we can delete the container. Does this cause some minor regressions? Yes: Tumblr uses container queries to make smaller text in footer buttons in masonry view in peepr.
 
Resolves #1876.

### Testing steps

todo

### DIFF
--- a/src/features/quick_reblog/index.js
+++ b/src/features/quick_reblog/index.js
@@ -83,6 +83,17 @@ ${keyToCss('engagementAction', 'targetWrapperFlex')}:has(> #quick-reblog) {
 ${keyToCss('engagementAction', 'targetWrapperFlex')}:has(> #quick-reblog) ${keyToCss('tooltip')} {
   display: none;
 }
+
+footer${keyToCss('postFooter')} {
+  /**
+   * Prevents a stacking context being created here and breaking layering in older browsers.
+   * This unfortunately breaks Tumblr's breakpoint-specific styling (smaller text in footer buttons in masonry view in peepr).
+   *
+   * @see https://github.com/w3c/csswg-drafts/issues/10544
+   * @see https://github.com/AprilSylph/XKit-Rewritten/issues/1876
+   */
+  container-type: unset;
+}
 `);
 
 const onBlogSelectorChange = () => {
--- a/src/features/quick_tags/index.js
+++ b/src/features/quick_tags/index.js
@@ -1,6 +1,7 @@
 import { cloneControlButton, createControlButtonTemplate, insertControlButton } from '../../utils/control_buttons.js';
+import { keyToCss } from '../../utils/css_map.js';
 import { dom } from '../../utils/dom.js';
-import { appendWithoutOverflow, filterPostElements, getTimelineItemWrapper, postSelector } from '../../utils/interface.js';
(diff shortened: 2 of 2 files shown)


src/utils/post_popovers.js
```diff
@@ -0,0 +1,26 @@
...
+ * stacking context, so the popup z-index does not apply outside of the footer.
+ *
+ * Removing the container fixes this, but causes a minor regression: disabling Tumblr's
+ * breakpoint-specific styling (smaller text in footer buttons in masonry view in peepr).
+ * @see https://github.com/w3c/csswg-drafts/issues/10544
+ * @see https://github.com/AprilSylph/XKit-Rewritten/issues/1876
+ */
+export const popoverStackingContextFix = oldBrowser
+  ? `
+    footer${keyToCss('postFooter')} {
```
AprilSylph: **fix it and ship it**: I think we should use `postSelector` here as well, so that we only target things we insert popups into. Notably, the Radar doesn't match `postSelector`. although I don't think it uses any container query gated styles... yet.

src/utils/post_popovers.js
```diff
@@ -0,0 +1,26 @@
+import { keyToCss } from './css_map.js';
+
+/**
+ * Detects Chromium <130, Firefox <140, Safari <26, or another old or non-compliant browser.
+ */
+const oldBrowser =
```
AprilSylph: **nit**: differentiating between nouns and answers to questions is good... obviously from context we wouldn't be saving an entire browser to a JavaScript constant, but it's good to flex the "good practice" muscle, I think.

You might think `isBrowserOld` would be a better name, but generally the `isSomething`/`isThingSomething` style is reserved for functions that test a parameter to output a boolean answer (e.g. `Array.isArray`).

For the answer itself, whether it's the result of a question-asking function or just calculated in-place like this, the preferred style puts "is" in the middle, so `browserIsOld` here.

marcustyphoon:
> My view is that the browsers that implemented container queries by creating stacking contexts did not actually implement container queries correctly, and so should not have done at all.

Oh—took me a minute, but I see. You're saying that this is equivalent in functionality to those browser versions simply not having supported container queries at all. Makes sense!

marcustyphoon:
Retested in Firefox 128 with/without classic footer, just in case something changed somehow.

PR #1884, opened 2025-09-24
TimeFormat: Update relative timestamps in real time

marcustyphoon:
### Description

I had such a neat idea to do this using `WeakRef`s. Then I realized that that was entirely unnecessary.

Anyway, this makes the relative timestamps added by TimeFormat, e.g. "12 minutes ago," update every minute.

### Testing steps

- Enable TimeFormat and its relative time option.
- Confirm that a post from less than one hour ago's relative timestamp updates to stay within a minute of the correct value.

### DIFF
--- a/src/features/timeformat/index.js
+++ b/src/features/timeformat/index.js
@@ -7,6 +7,8 @@ import { getPreferences } from '../../utils/preferences.js';
 let format;
 let displayRelative;
 
+let intervalId;
+
 export const styleElement = buildStyle(`
 [data-formatted-time] {
   font-size: 0px !important;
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Okay, no, you're right, that was better. I think I may have forgotten that there's a `characterData` option on the mutation observer.

AprilSylph:
> `I guess you like trimming *every* cycle, right`

"And for your information, I prefer hyper-optimised code with NO TRANSPILE STEPS!"
"I know it's a little messier! I like it!"

marcustyphoon:
Hm, if you basically write the correct function yourself and send it in to support with a cc0 declaration maybe they'll put it in :P

(Yeah, no, probably not for this. Not worth Staff's time any more than it's worth ours. I did get the new reblog menu unmounted, I assume, but that was a oneliner.)

PR #1889, opened 2025-09-29
refactor: Popover position css

marcustyphoon:
### Description

This changes the below/above position flag used for the quick reblog/quick tags popups from a class to a data attribute, which is semantically easier to toggle without interfering with any other classes on the element, and should cause no observable behavior change.

### Testing steps

Confirm that Quick Reblog and Quick Tags' popups still change positions when they would otherwise go off the top/bottom of the viewport.

### DIFF
--- a/src/features/quick_reblog/index.css
+++ b/src/features/quick_reblog/index.css
@@ -30,11 +30,11 @@
   }
 }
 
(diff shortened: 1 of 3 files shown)


marcustyphoon:
buttons are hard ok

PR #1894, opened 2025-10-02
Quick Tags; Quick Reblog: Prevent popovers overflowing off top of screen

### LINKED ISSUE #1836: Quick Reblog: Move popup to "below" position if it goes off the top of the screen
### Concept

We currently move the Quick Reblog popup to the "above" position if it would go below the bottom of the screen. We should also do the reverse, moving it to the "below" position if it would to above the top of the screen, because this is possible when a short post is viewed by itself in the blog view modal and it can make it impossible to change the selected blog.

marcustyphoon:
### Description

We currently move the Quick Reblog/Quick Tags popovers to the "above" position if they would go below the bottom of the screen. This also does the reverse, moving them to the "below" position if they would go above the top of the screen.

If both positions would exit the screen, this uses the default position.

Resolves #1836.

### Testing steps

- Confirm that the Quick Reblog and Quick Tags popovers change positions to avoid both the top and bottom edges of the screen.
- Confirm that the Quick Reblog and Quick Tags popovers use the correct default positions when neither position would hit either the top or bottom edge of the screen.

### DIFF
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -127,17 +127,22 @@ const getClosestWithOverflow = element => {
   }
 };
 
+const isVerticallyOverflowing = element => {
+  const elementRect = element.getBoundingClientRect();
+  return elementRect.bottom > document.documentElement.clientHeight || elementRect.top < 0;
+};
+
 export const appendWithoutOverflow = (element, target, defaultPosition = 'below') => {
   element.dataset.position = defaultPosition;
   element.style.removeProperty('--horizontal-offset');
 
   target.appendChild(element);
 
-  if (element.dataset.position === 'below' && element.getBoundingClientRect().bottom > document.documentElement.clientHeight) {
-    element.dataset.position = 'above';
+  if (isVerticallyOverflowing(element)) {
+    element.dataset.position = defaultPosition === 'below' ? 'above' : 'below';
   }
-  if (element.dataset.position === 'above' && element.getBoundingClientRect().top < 0) {
-    element.dataset.position = 'below';
+  if (isVerticallyOverflowing(element)) {
+    element.dataset.position = defaultPosition;
   }
 
   const preventOverflowTarget = getClosestWithOverflow(target);


src/utils/interface.js
```diff
@@ -133,13 +133,17 @@ export const appendWithoutOverflow = (element, target, defaultPosition = 'below'
 
   target.appendChild(element);
 
+  if (element.dataset.position === 'below' && element.getBoundingClientRect().bottom > document.documentElement.clientHeight) {
+    element.dataset.position = 'above';
+  }
+  if (element.dataset.position === 'above' && element.getBoundingClientRect().top < 0) {
+    element.dataset.position = 'below';
+  }
+
```
AprilSylph, line 147: **Question**: Can these conditionals not reuse `elementRect`, defined below? It kinda looks like the answer is no, but I don't know if the object returned by `getBoundingClientRect()` is actually static or if it updates with new paints.

AprilSylph: I guess if it does update, then the getter method isn't constructing a new object each time anyway, which would mean our saving the result to `elementRect` in the first place doesn't do anything except be an extra line of code. So either this is fine as-is because it needs to be like this, or it's fine as-is and the code below it could use cleaning up too.

marcustyphoon:
Oh. Uh. My "it returns to the original position" logic is straight-up wrong; nice.

src/utils/interface.js
marcustyphoon: Yeah, no, a `DOMRect` is just a static rectangle.

marcustyphoon:
Okay, take two! "If the default way is bad, try the other way; if that way is also bad, go back to the default." Which is what I intended but failed to do originally.

marcustyphoon:
Wait, really? I am on the same version.

AprilSylph (approved):
Apparently GitHub gave me the artifact for 20aa6e192e5719f878ba73f98dced244fecd42d9 from the Checks tab yesterday...? Sorry about that.

032a173d507e4084c3fa681f0e3a69166e1bf693 is working as described. 👍

marcustyphoon:
You know, I never asked what advantage you found in downloading zips like that over checking out PRs with github desktop/`gh` cli! Not that there has to be a specific one, of course.

AprilSylph:
> You know, I never asked what advantage you found in downloading zips like that over checking out PRs with github desktop/`gh` cli! Not that there has to be a specific one, of course.

Not having to clean my working tree is the main one, I think.

PR #1903, opened 2025-10-07
feat: add new element creation utils

AprilSylph:
### Description

Adds 1) a new version of `dom()` which only requires two arguments, and 2) a dozen shorthand functions for creating every type of element we create via `dom()` across the codebase.

The first thing I've been thinking of doing for a long time now, but only started on yesterday while looking at #1803; the per-element shorthands I've been thinking about since we merged #1831.

This PR doesn't implement any usages of these functions. That will be a followup chore.

### Testing steps

Just CI passing and code review suffices. Opinions matter here, since these will be used fairly widely—I would like to remove `dom()` entirely in favour of these new shorthands.

### DIFF
--- a/src/utils/dom.js
+++ b/src/utils/dom.js
@@ -5,8 +5,8 @@
 /**
  * Create elements with simple syntax
  * @param {string} tagName - Type of element to create
- * @param {Record<string, string>} [attributes] - Property-value pairs to set as HTML/XML attributes (e.g. { href: '/' })
- * @param {Record<string, () => void>} [events] - Property-value pairs to set as event listeners (e.g. { click: () => {} })
+ * @param {Attributes} [attributes] - Property-value pairs to set as HTML/XML attributes (e.g. { href: '/' })
+ * @param {Events} [events] - Property-value pairs to set as event listeners (e.g. { click: () => {} })
  * @param {Children} [children] - Zero or more valid children
  * @returns {Element} Element created to specification
  */


marcustyphoon:
Do we want children to be a named prop? Sounds like a nesting nightmare.

AprilSylph:
I've just realised that, likely because this branch was originally based off #1803, that there are a bunch of shorthand functions missing. Whoops.

Anyway, to your question... yeah, I wasn't thinking about our ESLint rules, actually.

Current form:
```js
const form = dom('form', { id: 'xkit-mass-deleter-delete-drafts' }, { submit: confirmDeleteDrafts }, [
  dom('label', null, null, [
    'Delete drafts before:',
    dom('input', { type: 'datetime-local', name: 'before', value: createNowString(), required: true })
  ])
]);
```

New form, children as named property:
```js
const deleteDraftsForm = form({
  id: 'xkit-mass-deleter-delete-drafts',
  submit: confirmDeleteDrafts,
  children: [
    label({
      children: [
        'Delete drafts before:',
        input({ type: 'datetime-local', name: 'before', value: createNowString(), required: true })
      ]
    })
  ]
});
```

New form, children as separate argument:
```js
const deleteDraftsForm = form({ id: 'xkit-mass-deleter-delete-drafts', submit: confirmDeleteDrafts }, [
  label({}, [
    'Delete drafts before:',
    input({ type: 'datetime-local', name: 'before', value: createNowString(), required: true })
  ])
]);
```

marcustyphoon:
- Had a random thought that the shorthands could be properties on a namespace object, like `dom.svg({ /* */ }, [/* */])`. Obviously, though, there's no real difference between that and just not using the shorthand, and it probably makes intellisense worse (also, one can namespace import this module if one really wants to do that). Kind of reads nicely, though.
- There are various sort-of-DRY things one could do here (`const shorthand = tagName => (...args) => element(tagName, ...args); export const a = shorthand('a');`?) but are any of them... good? Probably not.

marcustyphoon:
Apparently React.createElement has a [slightly different signature](https://react.dev/reference/react/createElement), so JSX compatibility might look something like

```js
export function element (tagName, properties = {}, ...children) {
/* */
  element.replaceChildren(children.flat(Infinity));
```

We would never use that, of course, but we could do it just for the sake of being... I dunno what the word would be. Applicable elsewhere?

AprilSylph:
Hmm, yeah, I don't like that. Who is actually using `React.createElement` instead of just using JSX, anyway? JSX is like 90% of the appeal of React to me.

marcustyphoon:
Other way around: you would do this if you wanted to make a function that *could* be used interchangeably as the back end for a JSX codebase, because JSX compiles to the React.createElement signature. (This would also allow you to use preact's htm, for example.)

src/utils/dom.js
```diff
@@ -1,9 +1,13 @@
...
+/** @typedef {(string | Node)[]} Children */
+
 /**
  * Create elements with simple syntax
  * @param {string} tagName - Type of element to create
- * @param {object} [attributes] - Property-value pairs to set as HTML/XML attributes (e.g. { href: '/' })
- * @param {object} [events] - Property-value pairs to set as event listeners (e.g. { click: () => {} })
- * @param {(Node|string)[]} [children] - Zero or more valid children
+ * @param {Record<string, string>} [attributes] - Property-value pairs to set as HTML/XML attributes (e.g. { href: '/' })
+ * @param {Record<string, () => void>} [events] - Property-value pairs to set as event listeners (e.g. { click: () => {} })
```
marcustyphoon: I assume these should uses the typedefs as well?

src/utils/dom.js
```diff
@@ -18,3 +22,57 @@ export function dom (tagName, attributes = {}, events = {}, children = []) {
...
+  const element = properties?.xmlns
+    ? document.createElementNS(properties.xmlns, tagName)
+    : document.createElement(tagName);
+
+  const attributes = Object.entries(properties).filter(([key, value]) => typeof value === 'string');
+  const events = Object.entries(properties).filter(([key, value]) => typeof value === 'function');
+
+  element.replaceChildren(children);
+  attributes.forEach(([name, value]) => element.setAttribute(name, value));
+  events.forEach(([type, listener]) => element.addEventListener(type, listener));
```
marcustyphoon, line 44: Do we want to consider trimming `/^on/` from these names?

`dom('button', { class: noteCountClass, click: onNoteCountClick }, [`
`dom('button', { class: noteCountClass, onclick: onNoteCountClick }, [`

src/utils/dom.js
AprilSylph: Whoops, yes, don't know how I missed them.

(Well, I do, it's because I already made this change multiple times while figuring out what I did and didn't like, but...)

src/utils/dom.js
AprilSylph: Eh, nah. I'm not trying to mimic React, really. And, forcing ourselves to omit `on*` when setting these makes it more explicit that we're not just doing `Object.assign` with the props; when we pass a callback as `click` we're definitely not assigning it to the created element's `onclick` event handler property.

src/utils/dom.js
```diff
@@ -18,3 +22,57 @@ export function dom (tagName, attributes = {}, events = {}, children = []) {
   element.normalize();
   return element;
 }
+
+/**
+ * Create elements with even simpler syntax
+ * @param {string} tagName Type of element to create
+ * @param {Attributes & Events} [properties] Property-value pairs to set as HTML/XML attributes or event listeners
+ * @param {Children} [children] Zero or more valid children
+ * @returns {Element} Element created to specification
```
marcustyphoon, line 31: Are the omitted ` - ` intentional? I know some people don't use them in jsdoc; we could standardize on either.

AprilSylph: Yeah, VS Code intellisense interprets `-` in JSdoc as bullet points, so I'm omitting them now. If you want to open a PR to remove them everywhere, I'll welcome it!

marcustyphoon: Hm, not sure what this is referring to! It does if you make a list, but does it here?

AprilSylph: It was making single-item unordered lists for me!

marcustyphoon: Interesting! Like in these popups?

AprilSylph: Yes, specifically when the individual parameters were hovered was where I was seeing the bullet point interpretation.

marcustyphoon: Aha, parameters. Yep, seeing it now.

marcustyphoon:
Realized we could in theory allow e.g. `div([these, are, children])` rather than `div(null, [these, are, children])` but it would be quite a mess to implement.

PR #1904, opened 2025-10-07
feat: add Classic Footer

AprilSylph:
### Description

TBD

### Testing steps

TBD


src/features/classic_footer/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+  ${footerContentSelector} ${keyToCss('targetWrapperFlex')}:has(svg use[href="#managed-icon__ds-reblog-24"]) {
+    flex: 0;
+  }
+  ${footerContentSelector} ${keyToCss('engagementCount')} {
+    display: none;
+  }
+  .${noteCountClass} {
+    padding: 8px;
+    margin-inline: 8px;
```
marcustyphoon: This could have e.g. `white-space: nowrap; overflow: hidden;` to slightly improve the narrow-post situation.

**edit:** Could also border-radius it to make the touch target not weirdly rectangular, but who's even gonna notice.

src/features/classic_footer/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+const noteCountClass = 'xkit-classic-footer-note-count';
+const footerContentSelector = `${postSelector} article footer ${keyToCss('footerContent')}`;
+const replyButtonSelector = `${footerContentSelector} button${keyToCss('engagementAction')}:has(svg use[href="#managed-icon__ds-reply-outline-24"])`;
+
+const locale = document.documentElement.lang;
+const noteCountFormat = new Intl.NumberFormat(locale);
+
+export const styleElement = buildStyle(`
+  ${postSelector} article footer {
```
marcustyphoon: Technically I see no place where this is breaking something yet but it should probably have a :has() and only apply if footerContent exists (or the note count we add in js, or this could all be added along with the note count by targeting `footer:has(> .${noteCountClass})`, whatever—actually I think that's the most elegant option). Right now we're styling community post footers, for example.

marcustyphoon:
```diff
diff --git a/src/features/classic_footer/index.js b/src/features/classic_footer/index.js
index 2f450cd5..e3f4036a 100644
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -12,13 +12,7 @@ const locale = document.documentElement.lang;
 const noteCountFormat = new Intl.NumberFormat(locale);
 
 export const styleElement = buildStyle(`
-  ${postSelector} article footer {
-    flex-direction: row;
-    justify-content: space-between;
-    align-items: center;
-  }
   ${footerContentSelector} {
-    justify-content: flex-end;
     gap: 0;
   }
   ${footerContentSelector} > div {
@@ -37,7 +31,8 @@ export const styleElement = buildStyle(`
   }
   .${noteCountClass} {
     padding: 8px;
-    margin-inline: 8px;
+    margin-right: auto;
+    order: -100;
 
     color: var(--content-fg-secondary);
     font-family: var(--font-family-modern);
@@ -61,7 +56,7 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
   ]);
 
   const footerContent = postElement.querySelector(footerContentSelector);
-  footerContent?.before(noteCountButton);
+  footerContent?.prepend(noteCountButton);
 });
 
 export const main = async function () {

```

One way to resolve the `xkit-controls-row` thing (and, I guess, the "styling community post footers" thing, too, technically).

**Edit:** Might need tweaks for mobile viewport.

marcustyphoon:
We could also include `footer:has(/* whatever */) { container-type: unset; }` because it's not like we're going to be making use of any container-specific Tumblr styles (so now we don't have to test them) and because it would fix #1876, but we need to fix #1876 when this isn't enabled, anyway, so it's arguably pointless.

marcustyphoon:
Maybe it's easier to just put `xkit-controls-row` before the actual footer, I dunno. Kind of semantically awkward and the aria-label is kind of wrong, but.

AprilSylph:
> Maybe it's easier to just put `xkit-controls-row` before the actual footer, I dunno. Kind of semantically awkward and the aria-label is kind of wrong, but.

I mean, my mental model is that tags are part of the footer too, but that's factually false. The footer doesn't have multiple rows. So if, visually, we're inserting a row above the footer... that's how we should be doing it from a DOM perspective as well.

AprilSylph:
Ready for re-review!

My latest changes also tweak our control buttons CSS to make our inserted buttons align with Tumblr's buttons by default, with Classic Footer tweaking that CSS further to match the styles it applies to the footer.

Classic Footer disabled | Classic Footer enabled
-|-
Screenshot 2025-10-08 at 10-44-14 Tumblr | Screenshot 2025-10-08 at 10-44-05 Tumblr

marcustyphoon:
How much bikeshedding am I supposed to be doing over style? The widened `xkit-controls-row` separating line looks worse to me, but like, eh.

AprilSylph:
Pointing things out isn't bikeshedding. Now that I see it, it's bothering me too... third try's the charm? One moment.

AprilSylph:
Hmm, this one is annoying. No such thing as negative padding to offset the needed increased margin... I guess I could use a pseudo-element to draw the border instead?

marcustyphoon:
Oh, whoops, I had assumed it was intentional because I hadn't realized that it's from the new footer, not the changes you just made :D Hm, that is kind of a pain.

AprilSylph:
Pseudo-element seems to work fine, and I don't think the changes mess with anything else...? I'll push what I've got so you can test for yourself.

---

Classic Footer disabled | Classic Footer enabled
-|-
Screenshot 2025-10-08 at 11-23-21 Tumblr | Screenshot 2025-10-08 at 11-23-30 Tumblr

marcustyphoon:
It used to be 16px; do you think 12px or 16px looks better?

(Are we bikeshedding now? :3)

AprilSylph:
Where are you getting 16px from? Is that what Tumblr's gap was?

This is current `master`:
https://github.com/AprilSylph/XKit-Rewritten/blob/df2ebcd59a0db0423d877fb8a94157069c9cbac3/src/content_scripts/control_buttons.css#L64-L70

marcustyphoon:
Yes, I'm comparing to the pre-rewrite footer.



Now that you mention it, I guess the 16px value would arguably only make sense with this feature enabled.

(Anyway, I really don't think it makes much difference.)

marcustyphoon (approved):
Either way: looks good!

AprilSylph:
What an adventure, thanks for joining me on it. I guess I'll merge this... whenever Mozilla approves v1.0.17? With the linked issue being so high-traffic I don't really want to have more time than necessary between it being marked as resolved and folks actually receiving this feature.

marcustyphoon:
I was wondering if the opposite was true and if we could sneak a fast review by having our old place in the queue apply to the new version if we released this really fast.

Can you retroactively make a commit close an issue? Probably no right. Hm.

marcustyphoon:
Oh, one other difference between this and the real old footer is that there's a bigger space above the old footer (below the notes) on posts without extra control buttons. I always kind of thought it looked wrong, at least as big as it was, but I'm not sure what the right value ought to be/if it ought to be more than it is now.

AprilSylph:
That increased vertical spacing was always a bug, as far as I'm concerned.
(It probably wasn't, but it always felt like one to me.)

AprilSylph:
FWIW, [Vision](https://vision-theme.tumblr.com/) has the same gap between the post body and tags as it has between tags and the note count, and pretty much all of its dimension values were lifted straight from the Tumblr dashboard at the time. I don't remember if that was pre- or post-Redpop though...

image

See also: https://www.tumblr.com/reblog/0/0

marcustyphoon:
> I was wondering if the opposite was true and if we could sneak a fast review by having our old place in the queue apply to the new version if we released this really fast.

~~Only piece of information I've found about this so far~~ conflicting info actually:

| | |
| - | - |
|  | |

marcustyphoon:
Never mind, not releasing was genius actually.

marcustyphoon:
one way to resolve the `postOwnerControls` thing. you know the drill by now

```diff
diff --git a/src/features/classic_footer/index.js b/src/features/classic_footer/index.js
index ae693b47..e36047f6 100644
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -7,28 +7,29 @@ import { timelineObject } from '../../utils/react_props.js';
 const noteCountClass = 'xkit-classic-footer-note-count';
 
 const footerContentSelector = `${postSelector} article footer ${keyToCss('footerContent')}`;
-const footerSelector = `${postSelector} article footer:has(> ${keyToCss('footerContent')})`;
 const replyButtonSelector = `${footerContentSelector} button${keyToCss('engagementAction')}:has(svg use[href="#managed-icon__ds-reply-outline-24"])`;
 
 const locale = document.documentElement.lang;
 const noteCountFormat = new Intl.NumberFormat(locale);
 
 export const styleElement = buildStyle(`
-  .xkit-controls-row {
-    column-gap: 0 !important;
-  }
-  .xkit-controls-row::after {
-    left: 16px !important;
-    right: 16px !important;
+  ${keyToCss('postOwnerControls')} {
+    position: relative;
+    gap: 0;
+
+    border: none;
   }
+  ${keyToCss('postOwnerControls')}::after {
+    position: absolute;
+    bottom: 0;
+    left: 16px;
+    right: 16px;
 
-  ${footerSelector} {
-    flex-direction: row;
-    justify-content: space-between;
-    align-items: center;
+    border-bottom: 1px solid rgba(var(--black), .13);
+    content: '';
   }
+
   ${footerContentSelector} {
-    justify-content: flex-end;
     gap: 0;
   }
   ${footerContentSelector} > div {
@@ -45,10 +46,14 @@ export const styleElement = buildStyle(`
   ${footerContentSelector} ${keyToCss('engagementCount')} {
     display: none;
   }
+  ${keyToCss('likesControl')}${keyToCss('withBlazeButton')} {
+    width: unset;
+  }
   .${noteCountClass} {
     padding: 8px;
     border-radius: 20px;
-    margin-inline: 8px;
+    margin-right: auto;
+    order: -100;
     overflow: hidden;
 
     color: var(--content-fg-secondary);
@@ -75,7 +80,7 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
   ]);
 
   const footerContent = postElement.querySelector(footerContentSelector);
-  footerContent?.before(noteCountButton);
+  footerContent?.prepend(noteCountButton);
 });
 
 export const main = async function () {
```

**edit:** Button vertical alignment is slightly off with this.

marcustyphoon:
Mildly interesting:

- The rightmost button (share, in the new footer; we replace it with like) has a bubble on hover of 8px padding and is 8px from the right edge and 12px from the bottom edge of the post. (This is a difference of 4px.)
- This makes the right edge of the button (non-hovered) line up with the edge of the line above it when it has 16px gap. Looks right!
- Giving the text-only note count 8px of padding and putting it 8px from the left edge and 12px from the bottom edge of the post makes it line up with the edge of the line above it when it has 16px gap. Looks... I'm going to say sort of right! There is greater than 4px more space below the text than to the left of it (because the text is 16px tall, not 24px like an icon would be—this is what the line height of 1.5rem reflects, I assume), which kind of makes it look too far to the left.
- Interestingly, if you add a bubble to the note count (say, `background-color: rgba(0,0,0,0.07);`), it looks wrong like this. There is 4px added to the top and bottom of the text due to the line height; we would need to add 4px to the left and right as well, i.e. `padding-inline: 12px`. The bubble now looks correct; it extends 12px from the text all around. (`padding: 12px; line-height: 1em;` does this also.) 
- If you take away the bubble after doing that, we now have a position where there is 4px more space below the text than to the left of it, which in a sense looks correct. But the fact that the text no longer lines up with the edge of the line above it when it has 16px gap doesn't look so correct.
- There's a part of me that thinks `padding-inline: 11px`/`padding: 11px; line-height: 1em;` actually looks the most correct, but I have absolutely no way of justifying that. `padding-inline: 10px`/`padding: 10px; line-height: 1em;` kind of makes sense as it is splitting the difference between the "correct" left edge gap and "correct" left/bottom edge gap difference.

AprilSylph:
Classic Footer disabled | Classic Footer enabled
-|-
Screen Shot 2025-10-09 at 10 32 03 | Screen Shot 2025-10-09 at 10 32 33
Screen Shot 2025-10-09 at 10 40 20 | Screen Shot 2025-10-09 at 10 40 26
Screen Shot 2025-10-09 at 10 40 49 | Screen Shot 2025-10-09 at 10 40 58

src/features/classic_footer/index.js
```diff
@@ -0,0 +1,97 @@
...
+    line-height: 1.5rem;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+`);
+
+const onNoteCountClick = ({ currentTarget }) => {
+  const postElement = currentTarget.closest(postSelector);
+  const closeNotesButton = postElement?.querySelector(closeNotesButtonSelector);
+  closeNotesButton?.click() ?? postElement?.querySelector(replyButtonSelector)?.click();
```
marcustyphoon: Apparently this works, obviously, but... this is clicking both buttons, right? I assume `.click()` returns undefined.

AprilSylph: I might be stupid. Of course it does.

src/features/classic_footer/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+export const styleElement = buildStyle(`
+  ${postOwnerControlsSelector} {
+    position: relative;
+    gap: 0;
+    border-bottom-color: transparent;
+  }
+  ${postOwnerControlsSelector}::after {
+    position: absolute;
+    bottom: -1px;
```
AprilSylph, line 26: If it's not clear, I did this so that the single pixel of vertical height for the border is still reserved. Without this, we put our own line in `postOwnerControls`' padding—which, extremely minor detail, but same LoC to avoid, so.

marcustyphoon:
My brain so wants to believe the control buttons are still misaligned. But `${postOwnerControlsSelector} > * { outline: 1px solid green; outline-offset: -8px; border-radius: 0; }` definitely shows they're not; I guess it's all down to the icon art.

AprilSylph:
Yeah. Maybe we should move to using Tumblr icons for our control buttons... 😏 

(Not in this PR, obviously.)

marcustyphoon:
This notably does leave in changes to `xkit-controls-row`, which won't get used unless Tumblr makes another change or a future feature (say that ten times fast) inserts a button into a non-editable post. Removing the `canEdit &&` from Quick Tags and Trim Reblogs to get a quick look, they're a) relatively fine, though b) missing the now-removed "make the gap zero if classic footer is enabled" css.

We could leave this as is, carefully tweak them to be pixel-perfect even though they're unused, revert the changes so this PR does not have an unused diff, or possibly straight up remove the whole thing in this PR if we want to have no fallback (but I kind of want the fallback).

src/features/classic_footer/index.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+`);
+
+const onNoteCountClick = ({ currentTarget }) => {
+  const postElement = currentTarget.closest(postSelector);
+  const closeNotesButton = postElement?.querySelector(closeNotesButtonSelector);
+
+  closeNotesButton
+    ? closeNotesButton.click()
+    : postElement?.querySelector(replyButtonSelector)?.click();
```
marcustyphoon, line 79:
```suggestion
  postElement?.querySelector(`${closeNotesButtonSelector}, ${replyButtonSelector}`)?.click();
```

If you desperately want to be succinct.

Yes, I messed this suggestion up a second ago.

AprilSylph: Wouldn't the ternary more performant? Since if it does find `closeNotesButton`, it doesn't query for `replyButtonSelector` at all, whereas in the succinct version it queries for both regardless...? Or, since this is `querySelector` and not `querySelectorAll`, would it automatically stop looking once it found a match?

marcustyphoon: I guess in the case where there's no close button, one checks every element up to the reply button against both selectors then stops, the other checks every element in the post against one selector, then every element up to the reply button against the other. Could try to make a microbenchmark, but since this'll take around zero milliseconds you'll have to be pretty careful :D

AprilSylph:
> This notably does leave in changes to `xkit-controls-row`, which won't get used unless Tumblr makes another change or a future feature (say that ten times fast) inserts a button into a non-editable post. Removing the `canEdit &&` from Quick Tags and Trim Reblogs to get a quick look, they're a) relatively fine, though b) missing the now-removed "make the gap zero if classic footer is enabled" css.
> 
> We could leave this as is, carefully tweak them to be pixel-perfect even though they're unused, revert the changes so this PR does not have an unused diff, or possibly straight up remove the whole thing in this PR if we want to have no fallback (but I kind of want the fallback).

I want to keep it in as a fallback, too. This leaves us with three options:
- OK fallback (revert changes made in this PR; achieve minimal diff)
- Better fallback (keep changes as they are)
- Perfect fallback (work further to make it pixel-perfect)

I don't think further work is warranted for a might-never-be-used fallback, but given that we already tested and verified the changes before the thing became a fallback, reverting them doesn't achieve anything except make the fallback worse on purpose. So, I vote we just leave them as-is.

marcustyphoon:
> I was wondering if the opposite was true and if we could sneak a fast review by having our old place in the queue apply to the new version if we released this really fast.

Replies to my message asking about this can be summarized as a) you probably keep your place in the queue if you push a second release, _probably maybe at least according to an educated guess by one person who knows_, and b) request from the team that you do push a second release in that case to reduce overhead; did not directly answer question.

src/features/classic_footer/index.js
AprilSylph: Wait a minute. There's no priority in selectors except DOM order...

The reply button comes first in the DOM, so the combined selector will always select that no matter what.

Ternary is needed.

marcustyphoon: Oh, right. (I actually don't know why the close button needs to be clicked, I guess I should have asked that.)

`(postElement?.querySelector(closeNotesButtonSelector) ?? postElement?.querySelector(replyButtonSelector))?.click()` is obviously doable but I generally prefer not to do those, and by the time you make another temp variable you may as well Not.

AprilSylph: Ah, right, sorry—it's so that clicking the notes button when the notes are opened to a non-Comments tab still closes the notes, instead of just switching you back to the Comments tab.

marcustyphoon: Oh, me stupid. Yep—knew there was a reason because I had done it, couldn't for the life of me figure out why later :D

marcustyphoon:
Minor bug; acceptable, I think, for the MVP:

- open the notes on a narrow post (e.g. https://www.tumblr.com/explore/trending or https://www.tumblr.com/april + masonry view)
- change tabs
- click the notes count
- notes do not close

Going to be annoying to fix. Now you're thinking with portals!

**edit:** wait now they do. what

**edit:** wow if you alternate between clicking the replies button and note count, the modal flickers. no idea at present how that would happen!

AprilSylph:
> > I was wondering if the opposite was true and if we could sneak a fast review by having our old place in the queue apply to the new version if we released this really fast.
> 
> Replies to my message asking about this can be summarized as a) you probably keep your place in the queue if you push a second release, _probably maybe at least according to an educated guess by one person who knows_, and b) request from the team that you do push a second release in that case to reduce overhead; did not directly answer question.

Hmm. I guess I would be willing to publish a release and just see what happens. The [recent announcement about version rollbacks](https://blog.mozilla.org/addons/2025/09/19/now-you-can-roll-back-to-a-previous-version-of-your-extension/) gives me hope that even if our submissions are approved out of order that I would be able to fix the wrong-version-marked-as-latest problem, if they haven't already fixed that themselves.

marcustyphoon:
I believe if you have two versions in the queue the earlier one gets auto disabled and the reviewer sees a combined diff.

AprilSylph:
> Minor bug; acceptable, I think, for the MVP:
> 
>     * open the notes on a narrow post (e.g. [www.tumblr.com/april](http://www.tumblr.com/april) + masonry view)
> 
>     * change tabs
> 
>     * click the notes count
> 
>     * notes do not close
> 
> 
> Going to be annoying to fix. Now you're thinking with portals!
> 
> **edit:** wait now they do. what

It's a race condition, I think? The notes floatie gets closed by any outside click occurring on the page—I guess if that happens first and our click handler happens second, the notes get closed and then immediately re-opened.

AprilSylph:
Only after I push a fix do I realise there's an extra layer: there are two click events! The original one from actually clicking the notes button, and the one we perform by triggering a click to the Reply button. By stopping propagation, the document body only sees one click (in the bubbling phase, at least). Seems to fix the issue, either way.

marcustyphoon:
Opening the floatie with the reply button and then pressing notes to close, or vice versa, still flicks it back up for me, but I didn't look at all closely why, and it's quite minor.

**edit:** Have now "looked at all closely." Conclusion: no idea.

marcustyphoon:
I'd love to center the position of the reblog modal (i.e. change from `bottom-start` to `bottom`) as a follow-up, but I don't see a way to do it in CSS that doesn't potentially mess with the automatic repositioning or a way to poke it in react either.

edit: Nope, I'm a liar.

marcustyphoon:
> Option to modify the Reblog button to link directly to the reblog form

Huh. Problem I just noticed with this: Quick Reblog users who don't want to use the classic with-note-counts footer should be able to access it. But if we...

- put it in Quick Reblog, it's not discoverable from Classic Footer (where most people will look)
- put it in Classic Footer, we'll also have to make the current Classic Footer functionality an (on-by-default, please) checkbox, which makes the feature description wrong and also makes it not discoverable from Quick Reblog (where it's obviously quite relevant)
- put it in Tweaks, it's not discoverable from either Classic Footer or Quick Reblog, and also everything in Tweaks has a discoverability problem inherently, because, you know, Tweaks
- put it in Classic Footer *and* Quick Reblog, our extension framework doesn't really allow for that unless the whole thing was in a util and both features used an iframe preference, which, uh.

edit: anyway here's wonderwall

```js
const modalReblogButtonSelector = `${keyToCss('menuItemsWrapper')} > a[href^="/reblog/"]:has(use[href="#managed-icon__ds-pencil-outline-24"])`;
const modalReblogButtonObserver = new MutationObserver(mutations =>
  mutations
    .flatMap(({ addedNodes }) => [...addedNodes])
    .filter(addedNode => addedNode.querySelector?.(modalReblogButtonSelector))
    .forEach(addedNode => {
      addedNode.setAttribute('hidden', '');
      setTimeout(() => addedNode.removeAttribute('hidden'), 500);
      addedNode.querySelector(modalReblogButtonSelector).click();
    })
);

/* if (preference) */ modalReblogButtonObserver.observe(document.body, { childList: true });
```

AprilSylph:
Reverting how the reblog button works when clicked is categorically a Classic Footer problem to solve. Quick Reblog is for adding the on-hover menu and nothing else.

I also don't see "people who do want the spread-out button layout, but don't want the new on-click reblog menu" as a demographic that actually exists.

marcustyphoon:
In practice that's likely true, yeah. Logically I don't think it makes as much sense; Quick Reblog cleanly replaces almost the entire functionality of the reblog menu*, and Classic Footer removes the split note counts, which is a sidegrade.

*Except opening the post notes to the reblogs view, which in a hypothetical case without Classic Footer but with the menu removed takes the same number of clicks—two—but is done in an illogical way, by first clicking on the replies button; this kind of illogical situation is one we have historically avoided creating and could be another reasonable argument against allowing the menu to be removed independently of Classic Footer.

PR #1911, opened 2025-10-10
chore: Update `xkit-controls-row` fallback styles

marcustyphoon:
### Description

In #1904, we made some changes to the `xkit-controls-row` element that our control button creates so that we can insert control buttons into editable posts without edit/delete buttons, i.e. those with the new footer. Tumblr then added back the edit/delete buttons in a new `postOwnerControls` element just when we thought we were done, making the `xkit-controls-row` element currently completely unused. We left it in as a fallback but didn't touch it further.

This updates it to match the current `postOwnerControls` element, both with and without Classic Footer enabled, so that if it does ever get used as a fallback (or if we want to add control buttons to a non-editable post), its appearance will be consistent.

Its position in the DOM, spacing, and bottom border line should be correct. Notably, this includes making the line span from post edge to post edge in non-classic-footer mode, which I personally think looks bad. But that's how Tumblr's element looks!

Tumblr's element was just changed today, illustrating that it's more likely to be changed again during this general timeframe, so in a sense this doesn't make much sense to merge now. Also, you know... at all, one could make the argument.

### Testing steps

- Un-revert the test commits and enable Quick Tags and Trim Reblogs so that control buttons appear using the `xkit-controls-row` on non-editable posts.
- Open your dash and your channel page side by side and confirm that the position in the DOM, spacing, and bottom border lines on the footer rows with the Quick Tags/Trim Reblogs buttons match.
- Enable Classic Footer and confirm that the position in the DOM, spacing, and bottom border lines on the footer rows with the Quick Tags/Trim Reblogs buttons match.


marcustyphoon:
I was debating if I think we should move our control buttons out of the queue footer; haven't decided, but either way, I feel like it would be nice not to have this bit of technical debt floating around.

Note that we could also make the `xkit-controls-row` styles not conditional on Classic Footer, but some of these changes would be needed whether we matched the native or classic footer versions; the current code matches neither. I can do that version on request.

PR #1912, opened 2025-10-10
feat(Classic Footer): Move reblog modal to the left instead of the right

### LINKED ISSUE #1924: Classic Footer: Improve reblog menu positioning
### Concept

Tumblr places the reblog button modal beneath and to the right of the reblog button (automatically changing if this would hit the screen edge). This makes sense, as the reblog button is on the left side of the post centerline. Classic Footer, however, puts the reblog button on the right side of the post centerline, so this should be reversed so that the modal isn't hanging off the right edge of the post.

| before | after |
| - | - |
| Image | Image |

marcustyphoon:
### Description

Tumblr places the reblog button modal beneath and to the right of the reblog button (automatically changing if this would hit the screen edge). This makes sense, as the reblog button is on the left side of the post centerline. Classic Footer, however, puts the reblog button on the right side of the post centerline, so this should be reversed so that the modal isn't hanging off the right edge of the post.

This does so by carefully targeting a div element containing the modal and reversing its direction. Unfortunately, there aren't a lot of super specific, semantically relevant class names/attributes in this area, so it's, you know... not the best selector in the world.

Resolves #1924.

### Testing steps

- Click a reblog button. Note that the reblog modal appears mostly to the right of the reblog button.
- Enable Classic Footer. (If you have the modal open, it won't move immediately, though it should if you resize the browser viewport or dismiss and reopen it.)
- Click a reblog button. Confirm that the reblog modal appears mostly to the left of the reblog button. Open the blog selector dropdown and confirm that its appearance and function are reasonable.
- Click a reblog button on a post with reblogs disabled. Confirm that the reblog modal appears mostly to the left of the reblog button. Open the blog selector dropdown and confirm that its appearance and function are reasonable.


src/features/classic_footer/index.js
```diff
@@ -74,6 +74,14 @@ export const styleElement = buildStyle(`
     white-space: nowrap;
   }
 
+  /* Move reblog modal to the left instead of the right */
+  body > div > div[style]:has(> ${keyToCss('withinModal')} > ${keyToCss('menuItemsWrapper', 'reblogsDisabledInfo')}) {
+    direction: rtl;
+  }
+  body > div > div[style] > ${keyToCss('withinModal')}:has(> ${keyToCss('menuItemsWrapper', 'reblogsDisabledInfo')}) {
+    direction: initial;
+  }
```
AprilSylph: * I feel like we should make `body > div` more specific, for the sake of clarity if not safety; we only care about portaled elements for this
* `withinModal` and `menuItemsWrapper` feel far too generic; if the post meatballs menu started using the same menu design, for example, these selectors might match that

```suggestion
  body > div[id^="portal/"] > div:has(a[role="menuitem"][href^="/reblog/"], ${keyToCss('reblogsDisabledInfo')) {
    direction: rtl;
  }
  body > div[id^="portal/"] > div > [role="menu"]:has(a[role="menuitem"][href^="/reblog/"], ${keyToCss('reblogsDisabledInfo')) {
    direction: initial;
  }
```

marcustyphoon:
Blocked, very mildly, by #2149; reviewable after that.

AprilSylph (approved):

marcustyphoon:
I would like to know if this works on trail items, but we don't have any right now, so ¯\\\_(ツ)\_/¯

PR #1914, opened 2025-10-10
fix(TimeFormat, Mutual Checker): Restore on-hover tooltips

marcustyphoon:
### Description

This uses the same tech as https://github.com/AprilSylph/XKit-Rewritten/issues/1654#issuecomment-2581887482 to make the timestamps that timeformat affects and mutual checker's mutual icons hoverable.

### Testing steps

- Confirm that hovering the mouse over a timeformat timestamp makes a tooltip appear.
- Confirm that hovering the mouse over a mutual checker mutual icon makes a tooltip appear.

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -32,6 +32,9 @@ const styleElement = buildStyle(`
     margin-top: 0;
     margin-left: 0;
     margin-right: 0.5ch;
+
+    /* fixes hover when covered by the "permalink" <a> element */
+    isolation: isolate;
   }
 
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I guess I should include mutual checker in this too.

marcustyphoon:
This could in theory be something like a `data-hoverable-through-overlay` attribute and an entry in one of our static stylesheets. If these were both constructed elements I'd probably propose that, but due to the way timeformat works it's less elegant there, and then it seems probably more trouble than it's worth.

PR #1917, opened 2025-10-11
 fix(Themed Posts): Adjust identical background/foreground colors

### LINKED ISSUE #1636: Themed Posts Making Poll Options Invisible
### Platform

Windows 11 (10.0.22621)

### Browser

Firefox Developer Edition 132.0b3 (64-bit)

### Addon version

v1.0.3

### Details

Having the "Themed Posts" plugin enabled seems to be making the options on a poll you haven't voted on yet invisible until hovered over. Not much more information to provide other than the screenshots below.

"Themed Posts" enabled:
Screenshot 2024-10-08 211107

"Themed Posts" disabled:
Screenshot 2024-10-08 211117

marcustyphoon:
### Description

This makes the text/accent colors used by Themed Posts a bit lighter or darker if they exactly match the background color (typically if a user doesn't have a blog description and thus has no reason to select a text color). This is a rare case, but one that makes Themed Posts quite problematic.

I put a lot of work (#1735) into using advanced color CSS for this, but it's pretty messy and prone to browser version bugs. This is what happens if you instead say, eh, doing arithmetic on rgb values is more than good enough.

Colors are lightened if they are dark and darkened if they're light, using a semi-arbitrary (color theory wise, incorrect, I believe) threshold of "do the rgb values average to 128" and an arbitrary reasonable-looking adjustment of "add/subtract 100."

Resolves #1636.

Closes #1735.

### Testing steps

- Make a blog with a black-on-black theme or view https://www.tumblr.com/transienttest (if I haven't done something else with it at the time).
- Enable Themed Posts and confirm that post text and UI elements are readable.
- Un-revert the test commit, which will make all blogs everywhere behave like they have title and link colors identical to the background color. Sample a selection of blogs with different background colors and confirm that post text and UI elements are readable.

### DIFF
--- a/src/features/themed_posts/index.js
+++ b/src/features/themed_posts/index.js
@@ -58,15 +58,11 @@ const processPosts = async function (postElements) {
       if (!blogs.has(name)) {
         blogs.add(name);
 
-        // const {
-        //   backgroundColor,
-        //   titleColor,
-        //   linkColor
-        // } = theme;
-
-        const backgroundColor = theme.backgroundColor;
-        const titleColor = theme.backgroundColor;
-        const linkColor = theme.backgroundColor;
+        const {
+          backgroundColor,
+          titleColor,
+          linkColor
+        } = theme;
 
         const backgroundColorRGB = hexToRGB(backgroundColor);
         const titleColorRGB = titleColor === backgroundColor ? createContrastingColor(titleColor) : hexToRGB(titleColor);


src/features/themed_posts/index.js
```diff
@@ -19,11 +19,24 @@ let blacklist;
 
 let reblogTrailTheming = true;
 
-const hexToRGB = (hex) => {
+const clamp = (number, lower, upper) => Math.min(Math.max(number, lower), upper);
+const average = array => array.reduce((prev, cur) => prev + cur, 0) / array.length;
```
marcustyphoon, line 23: I know you looooooooove reduce.

AprilSylph: I literally had to go and make sure there isn't such thing as `Math.mean()`. There isn't, and I'm so mad about it.

---

(Fun fact: _replying_ to code comments does _not_ count as a review!)

marcustyphoon: Just vendor lodash :3

PR #1918, opened 2025-10-11
chore: Add no-trigger option to pageModifications.register util function

marcustyphoon:
### Description

Right now, registering a callback with the pageModifications utility always runs it on existing elements on the page that match the given selector. We may not want to do this in some cases, only running code when elements are added to the page (I'm thinking of Classic Footer reblog-button-direct mode, where we may need to click the "reblog now" button in the popup when it appears, but shouldn't click if for some reason the preference becomes enabled with the popup open.)

Open to suggestions on variable name (it's local to this function, fortunately) and jsdoc.

### Testing steps

n/a

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -34,6 +34,11 @@ export default [
   jsdoc.configs['flat/recommended'],
 
   /**
+   * @see https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-hyphen-before-param-description.md
+   */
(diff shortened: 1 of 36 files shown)


marcustyphoon:
April, I assume you want to close this until it would actually be used.

(New policy idea: I actually use your name when using second person, as would be natural if we had more than two daily-active collaborators.)

AprilSylph:
> April, I assume you want to close this until it would actually be used.

Yes please.

> (New policy idea: I actually use your name when using second person, as would be natural if we had more than two daily-active collaborators.)

I like it!

PR #1927, opened 2025-10-17
feat(Classic Footer): Enable classic footer on radar posts

marcustyphoon:
### Description

Does what it says on the tin.

Fully reviewable, I think. Do we want to bother, though?

### Testing steps

Confirm that Classic Footer works on both regular and radar posts.


AprilSylph:
This is needed for #1931 since right now that breaks the reblog button on Radar posts.

marcustyphoon:
Oh, interesting!

I imagine we could util-ify some of this, of course. Haven't really gone through and looked at whether other features we have would make sense on radar as well.

PR #1931, opened 2025-10-20
feat(Classic Footer): "Turn reblog buttons back into links" 

### LINKED ISSUE #1920: Classic Footer: Add option to skip the reblog menu
### Concept

- relates to https://github.com/AprilSylph/XKit-Rewritten/issues/1901#issuecomment-3377835540

---

On-by-default option: "Turn reblog buttons back into links"

This would modify every reblog button on processable posts to:
- Take the post's reblog `<button>` element
- Clone the inner HTML of the button into our own `<a>` element
- Style the `<a>` to look the same as the `<button>` (`color`, `padding`, hover styles, active styles, focus ring)
- Set the `href` to the post's reblog URL (`/reblog/${blogName}/${idString}/${reblogKey}`)
- Set a `click` event listener on the `<a>` to call `preventDefault()` and `tumblr.navigate(currentTarget.getAttribute('href'))` if there are no modifier keys pressed
- Insert the `<a>` before the `<button>`
- Hide the `<button>`
- Hide the portaled reblog menu, in case it was open when the option was activated (not entirely sure how to do this at a glance...)

AprilSylph:
### To do
- [x] Hide the portaled reblog menu, in case it was already open
- [x] Use the portaled reblog menu to open the reblog form (restore green reblog button behaviour)
- [x] Turn reblog link green when reblog button is green
- [x] Update Quick Reblog so that the AlreadyReblogged tick doesn't move when this is activated
- [x] Properly account for <kbd>E</kbd>+<kbd>Click</kbd> shortcut
- [x] Properly account for <kbd>W</kbd>+<kbd>Click</kbd> shortcut
- [x] #1927

---

### Description

- resolves #1920

Adds an on-by-default option in Classic Footer to disable the new reblog menu.

This is achieved, as per the original issue text, creating our own anchor element with the cloned inner HTML of Tumblr's reblog button. This anchor has all the necessary CSS to perfectly visually mimic the original button.

When the reblog link is clicked without any modifier keys active, its event listener will prevent a browser-driven navigation to the reblog form, and instead open Tumblr's reblog menu in order to click the native reblog link; this is different from simply using `navigate()` in that a successful reblog from this state will turn the reblog button green.

When <kbd>E</kbd> or <kbd>W</kbd> is held down, the original reblog button is shown instead of the constructed reblog link, so that Fast Reblog or Quick Queue can be used unimpeded.

Quick Reblog's actually works on the reblog link without modification, but the AlreadyReblogged tick was in a slightly different position since the reblog link makes it look like the first variant of the split notes count post footer is active. This PR consolidates the AlreadyReblogged CSS to drop the styles specific to that variant, and make the selector for the shipped variant more greedy. (Notably, this does not actually break anything if that variant is enabled.)

#### Setup & basic functionality
1. Load the modified addon
2. Enable Classic Footer and Quick Reblog
3. Hover over a post's reblog button
    - **Expected result**: Your browser indicates that this element is an anchor linking to a `/reblog/` URL
    - **Expected result**: The Quick Reblog popup appears

#### Click behaviour
1. <kbd>Command</kbd>+<kbd>Click</kbd> (<kbd>CTRL</kbd>+<kbd>Click</kbd> on Windows/Linux) a reblog link
    - **Expected result**: The post's reblog form is opened in a new tab
2. <kbd>Shift</kbd>+<kbd>Click</kbd> a reblog link
    - **Expected result**: The post's reblog form is opened in a new window
3. Click a reblog link with no modifier keys
    - **Expected result**: The post's reblog form is opened in the current session

#### Button colour mimicking
1. With the new option enabled, finalise a reblog via the reblog form in the current session
    - **Expected result**: The reblog link for the post you reblogged turns green
2. Toggle the new option off
    - **Expected result**: The reblog button for the post you reblogged is still green

#### Reblog menu hiding/unhiding
1. With the new option still off, click a reblog button to open its reblog menu
2. Re-enable the new option
    - **Expected result**: The opened reblog menu is hidden
3. Disable the new option again
    - **Expected result**: The opened reblog menu is unhidden
4. Re-enable the new option, and find a post not processed by Classic Footer
    - **Expected result**: The reblog button is not a link
    - **Expected result**: Clicking the reblog button successfully opens its reblog menu

#### Keyboard shortcuts
1. With the new option enabled, attempt to use Tumblr's keyboard shortcuts to reblog/queue posts
    - **Expected result**: <kbd>E</kbd>+<kbd>Click</kbd> works as normal
    - **Expected result**: <kbd>W</kbd>+<kbd>Click</kbd> works as normal

#### Quick Reblog
1. Attempt to use Quick Reblog to reblog/queue/draft posts
    - **Expected result**: Reblogging turns the reblog link green, with a green tick
    - **Expected result**: Queueing turns the reblog link purple, with a purple tick
    - **Expected result**: Drafting turns the reblog link red, with a red tick
2. Refresh the feed and find the post(s) you used Quick Reblog on again
    - **Expected result**: The reblog links for the given posts are green, with a green tick
3. Toggle the new option off, and then back on again
    - **Expected result**: The position of the AlreadyReblogged tick stays the same

### DIFF
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -130,8 +130,8 @@ const processPosts = (postElements) => postElements.forEach(async postElement =>
 
   const { noteCount } = await timelineObject(postElement);
   const noteCountButton = button({ class: noteCountClass, click: onNoteCountClick }, [
-    span({}, [noteCountFormat.format(noteCount)]), ` ${noteCount === 1 ? 'note' : 'notes'}`]
-  );
+    span({}, [noteCountFormat.format(noteCount)]), ` ${noteCount === 1 ? 'note' : 'notes'}`
+  ]);
 
   const engagementControls = postElement.querySelector(engagementControlsSelector);
   engagementControls?.before(noteCountButton);


marcustyphoon:
Is there actually functionality added by making our own anchor instead of using redpop's button?

AprilSylph:
* You get to see the reblog form URL on hover (which lets you see plainly that the option is working)
* Assistive technology will correctly announce it as a link instead of a button

It's mostly to satisfy my own perfectionism, though.

marcustyphoon:
Oh, you know what:

- Right click -> open in new tab

Yeah, definitely no other way to do that.

(re: assistive technology, isn't it supposed to be a button? :P)

AprilSylph:
> Right click &rarr; open in new tab

Ah! That too.

> (re: assistive technology, isn't it supposed to be a button? :P)

No...? In the context of how it's visualised, it's a button by association because it's in a row of what are otherwise exclusively buttons... but what it does when clicked is to navigate you to a different URL, which makes it a link.

AprilSylph:
@marcustyphoon I'll work on this more tomorrow; in the meantime, feel free to conduct code review to see if this mess can be written more neatly. Maybe some util extractions would be in order...?

src/features/classic_footer/index.js
```diff
@@ -83,20 +122,99 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
...
+      }
+    }
+  });
+  mutationObserver.observe(document.body, { childList: true });
+
+  // Open the reblog menu for the observer to find.
+  reblogButton.click();
+
+  // If the reblog menu doesn't appear within 1 second, give up and do nothing.
+  sleep(1000).then(mutationObserver.disconnect);
```
marcustyphoon: Could just be a setTimeout if you want to save loc, though I don't imagine that's really what you meant by "can be written more neatly."

src/features/classic_footer/index.js
```diff
@@ -83,20 +122,99 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
...
+      .filter(addedNode => addedNode instanceof Element && addedNode.matches(reblogMenuPortalSelector));
+
+    for (const addedNode of addedNodes) {
+      const reblogMenuItem = addedNode.querySelector(reblogMenuItemSelector);
+      if (reblogMenuItem) {
+        observer.disconnect();
+        reblogMenuItem.click();
+        break;
+      }
+    }
```
marcustyphoon: refactor attempts

- could be an early return instead of a break, as there's nothing after this loop
- addedNodes isn't really an accurate variable name after that filter
- reblogMenuItemSelector includes reblogMenuPortalSelector so it might be possible to only check once (I guess if you are trying to hyperoptimize cycles you might not want to, would have to test though)
- it is, practically speaking, impossible for two reblog menu portal elements to be added to the page at the same time (and unlikely for there to be many added in succession), so I think the minimal performance impact of doing more work here is actually 0 more work (i.e. you could use forEach).

tried to cook something—a little overcooked, probably, but it's interesting (note this does nothing if a reblog menu modal appears but it's the wrong one, which... why would that happen)

```js
const getReblogMenuItem = (reblogButton, reblogMenuItemSelector) =>
  document.querySelector(reblogMenuItemSelector) ??
  new Promise(resolve => {
    const mutationObserver = new MutationObserver(mutations =>
      mutations
        .flatMap(({ addedNodes }) => [...addedNodes])
        .filter(addedNode => addedNode instanceof Element && addedNode.matches(reblogMenuPortalSelector))
        .forEach(addedNode => resolve(addedNode.querySelector(reblogMenuItemSelector)))
    );
    mutationObserver.observe(document.body, { childList: true });
    setTimeout(mutationObserver.disconnect, 1000);
    reblogButton.click();
  });
```

src/features/classic_footer/index.js
```diff
@@ -67,6 +71,41 @@ export const styleElement = buildStyle(`
...
+  }
+  .${reblogLinkClass}:focus-visible {
+    outline: 2px solid var(--brand-green);
+    outline-offset: -2px;
+  }
+  .${reblogLinkClass}:active {
+    background-color: var(--brand-green-tint-strong);
+    color: var(--brand-green);
+  }
+  footer:has(svg[style="--icon-color-primary: var(--brand-green);"] use[href="#managed-icon__ds-reblog-24"]) .${reblogLinkClass} {
```
marcustyphoon: maybe this should be `span:has(/* */) > .${reblogLinkClass}`? there aren't other green buttons in the footer, I assume, but

src/features/classic_footer/index.js
```diff
@@ -83,20 +122,99 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
...
 
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+const onReblogLinkClick = (event) => {
+  if (event.ctrlKey || event.metaKey || event.altKey || event.shiftKey) return;
+
+  event.preventDefault();
+
+  const reblogButton = event.currentTarget.parentElement.querySelector(reblogButtonSelector);
+  const href = event.currentTarget.getAttribute('href');
```
marcustyphoon, line 172: I've noticed you tend to do this kind of thing reasonably often (search the DOM for contextual information in a generic handler). Notable alternatives include a) inlining the handler so the relevant information is in-scope and b) `const onReblogLinkClick = href => event => {` /  `click: onReblogLinkClick(href),`; I don't think they're generally particularly better or worse, just different.

AprilSylph: Having a generic one is more memory-efficient! Or at least, I would hope it is. Surely there's a memory cost to constructing a new listener per post, instead of reusing the same one every time?

AprilSylph:
Labelled one of those commits as "style" when I meant "refactor" 🤦‍♀️

src/features/classic_footer/index.js
AprilSylph: I ❤️ PROMISES

marcustyphoon:
Have not checked this yet, but note to self and/or you: check behavior both on posts that cannot be reblogged which have no reblogs and posts that cannot be reblogged but which do have reblogs.

edit: looks fine 👍

src/features/classic_footer/index.js
marcustyphoon: Presumably, yep! Memory (1 closure) vs compute (DOM traversal).

src/features/classic_footer/index.js
```diff
@@ -83,20 +127,95 @@ const processPosts = (postElements) => filterPostElements(postElements).forEach(
...
+  const noteCountButton = button({ class: noteCountClass, click: onNoteCountClick }, [
+    span({}, [noteCountFormat.format(noteCount)]), ` ${noteCount === 1 ? 'note' : 'notes'}`]
+  );
 
   const engagementControls = postElement.querySelector(engagementControlsSelector);
   engagementControls?.before(noteCountButton);
 });
 
+const getReblogMenuItem = async (reblogButton, href) => {
+  const reblogMenuItemSelector = `${reblogMenuPortalSelector} a[href="${href}"]`;
```
marcustyphoon:
```suggestion
  const reblogMenuItemSelector = `${reblogMenuPortalSelector} a[href^="${href}"]`;
```

`instreamAd` sponsored posts' reblog menu items have, for example, `href="/reblog/regrettingyoumovie/797487584619184128/dlWx0nr4?ave=497962&cam=2727057&cre=41184822&agr=12922948&aid=41184822"`

src/features/classic_footer/index.js
```diff
@@ -67,6 +72,45 @@ export const styleElement = buildStyle(`
...
+  span:has(${quickActionsSelector}) > .${reblogLinkClass} {
+    display: none;
+  }
+
+  .${reblogLinkClass} ~ :is(${reblogButtonSelector}):not(:has(${quickActionsSelector})) {
+    display: none;
+  }
+  body:has(.${reblogLinkClass}) > ${reblogMenuPortalSelector} {
+    display: none;
+  }
```
marcustyphoon: Notably, this PR does not break posts in the tumblr tv interface exclusively because there are no normal posts on-screen at that time, so this selector does not apply. Not sure if we should do something more explicit (treating this like radar, for example; the big problem with that is that I hate `postOrRadarOrTumblrTvPostSelector`).

Just going for `article${keyToCss('post')}` might make sense.

(honestly truthfully having this css at all scares me—I'd be tempted to hide the popup in the click handler observer callback instead, and either manually handle the "feature is enabled with menu open" case or ignore it)

AprilSylph: I did have an idea to only hide the reblog menu of processed posts, which is a little messy code-wise (TWO style elements!) but would provide the level of safety we need.

AprilSylph: Wait, what am I talking about with just two style elements? We can add a style element inside each reblog link we create. Even if our targeted browsers had `@scope` support it's opt-in, not opt-out.

marcustyphoon: That's not functionally any different than the current code, right? Either way, it's hiding _all_ menus if _any_ reblog link is on-screen, not having the hidden state of any individual menu be tied to the existence of its reblog link.

marcustyphoon: (Unless you're talking about targeting using `:has([href="/* */"])`, which—I guess that could be what you meant by needing two style elements?)

AprilSylph: Yes, targeting each menu via `:has()` looking for the given processed post's reblog `href` so that it doesn't hide any other posts' reblog menus. By including the style element inside the reblog link, we don't need any additional logic for undoing this.

...ACTUALLY, no, not `href`, we should use the reblog button's ID to target specific post renders' reblog menus. Using just the `href` would have edge-cases like "what if there's an unprocessed version of a processed post on the page at the same time" which, granted, after #1927 hopefully will never happen, but it would be infinitely more robust for exactly the same code complexity.

(Not having merged that PR into this one yet is going to make this so easy to test...!)

AprilSylph: So it turns out the reblog button ID is only supplied to the portal if the reblog button is not hidden.

[This is extremely exploitable](https://github.com/AprilSylph/XKit-Rewritten/pull/1931/commits/e717cccbccbb490202cd3e3f761294ecd1d57234).

marcustyphoon:
`.xkit-classic-footer-reblog-link ~ :is(:is([tabindex="-1"][data-id], aside :is(.Q55_h)) article footer :is(.ndAE7, .OKxqZ, .IBIxS, .QF8_K) :is(.rtfvT, .fPlC8, .IivJ4) button:has(svg use:is([href="#managed-icon__ds-reblog-24"], [href="#managed-icon__ds-queue-add-24"]))):not(:has(svg[style="--icon-color-primary: var(--brand-blue);"], svg[style="--icon-color-primary: var(--brand-purple);"])) {`

Ah, selector composition :D

src/features/classic_footer/index.js
marcustyphoon: I tried to find the part of the AriaKit source that did this and failed completely. Seems to work, though!

marcustyphoon: Huh. It's not the button being hidden. Seems like it's the programmatic click.

src/features/classic_footer/index.js
```diff
@@ -84,20 +129,98 @@ const processPosts = (postElements) => postElements.forEach(async postElement =>
   postElement.querySelector(`.${noteCountClass}`)?.remove();
 
   const { noteCount } = await timelineObject(postElement);
-  const noteCountButton = dom('button', { class: noteCountClass }, { click: onNoteCountClick }, [
-    dom('span', null, null, [noteCountFormat.format(noteCount)]),
-    ` ${noteCount === 1 ? 'note' : 'notes'}`
-  ]);
+  const noteCountButton = button({ class: noteCountClass, click: onNoteCountClick }, [
+    span({}, [noteCountFormat.format(noteCount)]), ` ${noteCount === 1 ? 'note' : 'notes'}`]
+  );
```
marcustyphoon:
```suggestion
  const noteCountButton = button({ class: noteCountClass, click: onNoteCountClick }, [
    span({}, [noteCountFormat.format(noteCount)]), ` ${noteCount === 1 ? 'note' : 'notes'}`
  ]);
```

marcustyphoon:
Interesting oddity: things like right clicking a link in Firefox and selecting the option to search the web for the text contents or dragging a link in safari and observing the tooltip will use the contents of the contained style element as the link text. This is probably minor enough to make it not worth bothering moving the style element one level up and making cleanup more annoying, but we could.

marcustyphoon (approved):
Above lint suggestion aside:

- [x] listed tests work as expected in—for variety—desktop safari 18.5
- [x] works on sponsored posts (that have data ids; those without are ignored by this feature even though they have a footer and can be liked)
- [x] works on blazed posts
- [x] works on "review flagged posts" page
- [x] works on likes page in grid view
- [x] works on private posts as much as redpop itself does, which is to say it opens the reblog interface even though the reblog attempt will error
- [x] correct behavior on posts that cannot be reblogged with no reblogs
- [x] correct behavior on posts that cannot be reblogged that have reblogs
- [x] does not break tumblr tv

PR #1934, opened 2025-10-22
fix(Collapsed Queue): Infinite loop that freezes page

### LINKED ISSUE #1933: Collapsed Queue: use causes Firefox to become unresponsive
### Platform

Windows 10 22H2

### Browser

Firefox 144

### Addon version

1.1.0

### Details

As of this morning, attempts to load tumblr drafts or queue pages with Collapsed Queue enabled causes that Firefox tab to become unresponsive. Specifically, Firefox issues a warning at the top of that tab stating ""Xkit Rewritten" is slowing down Firefox. To speed up your browser, stop that extension." The posts in drafts/queue do not load, nor does the collapse get a chance to happen, as the tab is frozen at this point.

I tried disabling "run on the queue page" and opening a new tab with the queue: this loads the queue, but obviously without the collapse function. I did the same with drafts, disabling "run on the drafts page" and opening a new tab with the drafts: this loads the drafts, but obviously without the collapse function.

However, this led to the discovery that leaving Collapsed Queue enabled overall will cause Xkit Rewritten to stop working entirely on other tumblr tabs that do not even have the drafts or queue open. So, for now, I have disabled Collapsed Queue with great sadness. I did try a browser restart as well as a computer restart, but had no luck.

marcustyphoon:
### Description

Yikes.

This updates Collapsed Queue for the extra `div` that was added to the post DOM structure a few days ago. I thought I had tested this, but I misremembered; I tested Collapsed Queue with the new split-notes footer, but not the extra `div`.

Resolves #1933.

### Testing steps

Confirm that Collapsed Queue works on both the draft and queue pages.

### DIFF
--- a/src/features/collapsed_queue/index.js
+++ b/src/features/collapsed_queue/index.js
@@ -21,9 +21,9 @@ const processPosts = async function (postElements) {
     const container = Object.assign(document.createElement('div'), { className: containerClass });
     wrapper.append(container);
 
-    headerElement.after(wrapper);
-    while (wrapper.nextElementSibling && wrapper.nextElementSibling !== footerElement) {
-      container.append(wrapper.nextElementSibling);
+    footerElement.before(wrapper);
+    while (wrapper.previousElementSibling && wrapper.previousElementSibling !== headerElement) {
+      container.prepend(wrapper.previousElementSibling);
     }
   });
 };


marcustyphoon:
Also doable declaratively, though it's sort of a pain to implement the same logic. ex:

```js
    footerElement.before(
      div({ class: wrapperClass }, [
        div({ class: containerClass }, [
          ...footerElement.parentElement.querySelectorAll(
            `:scope > :not(:has(~ header), header, ${footerSelector}, ${footerSelector} ~ *)`
          )
        ])
      ])
    );
```

```js
    const elements = [...footerElement.parentElement.children];

    const headerIndex = elements.findIndex(element => element.matches('header'));
    const footerIndex = elements.findIndex(element => element === footerElement);

    footerElement.before(
      div({ class: wrapperClass }, [
        div({ class: containerClass },
          elements.slice(
            headerIndex !== -1 ? headerIndex : 0,
            footerIndex !== -1 ? footerIndex : undefined
          )
        )
      ])
    );
```

Or one could still use a loop but run it on an array of elements, avoiding the multiple DOM manipulations.

AprilSylph (approved):

PR #1935, opened 2025-10-23
refactor: Create and use generic portal traversal utility

marcustyphoon:
### Description

A little messy, but it works? Maybe helpful for #1931? I dunno. Possibly not worth it otherwise?

This:

- Places both `xkit-injection-request` and `xkit-injection-response` handlers on `document.documentElement` (yes, this could mean a few extra handlers firing once in a blue moon; _one could make `xkit-injection-element-response` and add the event listener twice to prevent this I guess but I don't think that would save any cycles_ **edit:** see comment)
- Allows injected functions to return DOM elements by passing them from the main world to the isolated world in the same way we pass the target DOM element from the isolated to the main world: event bubbling.
- Uses this ability to replace the overly-specific injected function in the meatballs menu util with, essentially a version of `.closest()` that traverses portals.

### Testing steps

- Confirm that meatballs menu items are still added to post meatball menus.

More tests could be run with #677/#1662, I guess.

### DIFF
--- a/src/utils/inject.js
+++ b/src/utils/inject.js
@@ -24,7 +24,7 @@ export const inject = (path, args = [], target = document.documentElement) =>
         reject(exception);
       } else if (type === 'xkit-injection-element-response') {
         resolve(target);
-      } else if (result) {
+      } else {
         resolve(result);
       }
     };


marcustyphoon:
Eh, you know, maybe this actually is more elegant. Leave the normal path's listeners where they are so calls don't trigger each other's listeners.

```diff
diff --git a/src/main_world/index.js b/src/main_world/index.js
index 329566b6..0e0a5858 100644
--- a/src/main_world/index.js
+++ b/src/main_world/index.js
@@ -19,10 +19,10 @@ document.documentElement.addEventListener('xkit-injection-request', async event
 
     if (result instanceof Element) {
       result.dispatchEvent(
-        new CustomEvent('xkit-injection-response', { detail: JSON.stringify({ id }), bubbles: true })
+        new CustomEvent('xkit-injection-element-response', { detail: JSON.stringify({ id }), bubbles: true })
       );
     } else {
-      document.documentElement.dispatchEvent(
+      target.dispatchEvent(
         new CustomEvent('xkit-injection-response', { detail: JSON.stringify({ id, result }) })
       );
     }
diff --git a/src/utils/inject.js b/src/utils/inject.js
index b0d49079..1591935a 100644
--- a/src/utils/inject.js
+++ b/src/utils/inject.js
@@ -13,20 +13,23 @@ export const inject = (path, args = [], target = document.documentElement) =>
     const requestId = String(Math.random());
     const data = { path: browser.runtime.getURL(path), args, id: requestId };
 
-    const responseHandler = ({ detail, bubbles, target }) => {
+    const responseHandler = ({ detail, type, target }) => {
       const { id, result, exception } = JSON.parse(detail);
       if (id !== requestId) return;
 
       target.removeEventListener('xkit-injection-response', responseHandler);
+      document.documentElement.removeEventListener('xkit-injection-element-response', responseHandler);
+
       if (exception) {
         reject(exception);
-      } else if (bubbles) {
+      } else if (type === 'xkit-injection-element-response') {
         resolve(target);
       } else if (result) {
         resolve(result);
       }
     };
-    document.documentElement.addEventListener('xkit-injection-response', responseHandler);
+    target.addEventListener('xkit-injection-response', responseHandler);
+    document.documentElement.addEventListener('xkit-injection-element-response', responseHandler);
 
     target.dispatchEvent(
       new CustomEvent('xkit-injection-request', { detail: JSON.stringify(data), bubbles: true })
```

src/utils/inject.js
```diff
@@ -13,14 +13,20 @@ export const inject = (path, args = [], target = document.documentElement) =>
     const requestId = String(Math.random());
```
AprilSylph, line 13: Unrelated to this PR, but this line makes me scared. Any reason we can't use [`randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) instead for extra entropy?

marcustyphoon:
The remaining parts of the above diff:

```diff
diff --git a/src/main_world/index.js b/src/main_world/index.js
index 08508f7e..0e0a5858 100644
--- a/src/main_world/index.js
+++ b/src/main_world/index.js
@@ -22,7 +22,7 @@ document.documentElement.addEventListener('xkit-injection-request', async event
         new CustomEvent('xkit-injection-element-response', { detail: JSON.stringify({ id }), bubbles: true })
       );
     } else {
-      document.documentElement.dispatchEvent(
+      target.dispatchEvent(
         new CustomEvent('xkit-injection-response', { detail: JSON.stringify({ id, result }) })
       );
     }
diff --git a/src/utils/inject.js b/src/utils/inject.js
index b29e2953..1591935a 100644
--- a/src/utils/inject.js
+++ b/src/utils/inject.js
@@ -17,7 +17,7 @@ export const inject = (path, args = [], target = document.documentElement) =>
       const { id, result, exception } = JSON.parse(detail);
       if (id !== requestId) return;
 
-      document.documentElement.removeEventListener('xkit-injection-response', responseHandler);
+      target.removeEventListener('xkit-injection-response', responseHandler);
       document.documentElement.removeEventListener('xkit-injection-element-response', responseHandler);
 
       if (exception) {
@@ -28,7 +28,7 @@ export const inject = (path, args = [], target = document.documentElement) =>
         resolve(result);
       }
     };
-    document.documentElement.addEventListener('xkit-injection-response', responseHandler);
+    target.addEventListener('xkit-injection-response', responseHandler);
     document.documentElement.addEventListener('xkit-injection-element-response', responseHandler);
 
     target.dispatchEvent(
```

src/utils/inject.js
```diff
@@ -13,14 +13,23 @@ export const inject = (path, args = [], target = document.documentElement) =>
...
-      exception ? reject(exception) : resolve(result);
+      document.documentElement.removeEventListener('xkit-injection-response', responseHandler);
+      document.documentElement.removeEventListener('xkit-injection-element-response', responseHandler);
+
+      if (exception) {
+        reject(exception);
+      } else if (type === 'xkit-injection-element-response') {
+        resolve(target);
+      } else if (result) {
+        resolve(result);
```
AprilSylph: I don't think this `resolve()` needs to become conditional:

```suggestion
      } else {
        resolve(result);
```

src/utils/inject.js
marcustyphoon: Just checked the profiler; it'd be like 0.1ms or whatever, so yep, seems fine.

I started trying to do the math for how likely it is that two of these values are the same in a short period, but that would require me to actually understand xorshift128+. I guess realistically one would just sample—no, that's not going to work, you'll never find one. Man... tom7 could do this in like 4 seconds, I just know it. Anyway, my basic understanding is that the type of psrng used in crypto.getRandomValues is probably reseeded every couple of seconds and in between that interval I don't see any reason to think both it and math.random() are pretty well behaved for this use case, so my guess is that it's a reasonable approximation to say the reduction in repetition chance is roughly equal to the 12 extra bits of precision we're getting.

Napkin math puts the error rate around "once every 25 years of every one of our users having one tab continuously open with scroll to bottom active" with Math.random(), with however many orders of magnitude in either direction due to How The Math Works, which is honestly a lot higher than I thought.

(It's possible that the error rate is actually "literally zero" within certain boundaries with math.random() and higher with crypto.getRandomValues, if the LFSR actually _can't_ give the same value within a few cycles, which I think might be true for nontrivial seeds? Okay, not that it has to be the same value, but the same top 52 bits, which—no, yeah, then that's probably possible. I dunno, I don't know math :D)

PR #1937, opened 2025-10-24
fix(Classic Footer): hide CSS from Firefox context menus

AprilSylph:
### Description

- relates to https://github.com/AprilSylph/XKit-Rewritten/pull/1931#issuecomment-3440323018

  > Interesting oddity: things like right clicking a link in Firefox and selecting the option to search the web for the text contents or dragging a link in safari and observing the tooltip will use the contents of the contained style element as the link text. This is probably minor enough to make it not worth bothering moving the style element one level up and making cleanup more annoying, but we could.

Alternate solution: use a `<link rel="stylesheet">` with a `data:` URI.

`link` elements are void, so rightfully ignored by Firefox's context menu.

This reduces the searchable text of the anchor element down to... the linked URL itself.

(I can't believe Chrome doesn't throw a fit about CSP for this???)

#### Fix testing – Firefox
1. Load the modified addon in Firefox
2. Open a new Tumblr tab
3. Open a post's reblog menu
4. Enable Classic Footer
5. Right-click a reblog link
    - **Expected result**: The "Search for" text contains none of our menu-hiding CSS

#### Fix testing – Safari
1. Load the modified addon in Firefox
2. Open a new Tumblr tab
3. Open a post's reblog menu
4. Enable Classic Footer
5. Drag-click a reblog link
    - **Expected result**: The tooltip text contains none of our menu-hiding CSS

#### Regression testing – Chrome
1. Load the modified addon in Chrome
2. Open a new Tumblr tab
3. Open a post's reblog menu
4. Enable Classic Footer
    - **Expected result**: The reblog menu is hidden
5. Disable the "Turn reblog buttons back into links" option
    - **Expected result**: The reblog menu is unhidden

### DIFF
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -1,5 +1,5 @@
 import { keyToCss } from '../../utils/css_map.js';
-import { a, button, span } from '../../utils/dom.js';
+import { a, button, span, link } from '../../utils/dom.js';
 import { buildStyle, postSelector } from '../../utils/interface.js';
 import { pageModifications } from '../../utils/mutations.js';
 import { getPreferences } from '../../utils/preferences.js';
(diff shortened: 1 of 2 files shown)


AprilSylph:
> Hm—I wonder if https://issues.chromium.org/issues/40654624 is relevant.

I suppose we'll find out after we release this!

marcustyphoon:
I mean, sort of—this is a feature precisely zero people, including us, would notice the breakage of in practice :D

PR #1941, opened 2025-10-25
chore: Improve some feature descriptions

marcustyphoon:
### Description

Here are some proposed feature description improvements. Edits welcome.

I was going to make this 5 PRs, but that seems like an unnecessary amount of overhead. We can just remove any you think are best unchanged.

### Testing steps

Make sure the updated feature descriptions look good in-situ. (Safari on iPad is the main place where they would probably look bad, as it has a narrow extension popup display, but I definitely don't think we care about that.)

### DIFF
--- a/src/features/classic_search/feature.json
+++ b/src/features/classic_search/feature.json
@@ -1,6 +1,6 @@
 {
   "title": "Classic Search",
-  "description": "Go to tagged pages easily",
+  "description": "Go to tagged pages more easily",
   "icon": {
     "class_name": "ri-search-line",
     "color": "white",


src/features/collapsed_queue/feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "Collapsed Queue",
-  "description": "Shorten posts in your queue",
+  "description": "Shorten posts in your queue (or drafts)",
```
AprilSylph: We're already using parentheses for (disabled) and (deprecated), so would prefer not to use them in feature descriptions as well

```suggestion
  "description": "Shorten posts in your queue or drafts",
```

src/features/notificationblock/feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "NotificationBlock",
-  "description": "Block a post's notifications",
+  "description": "Hide notifications from a post",
```
AprilSylph: I think "activity" is actually more accurate here; "notifications" could be construed as referring to the note view you can open on posts. (Yes, that component is called `PostActivity` internally; no, it's not an accurate name.)

```suggestion
  "description": "Hide activity about any post",
```

src/features/show_originals/feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "Show Originals",
-  "description": "Hide reblogged posts by default",
+  "description": "Add a no-reblogs tab to your timeline",
```
AprilSylph: I dunno...

We could go for utmost accuracy and readability, without necessarily clarifying what the antithesis of an original post _is_ (which notably, Tumblr seems to be fine with; we do lift this string from their translations):
```suggestion
  "description": "Add an “Original Posts” tab to your timelines",
```

Or we can try to tweak this wording to retain all the information within it, but try to make "reblogs" more obviously a noun:
```suggestion
  "description": "Add a reblogs-free tab to your timelines",
```

Or we could leave this one unchanged, hoping that it is juxtaposing "original" and "reblog" so effectively that it serves to educate people about the platform... that they already feel the need to add a bunch of complexity to... maybe not.
```suggestion
  "description": "Hide reblogged posts by default",
```

src/features/collapsed_queue/feature.json
marcustyphoon: "Shorten posts in your queue—or drafts!" Mm, possibly too colloquial.

src/features/notificationblock/feature.json
marcustyphoon: I like to use "activity items," would that fit here?

AprilSylph: :white_check_mark: 

```suggestion
  "description": "Hide activity items about any post",
```

src/features/collapsed_queue/feature.json
AprilSylph:
```suggestion
  "description": "Shorten posts in your queue, or drafts",
```
```suggestion
  "description": "Shorten posts in your queue/drafts",
```

src/features/show_originals/feature.json
marcustyphoon: I like the former two. I think "by default" doesn't really reflect the fact that the optional part is controlled using the UI; that makes me think the feature itself has multiple options and one is to not do what the description says.

Would it thus be "Add [] tabs to your timelines"?

src/features/notificationblock/feature.json
AprilSylph: Eeeeh actually I dunno, I like the terseness of just "activity".

When referring to just one activity item, it's "an activity item" for sure, but when it's multiple, it's just "activity" to me.

marcustyphoon: oops

src/features/collapsed_queue/feature.json
marcustyphoon: So far I think `queue or drafts` is clearest. Not seeing a good way to lampshade the fact that the feature name only references one of the two, but I think that's definitely optional.

src/features/show_originals/feature.json
AprilSylph: I've got a bias towards the very first option.

src/features/notificationblock/feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "NotificationBlock",
-  "description": "Block a post's notifications",
+  "description": "Hide activity items about any post",
```
AprilSylph: Yeah, no, it feels too off.


```suggestion
  "description": "Hide activity about any post",
```

src/features/collapsed_queue/feature.json
AprilSylph: Definitely. The beauty of the expand-preferences-on-enable behaviour is that features with few preferences don't _need_ to have that level of specificity in their descriptions. As long as it tells a prospective user that they might want to enable it, and a regretful user that it may be what they're looking to disable, it's fine. Anything else is a bonus.

src/features/show_originals/feature.json
marcustyphoon: `Add an “Original Posts” tab to your timelines` / `Add “Original Posts” tabs to your timelines`

Both seem correct, I think.

AprilSylph: Former seems... more correct... to me. The quantity per timeline is one; specifying "timelines", plural, makes it say we add one tab _per_ timeline.

"to each timeline" or "to every timeline" would make it even more clearer that it's a one-per-[thing that there are multiple of] deal, but the feature doesn't actually affect each and every timeline on the site, so that would be misleading.

And trying to list the timelines which it does/can run on would make the description far, far too long...

And "to some timelines" is just a weird way to advertise something.

marcustyphoon:
I guess for consistency, rather than

`Hide activity about any post` / `Disappear all instances of a post` / `Reblog posts with a single click` / `Trim threads after posting or drafting` / `Add tags to posts easily`

it should be either

`Hide activity about a post` / `Disappear all instances of a post` / `Reblog a post with a single click` / `Trim a thread after posting or drafting` / `Easily add tags to a post`

or

`Hide activity about any post` / `Disappear all instances of any post` / `Reblog any post with a single click` / `Trim any thread after posting or drafting` / `Easily add tags to any post`

or [the option where they're all plural but I don't think this one works].

I kind of hate "a thread" or "any thread" so personally I would break the rule on that one.

AprilSylph:
_It's not bikeshedding if it's marketing!_ and other slogans which I just made up

marcustyphoon:
_It's not bikeshedding if you long-ago made the decision to enforce your UI being really small, making painstakingly ascribing meaning to sets of about five words the only practical way to describe your large and complex set of features!_

marcustyphoon:
> Could we also trim the word "posts" from Vanilla Audio and Vanilla Video's descriptions? They're the only two features with two-line descriptions at 375[?]px, and "audio posts" and "video posts" aren't _really_ things anymore.

Hey look, now they aren't!

(I'd be somewhat tempted to re-add "players" but I think objectively the current versions are also correct.)

AprilSylph:
> Replying to https://github.com/AprilSylph/XKit-Rewritten/pull/1941#issuecomment-3447763356

There's not quite as much inconsistency as this, really... because the generic "posts"/"threads" verbiage is for when the feature applies to all relevant posts no matter what, while "any post"/"a post" denotes that which posts it applies to is user-selected. So there's an intentional divide here.

PostBlock with the description "Disappear all instances of posts" would imply it gets rid of all posts. No more posts on your dashboard, ever. A hilarious idea, but not at all what the feature does.

On the reverse, Quick Reblog with the description "Reblog a post with a single click" could conceivably leave someone wondering "What, just the one? How will I choose which one is worth this convenience?" which, granted, is an absurd interpretation given even basic background knowledge, but still.

+1 on avoiding "any post" for riding the fence between these two extremes, though.

(To be clear: I've only just consciously realised that this distinction exists and that I've been using it. If there are pre-existing exceptions to this rule, we should consider aligning those too.)

marcustyphoon:
I only mentioned the ones that you apply by clicking a post, yeah. The other ones are all plural, which makes sense. 

Hadn't thought about the fact that PostBlock applies to multiple posts once applied! Certainly true.

AprilSylph:
I think maybe more than anything, I've been subconsciously following the user story when writing feature descriptions.

Who needs PostBlock? Someone who just saw a post they hate so much that they want a way to destroy it forever. Give them the ability to "disappear all instances of a post", that user's problem is solved. They can do it again, of course, but the first event is singular.

Who needs Quick Reblog? Someone who reblogs a lot of posts and is tired of clicking a lot. Give them the ability to "reblog posts with a single click", that user's problem is solved. They already know that their problem is one of multiplicity; by using the plural "posts" we acknowledge their frustration more fully, thus making it a more effective advertisement.

<sub>Disclaimer: I have not actually studied marketing.</sub>

marcustyphoon:
Hm, I guess it's more about grammar than the singular/plural actually matching what happens.

You click a specific post to add it to the list of posts that always have an effect applied: `Hide activity about *a post*` / `Disappear all instances of a post`. "Posts" doesn't describe you choosing one, so it has to be singular, though functionally the feature is applying to multiple. (The singular is describing the feature working on the post you selected.)

You click a specific post once: `Reblog posts with a single click` / `Trim threads after posting or drafting`
`Add tags to posts easily`. "A post" is somewhat unclear, even though functionally the feature is only being used on one post at any given moment. (The plural, I guess, is describing the **ability to use the feature** being on multiple posts; we're not referring to an ongoing effect like we were in the above cases).

Confusing, but I think correct!

**edit:** User story concept is consistent with this, I think.

marcustyphoon:
`Add tags to posts easily` / `Easily add tags to posts` ?

marcustyphoon:
`Disappear avatars from specific blogs` / `Disappear avatars of specific blogs` ?

AprilSylph:
> `Add tags to posts easily` / `Easily add tags to posts` ?

image

Maybe `Add tags to posts more easily`...? It's 10:56pm on a Saturday, maybe my brain is made of dog turd at this minute.

> `Disappear avatars from specific blogs` / `Disappear avatars of specific blogs` ?

gif

`Disappear a blog's avatar`

Just kidding. If the way you added a blog to the hidelist was a menu item (hey, that would be neat, UUID-based avatar hiding...) then I would be serious, but it isn't, so `Disappear avatars of specific blogs` wins. Or maybe `Disappear avatars of specified blogs` if there's room.

PR #1942, opened 2025-10-25
feat(Tweaks): Restore new post indicator hiding option

marcustyphoon:
### Description

This reverts commit d51177d1f8b8b01c08a23953f8451e0c4679c8b5, restoring the tweak in question and updating it to reference the new `↑ New posts` indicator button text. The internal tweak name is left the same so that users who had it enabled before will still have it enabled.

### Testing steps

- With the restored tweak disabled, scroll down the dashboard following tab and wait until the floating new posts indicator button appears.
- Confirm that enabling the tweak hides the button.

To test preference inheritance:

- `git reset --hard v1.1.0`
- `npm start`
- enable Tweaks and the `Hide the "Now, where were we?" button` option
- pull the PR branch
- open a tumblr page and confirm that the "Hide the floating “New posts” indicator button on the dashboard" option is now enabled

### DIFF
--- a/src/features/tweaks/hide_new_posts_indicator.js
+++ b/src/features/tweaks/hide_new_posts_indicator.js
@@ -2,5 +2,7 @@ import { keyToCss } from '../../utils/css_map.js';
 import { buildStyle } from '../../utils/interface.js';
 
 export const styleElement = buildStyle(`
-${keyToCss('wrapper')} ${keyToCss('newPostIndicator')} { display: none; }
+${keyToCss('wrapper')} ${keyToCss('newPostIndicator')} {
+  display: none;
+}
 `);


marcustyphoon:
Ah, wait, hold on a second, the text is different now.

src/features/tweaks/feature.json
```diff
@@ -102,6 +102,11 @@
       "type": "checkbox",
       "label": "Hide the Activity notification badge",
       "default": false
+    },
+    "no_where_were_we": {
+      "type": "checkbox",
+      "label": "Hide the floating \"new posts\" indicator button on the dashboard",
```
AprilSylph: * Let's please match the text case for options like these, _especially_ inside quotation marks
* We could use real quotation marks instead of needing escape sequences, now that I think about it

```suggestion
      "label": "Hide the floating “New posts” indicator button on the dashboard",
```

PR #1945, opened 2025-10-26
feat(Tweaks): Accurately hide/show mobile menu notification badge

### LINKED ISSUE #1943: Tweaks: "Hide the ... notification badge / unread count" breaks all mobile burger menu badges
### Platform

Android 16 & Windows 11

### Browser

Firefox 144.0.1 (Android), 144.0 (Windows)

### Addon version

1.1.0

### Details

Enabling any of these three options:

Image

Globally (incorrectly) hides this generic badge that appears in mobile/narrow view:

Image

Example:

Image

Image

And yet:
Image

Edit: as I see #1256 

> given the choice between ghost notifications and having to click the menu to check if I have activity, I'd rather click

Personally, without the icon prompt to prompt me, I don't think to click at all. 

Perhaps a warning on these options for the unexpected behaviour would avoid confusion, if this isn't something feasible to solve?

marcustyphoon:
### Description

In #1393, we made the mobile menu badge hidden whenever a tweak is enabled that could make it incorrect; ideally a tweak would hide it only when the only reason it is shown is because of the notification type the tweak in question is about.

This... well, makes the tweaks hide it only when the only reason it is shown is because of the notification type the tweak in question is about. To do this, we have to extract the raw counts, remove the types we're ignoring, and add up the rest.

I don't know that the review effort to reward ratio is on this is worth it (it took me over three hours to write), but hey, it can be done!

Resolves #1943.

### Testing steps

todo; ping me and I'll add it

### DIFF
--- a/src/utils/mobile_menu_badge_hide.js
+++ b/src/utils/mobile_menu_badge_hide.js
@@ -4,7 +4,7 @@ import { buildStyle } from './interface.js';
 import { pageModifications } from './mutations.js';
 
 const mobileBadgeSelector = `header ${keyToCss('hamburger')} + ${keyToCss('notificationBadge')}`;
-const hideBadgeClass = 'xkit-mobile-menu-badge-hidden';
+const hideBadgeClass = 'xkit-hide-mobile-menu-badge';
 
 document.documentElement.append(
   buildStyle(`.${hideBadgeClass} ${mobileBadgeSelector} { transform: scale(0); }`)


src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,65 @@
...
+      const badgeData = await inject('/main_world/unbury_mobile_badge_data.js', [], mobileBadge);
+      if (badgeData) {
+        const countsByType = {
+          home: badgeData.unseenPostCount ?? 0,
+          communities: badgeData.unreadCommunityPostTotal ?? 0,
+          activity: badgeData.notificationCount ?? 0,
+          messages: badgeData.unreadMessagesCount ?? 0,
+          inbox: badgeData.inboxCount ?? 0,
+          account: badgeData.privateGroupBlogUnreadPostTotal ?? 0,
+        };
```
marcustyphoon, line 48: This mapping could certainly be in the injected function. In general I think it's nice to have the injected function contents be very generic and put as much meaningful logic in the regular extension files as is practical.

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,65 @@
...
+        mobileBadge.classList[total === 0 ? 'add' : 'remove'](hideBadgeClass);
+      }
+    } else {
+      mobileBadge.classList.remove(hideBadgeClass);
+    }
+  }
+});
+
+pageModifications.register(mobileBadgeSelector, () => mobileMenuBadgeHide.trigger());
+setInterval(() => mobileMenuBadgeHide.trigger(), 10 * 1000);
```
marcustyphoon: This was going to be a mutationobserver with `{ attributeFilter: ['class'] }` before I realized that, for example, going from the state of "I have unseen posts (ignored)" to "I have unseen posts (ignored) and an inbox message (not ignored, want to see this, please trigger a refresh)" doesn't result in anything happening to the badge.

AprilSylph: **nit**: prefer using underscore separators instead of in-place multiplication

```suggestion
setInterval(() => mobileMenuBadgeHide.trigger(), 10_000);
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,65 @@
...
+          account: badgeData.privateGroupBlogUnreadPostTotal ?? 0,
+        };
+
+        this.excludedTypes.forEach(type => { delete countsByType[type]; });
+
+        const total = Object.values(countsByType).reduce((a, b) => a + b, 0);
+
+        console.log({ excludedTypes: this.excludedTypes, badgeData, countsByType, total });
+
+        mobileBadge.classList[total === 0 ? 'add' : 'remove'](hideBadgeClass);
```
AprilSylph: `Array.prototype.reduce` detected blow it up NOW

(Since all this util does is hide or unhide the badge, we don't actually need the exact total, we just need to know if it's greater than zero.)

```suggestion
        const shouldHideBadge = Math.max(...Object.values(countsByType)) === 0;
        mobileBadge.classList.toggle(hideBadgeClass, shouldHideBadge);
```

I almost wrote a second version here using `Array.prototype.some` to stop iterating on the values array as soon as we see anything above zero, but like, come on, simple number comparison should be the literal cheapest thing a CPU can do.

src/utils/mobile_menu_badge_hide.js
marcustyphoon: Thought of it, wasn't sure if you were a fan!

src/utils/mobile_menu_badge_hide.js
marcustyphoon: `Object.values(countsByType).some(Boolean)`? I actually think that's more readable, no?

AprilSylph: You know what, why am I complicating things? We want to know if every value is zero, let's actually write it like that.

```suggestion
        const shouldHideBadge = Object.values(countsByType).every(value => value === 0);
        mobileBadge.classList.toggle(hideBadgeClass, shouldHideBadge);
```

marcustyphoon: Ah, there it is.

And yes, I copied code from another part of the extension and thus forgot about `.toggle()` again.

marcustyphoon:
an option:

```diff
-        this.excludedTypes.forEach(type => { delete countsByType[type]; });
+        const shouldHideBadge = Object.entries(countsByType)
+          .every(([key, value]) => this.excludedTypes.has(key) || value === 0);
 
-        const shouldHideBadge = Object.values(countsByType).every(value => value === 0);
         mobileBadge.classList.toggle(hideBadgeClass, shouldHideBadge);
       }
```

marcustyphoon:
also an option:

```diff
diff --git a/src/main_world/unbury_mobile_badge_data.js b/src/main_world/unbury_mobile_badge_data.js
index 4c0104ea..54fb096d 100644
--- a/src/main_world/unbury_mobile_badge_data.js
+++ b/src/main_world/unbury_mobile_badge_data.js
@@ -7,7 +7,7 @@ export default function unburyMobileBadgeData () {
     const { unseenPostCount } = fiber.memoizedProps || {};
     if (unseenPostCount !== undefined) {
       const memoizedPropsEntries = Object.entries(fiber.memoizedProps);
-      return Object.fromEntries(memoizedPropsEntries.filter(([key, value]) => typeof value === 'number'));
+      return memoizedPropsEntries.filter(([key, value]) => typeof value === 'number');
     } else {
       fiber = fiber.return;
     }
diff --git a/src/utils/mobile_menu_badge_hide.js b/src/utils/mobile_menu_badge_hide.js
index 2a6a1e82..55875a3b 100644
--- a/src/utils/mobile_menu_badge_hide.js
+++ b/src/utils/mobile_menu_badge_hide.js
@@ -12,14 +12,23 @@ document.documentElement.append(
 
 /** @typedef {'home' | 'communities' | 'activity' | 'messages' | 'inbox' | 'account'} NotificationType */
 
+const notificationKeysByType = {
+  home: 'unseenPostCount',
+  communities: 'unreadCommunityPostTotal',
+  activity: 'notificationCount',
+  messages: 'unreadMessagesCount',
+  inbox: 'inboxCount',
+  account: 'privateGroupBlogUnreadPostTotal',
+};
+
 export const mobileMenuBadgeHide = Object.freeze({
-  excludedTypes: new Set(),
+  excludedNotificationKeys: new Set(),
 
   /**
    * @param {NotificationType} type - Type of notification to exclude from mobile menu badge visibility
    */
   register (type) {
-    this.excludedTypes.add(type);
+    this.excludedNotificationKeys.add(notificationKeysByType[type]);
     this.trigger();
   },
 
@@ -27,7 +36,7 @@ export const mobileMenuBadgeHide = Object.freeze({
    * @param {NotificationType} type - Type of notification to stop excluding from mobile menu badge visibility
    */
   unregister (type) {
-    this.excludedTypes.delete(type);
+    this.excludedNotificationKeys.delete(notificationKeysByType[type]);
     this.trigger();
   },
 
@@ -35,21 +44,12 @@ export const mobileMenuBadgeHide = Object.freeze({
     const mobileBadge = document.querySelector(mobileBadgeSelector);
     if (!mobileBadge) return;
 
-    if (this.excludedTypes.size) {
+    if (this.excludedNotificationKeys.size) {
       const badgeData = await inject('/main_world/unbury_mobile_badge_data.js', [], mobileBadge);
       if (badgeData) {
-        const countsByType = {
-          home: badgeData.unseenPostCount ?? 0,
-          communities: badgeData.unreadCommunityPostTotal ?? 0,
-          activity: badgeData.notificationCount ?? 0,
-          messages: badgeData.unreadMessagesCount ?? 0,
-          inbox: badgeData.inboxCount ?? 0,
-          account: badgeData.privateGroupBlogUnreadPostTotal ?? 0,
-        };
-
-        this.excludedTypes.forEach(type => { delete countsByType[type]; });
-
-        const shouldHideBadge = Object.values(countsByType).every(value => value === 0);
+        const shouldHideBadge = badgeData
+          .every(([key, value]) => this.excludedNotificationKeys.has(key) || !value);
+
         mobileBadge.classList.toggle(hideBadgeClass, shouldHideBadge);
       }
     } else {
```

not as elegant as I'd hoped, really

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,62 @@
...
+  buildStyle(`.${hideBadgeClass} { transform: scale(0); }`)
+);
+
+/** @typedef {'home' | 'communities' | 'activity' | 'messages' | 'inbox' | 'account'} NotificationType */
+
+export const mobileMenuBadgeHide = Object.freeze({
+  excludedTypes: new Set(),
+
+  /**
+   * @param {NotificationType} type - Type of notification to exclude from mobile menu badge visibility
```
AprilSylph:
```suggestion
   * @param {NotificationType} type Type of notification to exclude from mobile menu badge visibility
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,62 @@
...
+   */
+  register (type) {
+    this.excludedTypes.add(type);
+    this.trigger();
+  },
+
+  /**
+   * @param {NotificationType} type - Type of notification to stop excluding from mobile menu badge visibility
```
AprilSylph:
```suggestion
   * @param {NotificationType} type Type of notification to stop excluding from mobile menu badge visibility
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,62 @@
+import { keyToCss } from './css_map.js';
+import { inject } from './inject.js';
+import { buildStyle } from './interface.js';
+import { pageModifications } from './mutations.js';
+
+const mobileBadgeSelector = `header ${keyToCss('hamburger')} + ${keyToCss('notificationBadge')}`;
+const hideBadgeClass = 'xkit-mobile-menu-badge-hidden';
+
+document.documentElement.append(
+  buildStyle(`.${hideBadgeClass} { transform: scale(0); }`)
```
AprilSylph:
```suggestion
  buildStyle(`.${hideBadgeClass} ${mobileBadgeSelector} { transform: scale(0); }`)
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,62 @@
...
+          activity: badgeData.notificationCount ?? 0,
+          messages: badgeData.unreadMessagesCount ?? 0,
+          inbox: badgeData.inboxCount ?? 0,
+          account: badgeData.privateGroupBlogUnreadPostTotal ?? 0,
+        };
+
+        this.excludedTypes.forEach(type => { delete countsByType[type]; });
+
+        const shouldHideBadge = Object.values(countsByType).every(value => value === 0);
+        mobileBadge.classList.toggle(hideBadgeClass, shouldHideBadge);
```
AprilSylph:
```suggestion
        document.documentElement.classList.toggle(hideBadgeClass, shouldHideBadge);
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    } else {
+      mobileBadge.classList.remove(hideBadgeClass);
```
AprilSylph:
```suggestion
      document.documentElement.classList.remove(hideBadgeClass);
```

marcustyphoon:
Oh yeah. I had it that way in an earlier iteration.

marcustyphoon:
finally bothered to test this.

this does result in an animate-out effect if you load the page in full width and then narrow it to tablet width. that I can think of (not exactly 100% atm), the cleanest way to fix this for real would be to add a second class to the page/button or a second style element in the pageModifications callback that turns off the transition, then remove it two rAFs later (kind of like `transitionStyleElement` in some of my other PRs, but in reverse: noTransitionStyleElement?). this would, I assume, make the change that puts the class on document.documentElement unnecessary. is there an advantage either way? I'm not sure.

of course, we could also just leave it, it's fairly minor.

AprilSylph:
Let's merge this as-is; we can decide if and how to tackle the viewport change issue afterwards.

marcustyphoon:
Oh yeah, it doesn't seem to be working the way I expect it to.

PR #1946, opened 2025-10-28
refactor: Store memoized data on dom objects

marcustyphoon:
### Description

You know what I just remembered? Using WeakMap for memoization is entirely unnecessary; the fact that we're in a content script sandbox effectively is in itself a WeakMap.

I would choose to do or not do this **entirely** based on what one thinks makes the code more clear, which I could see arguments for both sides on (if anything I kind of lean toward `weakMemoize`, but I'd love to know your thoughts). I'm intentionally not going to discuss the performance implications, which are well below the threshold which I think it's appropriate to optimize.

### Testing steps

Smoke test.

### DIFF
--- a/dev/check-outdated-css-keys.js
+++ b/dev/check-outdated-css-keys.js
@@ -4,7 +4,7 @@ import fs from 'node:fs/promises';
 
 const getCssMapKeys = async () => {
   const cssMapUrl = /(?<="cssMapUrl":")[^"]+.json(?=")/.exec(
-    await fetch('https://www.tumblr.com/').then(response => response.text())
+    await fetch('https://www.tumblr.com/').then(response => response.text()),
   )[0];
   const cssMap = await fetch(cssMapUrl).then(response => response.json());
   return new Set(Object.keys(cssMap));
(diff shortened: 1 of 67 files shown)


marcustyphoon:
Presumably also works:

```js
export const weakMemoize /* objectMemoize? */ = (func) => {
  const key = getRandomHexString();
  return (arg) => {
    arg[key] ??= func(arg);
    return arg[key];
  };
};
```

One could even do a universal memoize, but as that would include `typeof arg` and a bunch of bullshit and memoization is generally for performance, that would be a bit silly.

AprilSylph (commented):
I think this is much more readable, but I do think the JSdoc should be updated to make the "this runs in a content script so we're not doing pollution" thing clearer.

Also if we're touching any JSdoc with hyphen separators let's please remove them thank you.

marcustyphoon:
Hm.

(In all seriousness—okay, no, not all seriousness, but still—an [abandoned bug](https://github.com/microsoft/TypeScript/issues/20801) is still a bug, right? :3)

marcustyphoon:
I wonder if this saves a property lookup.

This is a joke. If we start trying to save one property lookup so help me god.

```js
export const timelineObject = postElement => {
  return postElement.timelineObjectPromise ?? (postElement.timelineObjectPromise = inject('/main_world/unbury_timeline_object.js', [], postElement));
};
```

AprilSylph:
> Hm.

That doesn't look like the JSdoc is parsing... at all? Needs `@type` or some other prefix.

PR #1947, opened 2025-10-28
chore: Remove jsdoc hyphens

marcustyphoon:
### Description

ok here u go

(see https://github.com/AprilSylph/XKit-Rewritten/pull/1903)

### Testing steps

- Run `npm run posttest`.
- Regex search `^ \* .* - ` and note the remaining entries.

### DIFF
--- a/src/utils/tumblr_helpers.js
+++ b/src/utils/tumblr_helpers.js
@@ -1,7 +1,7 @@
 import { inject } from './inject.js';
 
 /**
- * @param {...any} args - Arguments to pass to window.tumblr.apiFetch()
+ * @param {...any} args Arguments to pass to window.tumblr.apiFetch()
  * @see {@link https://github.com/tumblr/docs/blob/master/web-platform.md#apifetch}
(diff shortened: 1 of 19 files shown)


src/utils/tumblr_helpers.js
```diff
@@ -11,8 +11,8 @@ export const apiFetch = async (...args) => inject('/main_world/api_fetch.js', ar
  * Create an NPF edit request body.
  * @see https://github.com/tumblr/docs/blob/master/api.md#postspost-id---editing-a-post-neue-post-format
  * @see https://github.com/tumblr/docs/blob/master/api.md#posts---createreblog-a-post-neue-post-format
- * @param {object} postData - camelCased /posts/{post-id} GET request response JSON
- * @returns {object} editRequestBody - camelCased /posts/{post-id} PUT request body parameters
```
marcustyphoon, line 15: Note that eslint-plugin-jsdoc will not just remove the hyphen here; named return values are as far as I can tell Not Actually A Thing.

AprilSylph (approved):

PR #1953, opened 2025-10-29
refactor: Remove strict null check of react fibers

### LINKED ISSUE #1906: Remove null checks to ensure injected functions cannot infinite loop
### Concept

In the thankfully presumably unlikely case that changes to React internals make the fiber values we loop through in injected functions default to `undefined` instead of `null`, XKit Rewritten could as far as I know freeze Tumblr tabs instantly and permanently unless it was uninstalled in some cases. This would be, in a word, Bad.

We could change the `!== null` checks to simple truthiness checks to prevent this.

(I sometimes think there should be a lint rule straight up banning the character sequence`!== null` in eslint's recommended config, but that's just me.)

marcustyphoon:
### Description

As described in the linked issue, this expands the fiber prop values that will end the while loops in our injected functions that read react internals to include undefined, providing a stronger guarantee that the loops will end properly if our assumptions about react internals no longer hold.

This notably does precisely nothing, as we immediately read `fiber.something` on the next line, which will throw if `fiber` is not truthy. There will thus be no infinite loop in the case where react changes the default value of fiber.return.

In fact, `TypeError: can't access property "memoizedProps", fiber is undefined` is more informative than this promise just not resolving; this brings up the fact that we should probably manually throw something informative after these while loops.

Resolves #1906.

### Testing steps

Smoke test.

### DIFF
--- a/src/main_world/control_tags_input.js
+++ b/src/main_world/control_tags_input.js
@@ -7,7 +7,7 @@ export default function controlTagsInput ({ add, remove }) {
   const reactKey = Object.keys(selectedTagsElement).find(key => key.startsWith('__reactFiber'));
   let fiber = selectedTagsElement[reactKey];
 
-  while (fiber !== null) {
+  while (fiber) {
     let tags = fiber.stateNode?.state?.tags;
     if (Array.isArray(tags)) {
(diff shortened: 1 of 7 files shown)


marcustyphoon:
Cleanup: Closed because, eh. Reopenable.

PR #1956, opened 2025-10-30
chore: Format css files

marcustyphoon:
### Description

This uses prettier to format our CSS files.

I don't love every diff here, but they're mostly reasonable.

I was originally going to use biomejs for this (this was actually the result of trying it out), but it handled a few edge cases kind of poorly.

### Testing steps

I didn't test this.


src/action/backup.css
```diff
@@ -43,7 +43,7 @@
 
   background-color: rgb(var(--passive-grey));
   color: rgb(var(--black));
-  font-family: monospace, monospace;
```
marcustyphoon: I don't think there's a reason for this... right?

marcustyphoon: Aha: https://github.com/necolas/normalize.css/issues/519#issuecomment-197131966

AprilSylph (approved):

PR #1957, opened 2025-10-30
refactor: Misc js lint changes

marcustyphoon:
### Description

These are a few refactor suggestions from the default biomejs linter.

### Testing steps

I did not test this.


AprilSylph:
* [x] **Painter**: Colours are still applied if root post request fails
* [x] **Quick Tags**: Tag bundles can still be edited
* [x] **Quick Tags**: Tag bundles can still be deleted
* [x] **Scroll to Bottom**: Scrolling continues when there are more pages to load
* [x] **TimeFormat**, **Timestamps**: Relative times are correct

PR #1958, opened 2025-10-31
fix(Tweaks): Mobile menu notification badge transition

marcustyphoon:
### Description

As noted in https://github.com/AprilSylph/XKit-Rewritten/pull/1945#issuecomment-3470390403, if a tweak should hide the notification badge on the mobile menu, there is a brief moment where it's visible and then animates out right when it is added to the page, i.e. when the page is resized down through the mobile layout breakpoint.

This does pretty much what that comment suggested to prevent this: right when the badge element is added to the page, the transition CSS is removed. Also, to prevent the injected function from taking longer than usual if this occurs between zero and ten seconds from the page load, it is injected once as a module side effect to warm the injected function cache.

### Testing steps

- Ensure that, without tweaks, your account has a visible mobile menu badge.
- Enable the hide following/community/activity tweaks so that the badge is hidden.
- Open a new Tumblr tab in the desktop viewport width, count to ~5, and resize the window down to the mobile layout. Confirm that no mobile menu badge appears.
- Open a new Tumblr tab in the desktop viewport width, count to ~15, and resize the window down to the mobile layout. Confirm that no mobile menu badge appears.
- Confirm that toggling the tweaks in question off and on makes the mobile menu badge appear/disappear.

I have a test commit to simulate the badge naturally appearing/disappearing, but looking at it I don't think this code would really interact with it at all, so I think it must have been more relevant when I was trying something else that wound up not working or not being necessary.


src/utils/mobile_menu_badge_hide.js
```diff
@@ -5,9 +5,20 @@ import { pageModifications } from './mutations.js';
 
 const mobileBadgeSelector = `header ${keyToCss('hamburger')} + ${keyToCss('notificationBadge')}`;
 const hideBadgeClass = 'xkit-hide-mobile-menu-badge';
+const noTransitionClass = 'xkit-hide-mobile-menu-badge-no-transition';
+
+// load injected utility into module cache
```
AprilSylph: nitpick for re-triggering build

```suggestion
// Load injected utility into module cache
```

src/utils/mobile_menu_badge_hide.js
```diff
@@ -58,5 +69,15 @@ export const mobileMenuBadgeHide = Object.freeze({
   },
 });
 
-pageModifications.register(mobileBadgeSelector, () => mobileMenuBadgeHide.trigger());
+const waitForRender = () =>
+  new Promise(resolve => requestAnimationFrame(() => requestAnimationFrame(resolve)));
```
AprilSylph, line 75: Would be neat to have a comment here explaining why this resolves after two frames instead of one?

marcustyphoon: Honestly maybe we should util-ify that? I suppose that if I recall correctly you generally prefer to do so once a second file wants the import. Can always do it later.

src/utils/mobile_menu_badge_hide.js
marcustyphoon: You can also branch update, I assume? Or does it not let you do that if the branch is in my fork?

src/utils/mobile_menu_badge_hide.js
AprilSylph: I prefer to do so once we _anticipate a need_ for it in a second file. Where else would we use this? Anything you've got an open PR for, or any particularly exciting ideas not yet started on?

marcustyphoon: It's copied in #960, but — well, I see you're going through some of my "implemented this so we could if we wanted to, but do we want to? ¯\\\_(ツ)\_/¯" PRs at the moment, so I assume you'll opine on that one at some point :D

AprilSylph: Hmm... that one is definitely not as compelling now that Tumblr doesn't allow badges to wrap, like it did before.

marcustyphoon: Ah, actually, this is already in #1128. Not sure where it would go, though—I guess interface?

AprilSylph: I guess. I wish interface was less of a misc utils file, but I'm not sure how I would split it out.

marcustyphoon: It does feel like there's a "dom selectors/traversal utils that are about the specifics of the tumblr frontend markup" category of stuff that's actually supposed to be in there, and then like, buildStyle and postType that aren't, yeah. Dunno where else they would go though.

AprilSylph (approved):
Extracting `waitForRender` into a util can be a follow-up PR.

PR #1964, opened 2025-11-01
feat(Show Originals): Allow filtering blog search/tagged/filtered-by-type

marcustyphoon:
### Description

Currently, Show Originals has no effect if you search or filter a specific blog's posts in Peepr. This allows one to enable Show Originals in this case.

Internally, a separate `peeprFiltered` location is added to the code, as I don't think it makes sense to apply the remembered mode for viewing a specific blog when searching. This mode always defaults to off and is never remembered. The location is defined as "any peepr timeline that's not the main peepr posts view" so that exhaustive timeline id utils for the various options aren't necessary.

Somewhat related to, merge conflicts with, and is probably about as niche as #1378. (By which I mean I want both of them, but who knows if anyone else does.)

### Testing steps

Open a blog in peepr and turn Show Originals on on said blog.

Confirm that Show Originals shows a default-off but functional control element on that blog when:
- filtering by post type (nothing in search bar)
- searching for something
- searching for something and [filtering by post type or not, filtering by original posts or not, and viewing top posts first or not]; all 8 combinations should work
- viewing posts with a certain tag
- viewing posts with a certain tag and filtering by post type

Clear the search and confirm that Show Originals goes back to being on, as it remembered your preference for viewing blogs in peepr with no filter.

Add the blog in question to the "always show posts from these blogs" list and confirm that Show Originals appears disabled both with and without a filter.


marcustyphoon:
Related: #1985.

marcustyphoon:
I don't remember this PR well; I'll have to look back into it. Anyway, I assume that because I made it as nondraft it should be fully functional, but as it doesn't have a description written it should be draft.

marcustyphoon:
I'd drafted this because Tumblr added an original post filter to the timelines in question, which is obviously more efficient than applying Show Originals to them. However, that's only parity if the user doesn't have "always show reblogs with contributed content" enabled, so this feature would still be quite useful. However however, it would be pretty confusing looking to have two controls that do mostly or exactly the same thing. However however however, we already do have that on the unfiltered blog view, and I use that all the time so I refuse to remove it.

In theory, we could make the Show Originals control actually navigate to the Tumblr-native original posts view if and only if the feature's settings don't make it differ from the intent. This would, I guess, involve making Show Originals default-off in peepr, which I actually think is fine. But it would be a fair bit of work.

(I suppose I could throw it onto the _"I can justify spending an elevated amount of time on this complex PR without a huge user impact because it's for API efficiency"_ pile, which in reality is also a _"...but I don't consider it necessary for anyone else to spend a similarly elevated amount of time reviewing it unless they independently think it's worth it"_ pile and also an _"actually I lied about the first part: I can justify spending an elevated amount of time on this complex PR because it was fun and I learned something and I think it should be reviewed if and only if the reviewer also finds it fun/will learn something; I actually didn't and don't ever need a good reason, good reasons can suck it"_ pile.)

PR #1971, opened 2025-11-06
feat: Warn user if XKit fails to initialize

### LINKED ISSUE #1913: Consider adding a notification if extension startup fails but the user is logged in
### Concept

Network failures on our initial api requests will cause XKit Rewritten to silently not work until the user opens a new Tumblr tab or hard refreshes the tab (or certain specific actions like changing very specific tumblr settings, but I digress). This is not ideal, as users think the extension may be broken and do not know to try opening a new tab.

This behavior is fully intentional whenever the user is on tumblr.com and not logged in, though, so we must be somewhat careful if we want to implement a UI element indicating this situation in any way.

marcustyphoon:
### Description

Proof of concept; I'm not sure what we would want to actually show here so I copied something from one of my dev scripts. (Clickable button with more info would be cool? Of course, this is not worth much effort, so...)

This makes a UI element appear if we're confident the user is logged in, but the XKit initialization function hasn't succeeded after [arbitrary value, I put 10] seconds. This was motivated by yet another wave of posts I saw about a) XKit mysteriously not working, and b) the site being slow. I'd love to be able to be slightly more confident about whether b is the cause of a.

I dunno. What do you think?

Resolves #1913, I guess.

### Testing steps

Add an error to the init function in main.js, to one of the three critical utility functions that we preload, and/or to the injected functions they call using top level await. Observe that XKit modifications don't load and a warning message eventually appears.

Add a really long delay to the init function in main.js, to one of the three critical utility functions that we preload, and/or to the injected functions they call using top level await. Observe that a warning message appears, but then disappears as XKit modifications eventually load.

Repeat this when logged out and confirm that nothing happens.


AprilSylph:
I like the concept. I wouldn't have this extracted out to a util, though; we're not going to reuse it.

And if you're looking for a better arbitrary number than 10s, I think 60s gives us more confidence that something has gone awfully wrong.

marcustyphoon:
I feel like as a user I would want to be informed before an entire minute has passed :D

PR #1972, opened 2025-11-07
feat(Mutual Checker): Control UI for only-mutuals mode (copied from Show Originals)

marcustyphoon:
### Description

It looks really good until you also turn on Show Originals, at which point it immediately looks very stupid. Any ideas?

### DIFF
--- a/src/features/mutual_checker/feature.json
+++ b/src/features/mutual_checker/feature.json
@@ -10,7 +10,7 @@
   "preferences": {
     "showOnlyMutuals": {
       "type": "checkbox",
-      "label": "Only show posts from mutuals on the dashboard",
+      "label": "Add a mutuals-only tab to your timelines",
       "default": false
     },
     "showOnlyMutualNotifications": {


AprilSylph:
First idea that comes to mind is a new util that centralises custom tab adding.

marcustyphoon:
What would we want the resulting UI to look like?

AprilSylph:
One inserted tab bar. The util provides the "All Posts" tab, features can register additional tabs (inserted to the left). If there are no additional tabs registered, the util removes the tab bar and becomes inactive.

Screen Shot 2025-11-07 at 12 56 52

CSS would need to be adjusted to align text to vertical middle for better mobile compatibility.

We don't have to figure out how to make this scrollable when the maximum number of tabs is three. XKit Rewritten is a packaged add-on, so we don't need to build extensibility until we need to use that extensibility.

marcustyphoon:
Oh—right, that wouldn't also provide "original posts from mutuals." I actually had not thought about that possibility (as I would actually use "original posts from mutuals" sometimes), but objectively the simplicity could reasonably be argued to be worth it.

AprilSylph:
I'm sure we could figure out a more complicated UI that lets you layer the filters instead, but I can't envision it right now.

marcustyphoon:
Closest thing in the actual Tumblr UI is this, which I don't think we really want to do anything related to:

image

Yeah, nothing comes to mind. I think if we can think of any other hypothetical custom dashboard tabs we would theoretically make, it could make sense to put as much thought as possible into the best version of this, as ideally we wouldn't want to change it again later. If it's just this, then... ehh.

Anyway, note to self: #1378 will pretty much inevitably wind up part of this, as it makes little sense to implement the ability to disable tabs.

marcustyphoon:
Cleanup: Closed in favor of issue #1973.

PR #1969, opened 2025-11-06
refactor(Classic Footer): Scope modifications to affected footer

marcustyphoon:
### Description

~~Stacked PR on #1968.~~

This tweaks Classic Footer's modifications to ensure that they can't affect anything outside of the processed `footer` element.

There are, obviously, many ways to factor this; I went through a few totally working, justifiable options. As committed, this adds a data attribute to the footer element, which certainly isn't necessary but I think makes the CSS the most readable and uses the least `:has()` selectors. I avoided extracting more selector variables to avoid having to distinguish between which are only referring to active footers and which aren't, but again, could certainly do that also.

### Testing steps

Expect no functional changes. ~~_todo: write this section._~~

Original Classic Footer PR:

> 1. Load the modified addon
> 2. Enable the Classic Footer feature
>     - [ ] **Expected result**: Control buttons in post footers are right-aligned and reordered
>     - [ ] **Expected result**: Total note counts are added to post footers, on the left
> 3. Make sure nothing in the post footer is broken in this state
>     - [ ] **Expected result**: Clicking the note count opens/closes the post's notes (defaulting to replies)
>     - [ ] **Expected result**: Clicking the Share button opens/closes the share sheet
>     - [ ] **Expected result**: Clicking the Reply button opens/closes the post's notes (defaulting to replies)
>     - [ ] **Expected result**: Clicking the Reblog button opens/closes the post's reblog menu
>     - [ ] **Expected result**: Clicking the Like button likes/unlikes the post
> 4. Find a post of your own that can be Blazed
>     - [ ] **Expected result**: The Blaze button is positioned correctly
>     - [ ] **Expected result**: Clicking the Blaze button opens the Blaze modal

Tested on regular dashboard, on peepr, on peepr masonry, and on radar.

Tests from #1931 performed on regular dashboard or "channel."

PR #1975, opened 2025-11-09
chore(Quick Tags): Improve footer targeting

marcustyphoon:
### Description

This tweaks Quick Tags' fake tags insertion to ensure that it can't target the wrong footer element.

Related: #1969

Obsoleted by merging https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon/quick-tags-really-refresh, which I didn't make a PR for, but could; would we rather do this?

### Testing steps

Confirm that Quick Tags functions normally on posts with and without tags.

### DIFF
--- a/src/features/quick_tags/index.js
+++ b/src/features/quick_tags/index.js
@@ -175,7 +175,7 @@ const addFakeTagsToFooter = (postElement, tags) => {
   );
   const tagsElement = dom('div', { class: tagsClass }, null, [dom('div', null, null, fakeTags)]);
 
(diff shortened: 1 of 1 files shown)


AprilSylph:
> Obsoleted by merging [master...marcustyphoon/quick-tags-really-refresh](https://github.com/AprilSylph/XKit-Rewritten/compare/master...marcustyphoon/quick-tags-really-refresh), which I didn't make a PR for, but could; would we rather do this?

Yes please. I would love that a lot.

marcustyphoon:
Closed in favor of #1980.

PR #1976, opened 2025-11-09
fix(AccessKit Disable GIFs): Label not being applied in very specific scenarios

marcustyphoon:
### Description

In some scenarios (recommendation sidebar item on https://www.tumblr.com/excitementshewrote/757829278953062400/best-marge-simpson-quote-i-thought-googling, all non-full-width gifs if Staff deploys some currently unused code in the front end), paused GIFs won't have a label, because you can't have two `:after` pseudo elements on the same DOM element, and when I changed the label to be a pseudo, I made the fallback behavior not showing a label.

This is one way to make the fallback actually work: still use a pseudo.

This really feels very inelegant and we might just want to switch back to using a dom element label instead... although, last I recall, the code that did that was also very inelegant, so maybe not? Help.

### Testing steps

Confirm that AccessKit Disable GIFs shows labels in both regular cases and the cases mentioned above, and that they hide on image hover correctly.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -9,6 +9,7 @@ const canvasClass = 'xkit-paused-gif-placeholder';
 const pausedPosterAttribute = 'data-paused-gif-use-poster';
 const hoverContainerAttribute = 'data-paused-gif-hover-container';
 const labelAttribute = 'data-paused-gif-label';
+const labelSizeAttribute = 'data-paused-gif-label-size';
 const containerClass = 'xkit-paused-gif-container';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Another thing that might be somewhat more elegant is using two data attributes: one whose value is whether the label is before or after, another whose value is how small the label should be, if it needs to be smaller than normal. The loc is presumably basically the same, but the selectors might look less stupid.

Edit: Did this. It's... still inelegant.

marcustyphoon:
God I hate this diff. But the feature is in a/b testing, so... well. If it works, I guess.

PR #1983, opened 2025-11-11
feat(Tweaks): "Make the “Create” button start a new text post"

### LINKED ISSUE #1769: create button opens text post
### Concept

would like a tweak to make the create button just open a text post modal instead of the post type picker. this is closer to mobile parity, and text posts support adding any post type anyway so its just quicker and more convenient for me

thank you!

AprilSylph:
### Description

- resolves #1769

Adds a new option to Tweaks that modifies the "Create" button on the desktop layout to navigate straight to `/new/text` instead of the `/new` interstitial with the post type bubbles.

To help achieve this elegantly, the reusable `onClickNavigate` handler has been modified to stop event propagation when a modifier key is detected, or we have entered the codepath that calls `navigate()`. This hides the element from React's event processing, making the element behave more like one that we've created ourselves—forcing it to follow what the DOM `href` is, and not the React virtual DOM link target.

We only use `onClickNavigate` in one other place, Tag Tracking+, where this change has no effect at all. If we do ever need to build something where we do need click events on internal links to bubble... we can worry about that when we get there! (Probably just two variations of the same reusable handler; the only unknown there would be the naming.)

#### Setup
1. Load the modified addon
2. Enable Tweaks
3. Enable the new option

#### Feature testing – Desktop layout
1. Hover the Create button
    - **Expected result**: The link target is `https://www.tumblr.com/new/text`
2. Click the Create button
    - **Expected result**: The post form is opened immediately, with no interstitial
    - **Expected result**: The post form is empty, with no media upload placeholders
3. <kbd>Control</kbd>+click the Create button (<kbd>Command</kbd>+click on macOS)
    - **Expected result**: `https://www.tumblr.com/new/text` is opened in a new tab
4. <kbd>Shift</kbd>+click the Create button
    - **Expected result**: `https://www.tumblr.com/new/text` is opened in a new window

#### Regression testing – Mobile layout

1. Open the navigation drawer
2. Hover the Create button
    - **Expected result**: The link target is `https://www.tumblr.com/new`
3. Click the Create button
    - **Expected result**: The navigation drawer is closed
    - **Expected result**: The post form is opened immediately, with no interstitial
    - **Expected result**: The post form is empty, with no media upload placeholders

### DIFF
--- a/src/features/tweaks/create_button_no_bubbles.js
+++ b/src/features/tweaks/create_button_no_bubbles.js
@@ -13,7 +13,7 @@ const processCreateButtons = (createButtons) => {
     createButton.dataset.href ??= createButton.getAttribute('href');
 
     createButton.setAttribute(modifiedAttribute, '');
-    createButton.setAttribute('href', createButton.dataset.href.replace('/new', '/new/text'));
+    createButton.setAttribute('href', createButton.dataset.href.replace(/\/new$/, '/new/text'));
     createButton.addEventListener('click', onClickNavigate);
   });
 };


marcustyphoon:
Do we want to make the button continue to target the current community when on a communities page?

AprilSylph:
> Do we want to make the button continue to target the current community when on a communities page?

I had absolutely no idea that that was a thing. That's fun.

marcustyphoon:
We could make the selector `${keyToCss('createPostButton')}[href$="/new"]`, which sort of seems like an elegant anti-weird-behavior guard, but I don't imagine it's likely to make any difference.

AprilSylph:
> We could make the selector `${keyToCss('createPostButton')}[href$="/new"]`, which sort of seems like an elegant anti-weird-behavior guard, but I don't imagine it's likely to make any difference.

We actually _want_ to re-process buttons (which this would prevent) when the addon restarts, since `onClickNavigate` expires if the previous addon sandbox is thrown away. But we can do `.href.replace(/\/new$/, '/new/text')` instead, which I think is a good idea.

marcustyphoon:
Oh, yeah, I guess that would prevent reprocessing! Can't have that.

PR #1984, opened 2025-11-11
feat(Tweaks): "Hide all posts from blogs I've blocked"

### LINKED ISSUE #1776: [Tweaks] New option: "Hide all posts from blogs I've blocked"
### Discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/1772

> <sup>Originally posted by **githubguyyyy** April 10, 2025</sup>
> when you block someone on tumblr, you won't see them on the dash or in search/tag results, but will still if you go to another person's blog and they reblogged the blocked user. could you add a feature that gets rid of that?

AprilSylph:
### Description

- resolves #1776 

Adds an option to Tweaks that hides all posts containing blogs that the user has blocked from their primary blog.

When hiding a post, a check is performed to make sure that we're not just looking at a blocked blog's timeline. If we are, the post is not hidden unless a different blog within the post's reblog trail has also been blocked. This prevents this tweak from spamming Tumblr's API just by visiting a blocked blog's Peepr page.

### Testing steps

1. Load the modified addon
2. Block the `@fandom` blog via your primary blog's settings
3. Go to https://www.tumblr.com/kpop and scroll down to its sixth post
    - **Expected result**: By default, you can see the post reblogged from `@fandom`, despite having blocked that blog
4. Enable Tweaks &rarr; "Hide all posts from blogs I've blocked"
    - **Expected result**: The reblogged post on `@kpop` disappears!
5. Go to https://www.tumblr.com/fandom
    - **Expected result**: You are shown the "You have this Tumblr blocked" interstitial
6. Click "View this Tumblr"
    - **Expected result**: You can see the posts on the blog
    - **Expected result**: You can see that the latest post published on it is a reblog from `@tumblr`
7. Block the `@tumblr` blog via your primary blog's settings
8. Reload https://www.tumblr.com/fandom and dismiss the interstitial again
   - **Expected result**: The post that `@fandom` reblogged from `@tumblr` is gone, but you can still see the rest of the posts on `@fandom`

### DIFF
--- a/src/features/show_originals/index.js
+++ b/src/features/show_originals/index.js
@@ -7,8 +7,8 @@ import { translate } from '../../utils/language_data.js';
 import { userBlogs } from '../../utils/user.js';
 import {
   followingTimelineFilter,
-  anyBlogTimelineFilter,
-  blogTimelineFilter,
+  anyBlogPostsTimelineFilter,
+  blogPostsTimelineFilter,
   blogSubsTimelineFilter,
   timelineSelector,
   anyCommunityTimelineFilter,
@@ -74,11 +74,11 @@ const addControls = async (timelineElement, location) => {
 
 const getLocation = timelineElement => {
   const isBlog =
-    anyBlogTimelineFilter(timelineElement) && !timelineElement.matches(channelSelector);
+    anyBlogPostsTimelineFilter(timelineElement) && !timelineElement.matches(channelSelector);
 
   const on = {
     dashboard: followingTimelineFilter(timelineElement),
-    disabled: isBlog && disabledBlogs.some(name => blogTimelineFilter(name)(timelineElement)),
+    disabled: isBlog && disabledBlogs.some(name => blogPostsTimelineFilter(name)(timelineElement)),
     peepr: isBlog,
     blogSubscriptions: blogSubsTimelineFilter(timelineElement),
     community: anyCommunityTimelineFilter(timelineElement) || communitiesTimelineFilter(timelineElement)
--- a/src/utils/timeline_id.js
+++ b/src/utils/timeline_id.js
@@ -2,12 +2,14 @@ const createSelector = (...components) => `:is(${components.filter(Boolean).join
 
 export const timelineSelector = ':is([data-timeline], [data-timeline-id])';
 
+const startsWith = string => `^${string}`;
 const exactly = string => `^${string}$`;
-const anyBlog = '[a-z0-9-]{1,32}';
+
+const anyBlogName = '[a-z0-9-]{1,32}';
 const uuidV4 = '[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}';
 
-const peeprPostsTimelineId = ({ blog, postId, maybeOriginal, maybeTop, search, postType, tag }) =>
-  `peepr-posts-${blog}-${postId}-${maybeOriginal}-${maybeTop}-${search}-${postType}-${tag}`;
+const peeprPostsTimelineId = ({ blogName, postId, postRole, searchMode, searchTerm, postType, tag }) =>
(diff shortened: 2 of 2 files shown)


src/features/tweaks/hide_blocked_blogs.js
```diff
@@ -0,0 +1,48 @@
...
+      const timelineElement = postElement.closest(timelineSelector);
+      if (isTimelineExempt(timelineElement)) {
+        // This blog is blocked, but we're on that blog's timeline.
+        // Consider Tumblr's own warning interstitial to be sufficient.
+        continue;
+      }
+
+      getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
+      break;
+    }
```
marcustyphoon, line 37: On the one hand, this definitely could be an `if (blockedBlogNames.some()) {`, which is clearer logic, but there's no elegant place to put the callback and to make that comment structure (which I like) work, you would have to do `if () { return false; } return true;` which I never really like. So this might be more elegant, and it's probably understandable enough.

marcustyphoon: Ehhhhh.

```js
    const shouldFilter = blogName => {
      const isTimelineExempt = blogTimelineFilter(blogName);
      const timelineElement = postElement.closest(timelineSelector);
      if (isTimelineExempt(timelineElement)) {
        // This blog is blocked, but we're on that blog's timeline.
        // Consider Tumblr's own warning interstitial to be sufficient.
        return false;
      }
      return true;
    };

    if (blockedBlogNames.some(shouldFilter)) {
      getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
    }
```

AprilSylph: There's also the little detail that this is a `Set` for de-duplication purposes (don't want to run more RegExp matches than necessary!), so to use `Array.prototype.some` we would need to convert it back to an array.

We could of course just use `.filter((value, index, array) => array.indexOf(value) === index)` for de-duplication, but if we're talking about elegance, that doesn't hold a candle to `new Set()`.

marcustyphoon: Hey, want to see a piece of code that relies on the little-known fact that `[].every(() => false) === true`?

```js
    const shouldIgnore = blogName => {
      const isTimelineExempt = blogTimelineFilter(blogName);
      const timelineElement = postElement.closest(timelineSelector);
      if (isTimelineExempt(timelineElement)) {
        // This blog is blocked, but we're on that blog's timeline.
        // Consider Tumblr's own warning interstitial to be sufficient.
        return true;
      }
    };
    if (blockedBlogNames.every(shouldIgnore)) return;

    getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
 ```

marcustyphoon: Oh, yeah, that too.

AprilSylph: > `[].every(() => false) === true`

Ewww, that's disgusting!

marcustyphoon: You would, practically speaking, add another length check before getting to that part, but yeah, fun interaction :D

marcustyphoon:
I guess if you really want to shave cycles, you only need to do the exempt check if `blockedBlogNames.size === 1`, and then you can avoid looping at all, but.

edit: Today on the "things which are still stupid because Set utils are mediocre" show:

```js
    if (blockedBlogNames.size === 0) return;
    if (blockedBlogNames.size === 1) {
      const isTimelineExempt = blogTimelineFilter(blockedBlogNames.values().next().value);
      const timelineElement = postElement.closest(timelineSelector);
      if (isTimelineExempt(timelineElement)) {
        // This blog is blocked, but we're on that blog's timeline.
        // Consider Tumblr's own warning interstitial to be sufficient.
        return;
      }
    }
    getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
  ```

PR #1985, opened 2025-11-11
fix(Tweaks): match subpages when testing for Peepr

AprilSylph:
### Description

- blocks #1984

I have some nitpick-type naming problems with our blog timeline matching utils:
1. `anyBlogTimelineFilter` doesn't actually match any blog timeline, only any blog's main "Posts" timeline
2. `blogTimelineFilter`, similarly, only matches the given blog's main "Posts" timeline

Of course, this also means we don't have any filters that match any blog's or a given blog's myriad of different timelines, which is needed for the linked PR:

> When hiding a post, a check is performed to make sure that we're not just looking at a blocked blog's timeline. If we are, the post is not hidden unless a different blog within the post's reblog trail has also been blocked. This prevents this tweak from spamming Tumblr's API just by visiting a blocked blog's Peepr page.
>
> As of writing, this check is imperfect, and posts on a blocked blog's Peepr tag/search pages are still hidden.

This PR:
* renames our existing `anyBlogTimelineFilter` matcher to `anyBlogPostsTimelineFilter`
* renames our existing `blogTimelineFilter` matcher to `blogPostsTimelineFilter`
* introduces a new `anyBlogTimelineFilter` matcher, which also matches blogs' secondary timelines
  (e.g. queue, drafts, in-blog search, blogs' tagged pages)
* introduces a new `blogTimelineFilter` matcher, which also matches the given blog's secondary timelines (ibid)

If you have anything in progress that might get tripped up by this new-thing-with-old-name thing, we could disambiguate by calling the new matchers `anyBlogAnyTimelineFilter` and `blogAnyTimelineFilter`... what do you think?

Completely different solutions are, of course, welcome. I didn't write the timeline IDs util, so maybe I'm going about this the wrong way.

### Testing steps

1. Load the modified addon
2. Enable Tweaks &rarr; "Hide all posts from blogs I've blocked"
3. Block the `@fandom` blog from your primary blog's settings
4. Go to https://www.tumblr.com/fandom
    - **Expected result**: You are shown the "You have this Tumblr blocked" interstitial
5. Click "View this Tumblr"
    - **Expected result**: The timeline isn't blanked by Tweaks; you can see the posts
6. Go to https://www.tumblr.com/fandom/tagged/tumblr
    - **Expected result**: You are shown the "You have this Tumblr blocked" interstitial
7. Click "View this Tumblr"
    - **Expected result**: The timeline isn't blanked by Tweaks; you can see the one post on it

### DIFF
--- a/src/utils/timeline_id.js
+++ b/src/utils/timeline_id.js
@@ -27,17 +27,6 @@ export const forYouTimelineFilter = ({ dataset: { timeline, timelineId } }) =>
   timelineId === '/dashboard/stuff_for_you' ||
   timelineId?.startsWith('for-you-');
 
-// Matches any blog's timelines, including subpages such as drafts or in-blog searches.
-export const anyBlogTimelineFilter = ({ dataset: { timeline, timelineId } }) =>
-  timeline?.match(startsWith(`/v2/blog/${anyBlogName}/posts`)) ||
-  timeline?.match(startsWith(`/v2/user/inbox/${anyBlogName}`)) ||
-  timelineId?.match(startsWith(`peepr-posts-${anyBlogName}-`)) ||
-  timelineId?.match(startsWith(`blog-view-${anyBlogName}`)) ||
-  timelineId?.match(startsWith(`drafts-${anyBlogName}`)) ||
-  timelineId?.match(startsWith(`blog-${uuidV4}-${anyBlogName}`)) ||
-  timelineId?.match(startsWith(`drafts-${uuidV4}-${anyBlogName}`)) ||
-  timelineId?.match(startsWith(`queue-${uuidV4}-${anyBlogName}`));
-
 // Matches the given blog's timelines, including subpages such as drafts or in-blog searches.
 export const blogTimelineFilter = blogName =>
   ({ dataset: { timeline, timelineId } }) =>
@@ -50,12 +39,8 @@ export const blogTimelineFilter = blogName =>
     timelineId?.match(startsWith(`drafts-${uuidV4}-${blogName}`)) ||
     timelineId?.match(startsWith(`queue-${uuidV4}-${blogName}`));
 
-// Matches any blog's main posts timeline, not including subpages such as drafts or in-blog searches.
-export const anyBlogPostsTimelineFilter = ({ dataset: { timeline, timelineId } }) =>
-  timeline?.match(exactly(`/v2/blog/${anyBlogName}/posts`)) ||
-  timelineId?.match(exactly(peeprPostsTimelineId({ blogName: anyBlogName }))) ||
-  timelineId?.match(exactly(`blog-view-${anyBlogName}`)) ||
-  timelineId?.match(exactly(`blog-${uuidV4}-${anyBlogName}`));
+// Matches any blog's timelines, including subpages such as drafts or in-blog searches.
+export const anyBlogTimelineFilter = blogTimelineFilter(anyBlogName);
 
 // Matches the given blog's posts timeline, not including subpages such as drafts or in-blog searches.
 export const blogPostsTimelineFilter = blogName =>
@@ -64,6 +49,9 @@ export const blogPostsTimelineFilter = blogName =>
     timelineId?.match(exactly(peeprPostsTimelineId({ blogName }))) ||
     timelineId?.match(exactly(`blog-view-${blogName}`)) ||
     timelineId?.match(exactly(`blog-${uuidV4}-${blogName}`));
+
+// Matches any blog's main posts timeline, not including subpages such as drafts or in-blog searches.
+export const anyBlogPostsTimelineFilter = blogPostsTimelineFilter(anyBlogName);
 
 export const blogSubsTimelineFilter = ({ dataset: { timeline, which, timelineId } }) =>
   timeline === '/v2/timeline?which=blog_subscriptions' ||


marcustyphoon:
> Completely different solutions are, of course, welcome. I didn't write the timeline IDs util, so maybe I'm going about this the wrong way.

In the mentioning PR, I tried to do a related thing except that I did want to distinguish between "is search/tagged" and "is not" and... yeah, it's pretty yikes. If we're sticking with this general framework, the code here seems at least reasonable.

I've been thinking maybe we should just give up on parsing the ids and use an injected function; as you're no longer staff I can no longer hold out hope for an unambiguous separator/format, and also, like, why _would_ there be an unambiguous separator/format; these ids are not intended for the purpose we're using them for. Of course, that would mean making every timeline-querying function async and awaiting them, which... fun. And we can't cache that value because the same element gets reused on navigation in peepr, whee.

marcustyphoon:
Relatedly: no, you indeed did not write the timeline IDs util, so if you have any ideas about how to make it not, uh, kind of complete garbage besides the inject thing, I'm all ears.

src/utils/timeline_id.js
```diff
@@ -25,34 +27,49 @@ export const forYouTimelineFilter = ({ dataset: { timeline, timelineId } }) =>
   timelineId === '/dashboard/stuff_for_you' ||
   timelineId?.startsWith('for-you-');
 
-// includes "channel" user blog view page
+// Matches any blog's timelines, including subpages such as drafts or in-blog searches.
```
marcustyphoon: This doesn't match the drafts page, though, correct?

marcustyphoon:
Was this a waste of half an hour? Eh, probably. But:

```js
  import(browser.runtime.getURL('/utils/timeline_id.js')).then(timelineIdModule => {
    const genericFilterEntries = Object.entries(timelineIdModule).filter(
      ([keyToCss, value]) => {
        if (typeof value === 'function') {
          try {
            const result = value({ dataset: {} });
            if (typeof result !== 'function') {
              return true;
            }
          } catch {}
        }
      }
    );

    console.log('genericFilterEntries', genericFilterEntries);

    document.documentElement.append(
      buildStyle(`
        [data-matches-timeline-filters] {
          border-top: 2px dashed red;
        }
        [data-matches-timeline-filters]::before {
          display: block;
          content: 'matches timeline filters: ' attr(data-matches-timeline-filters);
          color: red;
          margin: 1ch;
        }
      `)
    );

    onNewPosts.addListener(postElements =>
      [...new Set(postElements.map(postElement => postElement.closest(timelineIdModule.timelineSelector)))]
        .forEach(timelineElement => {
          delete timelineElement.dataset.matchesTimelineFilters;
          genericFilterEntries.forEach(([key, value]) => {
            if (value(timelineElement)) {
              timelineElement.dataset.matchesTimelineFilters ??= '';

              if (timelineElement.dataset.matchesTimelineFilters) {
                timelineElement.dataset.matchesTimelineFilters += ', ';
              }

              timelineElement.dataset.matchesTimelineFilters += key;
            }
          });
        })
    );
  });
};

```

src/utils/timeline_id.js
AprilSylph: https://github.com/AprilSylph/XKit-Rewritten/blob/ab95919ce97d68ac129beb5441cbda3460b19ccc/src/utils/timeline_id.js#L30-L32
https://github.com/AprilSylph/XKit-Rewritten/blob/ab95919ce97d68ac129beb5441cbda3460b19ccc/src/utils/timeline_id.js#L65-L66

It would if the Drafts page was still using Timeline v1, but... I'm pretty sure I was the one to convert that one to v2.

marcustyphoon: Right. So, it... doesn't, in practice, then? Should probably use a different example for the comment.

AprilSylph: I don't think there are ... nevermind, the queue was never converted, was it?

AprilSylph: You know what, why not just make the code match the comment? I'm here because I need a "does this timeline belong, in some manner, to this blog" matcher—just because right now I only need to match blockable blogs doesn't mean the util should do that.

9am is one hell of a drug.

marcustyphoon: Correct.

Maybe it might be more consistent to make anyBlogTimelineFilter have the same v1 filters as blogTimelineFilter, explicitly excluding queue and drafts? It won't match them in patio, anyway.

If I'm not mistaken this shouldn't affect the upstream PR either way as it excludes your own posts and—uh... that should still be true on group blogs, right?

Edit: For clarity, this was posted simultaneously with the above comment. Yeah, that, alternatively.

marcustyphoon:
Not sure if it would make sense to make draftsTimelineFilter, queueTimelineFilter, inboxTimelineFilter, etc and call them to compose. Certainly seems less worthwhile if we think we'll eventually move away from parsing these ids, but.

AprilSylph:
> Not sure if it would make sense to make draftsTimelineFilter, queueTimelineFilter, inboxTimelineFilter, etc and call them to compose. Certainly seems less worthwhile if we think we'll eventually move away from parsing these ids, but.

I thought about that, but it seemed annoying.

AprilSylph:
I believe the new filters now truly do match any timeline that belongs to any blog / a given blog. If there are any other blog-specific Patio columns, I'm not privy to them.

marcustyphoon:
Ignore everything I said about replacing this with injecting something; the timeline component appears to have no additional information about what it is. Neat.

marcustyphoon:
```js
export const anyBlogTimelineFilter = blogTimelineFilter(anyBlogName);

export const anyBlogPostsTimelineFilter = blogPostsTimelineFilter(anyBlogName);
```

This should work... right?

(Cannot do this as a suggestion; it would trigger use-before-define)

marcustyphoon:
(...Actually. `anyBlogName` could be the default argument, and we could call `blogTimelineFilter()(timelineElement)` in features—eh, that doesn't really parse, never mind.)

AprilSylph:
> This should work... right?

Doesn't seem to break Show Originals, so yes!

marcustyphoon (approved):

PR #1989, opened 2025-11-13
feat(Classic Footer): add option to use modern note count style

### LINKED ISSUE #1922: Classic Footer: Add option to use modern styles for the notes button
### Concept

- relates to https://github.com/AprilSylph/XKit-Rewritten/issues/1901#issuecomment-3392296634

---

Off-by-default option: "Use modern styles for the notes button"

This would essentially just copy the button styles from the Comments button on community posts and slap them on the notes button inserted by Classic Footer. The only difference would be that the inline padding should be symmetrical at... 12px, I think?

For transparency's sake, this would mostly be for myself. #justice for regularPostsPostChromeFooter

@Pixielart Let me know if you want to be looped into the implementation of this; it would be awesome if our styles didn't conflict here! (Although, I'm not sure which one should win if one has both Old Tumblr Dashboard _and_ this option enabled... 🤔)

AprilSylph:
### Description

- resolves #1922

Adds an off-by-default option to Classic Footer: "Use a modern look for the note count".

When enabled, the note count buttons created by Classic Footer gain a new classname which styles them basically identically to the "comments" buttons on posts in communities. `#justice for regularPostsPostChromeFooter`

This PR also adds more proper focus styling for the non-modern note count button. Instead of a focus ring highlighting the button's otherwise-invisible bounding box, the text gains an accent-coloured underline—just like the focusable text parts in the post header.

Option disabled | Option enabled
-|-
default | modern

### Testing steps

1. Load the modified addon
2. Enable Classic Footer
3. <kbd>Tab</kbd> through the focusable elements in the footer of an affected post
    - **Expected result**: The note count button's text gains a two-pixel-thick accent-coloured underline
    - **Expected result**: The note count button does not gain any focus ring
4. Enable Classic Footer &rarr; "Use a modern look for the note count"
    - **Expected result**: Note count buttons gain an outline
    - **Expected result**: Note counts gains a stronger colour
    - **Expected result**: Note count labels ("note"/"notes") become a lighter font weight
5. Hover over a note count button
    - **Expected result**: The colour of the outline changes _extremely_ subtly
6. <kbd>Tab</kbd> to focus it again
    - **Expected result**: The note count button gains a two-pixel-thick accent-coloured focus ring
    - **Expected result**: The note count button's text does not gain any underline

### DIFF
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -142,6 +142,9 @@ export const styleElement = buildStyle(`
     .${noteCountClass}, .${reblogLinkClass} {
       padding: 6px;
     }
+    .${noteCountClass}.${modernStyleClass} {
+      margin-left: 6px;
+    }
   }
 
   span:has(svg[style="--icon-color-primary: var(--brand-green);"]) > .${reblogLinkClass} {


src/features/classic_footer/index.js
```diff
@@ -7,6 +7,7 @@ import { timelineObject } from '../../utils/react_props.js';
 
 const activeAttribute = 'data-classic-footer';
 const noteCountClass = 'xkit-classic-footer-note-count';
+const modernStyleClass = 'xkit-classic-footer-ds-look';
```
marcustyphoon, line 10: Does this stand for something? I can't tell if it's a joke I'm missing :D

src/features/classic_footer/index.js
```diff
@@ -138,7 +167,7 @@ const processPosts = (postElements) => postElements.forEach(async postElement =>
   const { noteCount } = await timelineObject(postElement);
   const noteCountButton = button({
     'aria-hidden': noteCount === 0 && noZeroNotes,
-    class: noteCountClass,
+    class: `${noteCountClass} ${modernButtonStyle ? modernStyleClass : ''}`,
```
marcustyphoon, line 182: I shouldn't, but I'm so tempted to special case this in the element util to allow both strings and arrays :P

marcustyphoon:
Oh, fun. Favorit modern 350 strikes again.

Edit: How stupid would it be if we did this as a link element in the head with a data uri source so that Palettes picked it up? Pretty stupid, right?

marcustyphoon:
Okay, far be it from me to... look: I realize a nonzero portion of this is fixating on the old design. But font weight aside, I do think there's something to "0.875rem is too small; the comment button is intentionally downsized from a natural font size because the word is long."

Edit: Yeah, I mean, that's the font size for timestamps. That can't also be the font size for primary footer row button text, right?

src/features/classic_footer/index.js
AprilSylph: Design System!

AprilSylph:
> Oh, fun. Favorit modern 350 strikes again. How stupid would it be if we did this as a link element in the head with a data uri source so that Palettes picked it up? Pretty stupid, right?

Hmmm... we could detect if Palettes is overriding the font family, and bump the weight up to "normal" if it is?

marcustyphoon:
Hm... could, yeah. No race conditions with

```css
  :root:has(> #palettes-for-tumblr-override) .${noteCountClass}.${modernStyleClass} {
    font-weight: normal;
  }
```

AprilSylph:
> "0.875rem is too small; the comment button is intentionally downsized from a natural font size because the word is long."

Nope, that really is what the original designs ordained for both! It does look a little small at first, but it can be gotten used to. Or, if not, one can just turn the option off again :P

src/features/classic_footer/index.js
```diff
@@ -76,10 +78,47 @@ export const styleElement = buildStyle(`
...
 
+  :root[style*="--font-family-modern"] .${noteCountClass} {
+    font-weight: bold;
+  }
+  :root[style*="--font-family-modern"] .${noteCountClass}.${modernStyleClass} {
+    font-weight: normal;
+  }
+  :root[style*="--font-family-modern"] .${noteCountClass}.${modernStyleClass} > span {
+    font-weight: bold;
+  }
```
marcustyphoon: Probably should comment these; it's not obvious what they're for from the text!

src/features/classic_footer/index.js
```diff
@@ -76,10 +78,47 @@ export const styleElement = buildStyle(`
...
+    outline-offset: -2px;
+    text-decoration: none;
+  }
   .${noteCountClass}[aria-hidden="true"] {
     visibility: hidden;
   }
 
+  :root[style*="--font-family-modern"] .${noteCountClass} {
+    font-weight: bold;
+  }
```
marcustyphoon: This seems out of line with the line weights of the icons, no?

src/features/classic_footer/index.js
```diff
@@ -76,10 +78,47 @@ export const styleElement = buildStyle(`
...
+  .${noteCountClass}:focus-visible {
+    outline: none;
+    text-decoration: 2px solid var(--accent) underline;
+  }
+  .${noteCountClass}.${modernStyleClass} {
+    padding-block: 10px;
+    padding-inline: 14px;
+    outline: 1px solid var(--content-tint-strong);
+    outline-offset: -1px;
+    margin-left: 8px;
```
marcustyphoon, line 90: The non-modern style drops the equivalent of this to 6px in peepr masonry; do we want to do this as well?

src/features/classic_footer/index.js
marcustyphoon: Aha! Yeah, I guess.

src/features/classic_footer/index.js
AprilSylph: Now that you point it out... yes.

The font weight the pre-React dashboard uses for this is 700, so I was matching that—but it also uses a font size of 13px, which we're definitely not doing here, so.

src/features/classic_footer/index.js
AprilSylph: Yes! Good catch.

AprilSylph:
The final interpolated CSS for this option, for @Pixielart's reference:

```css
.xkit-classic-footer-note-count.xkit-classic-footer-ds-look {
  padding-block: 10px;
  padding-inline: 14px;
  outline: 1px solid var(--content-tint-strong);
  outline-offset: -1px;
  margin-left: 8px;

  font-size: .875rem;
  font-weight: 350;
  line-height: 1.25rem;
}
.xkit-classic-footer-note-count.xkit-classic-footer-ds-look > span {
  color: var(--content-fg);
  font-weight: 500;
}
.xkit-classic-footer-note-count.xkit-classic-footer-ds-look:is(:hover, :active) {
  outline-color: var(--content-tint-heavy);
}
.xkit-classic-footer-note-count.xkit-classic-footer-ds-look:focus-visible {
  outline: 2px solid var(--accent);
  outline-offset: -2px;
  text-decoration: none;
}

/* If fonts other than Favorit Modern are allowed to use a weight of 350, it looks really bad. */
/* Use heavier weights to preserve the overall look when another addon is overriding the font. */
:root[style*="--font-family-modern"] .xkit-classic-footer-note-count.xkit-classic-footer-ds-look {
  font-weight: normal;
}
:root[style*="--font-family-modern"] .xkit-classic-footer-note-count.xkit-classic-footer-ds-look > span {
  font-weight: bold;
}
```

src/features/classic_footer/index.js
```diff
@@ -76,10 +78,46 @@ export const styleElement = buildStyle(`
     text-overflow: ellipsis;
     white-space: nowrap;
   }
+  .${noteCountClass}:focus-visible {
+    outline: none;
+    text-decoration: 2px solid var(--accent) underline;
+  }
+  .${noteCountClass}.${modernStyleClass} {
+    padding-block: 10px;
+    padding-inline: 14px;
```
marcustyphoon, line 87: oooo it's sort of halloween season sort of not really ooooo I can bikeshed from beyond the graaaaave

anyway do you think this should be `padding-inline: 16px;` to match the buttons in post activity?

AprilSylph: The original designs were 14px! The communities comment button is actually the same component—just remove the `.hasLeadingIcon` classname (`aJceP`) and you'll get the dimensions I was copying here.

PR #1993, opened 2025-11-18
feat: Show informational modal when content script is disabled

marcustyphoon:
### Description

What https://www.tumblr.com/transienturl/800535848005271552 was about :3

When XKit Rewritten is disabled in Firefox, static CSS that's linked in `manifest.json` is instantly deactivated in every open tab along with any content script js. DOM elements added by a content script are, however, not removed. This means that XKit-added DOM elements are still on the page, but have no styles (i.e. post control button SVGs get frickin' huge, sidebar items look real bad) or click handlers. This is, presumably, not a big issue during automatic updates, as the new version of the extension should be applied immediately or shortly (I have not experienced it so I don't know for sure), but if Firefox does not in fact apply the new version/is disabling the extension or the user does so themself, this looks pretty bad.

We could improve the style situation somewhat (either use link/style elements instead of `manifest.json` for static CSS, or just style-prop some critical things like control button svg size so they at least don't look awful), but we can't so easily remove the nonfunctional elements on disable, as we can run no javascript in this case.

But it's not like one needs javascript to make things occur on a web page. This is an MVP that adds a dismissable informational modal telling the user they may want to refresh the page, exclusively using the DOM. I don't remember where I first saw the pure-css dismissable popup thing, unfortunately, but I'm not sure I would have thought of it myself.

Kind of a joke PR, kind of not. (If actually doing this it probably makes sense to modify the modals css to work without modification, putting both an id and class on modals so that the util can still hide its own modals without affecting this.)

The cleaner version of this would be something like adding a data attribute to every DOM element we add, setting `display: none` with 0 specificity in a style element, and setting `display: block` with very low specificity in a static stylesheet. I don't actually know if that would work, but if it would it could make most of our modifications vanish when the static styles are removed, if implemented carefully.

### Testing steps

- Load XKit Rewritten in Firefox. (In Chromium this does nothing.)
- Disable XKit Rewritten. Confirm that an informational modal appears.
- Enable XKit Rewritten.  Confirm that the informational modal disappears.
- Disable XKit Rewritten. Confirm that an informational modal reappears, that it can be dismissed, and that it does not interfere with the page once dismissed.

### DIFF
--- a/src/utils/disabled_info_modal.js
+++ b/src/utils/disabled_info_modal.js
@@ -1,7 +1,7 @@
 import { br, div, h3, input, label, style } from './dom.js';
 
 const modalId = 'xkit-disabled-info-modal';
-const hiddenCheckboxId = 'xkit-disabled-info-dismiss-checkbox';
+const dismissCheckboxId = 'xkit-disabled-info-dismiss-checkbox';
 
 fetch(browser.runtime.getURL('/content_scripts/modals.css'))
   .then(result => result.text())
@@ -22,7 +22,7 @@ fetch(browser.runtime.getURL('/content_scripts/modals.css'))
         [
           style({}, [
             `
-              #${hiddenCheckboxId}, #${modalId}:has(> #${hiddenCheckboxId}:checked) {
+              #${dismissCheckboxId}, #${modalId}:has(> #${dismissCheckboxId}:checked) {
                 display: none;
               }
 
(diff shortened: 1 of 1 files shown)


AprilSylph:
> The cleaner version of this would be something like adding a data attribute to every DOM element we add, setting `display: none` with 0 specificity in a style element, and setting `display: block` with very low specificity in a static stylesheet. I don't actually know if that would work, but if it would it could make most of our modifications vanish when the static styles are removed, if implemented carefully.

Yeah... I would prefer to do that. We can do it one piece at a time—control buttons would be a fantastic place to start, as they're the most egregious when broken.

marcustyphoon:
Indeed. Off the top of my head, a few things will always be broken, as we have no way of making the act of disabling the extension affect non-static styles we apply to Tumblr's DOM elements—shorten posts is deprecated but surely we have combination-of-styles-and-elements features like that—but a reasonably consistent-looking state is probably possible for most things.

marcustyphoon:
Tracked in #1994.

(I imagine the work of implementing this is 90% project structure preference/10% actually doing the thing, so I will probably leave it to you.)

PR #1997, opened 2025-11-20
fix(Seen Posts): restore "only dim avatars" option functionality

### LINKED ISSUE #1951: Seen Posts: "Only dim avatars" mode does nothing
### Platform

macOS 14.8.1 arm64

### Browser

Firefox 144.0

### Addon version

v1.1.1

### Details

In "only dim avatars" mode, Seen Posts does nothing; this selector is incorrect:

https://github.com/AprilSylph/XKit-Rewritten/blob/9db84d0eadcde1f3c2cf73eff0d8b27a61114547/src/features/seen_posts/index.css#L2

AprilSylph:
### Description

- fixes #1951

Before | After
-|-
Screen Shot 2025-11-20 at 11 16 30 | Screen Shot 2025-11-20 at 11 16 38

Dimming the entire `figure` also dims whatever badge is attached to the avatar, which I think looks quite good.

I've opted to continue using `article > :first-child` instead of `article header` so that anyone using Dashboard Plus to bring back sticky avatars can continue to use this option.

(Wasn't there a `<div>` wrapper between the `<article>` and its `<header>` added recently? It seems to have gone away now. Maybe someone realised that `<header>`'s ARIA role changes based on whether it's a child of an `<article>` or not...)

### Testing steps

1. Load the modified addon
2. Open https://www.tumblr.com/dashboard and scroll until you find a post with a badge on the avatar in the header
3. Enable Seen Posts &rarr; "Only dim avatars on seen posts"
4. Refresh the feed
    - **Expected result**: The aforementioned post's avatar is now dimmed
    - **Expected result**: The avatar badge (or sub-avatar) is also dimmed to the same extent
    - **Expected result**: Original posts with dimmed avatars do not also have dimmed media

#### Compatibility testing (optional)
1. Install Dashboard Plus
2. Enable Dashboard Plus &rarr; posts &rarr; "floating avatars"
3. Reload your Tumblr tab
    - **Expected result**: The floating avatars of seen posts are dimmed
    - **Expected result**: The floating avatars of unseen posts are not dimmed

### DIFF
--- a/src/features/seen_posts/index.css
+++ b/src/features/seen_posts/index.css
@@ -1,5 +1,6 @@
-body:not(.xkit-seen-posts-only-dim-avatar) [data-seen-posts-seen]:not(:hover),
-body.xkit-seen-posts-only-dim-avatar [data-seen-posts-seen]:not(:hover) article > :first-child figure:has(img[loading="eager"]) {
+body.xkit-seen-posts-only-dim-avatar [data-seen-posts-seen]:not(:hover) article header figure, /* Unmodified Tumblr DOM structure */
+body.xkit-seen-posts-only-dim-avatar [data-seen-posts-seen]:not(:hover) article .dbplus-stickyContainer figure, /* Dashboard+ floating avatars */
+body:not(.xkit-seen-posts-only-dim-avatar) [data-seen-posts-seen]:not(:hover) {
   opacity: 0.5;
 }
 


marcustyphoon:
> (Wasn't there a `<div>` wrapper between the `<article>` and its `<header>` added recently? It seems to have gone away now. Maybe someone realised that `<header>`'s ARIA role changes based on whether it's a child of an `<article>` or not...)

Oh, interesting. Yeah. It might have become the `.eA_DC` postContent div that doesn't contain the header anymore.

AprilSylph:
image

src/features/seen_posts/index.css
```diff
@@ -1,5 +1,5 @@
 body:not(.xkit-seen-posts-only-dim-avatar) [data-seen-posts-seen]:not(:hover),
-body.xkit-seen-posts-only-dim-avatar [data-seen-posts-seen]:not(:hover) article > :first-child figure {
+body.xkit-seen-posts-only-dim-avatar [data-seen-posts-seen]:not(:hover) article > :first-child figure:has(img[loading="eager"]) {
```
AprilSylph: One might, at a glance of this, think "why even have the `> :first-child` here at all if we're already targeting avatars with `:has(img[loading="eager"])`?" to which the answer is "so we don't dim trail item avatars, obviously".

marcustyphoon: Uh... hm. I guess it's true that only header images have `loading="eager"` in this context (tumblr tv uses it but this feature is inactive there). `:is(article > header:first-child, article > div:first-child > header:first-child) figure` would work by itself, but that's kind of a lot. `header figure:is(article > :first-child *)` or `article > :first-child figure:is(header *)` also work, and feel a bit more semantic, but are perhaps a bit hard to parse. Or we could just leave it.

marcustyphoon: (If we really care about being semantic, we just querySelector `'header figure'` in the javascript, but I don't know that that's necessary.)

AprilSylph: I suppose the simplest and most semantic thing (without moving this to JS) would be just to have one selector for Tumblr's DOM and another for Dashboard+'s, huh. I should just do that instead of trying to be clever.

I do want to point out, though, that saying "we could just leave it" without also approving the PR... does feel like bikeshedding of a sort? This is probably not intentional on your side; there's no indication that you actually did test the latest changes—but all code comments are treated by GitHub as reviews, so it does put me in the position that I have to re-request review because of a non-blocking comment.

Not a big deal by any stretch, to be clear! I just know I would want it pointed out to me when there's an opportunity to be more cordial than I thought was possible in a given medium.

marcustyphoon: Interesting, I guess there's a difference between a comment that references a line and one that doesn't, in github's system, yeah. That is kind of annoying. Why give the option to request changes explicitly if not doing that has the same effect? I suppose I can make that kind of comment using a—hm, what happens if you paste a line number reference into the github comment box?

Anyway, that aside, yes, I do have some work to do in dialing in my—how do you put it? Code review... sensibilities? Not sure what the word is. As I mentioned in the other PR we're conversing in, my first question is always what we're trying to communicate. I would say in this case that, as a reviewer, I _was_ looking for something before I pressed approve, but it _wasn't necessarily a change_: it could have been that, or it could have been a comment from you acknowledging that you had thought of or had now considered alternative selectors, and were happy with the current one!

Is that what the "comments are treated as reviews" system is supposed to represent? No idea. Is that what we want it to represent? I dunno—that's for us to decide!

(Aside: backing up a bit, it might also be worth discussing what my review approval is supposed to _mean_, explicitly. It's a bit of a different dynamic when the reviewee, not the reviewer, is the repository owner! If we say the review system, when used in this direction, is about passing tests, that's totally coherent and I can stick to comments for refactoring-related discussions. That's of course not a rule we follow in the opposite direction, but there's no particular reason those have to be the same!)

marcustyphoon:
https://github.com/AprilSylph/XKit-Rewritten/blob/ce8406866e260282a42563129aa72c132d82ed21/src/features/seen_posts/index.css#L1

Generic as this is, yeah, I think I'm more confident in "only the header of a post is a header element, and it will always be a header element" than "avatars, and only avatars, have `loading="eager"`." I do like cleverly combining selectors, but that feels hard to do here without the ability to use mapped CSS.

(Yes, this comment is partially to test how it works if you substitute a permalink for commenting on a line.)

src/features/seen_posts/index.css
marcustyphoon: > saying "we could just leave it" without also approving the PR... does feel like bikeshedding of a sort?

Adding an acknowledgement here that, on a second read, you may be pointing to the fact that the comment may remove the PR from my review queue until you press to re-request (I do not actually know if this is the case, as I don't use my review queue that way), in which case it places the burden on you to press re-request so that the PR is shown to be in the correct state: waiting on me, not on you.

That's certainly a good reason not to do what I did here! (I wish there was a better alternative—I _did_ want to comment, I _did_ want that comment to point at a specific line, I _did not_ want to change the review status of the PR, and I _did not_ want to approve it at that moment as I indeed have not tested the code yet. Github apparently doesn't offer that combination of things, so that's... neat.)

AprilSylph: > I wish there was a better alternative—I _did_ want to comment, I _did_ want that comment to point at a specific line, I _did not_ want to change the review status of the PR, and I _did not_ want to approve it at that moment as I indeed have not tested the code yet. Github apparently doesn't offer that combination of things, so that's... neat.

What I like to do in this situation is leave my code-only review and then immediately re-request my own review. It's a very silly that I have to remember to do it, though.

marcustyphoon: Oh! Right, forgot one can do that.

(Does the UI actually change when you submit the... comment-that-is-apparently-a-review-I-guess? Or is there no indication that anything is different?)

AprilSylph: > Does the UI actually change when you submit the... comment-that-is-apparently-a-review-I-guess? Or is there no indication that anything is different?

Pending reviews are displayed in the sidebar as the reviewer's name and an orange dot; if there is no pending review from that person, their last most important review (changes requested = approved > commented) is displayed instead, alongside a re-review button.

AprilSylph: > Aside: backing up a bit, it might also be worth discussing what my review approval is supposed to _mean_, explicitly. It's a bit of a different dynamic when the reviewee, not the reviewer, is the repository owner!

You're a collaborator on this repository, so please try to think of yourself more as a co-owner! Ideally, even if you are still approving code where you somewhat disagree with the exact implementation, it should be because you're trusting me as a collaborator, as opposed to bowing to me as an authority figure.

PR #1999, opened 2025-11-21
fix(Quick Reblog): make tag bundle buttons accessible

AprilSylph:
### Description

- relates to #1611

Yet more things I found needed doing when trying to work on the above issue:
* Quick Reblog's tag bundle buttons don't have a fully visible focus outline (at least on Firefox)
* Quick Reblog's tag bundle buttons' checked state does not use the appropriate ARIA attribute for a toggle button

Yes, this is two fixes in one PR; yes, I am mildly ashamed.

Also includes CSS refactoring. Diff best viewed per-commit.

#### Regression testing
1. Load the modified addon
2. Enable Quick Reblog
3. Add at least five tag bundles to Quick Tags, including one with a really long title
4. Hover a post's reblog button to open the Quick Reblog form
    - **Expected result**: Exactly four tag bundles are visible without scrolling
    - **Expected result**: The tag bundles box can be scrolled vertically
    - **Expected result**: The tag bundles box cannot be scrolled horizontally
5. Look specifically at the tag bundle button with the really long title
    - **Expected result**: The tag bundle title is clipped with an ellipsis
    - **Expected result**: The tag bundle button reserves some space for a checkmark on the right
6. Click some tag bundle buttons in the Quick Reblog form
    - **Expected result**: Pressing a tag bundle button adds that bundle's tags to the tags input
    - **Expected result**: A pressed tag bundle gains a checkmark on the right
    - **Expected result**: Pressing a tag bundle button again removes that bundle's tags from the tags input
    - **Expected result**: Not-pressed tag bundles do not have a checkmark on them

#### Fix testing
1. Double-click the blog selector in the Quick Reblog form to focus it
2. Use <kbd>TAB</kbd> to move focus to the tag bundle buttons
    - **Expected result**: Focused tag bundles gain a fully visible focus outline
3. Press a tag bundle button to put it in its pressed state
4. Inspect the pressed tag bundle button's accessibility properties
    - **Expected result**: The pressed tag bundle has the `"pressed"` state
    - **Expected result**: Not-pressed tag bundles do not have the `"pressed"` state

### DIFF
--- a/src/features/quick_reblog/index.css
+++ b/src/features/quick_reblog/index.css
@@ -64,7 +64,8 @@
 
 #quick-reblog .select-container {
   display: flex;
-  padding: 4px;
+  gap: 2px;
+  padding: 6px;
   margin-bottom: 2px;
 
   background-color: inherit;
@@ -74,7 +75,6 @@
   flex: none;
   height: 26px;
   width: 26px;
-  margin-right: 2px;
 
   background-size: cover;
   border-radius: 2px;


AprilSylph:
It does sometimes feel a bit silly to write out these detailed tests/expectations to essentially verify that something visually works as it did before, but I feel like "open two browser profiles, load the `master` version of the addon in the first one, and load this modified version in the other, then mimic every action taken in the first one as precisely as possible in the second, and expect things to look more or less the same; maybe a little different but not fundamentally different" is probably an even taller order for a reviewer, so.

Oh, and I guess doing it this way trains my unit testing muscle, which would otherwise be so withered as to be dead.

src/features/quick_reblog/index.css
```diff
@@ -81,59 +81,69 @@
...
 
+#quick-reblog .quick-tags button::after {
   content: "\2713";
-  line-height: 1;
+  visibility: hidden;
+}
+
+#quick-reblog .quick-tags button[aria-pressed="true"]::after {
+  visibility: visible;
+}
```
marcustyphoon, line 142: I see this reserves the slot for the checkmark icon even when it's not there, so that the ellipsis is in the same place when you click; this doesn't have to be the case and is just preference, correct? (It seems slightly better to me the other way, but of course it makes essentially no difference.)

Aside: okay, let's see what the UI looks like once I submit this comment without having re-requested review yet...

marcustyphoon: > Aside: okay, let's see what the UI looks like once I submit this comment without having re-requested review yet...

Ah, I see.

marcustyphoon (approved):
Line-height Good, Tags Manipulatable ✅

src/features/quick_reblog/index.css
AprilSylph: > I see this reserves the slot for the checkmark icon even when it's not there, so that the ellipsis is in the same place when you click; this doesn't have to be the case and is just preference, correct? (It seems slightly better to me the other way, but of course it makes essentially no difference.)

Correct. I should at least see how it looks the other way...

AprilSylph: Yeah, I really don't like having the text clipping move to account for the tick. The visible text is the button's label, so it's distracting for it to change.

1ch (~10px) is too much of a gap, though. A space character's width is closer to ~4px, which feels more appropriate.

Also, `ch` is a weird unit... I want to move away from it.

marcustyphoon:
Was making every line taller intentional?

marcustyphoon:
Oh, I get it: it's the same with the standard font size, but not with modified font size (Palettes for Tumblr).

PR #2010, opened 2025-11-28
fix(Tweaks): add cleanup logic to attribution links option

### LINKED ISSUE #1087: Restore attribution tweak cannot be disabled
### Platform

MacOS 13.2.1 arm64

### Browser

Chrome 113.0.5672.126

### Addon version

12dcad4

### Details

The new tweak to make post header links point to individual posts does not turn itself off when it or the entire Tweaks script is disabled, as it does not have a fully implemented `clean()` function.

This is pretty minor, of course, but it's reasonably easy to implement (e.g. storing the original href value in a data attribute).

AprilSylph:
### Description

- fixes #1087

### Testing steps

1. Load the modified addon in Firefox via `about:debugging#/runtime/this-firefox`
2. Open a Tumblr tab, and find a reblogged post on your dashboard (not Peepr)
3. Enable Tweaks &rarr; "Restore links to individual posts in the post header"
    - **Expected result**: Post attribution blogname links become permalinks to the current post
    - **Expected result**: Reblog attribution blogname links become permalinks to the reblog parent post
4. Disable the option
    - **Expected result**: Links that were previously modified are restored to their default state, linking only to blogs
5. Enable the option again
6. Reload the addon via `about:debugging` to simulate an update or restart
7. Go back to the Tumblr tab
    - **Expected result**: Attribution links still have modified link targets, pointing to the post or parent as appropriate
    - **Expected result**: Clicking an attribution link still opens the post/parent in Peepr
8. Disable the option again
    - **Expected result**: Attribution links are once again restored to their original state—including those that were on the page at the time of the addon restart

### DIFF
--- a/src/features/tweaks/restore_attribution_links.js
+++ b/src/features/tweaks/restore_attribution_links.js
@@ -36,7 +36,7 @@ const processPosts = async function (postElements) {
     const reblogAttributionLink = postElement.querySelector(reblogAttributionLinkSelector);
 
     if (postAttributionLink && postAttributionLink.textContent === blogName) {
-      postAttributionLink.dataset.href ??= postAttributionLink.getAttribute('href');
+      postAttributionLink.dataset.originalHref ??= postAttributionLink.getAttribute('href');
       postAttributionLink.href = postUrl;
       postAttributionLink.dataset.blogName = blogName;
       postAttributionLink.dataset.postId = id;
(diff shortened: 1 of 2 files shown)


src/features/tweaks/restore_attribution_links.js
```diff
@@ -20,6 +20,8 @@ const onLinkClick = event => {
   }
 };
 
+const useCapture = { capture: true };
```
marcustyphoon: The fact that the addEventListener/removeEventListener signature allows this to be a boolean value (useCapture?) or an options object (listenerOptions?) makes this variable name slightly odd, but I think it's fine :D

marcustyphoon:
Something like `data-original-href` might be a bit more descriptive than `data-href`.

If making this change, it should of course be made to the create button tweak as well.

marcustyphoon (approved):

src/features/tweaks/restore_attribution_links.js
AprilSylph: I was trying to be cute, but I should just call it `listenerOptions`.

marcustyphoon: `const useCapture = true;` presumably also works.

PR #2018, opened 2025-12-02
 feat(Tweaks): apply restore attribution tweak to trail items 

### LINKED ISSUE #2008: Tweaks: apply restore attribution tweak to trail items
### Description

- blocks #2007
- see https://github.com/AprilSylph/XKit-Rewritten/issues/2007#issuecomment-3585077863

---

Treating the post header and trail item headers any differently without good reason is, in my opinion, a mistake.

The "Restore links to individual posts in the post header" Tweaks option should apply to both, not just the former.

AprilSylph:
### Description

- resolves #2008 

Adds the necessary util to unbury the `trailItem` context value from `trailContextProvider`, which wraps every reblog trail item element. I found this necessary to achieve a level of robustness that I am satisfied with; the various ways that the reblog trail can be modified (the editor, Trim Reblogs) gives me low confidence that the rendered DOM will always match what we see in a timeline object's `trail`, especially given that we cache post timeline objects.

Also adds a `trailItemSelector` export to `interface.js`, to match the symbiosis between `postSelector` and the `timelineObject` util.

The actual change to the Tweaks option in question is better explained by the diff.

### Testing steps

1. Load the modified addon
2. Enable Tweaks &rarr; "Restore links to individual posts in the post header"
3. Find a reblogged post with several different trail items
    - **Expected result**: Every blogname in the trail links to their associated permalink correctly
      - When the permalink is a `*.tumblr.com` blog network URL
      - When the permalink is a custom domain blog network URL
      - When the permalink is a Peepr URL
    - **Expected result**: A normal click to modified blogname links opens that permalink in the Redpop UI
      - When the permalink is a `*.tumblr.com` blog network URL (should open Peepr)
      - When the permalink is a custom domain blog network URL (should open Peepr)
      - When the permalink is a Peepr URL (should open Peepr)
4. Find a reblogged post with contributed content
    - **Expected result**: The last blogname in the trail links to the permalink of the post itself, matching the main post attribution
      - When the permalink is a `*.tumblr.com` blog network URL
      - When the permalink is a custom domain blog network URL
      - When the permalink is a Peepr URL
      - When the permalink is a communities URL
    - **Expected result**: A normal click to the modified blogname link opens the permalink in the Redpop UI
      - When the permalink is a `*.tumblr.com` blog network URL (should open Peepr)
      - When the permalink is a custom domain blog network URL (should open Peepr)
      - When the permalink is a Peepr URL (should open Peepr)
      - When the permalink is a communities URL (should open Communities)
5. Find a reblogged post with broken blogs in the reblog trail
    - **Expected result**: All active blognames in the reblog trail link to their associated permalinks
    - **Expected result**: All inactive blognames in the reblog trail are not affected

### DIFF
--- a/src/features/tweaks/restore_attribution_links.js
+++ b/src/features/tweaks/restore_attribution_links.js
@@ -38,9 +38,10 @@ const processPosts = async function (postElements) {
       slug,
     } = await timelineObject(postElement);
     const postAttributionName = (!!community && authorBlog?.name) || blogName;
-    const postRouterUrl = new URL(postUrl).hostname === 'www.tumblr.com'
-      ? postUrl
-      : `${blog.blogViewUrl.replace(/\/$/, '')}/${id}${slug ? `/${slug}` : ''}`;
+    const hasCustomTheme = new URL(postUrl).hostname !== 'www.tumblr.com';
+    const postRouterUrl = hasCustomTheme
+      ? `${blog.blogViewUrl.replace(/\/$/, '')}/${id}${slug ? `/${slug}` : ''}`
+      : postUrl;
 
     const postAttributionLink = postElement.querySelector(postAttributionLinkSelector);
     const reblogAttributionLink = postElement.querySelector(reblogAttributionLinkSelector);


src/features/tweaks/restore_attribution_links.js
```diff
@@ -1,13 +1,12 @@
...
+import { timelineObject, trailItem } from '../../utils/react_props.js';
 import { navigate } from '../../utils/tumblr_helpers.js';
+import { trailItemSelector } from '../../utils/interface.js';
 
 const postAttributionLinkSelector = 'header a[rel="author"]';
-const reblogAttributionLinkSelector = `
-  header ${keyToCss('rebloggedFromName')} a,
-  header ${keyToCss('subheader')} a${keyToCss('blogLink')}
-`;
+const reblogAttributionLinkSelector = `header ${keyToCss('subheader')} a${keyToCss('blogLink')}`;
```
marcustyphoon, line 8: Not related to this PR, but I was worried that this would include "reblogged to [community name]. It doesn't seem to.

src/features/tweaks/restore_attribution_links.js
```diff
@@ -50,6 +50,19 @@ const processPosts = async function (postElements) {
...
       reblogAttributionLink.addEventListener('click', onLinkClick, listenerOptions);
     }
+
+    trailItemElements.forEach(async trailItemElement => {
+      const { blog, post } = await trailItem(trailItemElement);
+      const trailAttributionLink = trailItemElement.querySelector(trailAttributionLinkSelector);
+
+      if (trailAttributionLink && trailAttributionLink.textContent === blog?.name && post?.id) {
+        trailAttributionLink.dataset.originalHref ??= trailAttributionLink.getAttribute('href');
+        trailAttributionLink.href = `https://${blog.name}.tumblr.com/post/${post.id}`;
```
marcustyphoon: Using `blog.url` here would skip one redirect on blogs with custom domains, but I don't think that matters? I imagine the robustness of both would be the same.

marcustyphoon (changes requested):
Does not work on posts reblogged into communities with contributed content. Yeah, I don't think I would have thought of that either :D

src/features/tweaks/restore_attribution_links.js
AprilSylph: Yeah, blog links are their own thing because of blog cards.

This _does_ match author/submitter names on non-reblogged posts, but that's what the `textContent` check is for.

src/features/tweaks/restore_attribution_links.js
AprilSylph: `blog.url` could also be a `www.tumblr.com` address, for blogs without a custom theme. We would then need to test for that and only situationally include `/post` if we didn't find it, and... well, we're already basically expecting a redirect because of the lack of slug. If Tumblr's redirects are inefficient, that's their problem.

AprilSylph:
I fail to escape special-casing things yet again.

marcustyphoon:
It's too bad the a/b test isn't launched so that we could just simulate a click on the post body and let tumblr deal with it lmao (yeah, no, let's not)

src/features/tweaks/restore_attribution_links.js
```diff
@@ -25,31 +29,57 @@ const listenerOptions = { capture: true };
...
+        trailAttributionLink.dataset.originalHref ??= trailAttributionLink.getAttribute('href');
+        trailAttributionLink.dataset.blogName = blog.name;
+        trailAttributionLink.dataset.postId = post.id;
+
+        if (isContributedContent) {
+          trailAttributionLink.href = postUrl;
+        } else if (blog.url.startsWith('https://www.tumblr.com')) {
+          trailAttributionLink.href = `${blog.url.replace(/\/$/, '')}/${post.id}`;
+        } else {
+          trailAttributionLink.href = `${blog.url.replace(/\/$/, '')}/post/${post.id}`;
```
AprilSylph: I don't like trusting that one format will always end with a trailing `/` and the other always won't.

marcustyphoon:
Wow, I see why we _don't_ use `blog.url` :D

src/features/tweaks/restore_attribution_links.js
```diff
@@ -1,20 +1,24 @@
...
   event.stopPropagation();
   if (event.ctrlKey || event.metaKey || event.altKey || event.shiftKey) return;
 
+  const { href } = event.currentTarget;
   const { blogName, postId } = event.currentTarget.dataset;
-  if (blogName && postId) {
+
+  if (href.startsWith('https://www.tumblr.com/')) {
+    event.preventDefault();
+    navigate(href.replace(/^https:\/\/www\.tumblr\.com\//, '/'));
```
marcustyphoon:
```suggestion
    navigate(new URL(href).pathname);
```

I mean, in theory we could...

src/features/tweaks/restore_attribution_links.js
```diff
@@ -25,31 +29,57 @@ const listenerOptions = { capture: true };
...
+        trailAttributionLink.dataset.blogName = blog.name;
+        trailAttributionLink.dataset.postId = post.id;
+
+        if (isContributedContent) {
+          trailAttributionLink.href = postUrl;
+        } else if (blog.url.startsWith('https://www.tumblr.com/')) {
+          trailAttributionLink.href = `${blog.url.replace(/\/$/, '')}/${post.id}`;
+        } else {
+          trailAttributionLink.href = `${blog.url.replace(/\/$/, '')}/post/${post.id}`;
+        }
```
marcustyphoon, line 78:
```suggestion
          trailAttributionLink.href = postUrl;
        } else {
          const pathname = new URL(blog.url).origin === 'https://www.tumblr.com' ? `/${post.id}` : `/post/${post.id}`;
          trailAttributionLink.href = new URL(pathname, blog.url);
        }
```

...use native URL parsing, right? (But yeah, I would also be tempted to just let Tumblr handle it; we already are doing that with the `tumblr.com/@[blogname]` trick.)

AprilSylph: I forgot about the second argument to `new URL`. Yeah, I like this enough.

src/features/tweaks/restore_attribution_links.js
```diff
@@ -25,31 +29,56 @@ const listenerOptions = { capture: true };
...
+      if (trailAttributionLink && trailAttributionLink.textContent === blog?.name && post?.id) {
+        trailAttributionLink.dataset.originalHref ??= trailAttributionLink.getAttribute('href');
+        trailAttributionLink.dataset.blogName = blog.name;
+        trailAttributionLink.dataset.postId = post.id;
+
+        if (isContributedContent) {
+          trailAttributionLink.href = postUrl;
+        } else {
+          const pathname = new URL(blog.url).hostname === 'www.tumblr.com' ? `/${post.id}` : `/post/${post.id}`;
+          trailAttributionLink.href = new URL(pathname, blog.url);
```
marcustyphoon:
```suggestion
          const hasCustomTheme = new URL(blog.url).hostname !== 'www.tumblr.com';
          trailAttributionLink.href = new URL(hasCustomTheme ? `/post/${post.id}` : `/${post.id}`, blog.url);
```

Okay, I'll admit my own mistake. This should probably be self-documenting.

marcustyphoon (changes requested):
Ohhhh fuck that's not how `new URL(url, base)` works unless there's a trailing slash on the base.

My bad. Uh, maybe that's not worth it.

Comment I was going to make before I noticed this:

> If we're bothering to get rid of one redirect, do we rely on the existence of `parentPostUrl` (and `blog.blogViewUrl`, and an override that selects `blog.url` instead if it's a community, because `blog.blogViewUrl` in communities is literally broken, whee), replace `dataset.blogName` and `dataset.postId`  with `dataset.blogViewHref`, and get rid of them all?
> 
> On the one hand, that obviously doesn't seem necessary. On the other hand, the code would be a lot more readable if we did do it that way; right now without comments there's no reason an uninformed reader could conclude for three DOM properties existing. We could add comments or factor more, but

marcustyphoon:
So, on an affected link element...

- `href` is where we want a command-click to go, which will be the blog network if the post in question is on a blog with a custom theme, otherwise peepr (or the community post thing)
- `data-original-href` is what `href` used to be, for cleanup
- `data-router-url` is where we want a click to go, which will be peepr (or the community post thing); if this is not a www.tumblr.com url... something has gone wrong; we can't soft-navigate to that; fall back to letting the click hit the anchor, which will hard navigate to the href (reasonable)

marcustyphoon:
We don't clean up `data-router-url`, though we also shouldn't really need to.

src/features/tweaks/restore_attribution_links.js
```diff
@@ -1,55 +1,84 @@
...
-      rebloggedFromId,
       rebloggedFromName,
-      rebloggedFromUrl
+      rebloggedFromUrl,
+      slug,
     } = await timelineObject(postElement);
+    const postAttributionName = (!!community && authorBlog?.name) || blogName;
+    const routerPostUrl = new URL(postUrl).hostname === 'www.tumblr.com'
+      ? postUrl
+      : `${blogViewUrl.replace(/\/$/, '')}/${id}${slug ? `/${slug}` : ''}`;
```
marcustyphoon: I think it would be slightly clearer to only destructure `blog` and do `blog.blogViewUrl` here (else it looks like it might be the blog view url for the post until you scroll back up).

(Test comment from within the VS Code Github integration)

marcustyphoon: Also, this variable seems like a `postRouterUrl`, as it's the router url for the post.

src/features/tweaks/restore_attribution_links.js
```diff
@@ -1,55 +1,84 @@
...
+      parentPostUrl,
       postUrl,
-      rebloggedFromId,
       rebloggedFromName,
-      rebloggedFromUrl
+      rebloggedFromUrl,
+      slug,
     } = await timelineObject(postElement);
+    const postAttributionName = (!!community && authorBlog?.name) || blogName;
+    const routerPostUrl = new URL(postUrl).hostname === 'www.tumblr.com'
```
marcustyphoon: The logic below it isn't quite the same as the below use for trail items, but would it still be accurate to call this hasCustomTheme?

(I guess that means the bottom one would shadow, which is annoying.)

AprilSylph: No, this covers the community post case as well.

marcustyphoon: Yeah—for clarity, it would be `!==`, so the community post case would accurately be described as "does not have a custom theme; we use unmodified `postUrl`."

AprilSylph: We're already shadowing `blog`, so why not.

marcustyphoon: "If the blog has a custom theme, `postUrl` isn't what we would want to use for `postRouterUrl`, so construct it from the blog's `blogViewUrl`, the `id`, and the `slug`"

Yep, that's the good stuff.

Issue #2001, opened 2025-11-24
Themed Posts: Theming trail items clashes with new Tumblr A/B test


marcustyphoon:
### Platform

MacOS 14.8.1 arm64

### Browser

Firefox 145.0.1 

### Addon version

1.1.4

### Details

With the new Tumblr a/b test, hovered trail items can have `background-color: var(--content-tint)`. This overrides, rather than adding to, the background color we set on trail items with the "Theme every reblog trail item individually" option, so trail items go "transparent" when hovered and their text colors can be unreadable.

A hotfix for this would be straight-up disabling that option if the a/b test is detected (not immediately sure of a clean way to do this; the CSS map has 53 entries for the obvious string). If we use color-mix to fix it I am going to advocate that we do not change minimum browser versions and leave the regression on Chrome 109 intentionally.

**Edit:**

> A hotfix for this would be straight-up disabling that option if the a/b test is detected

This has been done in #2026, so the new behavior is that the "Theme every reblog trail item individually" option just does nothing in this case for now.

marcustyphoon:
Well... this is uh... one way to do it.

Background color is themed post coloration; inset box shadow is hover effect:

```diff
diff --git a/src/features/themed_posts/index.js b/src/features/themed_posts/index.js
index 7a42fac1..030f2336 100644
--- a/src/features/themed_posts/index.js
+++ b/src/features/themed_posts/index.js
@@ -115,6 +115,16 @@ export const main = async function () {
       ${postSelector} ${reblogSelector}:not(:last-child) > :last-child {
         margin-bottom: 15px;
       }
+
+      @media (hover: hover) and (pointer: fine) {
+        [data-xkit-themed]${keyToCss('reblog')}${keyToCss('withTrailItemPermalink')} {
+          transition: box-shadow 0.2s linear;
+        }
+        [data-xkit-themed]${keyToCss('reblog')}${keyToCss('withTrailItemPermalink')}:hover {
+          background-color: var(--content-panel);
+          box-shadow: inset 0 0 0 9999px var(--content-tint);
+        }
+      }
     `;
     if (missingPostMode === 'palette') {
       styleElement.textContent += `
```

marcustyphoon:
Also works. Background color is themed post coloration; pseudo element is hover effect:

```diff
diff --git a/src/features/themed_posts/index.js b/src/features/themed_posts/index.js
index 7a42fac1..d427b027 100644
--- a/src/features/themed_posts/index.js
[...]

PR #2026, opened 2025-12-04
fix(Themed Posts): Temporarily disable reblog trail theming with new Tumblr A/B Test

marcustyphoon:
### Description

As noted in #2001, on accounts with the new Tumblr A/B test, hovered reblog trail items change background color. This overwrites the Themed Posts reblog trail theming background color, but not the foreground, often making text unreadable.

Options to fix this properly are discussed in that PR. If we want to do a release without fixing that, I think it makes sense to do one in which the subfeature is nonfunctional if it would be buggy, which is quite easy. This will also mean #2024 will not cause the same bug (it only affects `withTrailItemPermalink`) until we revert this and do it all properly.

Naturally, if you intend to look into resolving #2001 with full functionality, April, close this instead of reviewing it.

### Testing steps

- Open https://www.tumblr.com/april/801912410171719680 on an account with `postChromeBodyNavigationEventsRedesign`.
- Enable Themed Posts.
- Enable the "Theme every reblog trail item individually" option and confirm that... nothing happens.
- Hover over the text at the bottom of the reblog trail item and confirm that the text is still readable, as there is a normal/subtle hover effect.
- Open https://www.tumblr.com/april/801912410171719680 on an account without `postChromeBodyNavigationEventsRedesign`.
- Enable Themed Posts.
- Enable the "Theme every reblog trail item individually" option and confirm that the reblog trail item is themed blue as normal.

### DIFF
--- a/src/features/themed_posts/index.js
+++ b/src/features/themed_posts/index.js
@@ -8,7 +8,11 @@ export const styleElement = buildStyle();
 
 const blogs = new Set();
 const groupsFromHex = /^#(?<red>[A-Fa-f0-9]{1,2})(?<green>[A-Fa-f0-9]{1,2})(?<blue>[A-Fa-f0-9]{1,2})$/;
-const reblogSelector = keyToCss('reblog');
+
+// Prevent unreadable text by not theming trail items if they'll change
+// background color on hover until we implement compatibility with
+// postChromeBodyNavigationEventsRedesign.
+const reblogSelector = `${keyToCss('reblog')}:not(${keyToCss('withTrailItemPermalink')})`;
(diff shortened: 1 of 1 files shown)


AprilSylph:
> Naturally, if you intend to look into resolving #2001 with full functionality, April, close this instead of reviewing it.

This isn't an either/or situation! Mitigation before resolution is encouraged.

...you probably mean "intend" with immediacy. Yeah, I don't.

marcustyphoon:
If you intend to—hm, there's indeed no way to make that sentence work as intended, is there :D I guess "if you instead intend to" would do the job!

PR #2034, opened 2025-12-15
docs: document `src/main_world/`

### LINKED ISSUE #1519: Documentation: Document `main_world` directory
### File

docs/Chapter 3 - Repository structure.md

### Details

We moved some files into a new directory in `src` in #1514.

AprilSylph:
### Description

- resolves #1519

### Testing steps

1. Take a look at the modified [docs/Chapter 3 - Repository structure.md](https://github.com/AprilSylph/XKit-Rewritten/blob/c21274fea92bc35561599cfac5343a3d11145563/docs/Chapter%203%20-%20Repository%20structure.md)
    - Does it make sense? Could the verbiage be improved somehow?
    - Are there any spelling or grammatical errors?
    - **Expected result**: The inline link to `src/utils/inject.js` works

### DIFF
--- a/src/utils/inject.js
+++ b/src/utils/inject.js
@@ -3,7 +3,7 @@
  * its result. This permits access to variables exposed by the Tumblr web
  * platform that are normally inaccessible in the content script sandbox.
  * @see [src/main_world/index.js](../main_world/index.js) and named scripts in the same directory
- * @param {string} path Absolute path of script to inject; will be fed to {@linkcode https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL|browser.runtime.getURL()}
+ * @param {string} path Absolute path of script file to inject; will be fed to {@linkcode https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL|browser.runtime.getURL()}
  * @param {Array} [args] Array of arguments to pass to the injected function
  * @param {Element} [target] Target element; will be accessible as the `this` value in the injected function
  * @returns {Promise<any>} The transmitted result of the function call


src/utils/inject.js
```diff
@@ -1,12 +1,12 @@
...
- * in the content script sandbox.
- * See the src/main_world directory and [../main_world/index.js](../main_world/index.js).
- * @param {string} path Absolute path of script to inject (will be fed to `runtime.getURL()`)
- * @param {Array} [args] Array of arguments to pass to the script
- * @param {Element} [target] Target element; will be accessible as the `this` value in the injected function.
- * @returns {Promise<any>} The transmitted result of the script
+ * Runs a function in the page's "main" execution environment and returns
+ * its result. This permits access to variables exposed by the Tumblr web
+ * platform that are normally inaccessible in the content script sandbox.
+ * @see [src/main_world/index.js](../main_world/index.js) and named scripts in the same directory
```
marcustyphoon, line 5: This link doesn't work in my vs code if I mouse over `inspect()` in another file, but I think that's just a vs code limitation/bug.

src/utils/inject.js
```diff
@@ -1,12 +1,12 @@
...
+ * @param {string} path Absolute path of script to inject; will be fed to {@linkcode https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL|browser.runtime.getURL()}
```
marcustyphoon:
```suggestion
 * @param {string} path Absolute path of script file to inject; will be fed to {@linkcode https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL|browser.runtime.getURL()}
```

Either way.

docs/Chapter 3 - Repository structure.md
```diff
@@ -27,6 +28,7 @@ src/
 ├── features/             User-facing features
 ├── icons/                Extension icons
 ├── lib/                  External libraries
+├── main_world/           Scripts to execute in the webpage context
```
marcustyphoon, line 31: Should this also be "functions"?

src/utils/inject.js
AprilSylph: I feel like if GitHub can accurately transform relative paths to link to the right thing in Markdown files, VS Code should be able to do the same for Markdown-formatted JSdoc. Their loss.

Also we already had this link in this form, it just wasn't in a `@see` block before.

marcustyphoon: Yeah, to be clear, it works fine in the file itself. It's just that if you mouse over the imported function in another file, vs code tries to follow the relative path from the file you're in, not from the location of the jsdoc in question. Presumably an oversight.

docs/Chapter 3 - Repository structure.md
AprilSylph: The distinction between script and function I make in the `inject()` JSdoc IMO only really matters when trying to communicate "the function lives in a file; you need the file name in order to execute the function", which isn't a factor here.

- "Scripts to _evaluate_ in the webpage context"—technically the most correct, but doesn't communicate the actual usage (running the functions that are evaluated)
- "_Functions_ to execute in the webpage context"—Communicates usage perfectly, but glosses over the existence of `index.js`, which seems... wrong? More wrong than the ambiguity, at least.

PR #2036, opened 2025-12-15
feat(UI): hide help links on deprecated features

### LINKED ISSUE #2035: Hide help links on deprecated features
### Description

I've been removing deprecated features' help sections from https://github.com/AprilSylph/XKit-Rewritten/wiki/Features, but their help links remain in the addon UI.

If we were to simply remove their help links from their feature definitions, they would show up as both deprecated and "New", so I'm not proposing that that be done.

If a feature is marked as deprecated, it should not show a help link or a "New" badge. If it should be replaced by anything, maybe it should be a little warning symbol...?

AprilSylph:
### Description

- resolves #2035

Replaces help buttons on deprecated features with orange warning icons:

Before | After
-|-
Screen Shot 2025-12-15 at 16 15 29 | Screen Shot 2025-12-22 at 11 12 55

This PR also changes the colour of the "New" badge to be pink instead of orange, since orange is more of a warning-type colour, while pink is more... celebratory? They definitely shouldn't be the same colour, anyway.

Before | After
-|-
Screen Shot 2025-12-15 at 16 17 51 | Screen Shot 2025-12-22 at 11 13 36

### Testing steps

1. Load the modified addon
2. Restore this config: [XKit Backup @ 2025-12-15.json](https://github.com/user-attachments/files/24171001/XKit.Backup.%40.2025-12-15.json)
3. Open the XKit control panel
    - **Expected result**: Shorten Posts and Timestamps show warning icons where the help icon usually is
    - **Expected result**: Hovering over a warning icon shows a "This feature is deprecated." tooltip
    - **Expected result**: Clicking a warning icon does nothing

### DIFF
--- a/src/action/components/xkit-feature/index.css
+++ b/src/action/components/xkit-feature/index.css
@@ -46,7 +46,7 @@ summary:focus {
 
 .icon,
 .meta,
-.help {
+.badge {
   margin-right: 1ch;
 }
 
@@ -62,47 +62,23 @@ summary:focus {
   margin-left: auto;
 }
 
-.help {
+.badge ::slotted {
   display: flex;
   justify-content: center;
   align-items: center;
+}
 
+.badge ::slotted(a) {
   text-decoration: none;
 }
 
-.help i {
-  font-size: 1.25rem;
-}
-
-[data-deprecated="false"] .help[href]::before {
-  content: '' / 'Help';
-}
-
-[data-deprecated="false"] .help:not([href]) i {
-  display: none;
-}
-
-[data-deprecated="false"] .help:not([href])::before {
-  padding: 0 1ex;
+.badge ::slotted(span) {
+  padding: 2px 6px;
   border-radius: 1em;
 
   background-color: #ff62ce;
   color: rgb(var(--white));
-  content: 'New!';
   line-height: 1.5;
-}
-
-[data-deprecated="true"] .help {
-  pointer-events: none;
-}
-
-[data-deprecated="true"] .help::before {
-  content: '' / 'Deprecated';
-}
-
-[data-deprecated="true"] .help i::before {
-  content: '\EA20';
-  color: #ff8a00;
 }
 
 .title,
--- a/src/action/components/xkit-feature/index.js
+++ b/src/action/components/xkit-feature/index.js
@@ -14,7 +14,9 @@ const templateDocument = new DOMParser().parseFromString(`
           <p class="description"><slot name="description"></slot></p>
         </div>
         <div class="buttons">
-          <a class="help" target="_blank"><i aria-hidden="true" class="ri-fw ri-question-fill" style="color:rgb(var(--black))"></i></a>
+          <div class="badge">
+            <slot name="badge"></slot>
+          </div>
           <input type="checkbox" checked class="toggle-button" aria-label="Enable this feature">
         </div>
       </summary>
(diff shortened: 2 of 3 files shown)


marcustyphoon:
We increased the minimum Firefox version to 128, so miraculously for you, you actually do have a [way to make pseudos accessible](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/content#alternative_text_string_counter)!

marcustyphoon:
Naturally, being me, I was immediately like "there is clearly slightly too much white in that pink," then stared at my screen moving the color picker until I got eyestrain, settled on somewhere around `rgb(236, 92, 190)`, checked, and noted that no, it really doesn't look all that different. I think to truly nail the design you probably have to pick different tones in dark and light modes. Pretend I said something here involving terms like luminance and the word "oklch" that belied a history as a trained graphic designer.

marcustyphoon:
> We increased the minimum Firefox version to 128, so miraculously for you, you actually do have a [way to make pseudos accessible](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/content#alternative_text_string_counter)!

(Do note that using this syntax in an unsupported browser completely breaks the pseudo, iirc. I assume you could probably supports-block it. But anyway.)

src/action/components/xkit-feature/index.css
```diff
@@ -70,22 +70,39 @@ summary:focus {
...
 }
 
-.help i {
-  font-size: 1.25rem;
+[data-deprecated="true"] .help {
+  pointer-events: none;
+}
+
+[data-deprecated="true"] .help::before {
+  content: '' / 'Deprecated';
```
marcustyphoon: Interesting—I assume this is for parity with the help version? Or is there a specific reason to put the alt text on a different pseudo?

marcustyphoon:
Actually—there's no hover text on the deprecated icons and they don't link anywhere like the help icons do, so it's not obvious what they mean. Should they link to... I guess a wiki page?

src/action/components/xkit-feature/index.css
AprilSylph: It's so that the alt text is still exposed to the accessibility tree when the `<i>` elements have `aria-hidden="true"` on them, which I did so that the icon glyph codes / unknown characters aren't read aloud.

AprilSylph:
> Actually—there's no hover text on the deprecated icons and they don't link anywhere like the help icons do, so it's not obvious what they mean. Should they link to... I guess a wiki page?

Yeah, probably.

src/action/components/xkit-feature/index.css
marcustyphoon: Ah, gotcha.

marcustyphoon: ...Now that I think about it, is this really doing anything different than just setting a relevant aria- attribute on the parent?

AprilSylph: Nope! It's just doing it with CSS instead of JS because I am lazy.

Commits to remedy this forthcoming.

marcustyphoon: Hey, when all you have is a hammer—nope, that's not an anecdote about—nope, anecdote is not the word I'm looking for there either.

Correction: "Hey, when all you have is a hammer—oh wait, no, that's not a proverb about the tool in question being new. Well, when you have a new hammer."

AprilSylph:
Actually, I'm not at all motivated to add a page to the wiki right now—maybe if we ship this and it generates questions, we add it then?

marcustyphoon:
What do you think about putting a title attribute—although, it does say "deprecated" a few pixels to the left, I know that's kind of weird. On the other hand, "(deprecated)" and "(disabled)" look essentially the same. Yeah, I dunno, maybe there's nothing to do to improve it. I am just somewhat concerned that this reads more as "action required" or "this feature is currently nonfunctional."

**edit:** Ignore this, I was on my phone and saw the comment but not the commit, so I thought you had made no changes!

PR #2040, opened 2025-12-18
chore: configure Dependabot version updates

AprilSylph:
### Description

I would like to migrate away from Renovate so that I can uninstall it from my account.

Now that Dependabot can do grouping, I don't see a compelling reason to stay on Renovate.

### Testing steps

Just code review. See [Dependabot options reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference) for comparison with our Renovate config.


marcustyphoon:
There is a part of me that's tempted to suggest switching to cron (e.g. `0 0 5 */4 *`) to be able to force these off the first of the month, as I've experienced quite slow dependabot jobs on first-of-month and I figure being a few days late to the party is always kind of nice when it means other people will be finding bugs instead of you. But it really doesn't matter if they're slow unless you have a lot of repos and mind the PRs trickling in over a few hours, and the descriptive interval value is certainly clearer.

AprilSylph:
So, using an empty string for labels doesn't work. My backup plan was to add a "bot" label and set that as the dependabot label, which does seem to work.

- See https://github.com/AprilSylph/Filtering-Plus/pull/37#issuecomment-3873245786
- See AprilSylph/Palettes-for-Tumblr@e2613be1d157d9f0c4986c3a7538dbf3fb323e46
- See https://github.com/AprilSylph/Palettes-for-Tumblr/pull/281

marcustyphoon (approved):

PR #2046, opened 2025-12-27
feat(Themed Posts): Adjust nearly-identical background/foreground colors

marcustyphoon:
### Description

This makes the text/accent colors used by Themed Posts a bit lighter or darker if they're _close to_ matching the background color, expanding on #1917.

The threshold used here is of course pretty arbitrary. I'm generally inclined to be fairly conservative with it, because due to the fact that for browser support reasons we're not doing color manipulation anywhere close to correctly, the results of this intervention (and the consistency of the computation of the threshold, probably) are quite low quality. Readable is better than not, of course, but readable and bad looking is not necessarily better than marginally readable and good looking, in the case of this specific feature (there's a clear consensus that the jankiness is clearly part of the charm).

### Testing steps

- Enable Themed Posts and post theming in the blog view.
- Confirm that https://www.tumblr.com/markovchaingang has post text increased in contrast.
- Confirm that https://www.tumblr.com/targetedknowledge has post text increased in contrast.
- Confirm that https://www.tumblr.com/eregar does **not** have post text increased in contrast (should we bump the threshold to e.g. 48 so this one is affected? I'm torn.)
- Confirm that https://www.tumblr.com/robo-nonagon does **not** have post text increased in contrast.

### DIFF
--- a/src/features/themed_posts/index.js
+++ b/src/features/themed_posts/index.js
@@ -34,6 +34,14 @@ const hexToRGBComponents = hex => {
 };
 const hexToRGB = hex => hexToRGBComponents(hex).join(', ');
 
+const colorsAreSimilar = (hexA, hexB) => {
+  const componentsA = hexToRGBComponents(hexA);
+  const componentsB = hexToRGBComponents(hexB);
+  return Object.keys(componentsA).every(
+    i => Math.abs(componentsA[i] - componentsB[i]) < 32
+  );
+};
+
 const createContrastingColor = hex => {
   const components = hexToRGBComponents(hex);
   const isLight = average(components) > 128;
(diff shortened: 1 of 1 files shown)


AprilSylph:
> because due to the fact that for browser support reasons we're not doing color manipulation anywhere close to correctly

Are we being limited by old browsers here, or Safari?

marcustyphoon:
Old browsers. Well... I don't have a flawless entirely browser-native version of this logic at all, but we could certainly do _somewhat better than the current one_ using color-mix, put it behind a supports query, and just do no intervention here on our oldest supported browsers.

**edit:** Er... by color-mix do I mean color from? I don't actually remember.

**edit 2:** Okay, looking back at #1735: yeah, maybe we couldn't do better than the current one, actually. Correction: for **vendoring-colorjs-just-for-this-is-probably-not-worth-it reasons**, we're not doing color manipulation anywhere close to correctly.

AprilSylph:
> (should we bump the threshold to e.g. 48 so this one is affected? I'm torn.)

I think it's fine; I can read it (assuming the blog colours haven't changed since this PR was posted; the blog URL did). Those with sight issues that need text to meet WCAG standards for contrast probably shouldn't be using this feature at all.

marcustyphoon:
Oh yeah, that one's not noticeably tweaked by the Tumblr automatic contrast increaser, I think, so it makes sense not to touch it.

(I'm honestly thinking about PRing a vendored colorjs version of this, at this point. Not really worth the effort, though.)

PR #2051, opened 2026-01-06
feat: add VS Code extension recommendations

AprilSylph:
### Description

Another little thing in my "hey let's make the repository more friendly to new contributors" crusade.

These are all the VS Code extensions I use when developing addons. Yes, all six of them.

I think being able to trivially configure your workspace to match that of the author-maintainer is useful, no?

This is committed via `git add --force` rather than modifying our gitignore file, so other changes to files in the `.vscode/` directory remain untracked by default. This allows individuals to customise their workspace for this repository without their `settings.json` getting picked up by git and unintentionally pushed as part of a PR.

### Testing steps

1. Open this branch in VS Code
2. Go to Extensions &rarr; "Search Extensions in Marketplace"
3. Search for `@recommended`
    - **Expected result**: Six items appear under "Workspace Recommendations": ESLint, EditorConfig for VS Code, GitHub Theme, GitHub Actions, Git Graph, GitHub Pull Requests

(I assume that a fresh VS Code install would, upon starting up with this file committed, prompt to install these if they're not already installed; I can make no assumptions about your install, so I don't know if you'll be prompted!)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,10 @@
+{
+    "recommendations": [
+        "dbaeumer.vscode-eslint",
+        "editorconfig.editorconfig",
+        "github.github-vscode-theme",
+        "github.vscode-github-actions",
+        "github.vscode-pull-request-github",
+        "mhutchie.git-graph"
+    ]
+}


marcustyphoon:
I don't know offhand what `git add --force` does; I can look for documentation, but do you have any offhand?

Alternatively, does it make more sense to unignore this specific file in .gitignore?

AprilSylph:
`git add --force` just ignores gitignore. If anyone makes any changes to this file in order to customise their workspace, because the file is still gitignored, their changes won't be tracked by git unless they too use `git add --force`. It's useful for this exact scenario where we want to provide a default but don't care about the file beyond that.

marcustyphoon:
Hm. Unless github desktop is doing unusual checkout and stage commands, I don't think that works.

AprilSylph:
> Hm. Unless github desktop is doing unusual checkout and stage commands, I don't think that works.

How do you mean? Do the file changes not show up when checking out this branch?

marcustyphoon:
Didn't test that much yet, but I believe repo file changes overwrite existing content and local changes to files are tracked. That's been my experience in the past: .gitignore doesn't matter once a file is tracked. You can work around it on an individual basis (skip-worktree, or something?) but I never found a way to do it from the repo side.

AprilSylph:
Ah! That's... interesting, and unexpected. Let me play with it.

marcustyphoon:
I could definitely be wrong! But I'm trying to imagine how it would work if I am wrong. Presumably what you want is "a clone to an empty folder includes the file but a clone to a—

...okay, I'm not sure I actually know how to finish that sentence :D

AprilSylph:
What I was going for was a committed version of doing `git update-index --assume-unchanged .vscode/extensions.json`, but that's not how `.gitignore` works at all.

AprilSylph:
Hm, closing this was silly. I made a bad assumption about the details (#2052 deserves to stay closed) but this file in particular shouldn't need to be touched by individual developers unless they're planning to propose new recommended extensions anyway.

AprilSylph:
No changes to gitignore needed to continue to ignore other files in the `.vscode/` directory.

PR #2054, opened 2026-01-09
feat(UI): modernise feature summary styling

AprilSylph:
### Description

- relates to #1641 

The way that feature icons are aligned when feature descriptions become multi-line is bugging me. I think they should always be visually anchored to the feature name, rather than centred in the feature row. I think the same should apply to the help button and feature toggle, too.

And as soon as I started touching these styles... more things started to bug me. The line height of feature titles/descriptions is just a little bit over half the height of feature icons, by .1px, but the whole thing looks too cramped if that .1px is shaved off. We're using `margin` where flexbox `column-gap` would work identically. We're using `ch` units for a bunch of measurements instead of whole pixel values.

So, I took some pointers from Tumblr's design system and made feature summaries into a weird variant of the "user row" component, which resolves all my gripes and, in my opinion, looks slightly better overall:

Before | After
-|-
Screen Shot 2026-01-09 at 10 54 57 | Screen Shot 2026-01-09 at 10 55 26

This does mean that I'll need to redo the product screenshots again... oh well.

### Testing steps

1. Load the modified addon
2. Open the control panel via the browser toolbar
    - **Expected result**: All features have bigger, slightly more rounded icons
    - **Expected result**: Features with long descriptions still have their description text wrap onto new lines
    - **Expected result**: All feature icons and buttons are aligned to the top of the summary, regardless of description wrapping

### DIFF
--- a/src/action/components/xkit-feature/index.css
+++ b/src/action/components/xkit-feature/index.css
@@ -7,8 +7,8 @@ summary {
   flex-direction: row;
   justify-content: flex-start;
   align-items: flex-start;
-  column-gap: 1ch;
-  padding: 1ch;
+  column-gap: 8px;
+  padding: 8px;
 
   cursor: pointer;
 }
@@ -54,7 +54,7 @@ summary:focus {
   display: flex;
   flex-direction: row;
   align-items: center;
-  column-gap: 1ch;
+  column-gap: 8px;
   height: 36px;
   margin-left: auto;
 }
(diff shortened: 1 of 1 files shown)


AprilSylph:
Maybe worth mentioning that another thing I want to do to feature summaries in the future is make the toggle buttons a lot bigger, at least on touch devices—right now, I find them to be rather small on my phone.

marcustyphoon (approved):

PR #2055, opened 2026-01-12
feat: add `CheckboxPreference` component

AprilSylph:
### Description

- relates to #1831

Creates a `CheckboxPreferenceElement` class and `CheckboxPreference` shorthand function, for rendering checkbox-type preferences.

This just creates the component and doesn't implement it anywhere. I would prefer for all preferences to be switched to using slotted components at the same time, since that will require not-insignificant lift to make the control panel's search/highlight functionality work on componentised preferences.

Having unused files is far less messy than having mid-migration spaghetti code, in my opinion.

### Testing steps

1. Pull this branch locally
2. Apply this patch to the branch: [diff.patch](https://github.com/user-attachments/files/24562094/diff.patch)
3. Load the addon from `src/`
4. Open a Tumblr tab
5. Open the XKit control panel via the browser toolbar action
6. Enable a feature that has checkbox-type preferences, e.g. Classic Footer
    - **Expected result**: Checkbox preferences render an `<input type="checkbox">` and a `<label>`
    - **Expected result**: Checkbox preferences render the correct checked state for their `<input>`
    - **Expected result**: Checkbox preferences render the correct text content for their `<label>`
7. Toggle a checkbox-type preference on
    - **Expected result**: The preference change takes effect on the Tumblr tab immediately
8. Toggle a checkbox-type preference off
    - **Expected result**: The preference change takes effect on the Tumblr tab immediately
9. Open the XKit control panel in a new tab
10. Inspect the accessibility properties of checkbox-type preferences
    - **Expected result**: Every `checkbox` element in the accessibility tree has a `name` matching the visible label
    - **Expected result**: Every `label` element in the accessibility tree has the correct `label for` relation

### DIFF
--- a/src/action/components/checkbox-preference/index.css
+++ b/src/action/components/checkbox-preference/index.css
@@ -1,0 +1,6 @@
+li {
+  display: flex;
+  align-items: flex-start;
+  column-gap: 8px;
+  padding: 4px;
+}
--- a/src/action/components/checkbox-preference/index.js
+++ b/src/action/components/checkbox-preference/index.js
@@ -1,0 +1,61 @@
+import { CustomElement, fetchStyleSheets } from '../index.js';
+
+const localName = 'checkbox-preference';
+
+const templateDocument = new DOMParser().parseFromString(`
+  <template id="${localName}">
+    <li>
+      <input id="checkbox" type="checkbox">
+      <label for="checkbox"></label>
+    </li>
+  </template>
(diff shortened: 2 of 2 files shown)


marcustyphoon:
I'm soooo tempted to alt-click approve without checking this out :D

marcustyphoon (approved):
Works as described!

PR #2057, opened 2026-01-13
feat: Hide XKit-added elements when extension is disabled in Firefox

### LINKED ISSUE #1994: Hide XKit-added elements when extension is disabled in Firefox
### Concept

As discussed in #1993, we can use CSS specificity trickery to cause many XKit-added UI elements to disappear when the user disables the extension in Firefox, rather than just becoming unusable and/or badly styled.

marcustyphoon:
### Description

This creates an attribute that can be applied to DOM elements to make them vanish when the user disables the extension in Firefox, rather than staying around with invalidated click handlers and (if the styles in question are in static stylesheets) missing styling, and applies it to:

- XKit Rewritten control buttons
- XKit Rewritten meatballs menu buttons
- XKit Rewritten post action buttons (i.e. Quick Tags in the post form)
- XKit Rewritten sidebar items
- Quote Replies action buttons
- the Scroll to Bottom button
- Sidebar Blogs narrow sidebar icons

I put this in interface.js sort of arbitrarily. It could also be its own util, could be in dom.js (though I'll be honest, I don't want to put a side effect in dom.js), could be "built-in" (`index.css`?), or whatever.

Resolves #1994.

Prior art: #1993, #451. (For the record, I still don't personally believe making the extension function entirely correctly around update cycles in Firefox is achievable and thus that they should be hidden from the user.)

### Testing steps

todo: list this

Be sure to at least smoke test this in both Firefox and Chromium.

### DIFF
--- a/src/content_scripts/interface.css
+++ b/src/content_scripts/interface.css
@@ -1,5 +1,5 @@
 @layer xkit-hide-on-disable {
-  [data-xkit-hide-on-disable] {
+  :is([data-xkit-hide-on-disable], #extra#specificity) {
     display: revert-layer !important;
   }
 }


marcustyphoon:
Ah, nope, !important indeed doesn't work with revert-layer, like I thought.

marcustyphoon:
Yep, can't get it to work via specificity/cascade. Will have to use e.g. `display: var(--flex-if-xkit-enabled)` or `display: var(--none-if-xkit-disabled, flex)`. (Putting our styles in a cascade layer would work, except that we need to override redpop styles, like, all the time, so we really don't want to do that.)

Issue #2063, opened 2026-01-14
use `lit` for web component creation


AprilSylph:
### Description

Building custom boilerplate stuff is fun, but not extremely useful. Knowledge of our custom boilerplate is impossible for a new contributor to have coming into the project, and learning how to use it does not give anyone any transferable skills.

We should just be using [Lit](https://lit.dev/) instead.

AprilSylph:
Wow, this package really is not meant to be vendored, huh.

marcustyphoon:
Told you.

Yeah, you're kind of supposed to use import maps, but we can't (yet).

PR #2064, opened 2026-01-14
build: vendor `lit-core` via jsDelivr

AprilSylph:
### Description

- relates to #2063 

This feels really, really stupid. Lit doesn't distribute bundles by any means other than jsDelivr, and the actual `lit` package makes a hard assumption that if you are using npm, you are using your own bundler.

Am I going to start using Webpack to build a browser extension because of this? Nope.

There is a `lit-dist` package that someone set up to try to mirror the bundles to npm... but it doesn't actually auto-update like it claims to. See https://github.com/lit/dist/tags vs https://github.com/fserb/lit-dist/tags.

I don't really want to add a dependency to the `package.json` that isn't actually used—if anything, I would consider it a hindrance to have in the project since then the `node_modules` versions of its exports get picked up by code autocomplete tools, which would not actually work. So... no Dependabot update notifications for this. Oh well, it'll get picked up when we update other vendored dependencies.

Oh yeah, and I rearranged `dev/copy-libs.sh` for fun.

**I ❤️ ALPHABETISATION**

### Testing steps

1. Pull this branch locally
2. Apply this patch: [visible_alt_text.patch](https://github.com/user-attachments/files/24612506/visible_alt_text.patch)
3. Load the modified addon
4. Enable AccessKit &rarr; "Move alt text to captions below images"
5. Find a post with alt text
    - **Expected result**: The AccessKit option works as usual
    - **Expected result**: There are no weird errors or warnings reported in the console


AprilSylph:
If you find a better way to vendor this library—before, or after, this PR is merged—please let me know.

marcustyphoon:
We could submodule lit/dist; is that better?

marcustyphoon:
Ooh, dependabot has a gitsubmodule option. So, yes, that.

Have fun making the checkout documentation for ultra new developers not kind of annoying.

marcustyphoon:
Aside: we don't need to chmod libs, right?

marcustyphoon:
> Am I going to start using Webpack to build a browser extension because of this? Nope.

(I do think there's a difference between that, and having an esbuild oneliner in the vendor-libs script. I imagine many would consider gitignoring the vendored folder and running the vendor script on postinstall; then it's very transparent and you could even do a single lib package without having a "build step" you ever interact with. We should still avoid that for review reasons, but it's not like it fundamentally changes much.)

marcustyphoon:
Not actually related to the vendoring in this PR, but note that the linked patch does not function correctly; toggling the feature off and on breaks it.

AprilSylph:
...no, that's relevant, if we can't use Lit for `dom()` et al. replacement then its value goes down drastically, and we would need to reassess it versus using Preact or continuing to just use our custom boilerplate and utils.

AprilSylph:
Yeah, no, I was expecting `html`/`render` to not have any of this bullshit. The concept of Lit excited me because at first it looked like it was, at its core, _just_ web components with the boilerplate necessary to use React-like syntax, but if it's so far away from vanilla JS that you can't mix the two together cleanly then it's completely worthless to me here.

marcustyphoon:
I mean, I think using Lit for the entire preference UI is entirely reasonable; that's what it's for: React-like dynamism that's encapsulated by web components. Reusing its templating in content scripts where we cannot use web components is a nice consistency play, and we could extract the resulting elements by rendering to a div and pulling out the first child element, but it certainly would be throwing away some functionality. There are a few cases in the content script where it could make sense to make a "lit-html component that's not a registered web component but uses the 'UI is a function of data/vdom' functionality" like the entire XKit Rewritten sidebar, the Quick Tags modal, or the Show Originals controls UI, if that's a model that you enjoy/prefer. But yes, you kind of pick between imperative vanilla js and declarative "this bit is vdom now" and it 's not necessarily trivial to make that distinction clear in the code, I guess. But maybe there is some fundamental logic to that ("`render(html())` is for component-like logic, `vanillaHtml()` or whatever is a oneshot dom template", idk).

I can't remember if preact also uses a comment to store stuff for diffing purposes or not. Certainly we know we can hook htm up to the current dom utils to produce vanilla elements if it does and we don't want that; most of the above paragraph would apply if so (i.e. whynotboth, besides that we have fairly few dynamic UI elements and maybe that's more confusing).

PR #2065, opened 2026-01-14
chore: configure VS Code auto imports

AprilSylph:
### Description

- relates to #2064

> I don't really want to add a dependency to the package.json that isn't actually used—if anything, I would consider it a hindrance to have in the project since then the node_modules versions of its exports get picked up by code autocomplete tools, which would not actually work.

Importing directly from `node_modules` doesn't work anyway, stupid.

Stop allowing VS Code to suggest unusable exports.

### Testing steps

1. Pull this branch locally
2. Open any feature's `index.js` file in VS Code
3. Start typing "Sortable" somewhere in the file
    - **Expected result**: `../../lib/sortable.esm.js` is suggested
    - **Expected result**: 🆕 The bare `sortablejs` package is _not_ suggested
    - **Expected result**: 🆕 `SortablePlugin` is _not_ suggested

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "javascript.preferences.autoImportFileExcludePatterns": ["node_modules/**/*"],
+  "javascript.preferences.importModuleSpecifier": "relative",
+  "javascript.preferences.importModuleSpecifierEnding": "js",
+  "javascript.suggest.paths": true
+}
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,4 @@
+{
+  "include": ["src/**/*"],
+  "exclude": ["node_modules"]
+}


marcustyphoon:
Hm. Not sure if we should commit this as `settings.json.example` and have developers copy it if desired, as is sometimes used for repo env files.

AprilSylph:
I did think about that, but this project is small potatoes enough that I don't think anyone is heavily personalising their VS Code settings for it. I would rather be beginner-friendly and very firmly tell as many IDEs as possible "hey, don't import things we can't use in this project".

marcustyphoon:
Breaks dev scripts that are js files (#2061 and, I guess, a hypothetical esbuild script for vendoring), but I don't know if there's any way we can make this only apply to src/ to avoid that.

**edit:** Ha: make those all .ts, maybe that would work :P

AprilSylph:
@marcustyphoon Breaks how?

```
april@KW6GP44CTV ~/Desktop/XKit-Rewritten aprilsylph/vs-code-auto-import $ git merge pr/marcustyphoon/2061
Merge made by the 'ort' strategy.
 dev/check-outdated-css-keys.js | 37 +++++++++++++++++++++++++++++++++++++
 package-lock.json              | 96 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------
 package.json                   |  1 +
 3 files changed, 122 insertions(+), 12 deletions(-)
 create mode 100755 dev/check-outdated-css-keys.js
april@KW6GP44CTV ~/Desktop/XKit-Rewritten aprilsylph/vs-code-auto-import $ dev/check-outdated-css-keys.js
keyToCss is called with outdated/invalid key arguments:
- reblogTrailWrapper
- hasPermalink
- withTrailItemPermalink
- isReblogWithAddedContent
- heightRestrictorExpandButtonWrapper
- contentWarningCover
- expandTagsButtonWrapper
- notifications
- videoCTA
```

marcustyphoon:
Oh, just that you have to manually write the import of e.g. `glob` in a dev script file like that, because in that case one indeed does want to do a bare import from node_modules. Not a big deal.

(I should have written "breaks in.")

AprilSylph:
I think if you have the proficiency to be writing a dev script, you likely either don't need autoimports or know how to turn them back on temporarily...?

Or the dev scripts could be `.ts`, yeah.

marcustyphoon:
Oh, this overwrote my settings.json, huh. Fortunately I don't think I had anything important in there.

PR #2066, opened 2026-01-15
refactor: use slim build of jQuery

AprilSylph:
### Description

Yesterday, I learnt that jQuery has a "slim" build, which excludes the [ajax](https://api.jquery.com/category/ajax/) and [effects](https://api.jquery.com/category/effects/) modules.

We don't use jQuery for fetches or animations, so we can use it!

The filesize reduction in `src/` is only 16.9KB, and the build size of the addon is not reduced significantly, remaining at 2.2MB. But this is a trivial change to make and to verify, so why not?

### Testing steps

1. Load the modified addon
2. Enable AccessKit &rarr; "Move alt text to captions below images"
  (or another feature/option which adds elements that are removed via jQuery on cleanup)
3. Open a Tumblr tab and find the XKit-added elements on the page
4. Disable the feature/option
    - **Expected result**: The XKit-added elements are removed as usual
5. Re-enable the feature/option
    - **Expected result**: The elements are added back to the page without issue


AprilSylph:
Oh hey, per-commit builds outside of PRs comes in handy for addon build size comparison:
* https://github.com/AprilSylph/XKit-Rewritten/actions/runs/20989054455: 2.23 MB
* https://github.com/AprilSylph/XKit-Rewritten/actions/runs/21038243953: 2.22 MB

PR #2068, opened 2026-01-16
chore: thoroughly type-annotate `XKitFeatureElement` and `XKitFeature()`

AprilSylph:
### Description

You know what? There's actually nothing wrong with our web components. Let's invest in them more.

This PR brings `XKitFeatureElement`'s typing up to standard with the as-of-yet unused `CheckboxPreferenceElement` component. This is a lot of JSdoc! But damn, it's pretty nice. VS Code even understands that these two destructured constants are both strings now:

https://github.com/AprilSylph/XKit-Rewritten/blob/31de8ce642580de584720a989ee593a5bcd9c08f/src/action/components/xkit-feature/index.js#L173-L174

Oh, and static members are now actually `static` now. I assume this saves some memory?

### Testing steps

1. Pull this branch locally
2. Open `src/action/components/xkit-feature/index.js` in VS Code
3. Hover over everything that's been type-annotated
    - **Expected result**: Nothing is... wrong? There's still a lot missing, but when VS Code has the opportunity to infer something, it should infer correctly, at least.
4. Load the modified addon in the browser
5. Open the control panel
    - **Expected result**: All features render without issue
    - **Expected result**: All features' preferences render without issue
    - **Expected result**: Features can be enabled/disabled without issue
    - **Expected result**: Instant-save preference types (e.g. checkbox, select) can be changed without issue
    - **Expected result**: Debounced-save preference types (e.g. text, textarea) can be changed without issue

### DIFF
--- a/src/action/components/xkit-feature/index.js
+++ b/src/action/components/xkit-feature/index.js
@@ -35,54 +35,49 @@ const adoptedStyleSheets = await fetchStyleSheets([
 ].map(import.meta.resolve));
 
 /**
- * @typedef CheckboxPreferenceObject
+ * @typedef BasePreference
+ * @property {string} label Label displayed to the user to describe the preference.
+ * @property {string} [inherit] The storage key to inherit the value of, if the preference has not been set.
+ */
+
+/**
+ * @typedef Checkbox
  * @property {"checkbox"} type Type of preference.
- * @property {string} label Label displayed to the user to describe the preference.
  * @property {boolean} default Default value of the preference to display to the user.
- * @property {string} [inherit] The storage key to inherit the value of, if the preference has not been set.
  */
 
 /**
- * @typedef TextPreferenceObject
+ * @typedef Text
  * @property {"text"} type Type of preference.
- * @property {string} label Label displayed to the user to describe the preference.
  * @property {string} default Default value of the preference to display to the user.
- * @property {string} [inherit] The storage key to inherit the value of, if the preference has not been set.
  */
 
 /**
- * @typedef TextAreaPreferenceObject
+ * @typedef TextArea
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I think I've already used an "I ain't reading all that/I'm happy for you/or sorry that happened" image macro in a github comment, haven't I, darn :D

Yeah, no. Anyway. I feel like the `Preferences` types should probably be in another file, but we can always move them if we have an actual reason to.

(Also, the April in the back of my head from back in the day is like "surely if you name your variables correctly and structure your code carefully, you only need minimal prose comments" :D)

AprilSylph:
> I feel like the `Preferences` types should probably be in another file, but we can always move them if we have an actual reason to.

Yeah, I feel like that's a separate unit of work just to decide where that would go! I'm not sure extension store reviewers would particularly enjoy seeing a new file that consists entirely of JSdoc in the addon ZIP.

marcustyphoon:
I... guess? It's just a .d.ts file. Doesn't seem any harder to review?

AprilSylph:
No, I mean they might say "this isn't necessary to include, please remove it" and reject the version. I'm pretty sure that's happened to me before with a markdown file.

marcustyphoon:
Ah, extension review. I won't add a second sentence to this comment besides this one because I can't think of one that would actually be a good thing to say :3

marcustyphoon:
(Oh, right—no, I did have a relevant thought, which was that I assume we could easily exclude a different file type using web-ext options, and then like, oh, I don't know if you can actually comment a typescript interface in a way that gets picked up by jsdoc. `whatever.types.js` is an option, though an odd one. Anyway.)

PR #2070, opened 2026-01-18
chore(Tweaks): Slightly improve blocked blog hiding tweak performance

marcustyphoon:
### Description

This moves a closest() function call in the "Hide all posts from blogs I've blocked" tweak out of a loop, which won't be a measurable performance increase, but seems proper.

### Testing steps

See testing steps for #1984, I guess.


marcustyphoon:
Oh, right... this code is weird and `blockedBlogNames` is only the blocked blog names in the post, which is usually length zero or one. Eh, not really necessary.

PR #2078, opened 2026-01-28
perf: use element IDs and ARIA relations to initialise preference components

AprilSylph:
### Description

- relates to #2055 
- relates to #2075 

Using `querySelector` unnecessarily makes me feel like I'm doing something wrong.

These components already take advantage of their DOM being sandboxed in terms of how they use `id` and `for` attributes. We might as well make further use of this and avoid interacting with the CSS engine when initialising the element references.

At least, I think `getElementById` is faster than `querySelector`. At the scale we're working at, this shouldn't be a perceptible change to anyone either way.

### Testing steps

1. Pull this branch locally
2. Apply this patch to enable rendering components for checkbox/text-type preferences: [preference-components.patch](https://github.com/user-attachments/files/24910313/preference-components.patch)
3. Open the XKit control panel and expand Quick Reblog's and TimeFormat's preferences
    - **Expected result**: Checkbox-type preferences render with correct labels and initial checked states
    - **Expected result**: Changes to checkbox-type preferences are reflected in the Backup tab
    - **Expected result**: Text-type preferences render with correct labels and initial values
    - **Expected result**: Changes to text-type preferences are reflected in the Backup tab

### DIFF
--- a/src/action/components/checkbox-preference/index.js
+++ b/src/action/components/checkbox-preference/index.js
@@ -38,8 +38,7 @@ class CheckboxPreferenceElement extends CustomElement {
   set value (value = false) { this.#inputElement.checked = value; }
   get value () { return this.#inputElement.checked; }
 
-  /** @type {(event: Event) => void} */
-  #onChange = () => {
+  /** @type {(event: Event) => void} */ #onChange = () => {
     const storageKey = `${this.featureName}.preferences.${this.preferenceName}`;
     const storageValue = this.#inputElement.checked;
 
--- a/src/action/components/text-preference/index.js
+++ b/src/action/components/text-preference/index.js
@@ -39,8 +39,7 @@ class TextPreferenceElement extends CustomElement {
   set value (value = '') { this.#inputElement.value = value; }
   get value () { return this.#inputElement.value; }
 
-  /** @type {(event: InputEvent) => void} */
-  #onInput = () => {
+  /** @type {(event: InputEvent) => void} */ #onInput = () => {
     const storageKey = `${this.featureName}.preferences.${this.preferenceName}`;
     const storageValue = this.#inputElement.value;
 


src/action/components/checkbox-preference/index.js
```diff
@@ -38,9 +38,11 @@ class CheckboxPreferenceElement extends CustomElement {
   set value (value = false) { this.#inputElement.checked = value; }
   get value () { return this.#inputElement.checked; }
 
-  #onChange = async ({ currentTarget }) => {
+  /** @type {(event: Event) => void} */ #onChange = () => {
```
marcustyphoon, line 41: This may be the signature this function is called with, but it's not the signature of the actual function, right? :D

marcustyphoon:
I might consider this "refactor" more than "perf"—it may be technically correct that it's fewer cycles if browsers keep a hash table of id'd elements, or something, which I certainly would if I were writing a DOM implementation. But yeah, not measurable.

src/action/components/checkbox-preference/index.js
AprilSylph: Technically it's not, no. What I really want here is the vanilla JS equivalent of `React.ChangeEventHandler<HTMLInputElement>` and this is the shortest-form way of doing so.

It does document that if you add a parameter to this function, you will get a plain `Event` and not a `UIEvent` or `InputEvent`, so I think it's worth keeping regardless.

AprilSylph: Hm, maybe I should make typedefs for all the possible signatures of the handler, and then an overlap type of those signatures...?

marcustyphoon: Interesting. `/** @type {HTMLInputElement['onchange']} */` works (but has an overly generic `this` because of how lib.dom.d.ts is written); `/** @type {Parameters<HTMLInputElement['addEventListener']<'change'>>[1]} */` seems like it should work but doesn't intellisense for me.

AprilSylph: I think I'll leave them as-is.

Worth noting that `this` in these callbacks actually refers to the component class instance, since they need to access other class instance fields anyway, so declaring their type with an explicit `this` is incorrect anyway.

PR #2079, opened 2026-01-29
fix: assign default ARIA roles to preference components

AprilSylph:
### Description

- relates to #2055 
- relates to #2075

Something I didn't realise until yesterday: custom elements containers do actually exist in the accessibility tree! They just don't have any inherent semantic meaning, so it kinda just looks like they are their children... except, in our case, they're a barrier between the preference list `<ul>` elements and the shadowRoot-hosted `<li>` elements, which breaks the semantic meaning of `<li>`.

If the custom element is trying to render a list item, it actually needs to try to _be_ a list item.

Before | After
-|-
Screenshot 2026-01-29 at 10 06 54 am | Screenshot 2026-01-29 at 10 07 21 am

This also lets us remove the `<li>` elements from preference components entirely, since we can style the container via `:host` instead.

### Testing steps

1. Pull this branch locally
2. Enable rendering preference components: [preference-components.patch](https://github.com/user-attachments/files/24933470/preference-components.patch)
3. Load the modified addon in Firefox (does Chrome have A11Y tree devtools yet?)
4. Open the XKit control panel in a new tab
5. Browse the accessibility tree for features with checkbox/text-type preferences
    - **Expected result**: Every checkbox-type preference is detected as a list item
    - **Expected result**: Every text-type preference is detected as a list item

### DIFF
--- a/src/action/components/checkbox-preference/index.css
+++ b/src/action/components/checkbox-preference/index.css
@@ -1,4 +1,4 @@
-li {
+:host {
   display: flex;
   align-items: flex-start;
   column-gap: 8px;
--- a/src/action/components/checkbox-preference/index.js
+++ b/src/action/components/checkbox-preference/index.js
@@ -4,10 +4,8 @@ const localName = 'checkbox-preference';
 
 const templateDocument = new DOMParser().parseFromString(`
   <template id="${localName}">
-    <li>
-      <input id="checkbox" type="checkbox">
-      <label for="checkbox"></label>
-    </li>
+    <input id="checkbox" type="checkbox">
+    <label for="checkbox"></label>
   </template>
 `, 'text/html');
 
--- a/src/action/components/text-preference/index.css
+++ b/src/action/components/text-preference/index.css
@@ -1,4 +1,4 @@
-li {
+:host {
   display: flex;
   justify-content: space-between;
   align-items: center;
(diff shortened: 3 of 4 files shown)


marcustyphoon:
What's our policy on blind approving PRs?

marcustyphoon:
Screenshot 2026-01-29 at 9 31 17 AM

Like this?

AprilSylph:
> What's our policy on blind approving PRs?

Does "at least do code review" work for you? 😄

marcustyphoon:
Sure thing!

Anyway, there are differences between that Chrome screenshot and the Firefox screenshot, but I haven't familiarized myself with those systems, so I don't know if they are meaningful or point to something that could or should be changed.

AprilSylph:
> Anyway, there are differences between that Chrome screenshot and the Firefox screenshot, but I haven't familiarized myself with those systems, so I don't know if they are meaningful or point to something that could or should be changed.

Interesting that Chrome basically omits the `<label>` element. It kinda makes sense - that text leaf already serves as the accessible label for its associated `<input>` element, so it's a little redundant to read it out twice - but from cursory web searches I can find nothing to imply that this is a huge upside of Chrome or something we should be emulating in other browsers somehow.

I would just chalk this up to different browsers having different opinions, which definitely isn't a new concept :laughing:

PR #2080, opened 2026-01-30
build: use unminified version of Coloris library

AprilSylph:
### Description

This morning, Mozilla scheduled a takedown of v1.2.0 for including `lib/coloris.min.js` with no source information. I have provided source links via AMO portal replies for now, but I don't know if that will suffice for this version—I assume it will, since it's still in "pending review" state, although that probably still means another week before approval.

As a preventative measure against similar actions being taken against future versions of XKit Rewritten, this PR swaps out the minified NPM version of Coloris for the source NPM version, which is unminified and includes full copyright and version information.

### Testing steps

1. Load the modified addon
2. Open the XKit control panel &rarr; Painter
    - **Expected result**: Colour-type preferences render without issue
    - **Expected result**: Colour-type preferences open the Coloris colour picker when clicked
    - **Expected result**: Colour-type preferences save their values to local storage as normal


AprilSylph:
2.22MB &rarr; 2.23MB. Not nothing, but not terrible either.

marcustyphoon:
Interesting.

Yeah, since it's shipped gzipped, there shouldn't be a big difference (naturally this is also true on the open web, but).

PR #2084, opened 2026-02-02
refactor: DOMParser use

marcustyphoon:
### Description

This:
- Uses an `html` tagged template on html template strings in preference pane components, allowing them to be syntax highlighted as html via editor plugins. (There are a number of vs code plugins that do this, none of which seem super well-maintained; notably, despite more than one having preferences allowing the user to customize what the template string can be, none of them function. C'est la vie.)
- Makes that tagged template call DOMParser.parseFromString, because... why not, I guess.

One could also decline to do the latter and put `/* html */` comments in front of the template strings to accomplish the former, which perhaps makes each component file more comprehensible by itself. Or not do this at all, of course. Would I do it, if it were a solo project? Eh, I dunno, maybe.

Note that it's possible to make a caching version of this if you wanted to inline them into component classes; tagged templates are explicitly designed to work that way. If we were using this for the content script, in many cases we would not want this. (But then we would want a version that doesn't return `Documents`, so it would be different anyway. And we don't, so it's irrelevant.)

```js
const domParseCache = new Map();
export const domParseMemo = (strings, ...values) => {
  domParseCache.has(strings) || domParseCache.set(strings, domParse(strings, ...values));
  return domParseCache.get(strings);
};
```

### Testing steps

- Apply [preference-components.patch](https://github.com/user-attachments/files/25002075/preference-components.patch)
- Confirm preference pane functionality

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,6 @@
 {
     "recommendations": [
+        "bierner.lit-html",
         "dbaeumer.vscode-eslint",
         "editorconfig.editorconfig",
         "github.github-vscode-theme",


AprilSylph:
> There are a number of vs code plugins that do [HTML template string syntax highlighting], none of which seem super well-maintained; notably, despite more than one having preferences allowing the user to customize what the template string can be, none of them function.

Does that mean the plugin you chose will actually do syntax highlighting on invalid strings? If that's the case, I wouldn't merge this—new contributors would need to somehow be told "yes, we suggest a Lit HTML syntax highlighter; yes, Lit syntax appears to be valid in your IDE; no, Lit syntax does not actually work", which I think costs more than we gain from "HTML template strings look a bit nicer".

marcustyphoon:
Probably, yeah. There's definitely an advantage to leaving the new DOMParser.parseFromString() visible and not doing this at all, as it makes it unambiguous what's going on here (aside from the fact that the entire file is, imho, unreadable).

PR #2085, opened 2026-02-02
feat(Quick Tags): Disable UI while processing; adjust UI progress timing 

marcustyphoon:
### Description

This:
- Disables the Quick Tags UI after it has been submitted while it's processing. ~~(I considered putting a display: contents fieldset around the button elements or something, to be able to elegantly use the disabled attrubute, but I'm not sure if that actually better).~~
- Fades the Quick Tags UI after it has been submitted while it's processing. (Improvements welcome.)
- Changes the timing of the Quick Tags popup removal when clicking a bundle button so that it's after success/failure.
- Changes the timing of the Quick Tags success/failure notification so that it's simultaneous with the post appearing updated. (If the edit succeeds but updatePostOnPage fails, there will be an error modal covering the page anyway, so the change in notification behavior is probably not that consequential.)
- Changes the behavior on failure of a text input submission to not clear the text form, so the user can try again if it was just a network thing.

### Testing steps

- Add a tag to a post using the Quick Tags text input box. Confirm that the popup greys out, then after a moment, the post becomes updated and the success notification appears.
- Add a tag to a post using a Quick Tags button. Confirm that the popup greys out, then after a moment the post becomes updated, the success notification appears, and the popup vanishes.
- Disable networking. Add a tag to a post using the Quick Tags text input box. Confirm that an error modal appears, and after it's dismissed, the text input box still has the user-entered text.
- Disable networking. Add a tag to a post using a Quick Tags button. Confirm that an error modal appears.

### DIFF
--- a/src/features/quick_tags/index.js
+++ b/src/features/quick_tags/index.js
@@ -147,23 +147,24 @@ const addTagsToPost = async function ({ postElement, inputTags = [] }) {
 
   tags.push(...tagsToAdd);
 
+  let displayText = '';
+
   if (isNpfCompatible(postData)) {
-    const { response: { displayText } } = await apiFetch(`/v2/blog/${uuid}/posts/${postId}`, {
+    const { response } = await apiFetch(`/v2/blog/${uuid}/posts/${postId}`, {
       method: 'PUT',
       body: {
         ...createEditRequestBody(postData),
         tags: tags.join(','),
       },
     });
-
-    await updatePostOnPage(postElement, ['tags', 'tagsV2']);
-    notify(displayText);
+    ({ displayText } = response);
   } else {
     await megaEdit([postId], { mode: 'add', tags: tagsToAdd });
-
-    await updatePostOnPage(postElement, ['tags', 'tagsV2']);
-    notify(`Edited legacy post on ${blogName}`);
+    displayText = `Edited legacy post on ${blogName}`;
   }
+
+  await updatePostOnPage(postElement, ['tags', 'tagsV2']);
+  notify(displayText);
 };
 
 const processFormSubmit = function ({ currentTarget }) {


src/features/quick_tags/index.js
```diff
@@ -156,31 +156,43 @@ const addTagsToPost = async function ({ postElement, inputTags = [] }) {
...
+  addTagsToPost({ postElement, inputTags })
+    .then(() => currentTarget.reset())
+    .catch(showErrorModal)
+    .finally(() => popupElement.classList.remove('processing'));
 };
 
 const processBundleClick = function ({ target }) {
   if (target.tagName !== 'BUTTON') { return; }
+  if (popupElement.classList.contains('processing')) { return; }
+  popupElement.classList.add('processing');
```
AprilSylph: This wouldn't be exposed to the accessibility API; really, all the buttons should be wrapped in a `<fieldset>` which we then toggle the `disabled` attribute of.

marcustyphoon: Yeah, that is indeed the right way to do it.

src/features/quick_tags/index.css
marcustyphoon, line 84: Real quick... does anyone want to explain to me what this does?

marcustyphoon:
Would it be crazy to make the popup itself a fieldset element?

src/features/quick_tags/index.js
```diff
@@ -156,31 +156,41 @@ const addTagsToPost = async function ({ postElement, inputTags = [] }) {
       },
     });
 
+    await updatePostOnPage(postElement, ['tags', 'tagsV2']);
     notify(displayText);
   } else {
     await megaEdit([postId], { mode: 'add', tags: tagsToAdd });
+
+    await updatePostOnPage(postElement, ['tags', 'tagsV2']);
```
AprilSylph: **Nit**: I think this logic can be DRY'd some more...

```
let displayText = '';

if (isNpfCompatible(...)) {
  const { response } = await apiFetch(...);
  ({ displayText } = response);
} else {
  await megaEdit(...);
  displayText = `Edited legacy post on ${blogname}`;
}

await updatePostOnPage(...);
notify(displayText);
```

PR #2087, opened 2026-02-03
chore: Apply eslint rules to full repository

marcustyphoon:
### Description

This applies the eslint package scripts to repository files outside of src/ (we don't have to do this) and formats a dev script that's written in js.

### Testing steps

Run `npm run posttest` and confirm that nothing is reported.


AprilSylph:
> This applies the eslint package scripts to repository files outside of src/ (we don't have to do this)

Yeah, would prefer not to do this. Everything outside of `src/` is vestigial.

PR #2090, opened 2026-02-04
chore(UI): Standardise "colour" spelling in UI

marcustyphoon:
### Description

This uses "colour" as the spelling in visible UI, and makes sure that "color" can also be used as a search keyword.

This doesn't fix preference highlighting, which could be done by adding a relatedTerms field to individual options (I did not do this because you are refactoring them) or by building a dictionary of alternative search strings into the search function itself (a lot of code for not a lot of value).

### Testing steps

Confirm that "color" and "colour" searches cause the same features to be shown in the configuration UI.


marcustyphoon:
> Standardise

Ooooof course :D

PR #2092, opened 2026-02-05
style: alphabetise all imports

AprilSylph:
### Description

This is the result of setting `'import-x/order': ['warn', { alphabetize: { order: 'asc', caseInsensitive: true } }]` and running our autofix package script.

This does not actually commit any rule changes; figuring out if we should actually enforce this is a different matter to actually doing the formatting, which is already a lot of lines changed!

### Testing steps

CI passing and this being a net-neutral change is probably good enough as validation.
I've smoke-tested the addon and nothing jumps out to me as broken.


marcustyphoon:
...I should not have done all of those merge commits, huh :D

marcustyphoon (approved):

PR #2093, opened 2026-02-05
chore: enable `codeActionsOnSave` in VS Code

AprilSylph:
### Description

- relates to #2092 

Because typing `npm run autofix` is more effort than hitting <kbd>command</kbd>+<kbd>S</kbd>
Because I love to lower the barrier to entry for new contributors, even if it's just by an inch
Because why wouldn't anyone want this???

### Testing steps

1. Pull this branch locally
2. Open a `.js` file somewhere in the `src/` directory
3. Mess up its formatting somehow
4. Save the file
    - **Expected result**: Your formatting mistakes are fixed before they even become part of the working tree!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,6 @@
 {
+  "editor.codeActionsOnSave": { "source.fixAll.eslint": "explicit" },
+  "eslint.validate": ["javascript"],
   "javascript.preferences.autoImportFileExcludePatterns": ["node_modules/**/*"],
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I'll probably have to figure out how to override this locally for doing merge commit maintenance—it [sounds like there is a way](https://laurens.io/blog/changing-vs-code-settings-locally-when-its-checked-into-git/), so I think this is fine.

Issue #1697, opened 2025-01-11
Tweaks: UI error on blazed posts with "Hide mini follow button" (minor)


marcustyphoon:
### Platform

macOS 14.7.1 arm64

### Browser

Chrome 131.0.6778.205

### Addon version

v1.0.5

### Details

The "Hide mini-follow buttons on posts" tweak messes up a special follow button on blazed posts. This blue pill should be larger and say "follow":

<img width="559" src="https://github.com/user-attachments/assets/6b22b1c7-cc0b-44e2-943c-c19a722f8eeb">

`:not(${keyToCss('postMeatballsContainer')} *)` is probably the easiest way to fix this.

PR #2097, opened 2026-02-10
chore: Remove old follow button selector

marcustyphoon:
### Description

The "Hide follow buttons on posts" tweak originally hid elements using the mapped selector key `followButton`. This is also used for various other follow buttons around the site (11 classes), as came up in https://github.com/AprilSylph/XKit-Rewritten/issues/1697, and we then added entirely different code in https://github.com/AprilSylph/XKit-Rewritten/pull/1733 to handle the new header, which does not use this.

This removes the now-unused (that I know of) code, assuming that the old version of the header is long-gone (and that there aren't any tools in the tumblr-modification ecosystem to bring it back; I don't personally know of any).

### Testing steps

- Confirm that the "Hide follow buttons on posts" tweak works as expected.
- Check my logic—is there any other kind of follow button this tweak is supposed to hide? (Community posts have a different header; this didn't work on blazed posts before anyway.)

### DIFF
--- a/src/features/tweaks/hide_mini_follow.js
+++ b/src/features/tweaks/hide_mini_follow.js
@@ -5,9 +5,7 @@ import { pageModifications } from '../../utils/mutations.js';
 
 const hiddenAttribute = 'data-tweaks-hide-mini-follow-hidden';
 
-export const styleElement = buildStyle(`
-article ${keyToCss('followButton')}:not(${keyToCss('postMeatballsContainer')} *), [${hiddenAttribute}] { display: none; }
-`);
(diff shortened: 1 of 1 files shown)


AprilSylph:
> * is there any other kind of follow button this tweak is supposed to hide?

Yes, the follow buttons on Like notes:

Tweak disabled | Tweak enabled (259990ce73fb82b6c224f009ce7b54887236d63f)
-|-
Screen Shot 2026-02-10 at 09 10 47 | Screen Shot 2026-02-10 at 09 11 52

marcustyphoon:
Oh, huh. I had never even noticed those!

PR #2098, opened 2026-02-10
refactor(UI): use web components to render preferences

AprilSylph:
### Description

- relates to #1831
- relates to #2055
- relates to #2075
- relates to #2086
- relates to #2089
- relates to #2091
- relates to #2096

We now have a web component for every type of preference in the control panel!

Let's start using them, and clean up the old code.

-|-|-|-
Screen Shot 2026-02-10 at 12 38 33 | Screen Shot 2026-02-10 at 12 38 46 | Screen Shot 2026-02-10 at 12 45 25 | Screen Shot 2026-02-10 at 12 45 41

This will be further refactored in a later PR to move the preference-rendering code out of `XKitFeatureElement`, to utilise element slotting, and to re-simplify the now rather messy search code.

### Testing steps

1. Load the modified addon
2. Make sure every type of preference renders correctly
  (visual layout; correct default state for unset preferences; correct default state for saved preferences)
    - Checkbox
    - Color
    - Iframe
    - Select
    - Text
    - Textarea
3. Make sure every type of preference saves changes made to it
    - Checkbox
    - Color
    - Select
    - Text
    - Textarea
4. Search for "hi" (search term of less than three characters)
    - **Expected result**: Tweaks is counted as a search result, due to its preferences that contain the word "Hide"
    - **Expected result**: Because you are only using two characters in the search, no preferences are highlighted
5. Search for "media" (search term of three or more characters)
    - **Expected result**: AccessKit, CleanFeed, and Panorama are all search results due to their preferences
    - **Expected result**: The preferences "Download paused GIFs:", "Hide media from:", and "Expand media content in wide posts" are all highlighted
      <sub>**Info**: "Download paused GIFs:" is highlighted because of its options, rather than its label!</sub>

### DIFF
--- a/src/action/components/checkbox-preference/index.js
+++ b/src/action/components/checkbox-preference/index.js
@@ -25,7 +25,7 @@ class CheckboxPreferenceElement extends CustomElement {
     super(templateDocument, adoptedStyleSheets);
 
     this.#inputElement = this.shadowRoot.getElementById('checkbox');
-    this.#labelElement = this.#inputElement.labels[0];
+    this.#labelElement = this.shadowRoot.querySelector('label[for="checkbox"]');
   }
 
   /** @param {string} label Label displayed to the user to describe the preference. */
--- a/src/action/components/color-preference/index.js
+++ b/src/action/components/color-preference/index.js
@@ -29,7 +29,7 @@ class ColorPreferenceElement extends CustomElement {
     super(templateDocument, adoptedStyleSheets);
 
     this.#inputElement = this.shadowRoot.getElementById('color');
-    this.#labelElement = this.#inputElement.labels[0];
+    this.#labelElement = this.shadowRoot.querySelector('label[for="color"]');
 
     Coloris.init(); // eslint-disable-line import-x/no-named-as-default-member
   }
--- a/src/action/components/select-preference/index.js
+++ b/src/action/components/select-preference/index.js
@@ -33,7 +33,7 @@ class SelectPreferenceElement extends CustomElement {
     super(templateDocument, adoptedStyleSheets);
 
     this.#selectElement = this.shadowRoot.getElementById('select');
-    this.#labelElement = this.#selectElement.labels[0];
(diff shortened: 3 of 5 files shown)


marcustyphoon:
...Hm.

marcustyphoon:
https://github.com/AprilSylph/XKit-Rewritten/blob/42da0ea3f315facec0020568ed7b5b68fee273ae/src/action/components/checkbox-preference/index.js#L27-L28

In Safari (tested in desktop Safari 26.0.1), `this.#inputElement.labels` is an empty nodelist when queried at this moment, for some reason (so this.#labelElement doesn't get set and the code crashes later on). `queueMicrotask(() => console.log(this.#inputElement.labels));` does result in the correctly populated nodelist, so it's not like the property isn't ever functional, but Safari seems to not populate it in document fragments that aren't rendered in a document, or something.

Minimal repro:

```js
(() => {
  const templateDocument = new DOMParser().parseFromString(`
    <template>
      <input id="checkbox" type="checkbox">
      <label for="checkbox"></label>
    </template>
  `, 'text/html');

  const content = templateDocument.querySelector('template').content.cloneNode(true);

  const inputElement = content.getElementById('checkbox');

  console.log('inputElement', inputElement);
  console.log('inputElement.labels', inputElement.labels); // empty

  setTimeout(() => {
    console.log('inputElement', inputElement);
    console.log('inputElement.labels', inputElement.labels); // still empty
  }, 0);
})();

```

AprilSylph:
I was already considering giving up on Safari support. It's not like I'm ever realistically going to get around to it.

AprilSylph:
Okay fine I'll initialise stuff with `querySelector` after all.

marcustyphoon:
I mean, we don't currently publish for it! I don't think there's necessarily anything wrong with pushing something that would have to be changed if we did decide to make it a supported platform—sure, it's not much work to change this to support Safari, but that could be an argument in either direction (easy to do now, but easy to do in the hypothetical future, too).

PR #2111, opened 2026-02-13
refactor(UI): render all `XKitFeature` children in `renderFeatures()`

AprilSylph:
### Description

- relates to #2107

Makes `XKitFeatureElement` stop rendering any of its own children. This is how things are supposed to work, I think...

I actually didn't find keeping the preference JSdoc types easy to fit into `render_features.js`, so I have simply removed them. We can always bring them back later if we need them.

~~If you know how to get `eslint-plugin-jsdoc` to respect types from all files in the project and not just the current one, let me know! I would love to make a `types/` directory.~~ Actually no, that's silly, at that point we should just do TypeScript checking, shouldn't we?

### Testing steps

1. Load the modified addon
3. Expand some features that have preferences
    - **Expected result**: The preferences render without any visual issues
    - **Expected result**: The "No preferences available" message is not present
    - **Expected result**: Changes to preferences are reflected in local storage as usual
4. Expand some features that don't have preferences
    - **Expected result**: You are shown the usual "No preferences available" message
5. Use the search tool to find a preference
    - **Expected result**: Features with matching preferences are included in search results
    - **Expected result**: Preferences with matching text are highlighted

### DIFF
--- a/src/action/components/xkit-feature/index.js
+++ b/src/action/components/xkit-feature/index.js
@@ -36,53 +36,6 @@ const adoptedStyleSheets = await fetchStyleSheets([
   './index.css',
 ].map(import.meta.resolve));
 
-/**
- * @typedef BasePreference
- * @property {string} label Label displayed to the user to describe the preference.
- * @property {string} [inherit] The storage key to inherit the value of, if the preference has not been set.
- */
-
-/**
- * @typedef Checkbox
- * @property {"checkbox"} type Type of preference.
- * @property {boolean} default Default value of the preference to display to the user.
- */
-
-/**
- * @typedef Text
- * @property {"text"} type Type of preference.
- * @property {string} default Default value of the preference to display to the user.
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I thought briefly about this refactor, but yeah, I don't think it's an improvement.

```js
const Preference = {
  checkbox: CheckboxPreference,
  color: ColorPreference,
  iframe: IframePreference,
  select: SelectPreference,
  text: TextPreference,
  textarea: TextAreaPreference
}[type];

if (Preference) {
  preferenceElements.push(Preference({ featureName, preferenceName, label, options, value, src }));
} else {
  console.error()
}
```

marcustyphoon:
Does, though, feel like a bit of a  `featureElement.append(...Object.entries(preferences).map(([preferenceName, preference]) => {/* */}));` situation. The error state can return `''`!

PR #2112, opened 2026-02-16
feat(Quote Replies): Add action buttons to replies in post notes (visible)

marcustyphoon:
### Description

Adapted from #1662.

This implements a button at the bottom of certain replies in the post footer, allowing the user to activate Quote Replies directly from that location if they would have been able to from the relevant item in the activity notifications menu/page.

Note that the situations referenced by the three types of verbiage ("replied to your post", "replied to you in a post" and "mentioned you on a post") used in Quote Replies have overlap. It's common for a threaded reply to one of your comments on one of your posts to be all three of these! When using Quote Replies from the activity notifications page/popup, the verbiage used is of course the one from the notification tab you're on (the same notification will appear in multiple tabs). In the notes in a post footer, any of them could accurately be said to apply.

This PR (see `determineNoteReplyType`) prioritizes them in the order `"replied to you in a post" > "replied to your post" > "mentioned you on a post"`, which... I _think_ makes sense? Most specific to most general.

Hm—I guess we could add a "replied to you on your post" verbiage if we wanted to.

(aside: my understanding is that at some point, some of what we're talking about here may become officially referred to as "comments," not "replies." I have no idea what the story is on that so for now I'll keep saying "replies" I guess)

### Testing steps

Confirm that "quote this reply" buttons are added to replies in the post notes, and function when clicked, in these cases:

- [x] A user adds a root-level reply to your post. (Verbiage: "replied to your post")
- [x] A user adds a threaded reply to someone else's/their own reply, on your post. (Verbiage: "replied to your post")
- [x] A user adds a threaded reply to your reply, on your post. (Verbiage: "replied to you in a post"; see above)
- [x] A user adds a threaded reply to your reply, on someone else's/their own post. (Verbiage: "replied to you in a post")
- [x] A user mentions you in a root-level reply to someone else's/their own post. (Verbiage: "mentioned you on a post")
- [x] A user mentions you in a threaded reply to someone else's/their own reply, on someone else's/their own post. (Verbiage: "mentioned you on a post")

Check that the correct mentioned/replied-to blog is where the post draft is created.

also:

- [x] Confirm that "quote this reply" buttons are added to replies both in the "Viewing an individual reply thread. See all replies" permalink view linked from activity, and from the normal view of that post with all replies visible.
- [x] Confirm that "quote this reply" buttons are added to replies outside of the blog view (e.g. on the dash)

edit: 

- [x] Use the "quote this reply" button on a reply in the post notes when viewing a post that is not a reblog. Confirm that the hyperlink on the post summary on a produced post draft points to the correct root post.
- Use the "quote this reply" button on a reply in the post notes when viewing a post that **is** a reblog. Note that the hyperlink on the post summary on a produced post draft points to the reblog you were viewing. (The summary text is, I assume, generally the same on a reblog as on its root post? Right?)


marcustyphoon:
To do: see if `notePropsObjects` (which currently simultaneously grabs the data for replies and, if they have them, their parents) can and should be simplified and called on threaded replies' parents.

I can't remember if there was a technical reason I did this, or if it was just done this way because the meatballs-menu version was processing the meatballs menu itself and, pre-#1935, couldn't traverse the DOM tree to access the reply parent. Probably the latter, honestly.

marcustyphoon:
Thoughts for @AprilSylph consideration:

- Is this definitely better than putting something in the meatballs menu?
- What is unambiguous terminology to distinguish between activity notifications and post notes? I was doing a second pass on renaming functions for clarity and realized all of a sudden that I was carefully using "activity" to mean activity notifications, but the post notes area literally has the aria label "Post Activity," so officially that clears up precisely nothing.
- It winds up being more trouble than it's worth here, but: general thoughts on splitting a feature into multiple js files within its folder?
- Ideas for an updated feature description?
- Should these buttons be blue or match their companions?
- If one clicks the Quote Replies button while viewing a reblog, should the summary link in the resulting quoted post be to that reblog, or the root post?

src/features/quote_replies/index.js
```diff
@@ -166,38 +199,91 @@ const processReply = async ({ type, timestamp, targetPostId, targetTumblelogName
...
+const quoteNotificationReply = async (tumblelogName, notificationProps) => {
+  const data = notificationProps.type === 'generic'
+    ? await createGenericNotificationReplyData(notificationProps)
+    : await createNotificationReplyData(notificationProps);
+
+  openPostDraft(tumblelogName, data);
+};
+
+const determineNoteReplyType = ({ noteProps, parentNoteProps }) => {
+  if (userBlogNames.includes(noteProps.note.blogName)) return false;
```
marcustyphoon: Interesting note from testing: this line excludes quote replying yourself, which probably makes sense in general... but a) maybe it doesn't (something something group blogs), and b) that's inconsistent with the behavior on activity items, which totally does let you quote reply yourself!

(comment copied from previous PR)

marcustyphoon:
~~- [ ] Handle trail footer items correctly~~

**edit:** Well, they were removed, so there would currently be no way to test code that handled them correctly (for if they ever return).

marcustyphoon:
Hmmm... should these buttons be blue only when hovered...

PR #2113, opened 2026-02-16
ci: create new release notes script; use GitHub mentions

AprilSylph:
### Description

What if every release in this repo actually had the "contributors" footer, rather than just bare author names on the listed commits?

This completely redoes the release notes script/action to use the GitHub API to achieve just that, with commit author names swapped out for mentions. Since bot accounts cannot be mentioned, they are linkified manually instead.

I feel like this could conceivably provide extra motivation for someone capable to become a contributor...?
I know it would at least feed _my_ ego more to see my face†, not just my name, on a release's credits‡.

† <sup>Well, it's not my physical face, but it's the face I choose to represent myself online, so... analogy still stands!</sup>
‡ <sup>I already get this on every release here because I'm the one doing the releases, so this isn't a factor in my wanting to do this, to be clear!</sup>

This is primarily designed to be run via GitHub Actions, but it can still be ran locally. Without configuration, it makes unauthenticated API requests to the base repo to fetch commit details.

---

Example output:

```md
- 752a523d14e70f8d6d15502e41b6da3cc1f5a7b6 @marcustyphoon refactor: Store memoized data on dom objects (#1946)
- 6842642264f6a1c67a8244ce0500711ec96311a2 @AprilSylph feat: add `IframePreference` component (#2096)
- 42da0ea3f315facec0020568ed7b5b68fee273ae @marcustyphoon chore: Minor CSS performance optimization (#1420)
- e227b34da8bef4a53c42146aba0d4a36256e7a83 **[dependabot[bot]](https://github.com/apps/dependabot)** build: bump jquery from 3.7.1 to 4.0.0 (#2101)
- ef7a614e3bef863b699000d13aeb073f434cb2e2 @AprilSylph refactor(UI): use web components to render preferences (#2098)
- b17b2008eff44ae1e473615428e3b5db4b5c0550 @AprilSylph refactor(Tweaks): "Restore links to individual posts in the post header" (#2100)
- 36272a504ce932221817e9f36022b17965d978c6 @AprilSylph refactor(UI): use slots to render preferences (#2107)
- edd301a6b84f372973967e9dc5f5fd58d77f190c @AprilSylph Revert "fix(Themed Posts): Temporarily disable reblog trail theming with new Tumblr A/B Test" (#2108)
- 836a6648633017036ea5653f6cfbf8cb8c88104e @AprilSylph refactor(UI): render all `XKitFeature` children in `renderFeatures()` (#2111)
- 15cc897c73bf761c46218c19e80c154ecb91c352 @marcustyphoon fix(Panorama): Unexpanded media mode bugs (#2110)
- dda530baeb2b839e539b14b08e80bb94dc0149cd @AprilSylph fix(UI): unbreak feature toggling
- f89f177cacef7ca4332c24cc5afabd38647392d3 @marcustyphoon chore: Moderate CSS performance optimizations (#1644)
```

---

Example rendered output:

Screen Shot 2026-02-16 at 12 34 00

---

### Testing steps

Look at https://github.com/AprilSylph/XKit-Rewritten/actions/runs/22062698062.

In all seriousness, I expect this to be heavier on code review than my other PRs, since I barely know what I'm doing here.

### DIFF
--- a/dev/release-notes.js
+++ b/dev/release-notes.js
@@ -16,7 +16,7 @@ try {
     `${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/commits?per_page=100`,
     { headers: { Accept: 'application/vnd.github+json', 'X-GitHub-Api-Version': '2022-11-28', ...GITHUB_TOKEN && { Authorization: `Bearer ${GITHUB_TOKEN}` } } },
   );
-  if (!response.ok) { throw new Error(`🛑 Error ${response.status}: ${response.statusText}`); }
+  if (!response.ok) { throw new Error(`🛑 HTTP ${response.status} (${response.statusText})`); }
 
   const commits = new Map();
   for (const commit of await response.json()) {


dev/release-notes.js
```diff
@@ -0,0 +1,67 @@
...
+
+const {
+  GITHUB_API_URL = 'https://api.github.com',
+  GITHUB_REPOSITORY = 'AprilSylph/XKit-Rewritten',
+  GITHUB_TOKEN,
+} = process.env;
+
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+exec('git log $(git describe --tags --abbrev=0)..HEAD --reverse --pretty --format="%H" --follow src/', async (error, stdout, stderr) => {
```
AprilSylph: The `$()` syntax here probably breaks on Windows... right? Should we add a platform check?

(&uarr; girl with no convenient access to a Windows machine)

marcustyphoon: Indeed it does. I'm sure one could call powershell, but like you said, platform check.

Could also just do two commands:

```js
import { execSync } from 'node:child_process';

try {
  const latestTag = execSync('git describe --tags --abbrev=0', { encoding: 'utf8' }).trim();
  const newRefs = execSync(`git log ${latestTag}..HEAD --reverse --pretty --format="%H" --follow src/`, { encoding: 'utf8' });
```

```js
import childProcess from 'node:child_process';
import { promisify } from 'node:util';
const exec = promisify(childProcess.exec);

try {
  const latestTag = await exec('git describe --tags --abbrev=0').then(({ stdout }) => stdout.trim());
  const newRefs = await exec(`git log ${latestTag}..HEAD --reverse --pretty --format="%H" --follow src/`).then(({ stdout }) => stdout);
  ```

File: dev/release-notes.js [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  let fatalError = null;
+
+  console.log('```md');
+  for (const ref of stdout.split('\n')) {
+    if (!ref) { continue; }
+
+    const [response] = await Promise.all([
+      fetch(`${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/commits/${ref}`, {
```
marcustyphoon: I imagine it's possible to paginate the commits endpoint (`while (!newRefs.isSubsetOf(allRefs))`, if you like Sets) to save API requests, but that might not be worth the effort.

(It's pretty easy with octokit, but I assume we want to be able to run this locally and don't really want to install octokit. I've done `const releases = await github.paginate(github.rest.repos.listReleases, { owner: context.repo.owner, repo: context.repo.repo, });` in an `actions/github-script@v8` inline script in a github action before. ~~We could do `import { Octokit } from "https://esm.sh/octokit"`, but that's pretty ridiculous.~~ *ah right that's a deno thing*)

marcustyphoon: Eh, I mean.

```js
#!/usr/bin/env node

import { execSync } from 'node:child_process';
import { Octokit } from 'octokit';

try {
  const latestTag = execSync('git describe --tags --abbrev=0', { encoding: 'utf8' }).trim();
  const newRefs = execSync(`git log ${latestTag}..HEAD --reverse --pretty --format="%H" --follow src/`, { encoding: 'utf8' }).split('\n').filter(Boolean);

  const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
  const commits = await octokit.paginate(
    octokit.rest.repos.listCommits,
    {
      owner: 'AprilSylph',
      repo: 'XKit-Rewritten',
      headers: { 'X-GitHub-Api-Version': '2022-11-28' },
    },
    (response, done) => {
      if (!response.data.some(({ sha }) => newRefs.includes(sha))) { done(); }
      return response.data;
    },
  );

  console.log('```md');
  for (const ref of newRefs) {
    const { author, commit } = commits.find(({ sha }) => sha === ref) ?? {};
    if (!commit) { throw new Error(`Fatal: API found no commit info for ref ${ref}`); }

    console.log(`- ${
      ref
    } ${
      author.type === 'User'
        ? `@${author.login}`
        : `**[${author.login}](${author.html_url})**`
    } ${
      commit.message.includes('\n')
        ? commit.message.slice(0, commit.message.indexOf('\n'))
        : commit.message
    }`);
  }
  console.log('```');
} catch (error) {
  console.error(error);
}
```

AprilSylph: > but I assume we want to be able to run this locally

Yup. I like being able to test things prior to pushing! I like it a lot!

marcustyphoon: Because it was fun:

```js
#!/usr/bin/env node

import { execSync } from 'node:child_process';

const {
  GITHUB_API_URL = 'https://api.github.com',
  GITHUB_REPOSITORY = 'AprilSylph/XKit-Rewritten',
  GITHUB_TOKEN,
} = process.env;

const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));

try {
  const latestTag = execSync('git describe --tags --abbrev=0', { encoding: 'utf8' }).trim();
  const newRefs = new Set(execSync(`git log ${latestTag}..HEAD --reverse --pretty --format="%H" --follow src/`, { encoding: 'utf8' }).split('\n').filter(Boolean));

  const commits = new Map();

  for (let page = 0; page < 10; page++) {
    const response = await fetch(`${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/commits?page=${page}`, {
      headers: {
        Accept: 'application/vnd.github+json',
        ...GITHUB_TOKEN && { Authorization: `Bearer ${GITHUB_TOKEN}` },
        'X-GitHub-Api-Version': '2022-11-28',
      },
    });
    for (const commit of await response.json()) {
      commits.set(commit.sha, commit);
    }
    if (newRefs.isSubsetOf(commits)) break;
    await sleep(GITHUB_TOKEN ? 0 : 1000);
  }
  if (!newRefs.isSubsetOf(commits)) { throw new Error('Could not find commit info for one or more commits.'); }

  console.log('```md');
  for (const ref of newRefs) {
    const { author, commit } = commits.get(ref);

    console.log(`- ${
      ref
    } ${
      author.type === 'User'
        ? `@${author.login}`
        : `**[${author.login}](${author.html_url})**`
    } ${
      commit.message.includes('\n')
        ? commit.message.slice(0, commit.message.indexOf('\n'))
        : commit.message
    }`);
  }
  console.log('```');
} catch (error) {
  console.error(error);
}
```

dev/release-notes.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const response = await fetch(
+    `${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/commits?per_page=100`,
+    { headers: { Accept: 'application/vnd.github+json', 'X-GitHub-Api-Version': '2022-11-28', ...GITHUB_TOKEN && { Authorization: `Bearer ${GITHUB_TOKEN}` } } },
+  );
+  if (!response.ok) { throw new Error(`Error ${response.status}: ${response.statusText}`); }
+
+  for (const commit of await response.json()) {
+    commits.set(commit.sha, commit);
+  }
```
marcustyphoon, line 24: I was going to suggest this be a map constructor oneliner now, if we're only doing the one fetch, but `const commits = new Map((await response.json()).map(commit => [commit.sha, commit]));` actually looks really stupid, so never mind.

Anyway, yeah, I guess if we put more than 100 commits in a PR, we have bigger problems.

AprilSylph: :thinking: :question: 

```js
const commitItems = await response.json();
const commitEntries = commitItems.map(commit => [commit.sha, commit]);
const commits = new Map(commitEntries);
```

dev/release-notes.js
```diff
@@ -0,0 +1,47 @@
...
+      ref
+    } ${
+      author.type === 'User'
+        ? `@${author.login}`
+        : `**[${author.login}](${author.html_url})**`
+    } ${
+      commit.message.includes('\n')
+        ? commit.message.slice(0, commit.message.indexOf('\n'))
+        : commit.message
+    }`);
```
marcustyphoon, line 41:
```suggestion
    const authorLink = author.type === 'User'
      ? `@${author.login}`
      : `**[${author.login}](${author.html_url})**`;

    console.log(`- ${ref} ${authorLink} ${commit.message.replace(/\n.*/s, '')}`);
```

Entirely optional.

dev/release-notes.js
marcustyphoon: Saves a line with a single close brace, exciting :D Whichever you think is more clear—if anything, the current code might show what the map keys are slightly better? (If keeping the current code I might consider moving the map constructor call right before the loop.)

AprilSylph: Hmm, yeah no, current code with moved map constructor is probably clearest.

dev/release-notes.js
AprilSylph: I always prefer fewer regular expressions where possible :v:

dev/release-notes.js
marcustyphoon: I kept going, "okay, there must be something analogous to `object.groupby` for this... okay, no, why would there be, actually."

dev/release-notes.js
marcustyphoon: Yep. And `.split('\n')[0]` really not much of an improvement.

dev/release-notes.js
```diff
@@ -0,0 +1,46 @@
...
+    `${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/commits?per_page=100`,
+    { headers: { Accept: 'application/vnd.github+json', 'X-GitHub-Api-Version': '2022-11-28', ...GITHUB_TOKEN && { Authorization: `Bearer ${GITHUB_TOKEN}` } } },
+  );
+  if (!response.ok) { throw new Error(`Error ${response.status}: ${response.statusText}`); }
+
+  const commits = new Map();
+  for (const commit of await response.json()) {
+    commits.set(commit.sha, commit);
+  }
+  if (!refs.isSubsetOf(commits)) { console.warn('⚠️ Could not find commit info for one or more commits!'); }
```
marcustyphoon, line 25: There are definitely ways to make this not crash so much if you run it with a commit to src/ un-pushed—`const { author, commit } = commits.get(ref) ?? {}; if (!commit) { console.warn(/* */); break; }` / `console.log('```md');` before/at the beginning of the try, `console.log('```');` in a finally block / `for (const ref of refs.intersection(commits) {`.

AprilSylph: Doesn't this version output a single non-blocking warning at maximum, before the constructed markdown block? My intention was to do that... output regardless, but with a disclaimer of potential incompleteness.

marcustyphoon: `const { author, commit } = commits.get(ref);` will throw, printing an error (and omitting the end of the markdown block).

marcustyphoon:
Two slightly interesting notes from testing a #1843 merge:
- console.warn writes to stderr, so `dev/release-notes.js >> $GITHUB_STEP_SUMMARY` will not include the "Could not find commit info" message.
- Set.prototype.intersection() uses the order of the smaller set; this means that in the case where we create more than 100 commits between releases, the commits will be listed in reverse-chronological order.

PR #2114, opened 2026-02-18
build(UI): migrate from `normalize.css` to `modern-normalize`

AprilSylph:
### Description

- See https://github.com/sindresorhus/modern-normalize?tab=readme-ov-file#readme

We don't target older browsers, which `normalize.css` does; we do target new browser versions, which `normalize.css` doesn't.

The new hotness appears to be `modern-normalize`; that's what [Tailwind](https://tailwindcss.com/docs/preflight) uses, anyway.
The opinionated defaults are nice, too.

I've removed all the CSS that I know is redundant with this package.
I couldn't find anything in need of updating for compatibility, so no visual changes are expected.

Sadly, it looks like this NPM package doesn't include the minified version, same as the original; maybe one day we can rewrite `dev/copy-libs.sh` to a node script which fetches the correct minified version from jsDelivr?

### Testing steps

1. Load the modified addon in every browser you have access to
    - **Expected result**: Nothing in the control panel looks broken
    - **Expected result**: Nothing in the control panel has visual discrepancies between different browsers

### DIFF
--- a/src/action/backup.css
+++ b/src/action/backup.css
@@ -33,7 +33,6 @@
 
 #backup pre,
 #backup textarea {
-  box-sizing: border-box;
   max-height: calc(10em + 2ch);
   overflow: auto;
   overflow-wrap: break-word;
--- a/src/action/components/select-preference/index.css
+++ b/src/action/components/select-preference/index.css
@@ -13,7 +13,6 @@ label {
 select {
   flex-shrink: 1;
 
-  box-sizing: border-box;
   padding: 4px;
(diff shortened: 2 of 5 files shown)


marcustyphoon:
Well, I do see a regression in Firefox 147.0.3, 140, and 132—but not 128—where the quick tags iframe is too tall (its html element is taller than its body element, apparently). But it happens on 1.2.1 without this PR (and without the preference component refactor), so it's totally unrelated :D

PR #2115, opened 2026-02-18
fix(UI): resize PostBlock/Quick Tags preference frames more reliably

AprilSylph:
### Description

- relates to https://github.com/AprilSylph/XKit-Rewritten/pull/2114#issuecomment-3920233329

It looks like resizing frames by matching `document.documentElement.scrollHeight` is not working too good.
Let's use `document.documentElement.getBoundingRect()` instead.

Also includes some minor fixes/tweaks for the two iframe-type preferences we have; see commits.

### Testing steps

1. Load the modified addon in both Chrome and Firefox
2. Open the PostBlock UI with no blocked posts
    - **Expected result**: On both browsers, there is no huge empty space below the "0 blocked posts" heading
3. Open the Quick Tags UI with no tag bundles
    - **Expected result**: On both browsers, there is no huge empty space below the "No tag bundles are defined" message
4. Create some tag bundles
    - **Expected result**: On both browsers, the Quick Tags preference frame grows to accommodate the added tag bundles; you should not have to scroll in the frame itself
5. Delete some tag bundles
    - **Expected result**: On both browsers, the Quick Tags preference frame shrinks to fit the remaining tag bundles
6. Block a ton of posts on Firefox
    - **Expected result**: When next opened, the PostBlock preference frame grows to accommodate all the blocked posts; you should not have to scroll in the frame itself
7. Unblock a ton of posts on Firefox
    - **Expected result**: The PostBlock preference frame shrinks to fit the remaining blocked posts

### DIFF
--- a/src/features/quick_tags/options/index.css
+++ b/src/features/quick_tags/options/index.css
@@ -22,8 +22,6 @@ html {
 }
 
 body {
-  padding: 3px;
-
   background-color: rgb(var(--white));
   color: rgb(var(--black));
   font-family: "Helvetica Neue", "HelveticaNeue", Helvetica, Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


AprilSylph:
Tested on Firefox Nightly v149.0a1 on my Fairphone 4, behaves exactly the same as desktop Firefox.

PR #2116, opened 2026-02-19
feat(UI): use system fonts; only use relative font sizes

AprilSylph:
### Description

This needs testing across different operating systems.

- [x] macOS
- [x] SteamOS
- [ ] Android
- [ ] Ubuntu
- [ ] Windows

Right now this uses `font: menu;` with relative font sizes; it might be better to use `font: message-box;` with absolute font sizes.

### Testing steps

TBD

### DIFF
--- a/src/action/components/xkit-feature/index.css
+++ b/src/action/components/xkit-feature/index.css
@@ -11,6 +11,7 @@ summary {
   padding: 8px;
 
   font: caption;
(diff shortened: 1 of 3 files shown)


AprilSylph:
```css
body {
  font: menu;
  line-height: inherit;
}

nav a {
  font: caption;
  line-height: inherit;
}

...
```

PR #2117, opened 2026-02-20
feat(UI): use normalised `font-family` and `line-height`

AprilSylph:
### Description

- depends on #2114
- relates to #2116

-|-|-|-
Screen Shot 2026-02-23 at 09 35 55 | Screen Shot 2026-02-23 at 09 37 50 | Screen Shot 2026-02-24 at 09 25 26 | Screen Shot 2026-02-24 at 09 25 30

https://firefoxux.github.io/firefox-desktop-components/?path=/docs/docs-typography--docs appealed to my desire to achieve perfect visual harmony in software:

> In-content pages and the browser chrome follow different type scales due to the chrome relying on operating systems' font sizing, while in-content pages follow the type scale set by the design system.
> 
> We set `font: message-box` at the root of `common-shared.css` and `global.css` stylesheets so that both in-content and the chrome can have access to operating system font families.
> 
> We also don't specify line height units and rely on the default.

> The chrome solely relies on `font` declarations (it also relies on `font: menu` for panels) so that it can inherit the operating system font family and sizing in order for it to feel like it is part of the user's operating system. Keep in mind that font sizes and families vary between macOS, Windows, and Linux. Moreover, you will only see a difference between `font: message-box` and `font: menu` font sizes on macOS.

I tried using the [system font values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font#system_font_values) in the control panel—see related PR—but during my testing across all of my devices, I came to one conclusion: they're just too inconsistent. (The font size for `caption` on Firefox Android is tiny—not fit for purpose at all.)

And, of course, we can already access the system font _family_ via `font-family: system-ui`, which `modern-normalize` sets anyway. All we have to do is take our own `font-family` declarations away.

Elements that use a monospace font have also been updated to match this declaration from `modern-normalize`:

https://github.com/AprilSylph/XKit-Rewritten/blob/2281461a632cd5d201176f0035ec228883d2b8e6/src/lib/modern-normalize.css#L62-L72

Line height declarations that attempted to match Acorn have also been removed, in light of the disclaimer.

### Testing steps

1. Load the modified addon
2. Open the control panel
    - **Expected result**: The font is different. Nothing looks broken.

### DIFF
--- a/src/features/postblock/options/index.css
+++ b/src/features/postblock/options/index.css
@@ -34,16 +34,30 @@ header {
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
-  margin-bottom: 1em;
+  column-gap: 6px;
+  flex-wrap: wrap;
+}
+
+#posts-blocked-count,
+#posts-blocked-info {
+  margin: 0;
 }
 
 #posts-blocked-count {
-  font-weight: bold;
+  font-size: 15px;
+}
+
+#posts-blocked-info {
+  font-size: 11px;
 }
 
 #blocked-posts {
   padding: 0;
-  margin: 0;
+  margin: 1em 0 0;
+}
+
+#blocked-posts:empty {
+  display: none;
 }
 
 .blocked-post {
--- a/src/features/postblock/options/index.html
+++ b/src/features/postblock/options/index.html
@@ -20,8 +20,8 @@
     </template>
     <main>
       <header>
-        <span id="posts-blocked-count"></span>
-        <small>Unblocking a post will immediately unhide it.</small>
+        <h3 id="posts-blocked-count">Loading…</h3>
+        <p id="posts-blocked-info">Unblocking a post will immediately unhide it.</p>
       </header>
       <ul id="blocked-posts"></ul>
     </main>


AprilSylph:
The font size change might be contentious here; maybe that should be left for when more of the control panel has been given the Acorn-matching treatment? Up to you.

marcustyphoon:
Hmm.

AprilSylph:
Oh, that `body { font-size: 100%; }` was there for Chrome reasons, wasn't it...

marcustyphoon:
Looks reasonable in:

- [x] chrome, macos
- [x] firefox, macos
- [x] safari, macos
- [x] safari, ipad simulator
- [x] safari, iphone simulator

marcustyphoon:
- [x] chrome, windows
- [ ] firefox, windows (I have permissions issues running web-ext apparently)
- [ ] firefox, android

marcustyphoon:
Hmm, quick tags tag bundles look kind of huge, no? Particularly when compared to feature headers.

marcustyphoon:
I _would_ say that—I dunno, the 15px base font size seems distinctly larger than all of the browser UI elements, which I do find odd? If matching the Firefox UX guidelines, anyway. Does that seem to be the case for you as well?

edit: Although it does match `about:addons`, besides that the feature title and description are too small then.

AprilSylph:
Yeah... I'm going to revert the font size change here. I have my own observations, but let's save that discussion for a future PR.

marcustyphoon (changes requested):
Postblock is, unfortunately, still wrapping in my Firefox.

AprilSylph:
o.o it's not wrapping in mine! You're definitely testing the latest commit?

Screen Shot 2026-02-24 at 09 37 02

AprilSylph:
Maybe I'll have to change the font sizes _first_ and then revisit this afterwards...?

marcustyphoon:
Aha—it's because I have a 3-digit number of blocked posts. "Testing with your real installation" win!

AprilSylph:
Hm, how many digits of blocked posts do you need to break it on `master`? :D

marcustyphoon:
Approximately 5, which, uh, I hope nobody gets to :D

I think we could consider this an unrelated issue and file it separately, if you want—there will always be a number that wraps it, of course. But given that the old number is unrealistically high, and is now much lower, I think one could equally validly consider it a minor regression.

marcustyphoon (approved):

PR #2123, opened 2026-02-25
feat(UI): use dark theme checkboxes when appropriate

AprilSylph:
### Description

Adds [`<meta name="color-scheme">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme) to the UI's HTML.

Currently, this has the sole effect of making checkboxes match the dark theme, if enabled:

Before | After
-|-
Screen Shot 2026-02-25 at 09 55 16 | Screen Shot 2026-02-25 at 09 55 43

### Testing steps

1. Load the modified addon onto a browser that uses a light theme
    - **Expected result**: No visual changes can be observed
3. Load the modified addon onto a browser that uses a dark theme
4. Open the XKit control panel
    - **Expected result**: Checkboxes are dark now!
    - **Expected result**: No deleterious visual changes can be observed

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta name="color-scheme" content="light dark">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>XKit Control Panel</title>
     <link rel="icon" href="../icons/128.png" type="image/png">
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Also, the "add bundle" button in Quick Tags!

marcustyphoon:
The chromium dark mode UI elements have a warmer color temp than the rest of the UI, but like, what are we gonna do about it:

AprilSylph:
> what are we gonna do about it

We'll get to that later... :smirk:

PR #2124, opened 2026-02-25
feat(Mirror Posts): remove `archive.today` option

### LINKED ISSUE #2120: Mirror Posts: Remove `archive.today` option
### Description

See https://arstechnica.com/tech-policy/2026/02/wikipedia-bans-archive-today-after-site-executed-ddos-and-altered-web-captures/.

AprilSylph:
### Description

- resolves #2120 

Removes the `archive.today` option from Mirror Posts; makes the meatball button open the Wayback Machine without any interstitial; adds an info modal that is shown if the post can't be archived due to the blog or community's privacy settings.

This probably should have been two or three commits, but the file is so small to begin with...

### Testing steps

1. Load the modified addon
2. Enable Mirror Posts
3. Try to mirror a post made by a public blog without a custom theme
    - **Expected result**: `https://web.archive.org/save` is opened, prefilled with the post's `www.tumblr.com` address
4. Try to mirror a post made by a public blog with a custom theme
    - **Expected result**: `https://web.archive.org/save` is opened, prefilled with the post's `*.tumblr.com` address
5. Try to mirror a post made by a logged-in only blog
    - **Expected result**: You are shown the "Can’t mirror this post!" info modal
6. Try to mirror a post made by a password-protected blog (you have to be an admin or member to do this)
    - **Expected result**: You are shown the "Can’t mirror this post!" info modal
7. Try to mirror a post made in a public community
    - **Expected result**: `https://web.archive.org/save` is opened, prefilled with the post's address
8. Try to mirror a post made in a private community
    - **Expected result**: You are shown the "Can’t mirror this post!" info modal

### DIFF
--- a/src/features/mirror_posts/index.js
+++ b/src/features/mirror_posts/index.js
@@ -1,6 +1,7 @@
 import { button, form, input } from '../../utils/dom.js';
 import { registerMeatballItem, unregisterMeatballItem } from '../../utils/meatballs.js';
-import { showModal, modalCompleteButton } from '../../utils/modals.js';
+import { showModal, modalCompleteButton, showErrorModal } from '../../utils/modals.js';
+import { apiFetch } from '../../utils/tumblr_helpers.js';
 
 const meatballButtonId = 'mirror_posts';
 const meatballButtonLabel = 'Mirror this post';
@@ -16,20 +17,41 @@ const waybackMachineForm = form({
 ]);
 document.documentElement.append(waybackMachineForm);
 
-/** @type {(event: PointerEvent) => void} */
-function onButtonClicked (event) {
-  const { blog, community, postUrl } = event.currentTarget.__timelineObjectData;
-  const isLoggedInOnly = !!blog.isHiddenFromBlogNetwork;
-  const isPrivateBlog = !!blog.isPasswordProtected && !community;
-  const isPrivateCommunity = !!community && community.visibility !== 'public';
+const modalProps = { title: 'Can’t mirror this post!', buttons: [modalCompleteButton] };
 
-  if (isLoggedInOnly || isPrivateBlog || isPrivateCommunity) {
-    showModal({
-      title: 'Can’t mirror this post!',
-      message: [`This ${community ? 'community' : 'blog'}’s privacy settings does not allow archiving.`],
-      buttons: [modalCompleteButton],
-    });
+/** @type {(event: PointerEvent) => void} */
+async function onButtonClicked ({ currentTarget }) {
+  const { blog, community, postUrl } = currentTarget.__timelineObjectData;
+
+  const ownerIsLoggedInOnlyBlog = !community && !!blog.isHiddenFromBlogNetwork;
+  const ownerIsPasswordProtectedBlog = !community && !!blog.isPasswordProtected;
+  const ownerIsPrivateCommunity = !!community && community.visibility !== 'public';
+
+  if (ownerIsLoggedInOnlyBlog || ownerIsPasswordProtectedBlog) {
+    showModal({ ...modalProps, message: ['This blog’s privacy settings do not allow archiving.'] });
     return;
+  }
+
+  if (ownerIsPrivateCommunity) {
+    showModal({ ...modalProps, message: ['This community’s privacy settings do not allow archiving.'] });
+    return;
+  }
+
+  // `blog.isHiddenFromBlogNetwork` is only defined in the blog view; if we're somewhere else, we need to fetch it.
+  if (!community && blog.isHiddenFromBlogNetwork === undefined) {
+    try {
+      currentTarget.disabled = true;
+      const { response } = await apiFetch(`/v2/blog/${blog.uuid}/info?fields[blogs]=?is_hidden_from_blog_network`);
+      if (response.blog.isHiddenFromBlogNetwork) {
+        showModal({ ...modalProps, message: ['This blog’s privacy settings do not allow archiving.'] });
+        return;
+      }
+    } catch (exception) {
+      showErrorModal(exception);
+      return;
+    } finally {
+      currentTarget.disabled = false;
+    }
   }
 
   waybackMachineForm.elements.url_preload.value = postUrl;
--- a/src/content_scripts/meatballs.css
+++ b/src/content_scripts/meatballs.css
@@ -25,6 +25,12 @@
   background-color: rgb(var(--secondary-accent));
 }
 
+.xkit-meatball-button:disabled {
+  background-color: transparent;
+  cursor: not-allowed;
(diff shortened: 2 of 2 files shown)


marcustyphoon:
We may have to do a fetch to check `isHiddenFromBlogNetwork`—I don't always see it.

AprilSylph:
image

AprilSylph:
I was worried that `.requestSubmit()` might require [transient user activation](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/User_activation#transient_activation), but I can't see anything on MDN that points to this being the case, so a fetch is probably fine.

...then again, it wouldn't be a regression to keep the current imperfect behaviour; we were already just putting a disclaimer and letting the user try anyway. (Unless you want to count the removal of the disclaimer as a regression? Eh, yeah, I can see that.)

marcustyphoon:
`postFilter` can be async, so we can avoid this completely if we want!

edit: Er... no, that doesn't work. I dunno why I thought we could do arbitrary computation in it (putting a class on the meatball menu). Hm.

edit2: If really worried about it, we could implement a full popup ui and synchronously submit it, so that if it works the user never sees the popup but if it doesn't the feature is still fully functional, but I don't think that's really necessary.

edit3: Okay, I guess it does work: register 2 buttons; use a postfilter with a memoized fetch in both of them. That's about as much work as modifying the util or going around it by adding a pagemodifications callback targeting the added meatballs menu or something equally silly, but.

AprilSylph:
`try...catch...finally` my beloved

src/features/mirror_posts/index.js
```diff
@@ -16,22 +17,43 @@ const waybackMachineForm = form({
...
+      if (response.blog.isHiddenFromBlogNetwork) {
+        showModal({ ...modalProps, message: ['This blog’s privacy settings do not allow archiving.'] });
+        return;
+      }
+    } catch (exception) {
+      showErrorModal(exception);
+      return;
+    } finally {
+      currentTarget.disabled = false;
+    }
```
AprilSylph, line 54: Wait, this doesn't work at all, does it.

AprilSylph: ...it actually seems to...? Huh?

AprilSylph: I love JavaScript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_finally_block

marcustyphoon: I've always thought that was kind of weird, yeah. It makes sense, but still :D

marcustyphoon:
This is as far as I got with the dual buttons way in the allotted time :D

```js
import { button, form, input } from '../../utils/dom.js';
import { registerMeatballItem, unregisterMeatballItem } from '../../utils/meatballs.js';
import { memoize } from '../../utils/memoize.js';
import { showModal, modalCompleteButton } from '../../utils/modals.js';
import { apiFetch } from '../../utils/tumblr_helpers.js';

const meatballButtonId = 'mirror_posts';
const inactiveMeatballButtonId = 'mirror_posts_inactive';
const meatballButtonLabel = 'Mirror this post';

const waybackMachineForm = form({
  action: 'https://web.archive.org/save',
  hidden: true,
  method: 'post',
  target: '_blank',
}, [
  input({ name: 'url_preload', type: 'text' }),
  button({ type: 'submit' }),
]);
document.documentElement.append(waybackMachineForm);

/** @type {(event: PointerEvent) => void} */
function onButtonClicked (event) {
  const { postUrl } = event.currentTarget.__timelineObjectData;
  waybackMachineForm.elements.url_preload.value = postUrl;
  waybackMachineForm.requestSubmit();
}

/** @type {(event: PointerEvent) => void} */
function onInactiveButtonClicked (event) {
  const { community } = event.currentTarget.__timelineObjectData;
  showModal({
    title: 'Can’t mirror this post!',
    message: [`This ${community ? 'community' : 'blog'}’s privacy settings does not allow archiving.`],
    buttons: [modalCompleteButton],
  });
}

const getBlogData = memoize(blogName => apiFetch(`/v2/blog/${blogName}/info?fields[blogs]=is_hidden_from_blog_network,is_password_protected`).then(response => response.response.blog));

const postFilter = async ({ blogName, community }) => {
  const blog = await getBlogData(blogName);

  const isLoggedInOnly = !!blog.isHiddenFromBlogNetwork;
  const isPrivateBlog = !!blog.isPasswordProtected && !community;
  const isPrivateCommunity = !!community && community.visibility !== 'public';

  return !isLoggedInOnly && !isPrivateBlog && !isPrivateCommunity;
};

  registerMeatballItem({
    id: meatballButtonId,
    label: meatballButtonLabel,
    onclick: onButtonClicked,
    postFilter,
  });
  registerMeatballItem({
    id: inactiveMeatballButtonId,
    label: meatballButtonLabel,
    onclick: onInactiveButtonClicked,
    postFilter: data => postFilter(data).then(result => !result),
  });
};
export const clean = async () => {
  unregisterMeatballItem(meatballButtonId);
  unregisterMeatballItem(inactiveMeatballButtonId);
};
```

marcustyphoon:
I tried factoring this the way I find most natural; I can't say it really feels like an improvement.

```js
/** @type {(event: PointerEvent) => void} */
async function onButtonClicked ({ currentTarget }) {
  let { blog: { uuid, isPasswordProtected, isHiddenFromBlogNetwork }, community, postUrl } = currentTarget.__timelineObjectData;

  // `blog.isHiddenFromBlogNetwork` is only defined in the blog view; if we're somewhere else, we need to fetch it.
  if (!community && isPasswordProtected === undefined) {
    try {
      currentTarget.disabled = true;
      isHiddenFromBlogNetwork = await apiFetch(`/v2/blog/${uuid}/info?fields[blogs]=?is_hidden_from_blog_network`).then(({ response }) => response.blog.isHiddenFromBlogNetwork);
    } catch (exception) {
      showErrorModal(exception);
      return;
    } finally {
      currentTarget.disabled = false;
    }
  }

  if (community) {
    if (community.visibility !== 'public') {
      showModal({ ...modalProps, message: ['This community’s privacy settings do not allow archiving.'] });
      return;
    }
  } else {
    if (isPasswordProtected || isHiddenFromBlogNetwork) {
      showModal({ ...modalProps, message: ['This blog’s privacy settings do not allow archiving.'] });
      return;
    }
  }

  waybackMachineForm.elements.url_preload.value = postUrl;
  waybackMachineForm.requestSubmit();
}
```

marcustyphoon:
_Technically_ this allows the user to attempt to archive a draft post on a password-protected blog, which won't work, but that's niche enough that I'm not sure if it's worth spending time to fix.

PR #2132, opened 2026-03-03
feat: Show controls when hiding a post on its permalink page

### LINKED ISSUE #2029: Show information instead of hiding posts on permalink pages
> > Note to self: the "Hide all posts from blogs I've blocked" tweak currently hides posts unconditionally when they're viewed by themselves if they have a blocked blog in the trail, something which this PR (iirc) carefully avoids; consider making these consistent.
> 
> I think we should never be outright hiding posts if we're on their permalink page; maybe we should have a "hide this post" util which checks for that and inserts a signpost in the permalink case. (No plans to do this myself, please implement this if you want to.)

_Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/issues/677#issuecomment-3631162376_

marcustyphoon:
### Description

> > Note to self: the "Hide all posts from blogs I've blocked" tweak currently hides posts unconditionally when they're viewed by themselves if they have a blocked blog in the trail, something which this PR (iirc) carefully avoids; consider making these consistent.
> 
> I think we should never be outright hiding posts if we're on their permalink page; maybe we should have a "hide this post" util which checks for that and inserts a signpost in the permalink case. (No plans to do this myself, please implement this if you want to.)

_Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/issues/677#issuecomment-3631162376_
            
Resolves #2029.

~~Semi-blocked by #719 (these should both have consistent appearance, so whichever of them is reviewed first should be approved before working on the other).~~

Here's an experiment that moves our basic "hide this post" logic into a util and, as discussed above, implements a controls element instead of a mysterious blank space if the post is being viewed on its permalink page. UI copied from #677 (which probably can't actually use this without some more refactoring, but oh well) / #719. I'm not sure if it actually handles blog colors very well; I haven't looked at it in ages.

Features may specify the message for the modal; if they don't, posts are simply never hidden on permalink pages, again as discussed above.

If I did this correctly, it should work if a post is hidden by multiple utils, and there shouldn't be lifecycle bugs even in that case.

<details>
<summary>outdated</summary>

```js
const { hidePost, showPosts } = createPostHideFunctions({
  id: 'tweaks-hide-blocked-blogs',

  // Only applies to posts hidden by a blocked blog in the trail (see isTimelineExempt below)
  controlsOnPermalinkPage: {
    message: 'This post contains a blocked blog!',
    buttonText: 'show post anyway',
  },
});
```

</details>

Updated appearance, from #719:

```js
const { hidePost, showPosts } = createPostHideFunctions({
  id: 'tweaks-hide-blocked-blogs',

  // Only applies to posts hidden by a blocked blog in the trail (see isTimelineExempt below)
  controlsOnPermalinkPage: {
    message: 'This post contains a blocked blog.',
  },
});
```

### Testing steps

- On e.g. https://www.tumblr.com/search/evil%20science/recent?src=typed_query, confirm that No Recommended can hide posts from non-joined communities, and that they reappear when the feature is disabled.
- Block a post with PostBlock and confirm that it disappears, reappears when PostBlock is disabled, and is replaced with the new UI when viewed alone when PostBlock is enabled.
- Enable the "Hide all posts from blogs I've blocked" tweak and block a blog like tumblr.com/fandom. Confirm that a post containing content from that blog in the trail, like https://www.tumblr.com/azul-ocean24/803113081314099201, is replaced with the new UI. Confirm that a post by that blog, such as https://www.tumblr.com/fandom/803112786763907072, does not use the new UI (since you already click through "You have this Tumblr blocked, so we've hidden it from your view." to see it).
- Filter some content in a post, enable the "Hide filtered posts entirely" tweak, and confirm that the post disappears, reappears when the tweak is disabled, and is replaced with the new UI when viewed alone when the tweak is enabled.
- Use postblock on the post with a blocked blog in the trail. Filter a term in it, too, why not? Confirm that **all three** UI elements are displayed in place of the post, and that you must click all of them to make the post appear. (We could write code to make a single element appear in this case, which is probably better UX, but then we'd have to write the UI copy in a composable way, and does that seem worth it?)

- Like a post... look, test the "Hide posts that I've liked on the dashboard" and "Hide my own posts on the dashboard" tweaks, ok?


marcustyphoon:
I was having a discussion about whether "code can speak for itself" and thought it would be interesting to imagine a PR body that's basically just extracted representative/important diffs. Tried it here: I can't say it _totally_ doesn't work, but like... no, the prose is definitely more informative and more readable.

---

excerpts:

```diff
diff --git a/src/main_world/hide_posts.js b/src/main_world/hide_posts.js
new file mode 100644

+export const createPostHideFunctions = ({ id, permalinkPageControls, hideAutomatically = true }) => {
```

```diff
diff --git a/src/features/tweaks/hide_blocked_blogs.js b/src/features/tweaks/hide_blocked_blogs.js
index 7b6bcb98..ee9dc7d6 100644

-const hiddenAttribute = 'data-xkit-tweaks-hide-blocked-blogs-hidden';
-export const styleElement = buildStyle(`
-[${hiddenAttribute}] {
-  content: linear-gradient(transparent, transparent);
-  height: 0;
-}`);
+const { hidePost, showPosts } = createPostHideFunctions({
+  id: 'tweaks-hide-blocked-blogs',
+
+  // Only applies to posts hidden by a blocked blog in the trail (see isTimelineExempt below)
+  permalinkPageControls: {
+    message: 'This post contains a blocked blog!',
+    buttonText: 'show post anyway',
+  },
+});
 
 const processPosts = (postElements) => {
   filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
@@ -32,7 +36,7 @@ const processPosts = (postElements) => {
         continue;
       }
 
-      getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
+      hidePost(postElement);
       break;
     }
```

result:

marcustyphoon:
- [x] #719 merge; UI changes

marcustyphoon:
Could probably remove the button text override and just set it to "View post" unconditionally, really. This would go along with removing the ability to disable`hideManually` and use the `hiddenAttribute`, `controlledHiddenAttribute`, and `controlsAttribute` exports.

(I wrote these all in to be able to use the utility in the Mute feature, which has a mode where it hides an entire timeline with an informational message like this. It probably makes sense to export the UI at a lower level for this and handle any complexity in Mute, rather than adding a bunch of complexity to this code path, I imagine.)

src/features/tweaks/hide_filtered_posts.js
```diff
@@ -1,17 +1,18 @@
...
+  id: 'tweaks-hide-filtered-posts',
+  permalinkPageControls: {
+    message: 'This post contains filtered tags or content.',
+  },
+});
 
 const hideFilteredPosts = filteredScreens => filteredScreens
   .map(getTimelineItemWrapper)
-  .forEach(timelineItem => timelineItem.setAttribute(hiddenAttribute, ''));
+  .forEach(hidePost);
```
marcustyphoon, line 15: The `getTimelineItemWrapper` call here is actually unnecessary, though it looks kind of weird to call `hidePost` on a `filteredScreen`.

```js
const hideFilteredPosts = filteredScreens => filteredScreens.forEach(hidePost);
```

AprilSylph:
Ugh, just found a slight regression from https://github.com/AprilSylph/XKit-Rewritten/pull/719 in testing this PR: unblocking a post via the control panel doesn't visibly unblock the post on its permalink page anymore.

Before: post is hidden, then becomes visible when unblocked
After: post is collapsed; unblocking does not change this

src/utils/hide_posts.js
AprilSylph, line 1: `showPost` needs to undo everything that `hidePost` does, including the removal of permalink page controls (if present).

AprilSylph (changes requested):
I guess it's still valid to request it be fixed in this PR, since the same regression is built into the util that supersedes the PostBlock-specific functionality.

src/utils/hide_posts.js
marcustyphoon: Oh—er—oh yeah. I have no idea how I missed that.

AprilSylph:
That aside, I do want to say that the new util file is absolutely beautiful.

marcustyphoon:
A minor annoyance I've noticed while fixing that: if you have a filtered post hidden on a permalink page, you click "view post" to unhide the post, which shows you a filtered post (reasonable)... but then when you click "view post" in the filtered post it gets hidden by our util again, so you have to click one more time. So I guess... oh, this is easy to fix; check for controlledHiddenAttribute.

src/utils/hide_posts.js
```diff
@@ -0,0 +1,135 @@
...
+          // do nothing; avoid hiding single post and making permalink page look broken
+        }
+      } else {
+        getTimelineItemWrapper(postElement).setAttribute(hiddenAttribute, '');
+      }
+    },
+    showPost: postElement => {
+      getTimelineItemWrapper(postElement).removeAttribute(hiddenAttribute);
+      getTimelineItemWrapper(postElement).removeAttribute(controlledHiddenAttribute);
+      postElement.closest(timelineSelector).querySelector(`[${controlsAttribute}]`).remove();
```
AprilSylph: This needs optional chaining; throws console exceptions currently.


```suggestion
      postElement.closest(timelineSelector).querySelector(`[${controlsAttribute}]`)?.remove();
```

marcustyphoon: Whoops. I was going to put them on the QSA too, and I guess that means I should do it in more places as well.

marcustyphoon:
### Additional testing steps:

- Filter some content in a post, block it with postblock, and view it on a permalink page. Confirm that clicking through to view it takes two clicks (when it transitions from filtered view to not-filtered view, it's not hidden again; the fact that you clicked to view it persisted).
- Refresh. Unblock it from the control panel. Confirm that it's shown.
- Immediately re-block it. Confirm that it **is** hidden again (the showPost util call cleaned up everything, so it can be blocked again; the fact that you clicked to view it did not persist).

AprilSylph (commented):
I don't like that a single post hidden for multiple reasons shows two signposts at the same time, which must both be dismissed before the post becomes visible.

I'm not entirely sure what I want to happen instead, though. Maybe CSS that hides a signpost when preceded by another signpost, so that they become layered instead?
i.e. "This post contains filtered tags or content." &rarr; View post &rarr; "This post is hidden by PostBlock." &rarr; View post &rarr; (The post)

marcustyphoon:
Oh, if I'd thought of that as an implementation method for the layering I probably would have done it, yeah. I think that's an improvement on showing both. Not quite as elegant as combined text, but as mentioned I don't think that's worth implementing and I don't have any better ideas at the moment.

AprilSylph (changes requested):
I think this is a regression from 78c4300ba504f5a9fa74e1c638da9311c1d4c181: the util no longer behaves correctly on restarting the addon.

Post is collapsed with signpost &rarr; restart XKit Rewritten &rarr; post is shown

src/utils/hide_posts.js
```diff
@@ -0,0 +1,137 @@
+import { br, button, div } from './dom.js';
+import { buildStyle, getTimelineItemWrapper } from './interface.js';
+import { anyPostPermalinkTimelineFilter, timelineSelector } from './timeline_id.js';
+
+const controlsClass = 'xkit-hidden-post-controls';
+
+// Remove outdated elements when loading module
+$(`.${controlsClass}`).remove();
```
AprilSylph, line 8: I guess the correct fix would be to also remove the `-controlled` attributes here?

marcustyphoon:
Hmmmm interesting. So we could either fire all of `showPosts` on setup, or use a different method to handle the filtered post thing.

edit: jinx

edit 2: wait I guess you submitted the message first so I am the one who is jinxed or whatever. is that how jinx works? I forget, I am no longer twelve

src/utils/hide_posts.js
marcustyphoon: In `createPostHideFunctions`, presumably. We could either create all of the returnable functions, call `showPosts`, then return them, or—I mean that's certainly the simplest way.

marcustyphoon:
> Oh, if I'd thought of that as an implementation method for the layering I probably would have done it, yeah. I think that's an improvement on showing both. Not quite as elegant as combined text, but as mentioned I don't think that's worth implementing and I don't have any better ideas at the moment.

I guess technically the copy could be `This post ${elementsAsList(messages, 'and').join('')}.` That would elegantly create "This post contains a blocked blog and is hidden by PostBlock."

**edit:** Hm, this would be a util rewrite. Maybe I'll try it at some point. It's too bad you can't use `content: "This post " attr(something that grabs multiple attrs somehow) ".";` here; that would have been fun.

src/utils/hide_posts.js
AprilSylph:
```js
const hidePost = ...;
const showPost = ...;
const showPosts = ...;

showPosts();

return { hidePost, showPost, showPosts } ;
```

...yep.

marcustyphoon: Do I realize that this now looks like it should be a class? Yes. Yes I do. Do I want to make it a class? Eughhhhh I mean like...

(Also, do I want to make the posts actually unhide when the extension is disabled? Nah, that would be some pretty wild static CSS.)

marcustyphoon: ### Additional testing steps:

- Be smarter than me.

marcustyphoon: > (Also, do I want to make the posts actually unhide when the extension is disabled? Nah, that would be some pretty wild static CSS.)

Okay **technically** we could actually make `displayBlockIfDisabledAttr`, but that's extreme. (Or, to be precise, something like `content: var(--invalid-value-if-xkit-disabled, linear-gradient(transparent, transparent));`, which might or might not work).

src/utils/hide_posts.js
```diff
@@ -0,0 +1,142 @@
...
+
+  const addPermalinkPageControls = (postElement, timelineElement) => {
+    const timelineItemWrapper = getTimelineItemWrapper(postElement);
+    if (timelineItemWrapper.getAttribute(controlledHiddenAttribute) !== '') {
+      timelineItemWrapper.setAttribute(controlledHiddenAttribute, '');
+
+      const { message } = permalinkPageControls;
+      const controlsElement = div({ class: controlsClass, [controlsAttribute]: id }, [
+        message,
+        br(),
```
marcustyphoon: Er, this br doesn't do anything, does it.

marcustyphoon: Yep, copy failure from when I brought over [`8a25cdc` (#719)](https://github.com/AprilSylph/XKit-Rewritten/pull/719/changes/8a25cdc9b07c5e224389b85ab320df1b877b5ffc).

AprilSylph (approved):
Works the way I want it to! Huge UX improvement.

Do you still daily-drive `master`? I was thinking I could give you the weekend to beta-test this, and then do a release on Monday.

marcustyphoon:
I do indeed daily-drive master*, though I don't really use any of the post-hiding utils regularly. Still, may as well see if anything explodes.

*haha uh... plus a bunch of additional random junk like #1625.

PR #2137, opened 2026-03-05
feat(UI): use Acorn styling and `role="switch"` for feature toggles

AprilSylph:
### Description

- Cherrypicked from #2125
- Reference: https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-toggle--default

Updates our feature toggle switches to be shaped like Acorn toggle switches, and adds `role="switch"` to them so that accessibility tools (hopefully) announce their states as "on"/"off" rather than "checked"/"unchecked".

This does not update their colour scheme to be Acorn compliant. Right now, they just reuse our custom UI colours:

-|-|-|-
Screen Shot 2026-03-05 at 11 29 41 | Screen Shot 2026-03-05 at 11 30 13 | Screen Shot 2026-03-05 at 11 29 45 | Screen Shot 2026-03-05 at 11 30 18

This also applies the same 60% disabled opacity to `.buttons` as is applies to `.icon` and `.meta`, which I think just looks better.

Help links now also have rounded focus outlines.

Anything in the "badge" slot also now has the intended flex styles applied.

### Testing steps

Smoke-test the control panel. Toggling features should still work.

### DIFF
--- a/src/action/components/xkit-feature/index.css
+++ b/src/action/components/xkit-feature/index.css
@@ -72,7 +72,7 @@ summary:focus {
     - var(--toggle-dot-diameter)
   );
 
-  translate: 0;
+  transform: translateX(0);
 
   display: block;
   height: var(--toggle-dot-diameter);


src/action/components/xkit-feature/index.css
```diff
@@ -25,6 +25,75 @@ summary:focus {
...
+    - (var(--toggle-dot-margin) * 2)
+  );
+  --toggle-dot-transform-x: calc(
+    var(--toggle-width)
+    - (var(--toggle-border-width) * 2)
+    - (var(--toggle-dot-margin) * 2)
+    - var(--toggle-dot-diameter)
+  );
+
+  translate: 0;
```
marcustyphoon: Not sure if this does anything?

AprilSylph: It's for `transition` to take effect-- wait, yeah, this should be `transform`, huh.

PR #2140, opened 2026-03-06
fix: Erroneous `reblog_info` query parameter in util

marcustyphoon:
### Description

We don't currently rely on this, thankfully, but the `updatePostOnPage` util should attempt to fetch as much post data as possible; I thus gave it the `reblog_info=true` query parameter. This query parameter, however, does nothing on the `/v2/blog/{blog-identifier}/posts/{post-id}` [API endpoint](https://github.com/tumblr/docs/blob/master/api.md#postspost-id---fetching-a-post-neue-post-format); it only works on the  `/v2/blog/{blog-identifier}/posts?id={post-id}` [API endpoint](https://github.com/tumblr/docs/blob/master/api.md#posts--retrieve-published-posts). This switches to the correct endpoint.

### Testing steps

- Smoke test Quick Tags.
- If desired, open the developer console's network tab, use Quick Tags on a post which is a reblog, and confirm that the final network request contains fields like `rebloggedFromId`.

### DIFF
--- a/src/utils/react_props.js
+++ b/src/utils/react_props.js
@@ -76,7 +76,7 @@ export const editPostFormTags = async ({ add = [], remove = [] }) =>
  */
 export const updatePostOnPage = async (postElement, keys) => {
   const currentTimelineObject = await timelineObject(postElement);
-  const { response: newTimelineObject } = await apiFetch(`/v2/blog/${currentTimelineObject.blog.uuid}/posts/${currentTimelineObject.id}?reblog_info=true`);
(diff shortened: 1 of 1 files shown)


AprilSylph:
Shouldn't we want to fetch as _few_ fields as possible and retain as much of the "old" timeline object as possible?
The fewer bits transferred, the faster it can be. 

...I also can't think of any fields returned by `reblog_info` that can be changed by editing the post...?

marcustyphoon:
> Shouldn't we want to [...] retain as much of the "old" timeline object as possible?

Correctness >>> transferring a few bytes over the wire. The point of calling this util is to make the user see the updated post state. Redpop replaces the entire post data with the fetch contents.

> ...I also can't think of any fields returned by `reblog_info` that can be changed by editing the post...?

Yeah, no, that's true.

PR #2141, opened 2026-03-06
feat(Postblock): Make options UI items links if possible (shortUrl)

marcustyphoon:
### Description

TBD; see discussion in #719. In particular, we must test if this works if you change blog names.

### Testing steps

TBD.


marcustyphoon:
> In particular, we must test if this works if you change blog names.

[Appears to not.](https://github.com/AprilSylph/XKit-Rewritten/pull/719#issuecomment-4047211390)

PR #2144, opened 2026-03-12
feat(UI): implement `SponsorProgress` component

AprilSylph:
### Description

Adds a percent-funded progress bar to the bottom of the Links tab:

Screen Shot 2026-03-12 at 12 13 34 | Screen Shot 2026-03-12 at 12 13 37
-|-

Since this is somewhat hidden away, I don't expect this to push the needle at all—this PR mainly serves as a way to test the basic component, without any of the complicated CSS that will be needed to add it in the spot I _actually_ want to put it.

This shouldn't require any privacy policy or data collection declaration changes, since this data is fetched from GitHub Pages—which doesn't give me any connection logs or traffic stats.

### Testing steps

1. Load the modified addon
2. Open the addon manifest in a new tab
3. Open the Links tabpanel
    - **Expected result**: The progress bar is visible and working, as in the screenshots
    - **Expected result**: The progress bar and its text caption links to `https://github.com/sponsors/AprilSylph` in a new tab
4. Using your Network devtools, block requests to `https://aprilsylph.github.io/github-sponsors-listing/` temporarily
5. Reload the page and open the Links tabpanel again
    - **Expected result**: The progress bar is gone!

### DIFF
--- a/src/action/components/sponsor-progress/index.js
+++ b/src/action/components/sponsor-progress/index.js
@@ -41,7 +41,7 @@ class SponsorProgressElement extends CustomElement {
 
   async connectedCallback () {
     try {
-      const response = await fetch('https://aprilsylph.github.io/github-sponsors-listing/');
+      const response = await fetch('https://sponsors.aprilsylph.dev/');
       const data = await response.json();
 
       if (typeof data?.sponsorsListing?.activeGoal?.percentComplete !== 'number') {


marcustyphoon:
I assume not having this fire only when the links panel is navigated to is intentional, e.g. so that the load simulated is the same as if this was not only in that location?

src/action/components/sponsor-progress/index.js
```diff
@@ -0,0 +1,70 @@
...
+      this.#captionElement.textContent = `${percentComplete}% funded`;
+      this.#progressBarElement.style.width = `${percentComplete}%`;
+
+      const { url } = data.sponsorsListing;
+      if (SponsorProgressElement.canParseUrl(url)) {
+        this.#coverLinkElement.href = url;
+      }
+    } catch {
+      this.style.display = 'none';
+    }
```
marcustyphoon, line 64: If it were me, and the failure UI was "don't show anything," I might make the progress UI also "don't show anything," but it really makes no difference.

marcustyphoon:
> If I would have to pay to show this thing consistently, I would rather just abandon the idea.

Hmm, 1m/month seems pretty reasonable. Cloudflare will give you 3m/month, which is the same order of magnitude. A daily-scheduled (or, hell, hourly scheduled) github action updating a "static" json resource in any of the ones that give unlimited requests to static content would give you infinity, if that was needed for some reason. 

**edit:** Actually, we have like 100k users—most will be opening the preference UI very few times per month, so I'm sure that doesn't add up to anywhere near 1m/month, but it's not as many orders of magnitude larger than a realistic estimate, so maybe strike that "for some reason." Now I'm very curious what it'll actually be!

marcustyphoon:
I couldn't figure out an obvious way to make this element both take up space and stick to the bottom of the page when the UI is opened in a tab and does not have a height set by its contents. (`position: static fixed;` doesn't make any sense as a thing to want, right?) Overlaying two elements of the same height would work I guess, but ???

I suppose this is really only an issue if doing something like #1193, though, and if doing that, one could modify the display of the in-a-tab UI not to take up the whole page, which would make this irrelevant.

But if there actually is an obvious way to specify that in CSS, I think it would be an improvement.

src/action/components/sponsor-progress/index.css
```diff
@@ -0,0 +1,46 @@
+:host {
+  position: relative;
+
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  column-gap: 12px;
+  padding-block: 4px;
```
marcustyphoon, line 9: Can't decide if I think it looks better without this padding.

AprilSylph:
> I assume not having this fire only when the links panel is navigated to is intentional, e.g. so that the load simulated is the same as if this was not only in that location?

Correct. (This is another reason why having testing steps is so good: if I list something as an expected result, that tells you that it was intentional.)

> > If I would have to pay to show this thing consistently, I would rather just abandon the idea.
> 
> Hmm, 1m/month seems pretty reasonable. Cloudflare will give you 3m/month, which is the same order of magnitude. A daily-scheduled (or, hell, hourly scheduled) github action updating a "static" json resource in any of the ones that give unlimited requests to static content would give you infinity, if that was needed for some reason.

I mean, shit, couldn't I just put the output on GitHub Pages? New XKit already proved that that it's reliable as a static assets server. Why was my first thought to build a Vercel app, I wonder?

marcustyphoon:
Yep. Making a vercel app/cloudflare worker certainly is a useful skill to have in your pocket for things that you want to update frequently or dynamically, I think, but they aren't the only path here.

AprilSylph:
I'll keep this open while I work on the GitHub Pages version of the API. The only thing that should change here, when that's done, is the endpoint URL.

That's 117/118 lines of code that are still relevant!

src/action/components/sponsor-progress/index.js
AprilSylph: Pop-in is a significantly worse user experience than vanish-on-failure, in my opinion. Especially when the failure scenario should be in the extreme minority.

src/action/components/sponsor-progress/index.css
AprilSylph: I like the padding

marcustyphoon (dismissed):

PR #2150, opened 2026-03-16
feat(Quick Reblog): Enable feature on trail items

### LINKED ISSUE #2151: Quick Reblog: Enable feature on trail items
### Description

Quick Reblog does not currently function on the new footer elements on trail items.

I wrote most of this implementation (https://github.com/AprilSylph/XKit-Rewritten/compare/master...AprilSylph:XKit-Rewritten:marcustyphoon/quick-reblog-in-trail-footer — please feel free to steal), however, in places like the following timeline that use height restrictors, there's an `overflow: clip` here:

We may need to adjust where in the dom we add popup elements, which isn't something I was planning on investigating presently.

marcustyphoon:
### Description

This allows Quick Reblog to process the footers on trail items the same way it processes post footers.

Resolves #2151.

### Testing steps

Not yet written.


marcustyphoon:
Well, there's a problem.

PR #2153, opened 2026-03-17
feat(UI): move sponsor progress to Configuration tab

AprilSylph:
### Description

Moves the sponsor progress element from the Links tab to the Configuration tab, and makes it stick to the bottom of the viewport.

This was a lot less complicated than I thought it would be! This is the first time that `position: sticky; bottom: 0;` has done what I want it to do.

-|-|-|-
Screen Shot 2026-03-17 at 10 56 36 | Screen Shot 2026-03-17 at 10 56 48 | Screen Shot 2026-03-17 at 10 57 07 | Screen Shot 2026-03-17 at 10 57 13

### Testing steps

1. Load the modified addon
    - **Expected result**: The sponsor progress element is immediately visible, and loads my sponsor goal progress
3. Scroll the Configuration tabpanel
    - **Expected result**: The sponsor progress element keeps its position
    - **Expected result**: The sponsor progress element's presence does not prevent you from seeing the last feature in the list

The other half of validating this PR is your opinion on whether or not this is okay to add, or if you believe my own goal is unreasonable in the face of your (so far, unpaid) contributions. I will not make this change without your consent.

### DIFF
--- a/src/action/components/sponsor-progress/index.css
+++ b/src/action/components/sponsor-progress/index.css
@@ -1,12 +1,16 @@
 :host {
-  position: relative;
+  position: sticky;
+  bottom: 0;
 
   display: flex;
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
   column-gap: 12px;
-  padding-block: 4px;
+  padding: 8px;
+  border-top: 1px solid rgb(var(--active-grey));
+
+  background-color: rgb(var(--white));
 }
 
 #caption {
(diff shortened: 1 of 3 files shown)


marcustyphoon:
(Oh yeah. And if I thought that you were a) going to become independently wealthy off donations for a tool for a site the userbase won't fund, and that b) if that scenario occurred we wouldn't revisit this, then sure, maybe I'd have some reservations. But, you know, then you wouldn't be asking :D)

AprilSylph:
A tiny bit janky on Android because the features list actually takes a noticeable amount of time to load. Not bad enough to block this, but bad enough to be worth a followup PR.

PR #2159, opened 2026-03-19
feat(UI): show sponsor progress in all control panel tabs

AprilSylph:
### Description

- See https://github.com/AprilSylph/XKit-Rewritten/pull/2153#issuecomment-4081180708

Moves the sponsor progress bar to after the `<main>` element, so that it is visible on all control panel tabs; makes the popup a consistent height, so that it does not shrink when changing the active control panel tab; positions the sponsor progress bar consistently, at the bottom of the viewport.

Before | After
-|-
Screen Shot 2026-03-23 at 10 23 30 | Screen Shot 2026-03-23 at 10 25 09

Also, gives the sponsor progress component the `contentinfo` ARIA role to designate it as a footer.

Previously blocked by https://github.com/AprilSylph/XKit-Rewritten/pull/2161 over concerns that the sticky footer could cover the data overwrite warning.

#### Firefox for Android

1. Pull this branch locally
2. Attach an Android device that has Firefox installed
3. Enable USB debugging on the Android device
4. Run the modified addon on Firefox for Android:
    ```
    npm start -- --target=firefox-android
    ```
5. On Firefox for Android, open the XKit control panel
    - **Expected result**: There is no flash of the sponsor progress bar being out-of-position while the control panel loads (this may be easier to see when refreshing the control panel page, rather than when you first enter the screen)

#### Desktop

1. Load the modified addon in a supported desktop browser
    - [ ] Firefox
    - [ ] Chrome
    - **Expected result**: The Configuration tab loads with no apparent changes vs `master`
3. Switch to the Configuration tab
    - **Expected result**: The control panel popup height does not change
    - **Expected result**: The sponsor progress bar position does not change
4. Switch to the Links tab
    - **Expected result**: The control panel popup height does not change
    - **Expected result**: The sponsor progress bar position does not change

### DIFF
--- a/src/action/configuration.css
+++ b/src/action/configuration.css
@@ -36,7 +36,7 @@ label[for="filter"] {
   display: flex;
   justify-content: center;
   align-items: center;
-  height: 100px;
+  height: 100vw;
 
   color: rgb(var(--active-grey));
   content: "Loading…";


AprilSylph:
The "popup is a consistent height" part of this PR doesn't work on desktop Safari, but we don't publish for desktop Safari, so I don't care right now.

src/action/configuration.css
```diff
@@ -36,7 +36,7 @@ label[for="filter"] {
   display: flex;
   justify-content: center;
   align-items: center;
-  height: 100px;
+  height: 100vw;
```
marcustyphoon, line 39: This is intentionally "a square," I assume?

AprilSylph: Yep!

marcustyphoon:
Aside: I feel like a more accurate title for this would specify the sponsor bar appearing on all panels as the primary change.

PR #2161, opened 2026-03-19
feat(UI): add confirmation modal to Backup restore

AprilSylph:
### Description

The overwrite warning on the Backup tab could, as of a64298a2978c1e827219749db5aa4d276b953068, potentially not be seen by a user who would have to scroll to see it.

To fix this UX problem, this PR replaces the static warning with an interactive dialog, which gives the same warning, plus approximate filesizes of the current vs backup configs for comparison. This might help someone catch that they're about to overwrite a config that has 10,000 AlreadyReblogged entries with one that doesn't, for example.

Screen Shot 2026-03-22 at 12 53 32 | Screen Shot 2026-03-22 at 12 53 36
-|-

This dialog is skipped if the current configuration is empty.

### Testing steps

1. Copy an XKit Rewritten config to your clipboard
2. Load the modified addon in an isolated environment, so that it loads with a blank config
3. Open the XKit control panel &rarr; Backup tab
4. Import your config
    - **Expected result**: Your config is restored with no confirmation needed
5. Type `{}` into the "Import" textarea, then click "Restore"
    - **Expected result**: The blank config is _not_ restored immediately
    - **Expected result**: The confirmation dialog appears, showing the filesize of your current configuration and your backup configuration, both in kilobytes
6. Click "Cancel"
    - **Expected result**: The dialog closes, and nothing else happens
7. Click "Restore" again, and then "Overwrite" in the dialog this time
    - **Expected result**: Your config is overwritten with an empty one

### DIFF
--- a/src/action/popup.css
+++ b/src/action/popup.css
@@ -20,6 +20,7 @@
   --text-color: #15141a;
   --button-text-color-destructive: #ffffff;
   --focus-outline: 2px solid #0062f9;
+  --icon-color-warning: #b16a00;
 }
 
 @media (prefers-color-scheme: dark) {
@@ -45,6 +46,7 @@
     --text-color: #fbfbfe;
     --button-text-color-destructive: #15141a;
     --focus-outline: 2px solid #00cadb;
+    --icon-color-warning: #f5cc58;
   }
 }
 
@@ -251,6 +253,17 @@ h4 {
 ul {
   padding-inline-start: 16px;
   padding-inline-end: 0;
+}
+
+.acorn-icons {
+  display: inline-flex;
+  vertical-align: bottom;
+  margin: 0;
+}
+
+.acorn-icons svg {
+  height: 1.15em;
+  width: 1.15em;
 }
 
 .clr-picker {
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -70,7 +70,15 @@
           </form>
         </section>
         <dialog id="overwrite-confirmation">
-          <h4>⚠️&ensp;Overwrite current configuration?</h4>
+          <h4>
+            <figure class="acorn-icons" aria-hidden="true">
+              <!-- https://mozilla.org/MPL/2.0/ | https://github.com/FirefoxUX/acorn-icons/blob/e326587/icons/desktop/16/svg/warning-fill-16.svg -->
+              <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="var(--icon-color-warning)">
+                <path d="M15.245 12.028 9.713 2.081a1.98 1.98 0 0 0-1.747-1.028A1.98 1.98 0 0 0 6.213 2.09L.75 12.037C.018 13.37.982 15 2.503 15h10.994c1.525 0 2.489-1.639 1.748-2.972M8.75 12.5h-1.5V11h1.5zm0-3h-1.5v-4h1.5z" />
+              </svg>
+            </figure>
+            Overwrite current configuration?
+          </h4>
           <p>Restoring this backup will replace your existing data.</p>
           <ul>
             <li><span id="overwrite-size-old">…</span> of existing data will be deleted.</li>


src/action/render_backup.js
```diff
@@ -62,6 +64,36 @@ function onExportSubmit (event) {
...
+};
+
+const unitFormat = new Intl.NumberFormat('en-GB', {
+  style: 'unit',
+  unit: 'kilobyte',
+  minimumFractionDigits: 2,
+  maximumFractionDigits: 2,
+});
+
+const showoverwriteConfirmationDialog = (currentStorage, parsedStorage) => new Promise((resolve, reject) => {
```
marcustyphoon: Unusual camelcasing!

AprilSylph: Whoopsie

AprilSylph:
> Unfortunately, our browser support matrix doesn't include `oklch()`.

I NEED TO STOP TRUSTING INTELLISENSE

src/action/popup.css
marcustyphoon, line 1:
```suggestion
:root, ::backdrop {
```

https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::backdrop#browser_compatibility

Getting niche now; maybe this is not worth bothering with. But there's no backdrop color in older chromium versions unless the css variable applies to the backdrop element.

(This may be a broken suggestion, idk, I couldn't make it look right in the preview.)

marcustyphoon, line 27:
```suggestion
  :root, ::backdrop {
```

marcustyphoon: Yeah, not sure github's "review any source line" thing works.

src/action/render_backup.js
```diff
@@ -82,18 +117,23 @@ async function onImportSubmit (event) {
...
     document.getElementById('configuration-tab').classList.add('outdated');
+
+    await sleep(3000);
   } catch (exception) {
-    importSubmitButton.classList.add('failure');
-    importSubmitButton.textContent = exception instanceof SyntaxError
-      ? 'Failed to parse backup contents!'
-      : 'Failed to restore!';
-    console.error(exception);
+    if (exception.message !== 'Cancelled') {
```
marcustyphoon: Probably not worth it here, but we could do a custom error type (`class CancelledError extends Error {`) and test `instanceof CancelledError`, removing the magic string.

I'm kind of tempted to do this in content script code, where the error declaration could be in a util (I mean, I would be if we had instances where a finally block should be triggered on a cancel action there). I don't imagine we really want to share more util files between content script and option UI, though.

AprilSylph: I like this idea. I wonder if neostandard will let us one-line the definition...

```js
class UserInterrupt extends Error { name = 'UserInterrupt' }
```

marcustyphoon: I wonder if you can just straight up do an empty class body. I guess it's probably good practice to make the thing meaningfully loggable.

marcustyphoon: (Aside: I think all error objects should end with `Error`. `throw new UserInterrupt()` aa, no, my eyes :D)

AprilSylph: > I wonder if you can just straight up do an empty class body. I guess it's probably good practice to make the thing meaningfully loggable.

You can, but then yeah, it logs as `Error`. But, I guess we're not logging it...

AprilSylph: > (Aside: I think all error objects should end with `Error`. `throw new UserInterrupt()` aa, no, my eyes :D)

Fun fact: the first programming language I learnt was actually Python. [Most exceptions in Python](https://docs.python.org/3/library/exceptions.html) also end with `Error`, but some don't—such as `KeyboardInterrupt`.

I think I prefer not calling it an error if it's not actually an error. Extending from `Error` would just be to make ESLint happy—you can technically throw anything in JavaScript, after all!

src/action/popup.css
```diff
@@ -127,8 +159,7 @@ body {
 }
```
marcustyphoon, line 162: Hey, stupid question, because this is literally from a different PR that I already approved: what's this? What's it supposed to do? I ran into it blocking UI clicks after cancelling the dialog in Safari, assuming the problem was a Safari bug with backdrop elements persisting, but it's actually this, which I've now realized I don't understand.

marcustyphoon: > - **Expected result**: Focused tabpanels gain a fully visible focus outline, viewport cutoff notwithstanding

Ah, I see. Can one set `pointer-events: none;` on a pseudo?

marcustyphoon: Yep, that will fix it.

AprilSylph: Oh, I'm very silly for not having done that.

AprilSylph:
the sneaky post-approval commit:

PR #2166, opened 2026-03-24
feat(UI): use Acorn colours in control panel

AprilSylph:
### Description

Is this an improvement by itself? No.

Will it make future PRs that Acornify pieces of the control panel look less stupid? Yes.

Is this enough of a step backwards that it's worth creating a feature branch for Acornification PRs to target, starting with this one? I'm not entirely sure; you tell me!

-|-|-|-
Screen Shot 2026-03-24 at 11 42 40 | Screen Shot 2026-03-24 at 11 43 02 | Screen Shot 2026-03-24 at 11 42 43 | Screen Shot 2026-03-24 at 11 43 06

### Testing steps

1. Load the modified addon
    - **Expected result**: None of the colour changes result in unreasonably low contrast between any UI component's background and foreground
3. Change the colour scheme (light ↔ dark), and repeat step 2

### DIFF
--- a/src/action/popup.css
+++ b/src/action/popup.css
@@ -4,7 +4,7 @@
   --accent: 0, 98, 249;
   --passive-grey: 240, 240, 244;
   --active-grey: 191, 191, 201;
-  --search-highlight: 255, 233, 0;
+  --search-highlight: 226, 247, 255;
 
   /* Acorn Tokens */
   --card-background-color: #ffffff;
@@ -30,7 +30,7 @@
     --accent: 0, 202, 219;
     --passive-grey: 35, 34, 43;
     --active-grey: 91, 91, 102;
-    --search-highlight: 0, 62, 170;
+    --search-highlight: 0, 49, 126;
 
     /* Acorn Tokens */
     --card-background-color: #23222b;


marcustyphoon:
> Is this enough of a step backwards that it's worth creating a feature branch for Acornification PRs to target, starting with this one? I'm not entirely sure; you tell me!

Hm. In like 30 seconds of "I clicked on the screenshots and flipped between them," I don't think I have a particularly strong opinion on this.

As you know, one of the things I often think about when judging code changes is "how would the exact order and context in which this is presented to users affect their impressions of it," such as that I often wish we had deployed XKit feature fixes for Tumblr changes before the change actually goes out, if I think the Tumblr change deserves a fair shake in its XKit-modified form (so that users do not associate the change with "it broke XKit and I'm mad about it"). (Similar logic in various nominally-opposite cases, notably.)

In that vein, all else being equal I could see the logic in carefully not releasing a change that without not-yet-implemented development could be considered arbitrary. On the other hand, maybe that's good because people will get used to it over time if they initially don't notice so this way the later changes are less jarring and so that's good?

(Honestly not sure in retrospect why I typed the last two paragraphs when they ultimately keep my response pretty neutral :D But maybe there will be something to agree/disagree with in there, idk.)

AprilSylph:
No, that's a good thing to think about. I do think people will hate the end product more if they get just this change first and associate the change as a whole with this unfinished version; it would be too easy to see this as the intended final change and further changes as course-correction.

AprilSylph:
Created `acorn` feature branch and set up protection rules for it mirroring the default branch.

marcustyphoon:
Either way works for me! If we're doing a feature branch I figure I'm basically blind-approving, at least at first; I don't know what the eventual look will be!

AprilSylph:
> Either way works for me! If we're doing a feature branch I figure I'm basically blind-approving, at least at first; I don't know what the eventual look will be!

Fair! My screenshots can serve as a good-enough way to validate the change here, to be honest. I would advise against blind-approving anything further, though, as my coming PRs actually will aim to finalise the look of each part of the UI. (Until Firefox changes its design language again, of course.)

PR #2170, opened 2026-03-25
feat(UI): Acorn styling for Backup tab

AprilSylph:
### Description

- Storybook: [UI Widgets / Input Text -Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-input-text--default)
- Storybook: [UI Widgets / Multiline Editor - Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-multiline-editor--default)
- Storybook: [UI Widgets / Button - Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-button--default)
- Storybook: [UI Widgets / Button - Primary](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-button--primary)
- Storybook: [UI Widgets / Message Bar - Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-message-bar--default)

-|-|-|-
Screen Shot 2026-03-26 at 10 18 38 | Screen Shot 2026-03-26 at 15 37 32 | Screen Shot 2026-03-26 at 10 18 42 | Screen Shot 2026-03-26 at 15 37 57
Screen Shot 2026-03-26 at 11 46 25 | Screen Shot 2026-03-26 at 15 38 21 | Screen Shot 2026-03-26 at 11 46 30 | Screen Shot 2026-03-26 at 15 38 25
Screen Shot 2026-03-26 at 11 48 11 | Screen Shot 2026-03-26 at 15 39 08| Screen Shot 2026-03-26 at 11 48 17 | Screen Shot 2026-03-26 at 15 39 10
Screen Shot 2026-03-26 at 11 50 07 | Screen Shot 2026-03-26 at 15 39 56 | Screen Shot 2026-03-26 at 11 50 13 | Screen Shot 2026-03-26 at 15 39 59

### Testing steps

Smoke-test the Backup tab. It should still do what it needs to do.

Every UI element in the Backup tab should look and feel like it came out of the Acorn Storybook.

### DIFF
--- a/src/action/backup.css
+++ b/src/action/backup.css
@@ -18,5 +18,5 @@
 
 #backup-panel section fieldset,
 #backup-panel section [role="alert"] {
-  margin-block-start: 7.5px;
+  margin-block-start: 15px;
 }
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -66,9 +66,6 @@
           <form id="import" aria-labelledby="import-heading">
             <label for="import-value">Paste the contents of your saved backup here.</label>
             <textarea id="import-value" rows="8" spellcheck="false"></textarea>
-            <fieldset>
-              <button id="import-submit" type="submit" class="primary">Restore</button>
-            </fieldset>
             <div id="import-success" role="alert" class="success" hidden>
               <figure class="acorn-icons" aria-hidden="true">
                 <!-- https://mozilla.org/MPL/2.0/ | https://github.com/FirefoxUX/acorn-icons/blob/e326587/icons/desktop/16/svg/checkmark-circle-fill-16.svg -->
@@ -87,6 +84,9 @@
               </figure>
               <span id="import-error-message"></span>
             </div>
+            <fieldset>
+              <button id="import-submit" type="submit" class="primary">Restore</button>
+            </fieldset>
           </form>
         </section>
         <dialog id="overwrite-confirmation">


marcustyphoon:
Food for thought:
- Is there something missing re: spacing with the label, or the interaction between the header and the label, or something? Something about it feels off to me; not sure what.
- Should alerts have more margin (more margin-block?) than text boxes so they look more temporary/floating? (Maybe more margin-bottom also?
- I know you wouldn't asymmetrically pad the text box by accident, so that must be on purpose, but does it look a bit odd to you?
- I'm tempted to copy the importValueTextarea value into the exportValueTextarea synchronously with clearing it after the successful local storage set; it just looks slightly cleaner. (This would cause a flicker if the restored value wasn't formatted with `JSON.stringify(/* */, null, 2)` though.)

AprilSylph:
> * I know you wouldn't asymmetrically pad the text box by accident, so that must be on purpose, but does it look a bit odd to you?

Honestly, it doesn't. The whole box is already asymmetric by nature of having a leading icon, so increasing the inline end padding wouldn't truly achieve symmetry; and symmetric padding for prose is actually not desirable by default, since (to my understanding) it's rare for the line to ever meet the edge of the content box perfectly, so a little extra room to fit an extra letter or two actually helps with the text balancing across lines.

marcustyphoon:
Oops, sorry, by text box I meant textarea.

AprilSylph:
I see. The textareas do have cardinally symmetrical padding, so that threw me off.

It's intentional that they don't have diagonally symmetrical padding, to match [UI Widgets / Input Text - Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-input-text--default). It does look a little odd, but the point of a design system is to let it make as many decisions for you as possible, and I don't think it looks jarring enough to go against that.

AprilSylph:
> * Is there something missing re: spacing with the label, or the interaction between the header and the label, or something? Something about it feels off to me; not sure what.

Agree. I tried to base it off things I found in `about:preferences` since the Acorn Storybook doesn't really have anything for how to put forms together, but I should check my work there...

marcustyphoon:
> I see. The textareas do have cardinally symmetrical padding, so that threw me off.
> 
> It's intentional that they don't have diagonally symmetrical padding, to match [UI Widgets / Input Text - Default](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-input-text--default). It does look a little odd, but the point of a design system is to let it make as many decisions for you as possible, and I don't think it looks jarring enough to go against that.



Yeah, I feel like it makes more sense for a `input type="text"` than a textarea.

AprilSylph:
> Should alerts have more margin (more margin-block?) than text boxes so they look more temporary/floating?

They're not temporary or floating, though—they only appear or disappear in response to user input.

(Reference for persistent static message bars: [Domain-specific UI Widgets / Backup / Restore from Backup - Encrypted Backup Found](https://firefoxux.github.io/firefox-desktop-components/?path=/story/domain-specific-ui-widgets-backup-restore-from-backup--encrypted-backup-found))

marcustyphoon:
> > * Is there something missing re: spacing with the label, or the interaction between the header and the label, or something? Something about it feels off to me; not sure what.
> 
> Agree. I tried to base it off things I found in `about:preferences` since the Acorn Storybook doesn't really have anything for how to put forms together, but I should check my work there...

Yeah, the closest thing they seem to have is the "Save files to" label, I guess? But that seems like a lot of space...

AprilSylph:
Reference | Implementation
-|-
image | image

(You're definitely allowed to use a heading as a `moz-fieldset` legend, `about:preferences` does it a lot.)

AprilSylph:
Ah, wait, the spacing actually increases to 15px if the `moz-fieldset` has the `hasheading` attribute... ugh.

AprilSylph:
If we want to perfectly match any of the Storybook examples for form field spacing, we would also have to mimic the placement of success/error messages before the submit controls (examples: [one](https://firefoxux.github.io/firefox-desktop-components/?path=/story/domain-specific-ui-widgets-backup-restore-from-backup--backup-found), [two](https://firefoxux.github.io/firefox-desktop-components/?path=/story/domain-specific-ui-widgets-backup-turn-on-scheduled-backups--enable-error)), which would cause the "Restore" button to jump... but it would ensure that the message would be seen.

I guess if we're following a design system, the decision on whether it's okay for buttons to jump has been made for us. I don't super like it, even though I can see the rationale behind it (you can't click the submit button without it being on-screen; ergo, the only thing guaranteed to be on-screen is the place the submit button was; ergo, that's where important messages go).

AprilSylph:
[acorn-backup-strict-spacing.patch](https://github.com/user-attachments/files/26279888/acorn-backup-strict-spacing.patch)

Failure | Success
-|-
image | image

I can't decide if I hate this or not. I think I want to just leave it as it is so I can stop thinking about it.

AprilSylph:
...nope, I have to do it, don't I.

AprilSylph:
Updated screenshots in PR description.

marcustyphoon:
Perhaps I should note that my perspective is that using alerts like this, where they pop up in real time, is inherently off-book—that I know of, Firefox doesn't use these this way, and would not. That informs both my comment that the "correct" layout for a component used in this way might/would not be the same, and a comment I'm now making that the position doesn't necessarily make sense that way either.

Would it make more sense to use a modal?

edit: I do see the second example, but in that case I think the alert is still below its triggering element (the location selector, not the submit button).

AprilSylph:
> edit: I do see the second example, but in that case I think the alert is still below its triggering element (the location selector, not the submit button).

I'm interpreting as being triggered by the submit button by virtue of its verbiage—specifically, "try again" heavily implies to me that the user already attempted to finalise the action. I think the message would be different (and more obviously attached to the directory input) if it were triggered by just an input change.

---

EDIT: Even more specifically, "Choose a different folder and try again" is basically explicit that _changing the location input_ and _submitting the form_ are two distinct steps which have both already been attempted.

It does frustrate me that this form can't be played with live, which would allow us to come to an unambiguous conclusion immediately. Maybe one of us can find it hiding in Firefox Nightly somewhere, somehow?

marcustyphoon:
I found it live in Firefox 145 (https://ftp.mozilla.org/pub/firefox/releases/145.0.2/mac/en-US/, toggle `browser.backup.x` in `about:config`; see https://docs.google.com/document/d/e/2PACX-1vTkMmsafMbb5G139lbeE55YhgKYT1NJkFO7M1_VOp1awNdWQyExIHWYSzJ6IlIyt_20k9uV2zNCON76/pub / https://www.askvg.com/how-to-enable-or-disable-backup-feature-in-mozilla-firefox-all-versions/; document search credit: claude ai), but I couldn't get it to throw an error.

Of course it ultimately doesn't matter; I don't see this particular UI element's implementation representing some kind of definitive design system ruling (nothing wrong with doing that, it's just not something I personally see a reason to do).

What I will say is that if it were me, I would either have the submit button action result be a modal popup, or (perhaps) add a fold-down/vertical expand CSS animation on the alert—something to, in some way, acknowledge that whatever UI change occurs after you press this button is the result of an action being taken, as distinct from a mutation to a UI that, per the design system parameters, looks like a static pane.

I, however, don't feel at all strongly about this and think it's fine as is if that seems right to you.

AprilSylph:
Ah, an animation really would convey extra information here, huh. Yeah, let's add that.

AprilSylph:
Just tested it myself on Firefox 148; I couldn't get the error bar to appear, but pressing "Backup now" in the parent screen made an info-type message bar appear, with no animation, above the "Backup now" button.

...I've changed my mind, I don't want the hassle[^1] of making a height animation work, actually.

[^1]: Not to say that it would be difficult; it would be trivial to do with a wrapper element. But I love my clean HTML so much that adding such a thing would bother me immensely.

PR #2174, opened 2026-03-27
feat(UI): use Acorn typography and link styles in Links tab

AprilSylph:
### Description

- Storybook: [UI Widgets / Support Link - Primary](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-support-link--primary)

I couldn't find any examples of an unordered list within the Acorn design system, so I kinda just made it up (although still using legal Acorn spacing, of course).

-|-|-|-
Screen Shot 2026-03-27 at 14 53 07 | Screen Shot 2026-03-27 at 14 53 30 | Screen Shot 2026-03-27 at 14 53 11 | Screen Shot 2026-03-27 at 14 53 35

### Testing steps

Anchor elements in the Links tab should now have hover, active, and focus-visible styles matching those of the Storybook entry linked above.

Help links on feature elements should be unchanged, apart from the new focus outline.

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -111,31 +111,39 @@
         </dialog>
       </section>
       <section id="links-panel" role="tabpanel" aria-labelledby="links-tab" tabindex="0" hidden>
-        <h4>Help & information</h4>
-        <ul>
-          <li><a href="https://github.com/AprilSylph/XKit-Rewritten/wiki" target="_blank">User guide</a></li>
-          <li><a href="https://github.com/AprilSylph/XKit-Rewritten/discussions" target="_blank">Support & discussion</a></li>
-          <li><a href="https://github.com/AprilSylph/XKit-Rewritten/releases" target="_blank">Changelog</a><small id="version"></small></li>
-        </ul>
-        <h4>Donate</h4>
-        <ul>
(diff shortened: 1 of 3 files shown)


src/action/popup.html
AprilSylph, line 1: This diff best viewed excluding whitespace-only changes!

marcustyphoon:
Do I think link underlines are an improvement? No: they make sense on e.g. `about:preferences#ai` because they distinguish the link text from the non-link text; in a page that's entirely made of links, however, they look less refined than when they're omitted, as though the developer used wikitext rather than designing the UI based on the context. (Not sure what I imagine the Mozilla team using if they had to make this component—out of the options, probably the box group, but I don't particularly like this option either.)

AprilSylph:
Yeah, I'm not super huge on the underlines either. I definitely don't want to use box groups, since the Configuration tab consists of one huge box group (or, perhaps more accurately, it will once I'm done with it).

Although... maybe if each link was a box item with a non-clickable label, and bare hyperlinks as the end content, that would be different enough from unexpanded features? You know, like how in `about:addons` there are addon boxes and then detail boxes:

Screenshot 2026-03-30 at 11 02 27 am

Screenshot 2026-03-30 at 11 03 01 am

Okay, it's only the Homepage item that has a bare hyperlink, but I still think it would work.

...

As you've pointed out before, I don't think the Links tab gets many eyeballs on it, so I think I'd like to leave it as-is for now and revisit this idea at a later date.

PR #2179, opened 2026-03-31
feat(UI): Acorn-style search bar

AprilSylph:
### Description

- Storybook: [UI Widgets / Input Search - With Placeholder](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-input-search--with-placeholder)
- Storybook: [UI Widgets / Select - With Aria Label](https://firefoxux.github.io/firefox-desktop-components/?path=/story/ui-widgets-select--with-aria-label)

-|-|-|-
Screen Shot 2026-03-31 at 11 44 39 | Screen Shot 2026-03-31 at 11 45 00 | Screen Shot 2026-03-31 at 11 44 44 | Screen Shot 2026-03-31 at 11 45 05
Screen Shot 2026-03-31 at 11 45 46 | Screen Shot 2026-03-31 at 11 46 25 | Screen Shot 2026-03-31 at 11 45 50 | Screen Shot 2026-03-31 at 11 46 30
Screen Shot 2026-03-31 at 11 46 52 | Screen Shot 2026-03-31 at 11 47 21 | Screen Shot 2026-03-31 at 11 46 56 | Screen Shot 2026-03-31 at 11 47 25

I've chosen to remove the "Filter by" text label here because I couldn't find a good way to keep it without making the typeable area of the search bar really narrow. I've kept "Filter" as the ARIA label; for sighted users, I assume the functionality becomes obvious after opening the dropdown or, failing that, after first use. Its positioning next to the search box hopefully is a big enough context clue for most people to immediately understand that it is part of the search bar.

### Testing steps

Confirm that the search bar (incl. filtering) still works.

The accessibility tree for the search bar should communicate the reality of the thing.

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -50,9 +50,11 @@
         <div role="search">
           <input id="searchbox" type="search" placeholder="Search" aria-label="Search" autocomplete="off" spellcheck="false">
           <select id="filter" aria-label="Filter">
-            <option value="all" selected>All</option>
-            <option value="enabled">Enabled</option>
-            <option value="disabled">Disabled</option>
+            <optgroup label="Filter by…">
+              <option value="all" selected>All</option>
+              <option value="enabled">Enabled</option>
+              <option value="disabled">Disabled</option>
+            </optgroup>
           </select>
         </div>
         <div class="no-results">


marcustyphoon:
Too bad "show all" would fit but "show enabled" and "show disabled" wouldn't.

AprilSylph:
I did consider "Show All"/"Enabled"/"Disabled" but it just felt too weird.

AprilSylph:
One thing I thought of but didn't try was the trick where you make the first option a disabled one in order to use that as the label. I suppose I should see how that feels...

AprilSylph:
> One thing I thought of but didn't try was the trick where you make the first option a disabled one in order to use that as the label. I suppose I should see how that feels...

...I don't like it. The font size is so big that making it more than a single word makes it feel too much like a primary action, when it really isn't one.

Screen Shot 2026-03-31 at 12 18 59 | Screen Shot 2026-03-31 at 12 19 04
-|-

PR #2186, opened 2026-04-01
 feat(UI): use Acorn dimensions + miscellaneous visual fixes

AprilSylph:
### Description

Yep, it's a junk PR. Too many things to do in a single commit; each thing so tiny that it would be silly to open separate pull requests. Sorry.

* Updates all dimensions in the main control panel stylesheet space to be Acorn-legal
* Sets the base font size in the control panel to 15px, removing the need for per-area overrides (preferences still retain their font size of 14px, for now)
* Removes our custom scrollbar colours; we don't really need them now that we set `<meta name="color-scheme">`
* Uses an Acorn-style focus outline for focused tabpanels
* Fixes an oversight where the focus outline for the Configuration tabpanel would overlap the sponsor progress footer

-|-|-|-
Screen Shot 2026-04-01 at 11 39 18 | Screen Shot 2026-04-01 at 11 40 21 | Screenshot 2026-04-01 at 11 39 54 am | Screenshot 2026-04-01 at 11 40 39 am

My actual motivation for using the default scrollbars is to create a gutter that the tablist/searchbar/footer borders can run into, instead of being abruptly cut off by nothing. This could be achieved with Acorn colours, but... the Acorn design system doesn't seem to define any scrollbar styles! Plus, you know, platform and system setting differences makes standardising their look a literally impossible task anyway.

### Testing steps

Smoke-test the control panel. Have I broken anything?

### DIFF
--- a/src/action/components/sponsor-progress/index.css
+++ b/src/action/components/sponsor-progress/index.css
@@ -1,6 +1,7 @@
 :host {
   position: sticky;
   bottom: 0;
+  z-index: 1;
 
   display: flex;
   flex-direction: row;


marcustyphoon:
Not necessarily relevant to this PR's contents, but are the coloris popups styled as you'd like right now?

AprilSylph:
Ah right, those are not part of `ColorPreferenceElement`, huh.

Well, I need a new PR title.

AprilSylph:
> Not necessarily relevant to this PR's contents, but are the coloris popups styled as you'd like right now?

To be unfailingly explicit: no! I overlooked these, and now they suck. Fixing them is now the next thing on my to-do list.

marcustyphoon:
The scroll bar being on the right side of the "funded" bit looks kind of weird, but there's no way to do anything about that while leaving the nice partially-sticky header behavior, so, shrug.

Oddly, in Chromium (138, apparently), switching to the links tab means the scroll bar disappears... but switching from either other tab to backup means that the scroll bar stays the same as it was (so, whether there's a scroll bar on backup depends on which tab you came from). Mystifying.

I haven't used scroll bars that take up space in years, so I never think about this stuff.

AprilSylph:
> Oddly, in Chromium (138, apparently), switching to the links tab means the scroll bar disappears... but switching from either other tab to backup means that the scroll bar stays the same as it was (so, whether there's a scroll bar on backup depends on which tab you came from). Mystifying.

Haha, I noticed this too, it's so weird! `html { overflow-y: scroll; }` is there specifically to stop the scrollbar gutter from ever disappearing, so hopefully this is a single-version browser bug...

marcustyphoon:
Unsurprisingly, nope: tested 4 random versions; it's consistent.

AprilSylph:
That sucks.

marcustyphoon:
Seems like one of those "maybe it has to be on body for weird reasons" things, I dunno.

PR #2191, opened 2026-04-01
feat(UI): Acorn-style `TextAreaPreferenceElement`

AprilSylph:
### Description

Before | After | Before | After
-|-|-|-
Screen Shot 2026-04-01 at 21 56 10 | Screen Shot 2026-04-01 at 22 11 20 | Screen Shot 2026-04-01 at 21 56 23 | Screen Shot 2026-04-01 at 22 11 24

### Testing steps

Screenshot comparison and code review will suffice as validation.

One may also want to load the modified addon to ensure that the Backup tab is unchanged by the Acorn label styles.

PR #2192, opened 2026-04-01
feat(UI): Acorn-style `SelectPreferenceElement`

AprilSylph:
### Description

-|-|-|-
Screen Shot 2026-04-01 at 22 25 48 | Screen Shot 2026-04-02 at 10 02 14 | Screen Shot 2026-04-01 at 22 25 52 | Screen Shot 2026-04-02 at 10 02 19

### Testing steps

Screenshot comparison and code review should suffice as validation.

### DIFF
--- a/src/features/accesskit/feature.json
+++ b/src/features/accesskit/feature.json
@@ -16,7 +16,7 @@
     },
     "disable_gifs_loading_mode": {
       "type": "select",
-      "label": "Download paused GIFs:",
+      "label": "Download paused GIFs",
       "options": [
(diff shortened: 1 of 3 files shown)


marcustyphoon:
Should we do a pass on being consistent about whether there's a colon at the end of a label before a control that's on the next line?

AprilSylph:
That would be nice, yes.

marcustyphoon:
Similar to my note in #2191, the gap above a select preference when it's the first one feels maybe a little big.

PR #2196, opened 2026-04-03
fix(UI): standardise and adjust preference spacing

AprilSylph:
### Description

A.K.A. "Margin collapse is good, actually."

Fixes the issue where the first preference is spaced an unreasonable distance away from the feature summary.
(Previously 26.25px; now 18.75px.)

This also makes search highlights span the entire width of the feature box, which I think looks really nice.

-|-|-|-
Screen Shot 2026-04-03 at 10 16 32 | Screen Shot 2026-04-03 at 10 23 56 | Screen Shot 2026-04-03 at 10 16 46 | Screen Shot 2026-04-03 at 10 24 00
Screen Shot 2026-04-03 at 10 44 39 | Screen Shot 2026-04-03 at 11 23 48 | Screen Shot 2026-04-03 at 10 44 43 | Screen Shot 2026-04-03 at 11 23 52

~~The spacing on "No preferences available" might look a little odd when the feature summary isn't being hovered or focused, but you can only see this message by opening the preferences manually, which on a desktop computer necessitates the summary being hovered or focused. I dunno, maybe it should match preferences' margin anyway...?~~

### Testing steps

Smoke-test feature elements, especially feature preferences. Everything should be high-enough quality to ship, apart from Quick Tags and PostBlock's custom panes; those are next on my to-do list.

### DIFF
--- a/src/action/components/xkit-feature/index.css
+++ b/src/action/components/xkit-feature/index.css
@@ -203,6 +203,9 @@ summary:focus-visible {
 }
 
 slot[name="preferences"] p {
+  margin-block-start: 7.5px;
+  margin-block-end: 11.25px;
+
   text-align: center;
 }
 


marcustyphoon:
Not actually related, but the links tab padding feels a little low, just clicking around and focusing on spacing.

marcustyphoon:
> The spacing on "No preferences available" might look a little odd when the feature summary isn't being hovered or focused, but you can only see this message by opening the preferences manually, which on a desktop computer necessitates the summary being hovered or focused. I dunno, maybe it should match preferences' margin anyway...?

No opinion on whether it should or should not precisely match, but it does feel on the big side at the moment imo, yeah.

AprilSylph:
> Not actually related, but the links tab padding feels a little low, just clicking around and focusing on spacing.

Agreed, that's on my to-do list too!

> > The spacing on "No preferences available" might look a little odd when the feature summary isn't being hovered or focused, but you can only see this message by opening the preferences manually, which on a desktop computer necessitates the summary being hovered or focused. I dunno, maybe it should match preferences' margin anyway...?
> 
> No opinion on whether it should or should not precisely match, but it does feel on the big side at the moment imo, yeah.

Alright, I'll adjust it down.

AprilSylph:
Updated screenshots for latest commit. I'm pretty happy with this, merging...

PR #2197, opened 2026-04-03
feat(UI): Acorn-style "Add new tag bundle" form

AprilSylph:
### Description

Does some scaffolding work to make focus outlines in the Quick Tags and PostBlock iframe preferences actually fully visible, for real this time; adds the Acorn stylesheet to the Quick Tags iframe; starts work on migrating the Quick Tags iframe to adhere to Acorn, ditching our old CSS variables and fully converting the "Add new tag bundle" form.

This also does some basic styling on the tag bundles list, but not much. To truly Acornify it is a big enough task that it warrants its own PR, or maybe several? I'd love to use box list styling, use a `<dialog>` for editing, use a different `<dialog>` for delete confirmations...

-|-|-|-
Screen Shot 2026-04-03 at 13 42 58 | Screen Shot 2026-04-03 at 13 43 41 | Screen Shot 2026-04-03 at 13 43 02 | Screen Shot 2026-04-03 at 13 43 45

### Testing steps

Smoke-test the Quick Tags &rarr; "Manage tag bundles" iframe preference. Everything should still work.

The "Add new tag bundle" form should feel Firefox-native.

### DIFF
--- a/src/features/quick_tags/options/index.html
+++ b/src/features/quick_tags/options/index.html
@@ -34,7 +34,14 @@
     </main>
     <template id="bundle-template">
       <li class="bundle">
-        <button class="drag-handle" title="Drag to rearrange"><i class="ri-draggable"></i></button>
+        <button class="drag-handle" title="Drag to rearrange">
+          <figure class="acorn-icons" aria-hidden="true">
+            <!-- https://mozilla.org/MPL/2.0/ | https://github.com/mozilla-firefox/firefox/blob/5151f62/toolkit/themes/shared/icons/move-16.svg -->
+            <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="currentColor">
+              <path d="M7 14H5v-2h2v2zM11 14H9v-2h2v2zM7 9H5V7h2v2zM11 9H9V7h2v2zM7 4H5V2h2v2zM11 2v2H9V2h2z"/>
+            </svg>
+          </figure>
+        </button>
         <div>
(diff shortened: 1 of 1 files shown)


marcustyphoon:
The "add new" section kind of feels to me like it should be surrounded by something—well, no, okay, since you mentioned dialogs, it should obviously be a dialog.

Unrelated thought: I sort of feel like we need to scale every single thing in this popup down by a little bit; Firefox's UI elements simply have more space and are scaled for that amount of space.

AprilSylph:
> The "add new" section kind of feels to me like it should be surrounded by something—well, no, okay, since you mentioned dialogs, it should obviously be a dialog.

Yes!! The whole thing should really just be a box list, a button at the bottom, and dialogs for add/edit/delete.

> Unrelated thought: I sort of feel like we need to scale every single thing in this popup down by a little bit; Firefox's UI elements simply have more space and are scaled for that amount of space.

Hmm... I was going to suggest maybe we can make the popup bigger, since we do control its width... but we seem to have no control over the height beyond what it expands to automatically, so that's annoying.

AprilSylph:
I suppose we could swap all the dimensions to use `rem` values (as they're supposed to be; maybe actually use tokens for them) and then reduce the root font size back down to 14px, and everything should just work?

I dunno. I like the larger font size, personally. The current control panel is in no way difficult for me to read, yet somehow I feel like reading the new one is easier.

I have to do a little more scrolling with the new one, but that's also attributable to how much more spaced out everything is, which I also enjoy. The current one feels unnecessarily dense when I look back at it.

PR #2219, opened 2026-04-14
fix(Classic Footer): hide updated reblog menus

### LINKED ISSUE #2212: Classic Footer: Reblog menus are not hidden, show up in top-left of screen
### Platform

Linux

### Browser

Firefox 149.0.2 (64-bit)

### Addon version

v1.2.3

### Details

"Turn reblog buttons back into links" doesn't work properly. The button is turned into a link, but clicking it opens Tumblr's reblog menu, which is positioned incorrectly.

Image

Clicking the reblog link a second time correctly navigates to the reblog form, but this is not very discoverable behaviour.

AprilSylph:
### Description

- Fixes #2212 

Updates the targeting for the reblog menu portal and reblog link menu item.

This code aims to be compatible with the prior version of the reblog menu as well, in case this change is reverted.

Before | After
-|-
Screen Shot 2026-04-14 at 08 48 00 | Screen Shot 2026-04-14 at 08 49 19

### Testing steps

1. Load the modified addon
2. Open a Tumblr tab
3. Click a reblog button to make the reblog menu appear
4. Enable Classic Footer, leaving its preferences as their defaults
    - **Expected result**: The reblog menu disappears
5. Click a reblog icon link on a post
    - **Expected result**: The reblog form is opened, without the reblog menu being visible for even a single frame
6. (Optional) Click a reblog icon link on a post with advertising data (you may have to temporarily disable the feature to find one accurately; we're looking for a post whose reblog menu has a reblog link href that doesn't perfectly match `/reblog/${name}/${id}/${reblogKey}`
    - **Expected result**: The reblog form is opened, without the reblog menu being visible for even a single frame

### DIFF
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -186,11 +186,7 @@ export const styleElement = buildStyle(`
   span:has(${quickActionsSelector}) > .${reblogLinkClass} {
     display: none;
   }
-
   .${reblogLinkClass} ~ ${reblogButtonSelector}:not(:has(${quickActionsSelector})) {
-    display: none;
-  }
-  body:has(.${reblogLinkClass}) > ${reblogMenuPortalSelector}:not(:has([role="menu"][aria-labelledby])) {
     display: none;
   }
 `);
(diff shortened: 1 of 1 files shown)


marcustyphoon:
- needs to be tested post #2222 merge
- why in the actual hell does opening your blog in modal, viewing it in masonry, and clicking to reblog https://www.tumblr.com/april/813856070160236544 result in the blog view modal disappearing, no post form, and the page not being scrollable. what

marcustyphoon (changes requested):
It turns out that `#glass-container:has(div[role="menu"] /* */)` may be a bad selector actually.

marcustyphoon:
re: the last commit... huh. did we basically just forget to remove that in #1931?

marcustyphoon (approved):
Smoke tested, but as in like, "stood on the porch sniffing around for three seconds, said 'well I don't smell anything,' and went back inside."

AprilSylph:
> re: the last commit... huh. did we basically just forget to remove that in #1931?

Not really, I just didn't do any refactoring with c4a24d51245e7662234ec4d60afcef481eeea6c8 and bbf5fe6a084b12776ba2445a1f3fdb6dfd124bdf.

- Previously: main stylesheet hides menus for processed buttons on a global basis, per-link stylesheets hide already-open menus on a per-button basis
- Now: Per-link stylesheets hide all menus on a per-post basis

PR #2229, opened 2026-04-20
refactor: remove `"iframe"` preference type

AprilSylph:
### Description

Removes all traces of the `"iframe"` preference type in favour of the `"component"` preference type.

This includes the deletion of the now-unused `src/action/resize_frames.js`. It's kind of embarrassing that we submitted two versions with an unused file; hopefully Mozilla doesn't reject v1.3.1 for this sin.

### Testing steps

Load the modified addon and ensure that PostBlock and Quick Tags both still work correctly, both in the control panel and on Tumblr.


marcustyphoon (approved):

marcustyphoon:
> It's kind of embarrassing that we submitted two versions with an unused file; hopefully Mozilla doesn't reject v1.3.1 for this sin.

I can't be 100% sure if this was a joke or not, but just in case, I will note that if having code for a preference type that we don't currently have any of is considered a problem, I'd consider that a problem with Mozilla review. That definitely isn't, as the policy says, "redundant code or files".

PR #2230, opened 2026-04-24
fix: prevent unbounded recursion on boot

AprilSylph:
### Description

A theoretical improvement. Caps the maximum number of boot attempts at 3600, which is a very safe one minute on a 60Hz display, decreasing as refresh rate increases.

I think it's very safe to assume that devices able to output 360 frames per second can load React in a lot less than 10 seconds.

I would be very interested in revisiting #1971 after this.

### Testing steps

1. Load the modified addon
2. Enable some features, ideally ones that take visible effect immediately upon activating
3. Open a Tumblr tab
    - **Expected result**: XKit boots normally
4. Open devtools &rarr; Network and enable throttling
5. Reload the Tumblr tab
    - **Expected result**: The tab is sloooooow to load
    - **Expected result**: When the tab finally does load, React loads normally
    - **Expected result**: XKit boots normally after this
    (Proving that `"run_at": "document_end"` is working as intended)

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -1,6 +1,8 @@
 'use strict';
 
 {
+  const MAX_BOOT_ATTEMPTS = 3600; // 60 seconds on 60Hz displays; 10 seconds on 360Hz displays
+
   const enabledFeaturesKey = 'enabledScripts';
 
   const redpop = [...document.scripts].some(({ src }) => src.includes('/pop/'));
@@ -8,8 +10,7 @@
 
   const restartListeners = {};
 
-  // prevent referencing outdated resources after firefox extension update/restart
-  const timestamp = Date.now();
+  const timestamp = Date.now(); // Prevent referencing outdated resources after Firefox extension update/restart
 
   const runFeature = async function (name) {
     const {
@@ -121,7 +122,7 @@
     ]);
 
     /**
-     * fixes WebKit (Chromium, Safari) simultaneous import failure of files with unresolved top level await
+     * Fixes WebKit (Chromium, Safari) simultaneous import failure of files with unresolved top level await
      * @see https://github.com/sveltejs/kit/issues/7805#issuecomment-1330078207
      */
     await Promise.all(['css_map', 'language_data', 'user'].map(name => import(browser.runtime.getURL(`/utils/${name}.js`))));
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Ooh, good idea. I think the chances that this is addressing a thing which is in fact happening are low, but good safety.

marcustyphoon:
Some alternative (non-recursive) forms (not tested, may have typos):

```js
  const waitForReactLoaded = async () => {
    for (let attempts = 0; attempts < MAX_BOOT_ATTEMPTS; attempts++) {
      if (isReactLoaded()) return;
      await new Promise(window.requestAnimationFrame);
    }
    throw new Error('XKit Rewritten boot failed; React did not load after 10+ seconds.');
  };

  if (redpop) {
    waitForReactLoaded()
      .then(init)
      .catch(console.error);
  }
```

```js
  const waitForReactLoaded = async () => {
    let attempts = 0;

    while (!isReactLoaded()) {
      if (++attempts <= MAX_BOOT_ATTEMPTS) {
        throw new Error('XKit Rewritten boot failed; React did not load after 10+ seconds.');
      }
      await new Promise(window.requestAnimationFrame);
    }
  };

  if (redpop) {
    waitForReactLoaded()
      .then(init)
      .catch(console.error);
  }
```

edit: we could also call await waitForReactLoaded() in init if that feels cleaner.

marcustyphoon:
(How do I feel about this commit coauthor and its legal interaction with the gpl license? I'm officially Not Sure.)

marcustyphoon (approved):
Well... I can't get isReactLoaded() to ever return false in practice, so this code pah is untested. I hacked it to return false 90% of the time and it seems to work fine, so... success?

AprilSylph:
> I hacked it to return false 90% of the time and it seems to work fine, so... success?

Did the same for the latest commit; turns out you can't use `window.requestAnimationFrame` as a generic callback, it errors with a complaint about being called on an object that doesn't implement `Window`.

marcustyphoon:
Interesting. I guess I've always used plain requestAnimationFrame. Alternatively, if that also doesn't work, I've hardcore mandala effected myself.

marcustyphoon:
(Surely it's not a module vs script thing?)

AprilSylph:
* Works: `await new Promise((resolve) => window.requestAnimationFrame(resolve));`
* Works: `await new Promise(requestAnimationFrame);`
* Breaks: `await new Promise(window.requestAnimationFrame);`

PR #2236, opened 2026-04-29
feat(UI): add link to `tumblr.com/addons` to Links tab

AprilSylph:
### Description

-|-|-|-
Screen Shot 2026-04-29 at 10 28 49 | Screen Shot 2026-05-20 at 16 01 29 | Screen Shot 2026-04-29 at 10 28 52 | Screen Shot 2026-05-20 at 16 01 33

I landed on this as the best title/text verbiage combo I could think up. Title is worded like a CTA; link text matches the tab title of the linked page; the word "Tumblr" isn't repeated. Do you have any suggestions for improving it?

On my machine, this makes the Links tab just a bit too big to be contained within the Firefox popup, and allows me to scroll when I couldn't before. Is it worth removing any of the other links to prevent this? I'm leaning towards "no", but it is worth noting that the "Related projects" links are essentially duplicated between this panel and the pinned post on `@addons`, so I could be persuaded towards "yes".

### Testing steps

Share your opinions on the above points; approve when you're happy with the current changes.

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -123,6 +123,12 @@
       </section>
       <section id="links-panel" role="tabpanel" aria-labelledby="links-tab" tabindex="0" hidden>
         <section>
+          <h3>Follow the blog</h3>
+          <ul>
+            <li><a href="https://www.tumblr.com/addons" target="_blank">@addons on Tumblr</a></li>
+          </ul>
+        </section>
+        <section>
           <h3>Help & information</h3>
           <ul>
             <li><a href="https://github.com/AprilSylph/XKit-Rewritten/wiki" target="_blank">User guide</a></li>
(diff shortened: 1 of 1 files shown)


marcustyphoon:
> On my machine, this makes the Links tab just a bit too big to be contained within the Firefox popup, and allows me to scroll when I couldn't before.

We could cut some padding. Or, I dunno, do `Issue Tracker (Bugs / Features)`.

marcustyphoon (approved):
The Acorn thing doesn't let us move the huge bottom margin from the ul (??? ok) to the section so it could be `section:not(:last-child)`, I assume.

My honest opinion other than what I've said above is "there is almost certainly a way to improve this, but I don't have a thread to tug on about what precisely it is, and my strong belief that there are certain things you should try to avoid shipping and changing later classifies this as distinctly _not_ one of those times.

AprilSylph:
> The Acorn thing doesn't let us move the huge bottom margin from the ul (??? ok)

I should change that; it's not _really_ Acorn styling, it's custom styling that uses Acorn tokens.

---

Agh, but we kinda rely on those styles being there in other components. Nevermind.

AprilSylph:
Decided to cut the bug/enhancement issue label links entirely. The issue tracker link is more intended for prospective devs anyway, and we get duplicate bug reports for emergent issues consistently regardless. Updated screenshots to reflect.

PR #2254, opened 2026-05-18
chore: Improve utility documentation

marcustyphoon:
### Description

Adds additional documentation to many utility module exports.

Please feel free to commit improvements directly to this branch if appropriate, cherry pick these, etc, April.

### Testing steps

n/a

### DIFF
--- a/src/utils/memoize.js
+++ b/src/utils/memoize.js
@@ -1,9 +1,9 @@
 /**
  * Create a version of a function that returns results for repeated calls with
  * the same argument from a cache, improving performance.
- * @template T, R
- * @param {function(T): R} func A function with one argument
- * @returns {function(T): R} A memoized version of the function
+ * @template {function (T): R} Callback, T, R
+ * @param {Callback} func A function with one argument
+ * @returns {Callback} A memoized version of the function
  */
 export const memoize = (func) => {
   const cache = new Map();
--- a/src/utils/memoize.js
+++ b/src/utils/memoize.js
@@ -1,3 +1,10 @@
+/**
+ * Create a version of a function that returns results for repeated calls with
+ * the same argument from a cache, improving performance.
+ * @template T, R
+ * @param {function(T): R} func A function with one argument
+ * @returns {function(T): R} A memoized version of the function
+ */
 export const memoize = (func) => {
   const cache = new Map();
   return (arg) => {
--- a/src/utils/modals.js
+++ b/src/utils/modals.js
@@ -35,6 +35,9 @@ export const showModal = ({ title, message = [], buttons = [] }) => {
   modalElement.focus();
 };
 
+/**
+ * Hide the current takeover prompt.
+ */
 export const hideModal = () => {
   document.getElementById('xkit-modal')?.remove();
   lastFocusedElement?.focus();
@@ -44,6 +47,10 @@ export const hideModal = () => {
 export const modalCancelButton = dom('button', null, { click: hideModal }, ['Cancel']);
 export const modalCompleteButton = dom('button', { class: 'blue' }, { click: hideModal }, ['OK']);
 
+/**
+ * Show a takeover prompt to the user with details about a thrown exception.
+ * @param {Error} exception Thrown exception to describe
+ */
 export const showErrorModal = exception => {
   console.error('XKit Rewritten error:', exception);
 
@@ -63,6 +70,12 @@ export const showErrorModal = exception => {
   });
 };
 
+/**
+ * Create a version of a function that shows a takeover prompt to the user if it throws.
+ * @template T, R
+ * @param {function(...T): R} func A function
+ * @returns {function(...T): R} A version of the function with explicit error handling
+ */
 export const withModalOnError = func =>
   async (...args) => {
     try {
--- a/src/utils/modals.js
+++ b/src/utils/modals.js
@@ -72,9 +72,9 @@ export const showErrorModal = exception => {
 
 /**
  * Create a version of a function that shows a takeover prompt to the user if it throws.
- * @template T, R
- * @param {function(...T): R} func A function
- * @returns {function(...T): R} A version of the function with explicit error handling
+ * @template {function (...T): R} Callback, T, R
+ * @param {Callback} func A function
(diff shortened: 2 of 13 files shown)


marcustyphoon:
@AprilSylph what I ran into when working on this, and helpwanted this for, is potentially related to #2263/#2264: what, precisely, is the isMyPost utility supposed to represent? I recall us putting some thought into it, but couldn't find the discussion. My best read on it is that it identifies posts the user has *had a direct hand in authoring*, I guess? We should clearly document this (and should double check whether it's the right behavior where it's used).

src/utils/react_props.js
```diff
@@ -36,6 +36,10 @@ export const notificationObject = notificationElement => {
  */
 export const blogData = async (meatballMenu) => inject('/main_world/unbury_blog.js', [], meatballMenu);
 
+/**
+ * @param {Element} postElement An on-screen post element
+ * @returns {Promise<boolean>}
```
AprilSylph:
```suggestion
 * @returns {Promise<boolean>} Resolves to `true` if the post could reasonably be assumed to be published by the current user.
```

```suggestion
 * @returns {Promise<boolean>} Resolves to `true` if the post could be reasonably assumed to be published by the current user.
```

Not sure which of these is more grammatically correct...

marcustyphoon: "to have been published by"?

AprilSylph:
```suggestion
 * @returns {Promise<boolean>} Resolves to `true` if the post can reasonably be assumed to have been published by the current user.
```

```suggestion
 * @returns {Promise<boolean>} Resolves to `true` if the post can be reasonably assumed to have been published by the current user.
```

marcustyphoon: "Can reasonably be assumed" definitely sounds like the more common/correct phrasing to me, even though I think both are grammatically correct.

~~Random different example: "it can quickly be done"/"it can be quickly done"... nope, that's not equivalent; "it can be done quickly" is an option that reads as the same here but "if the post can be assumed reasonably to have been" does not read the same.~~

I guess it's not actually a matter of grammar rules, it's a matter of clarity. "Be assumed to have been published" is fairly complex as a construction; I think it's easier to follow if it's not further broken up. Or more directly, if I'm midsentence, "the post can reasonably be assumed..." makes me strongly identify "be assumed" as part of a passive-voice construction; "the post can be reasonably assumed" makes me think we're done with the sentence and that "assumed" is a verb referring to _the post_, which makes no sense until I read the rest of the sentence and backtrack.

If I were trying to force simple english rules, I guess I might go "Resolves to `true` if the post was published by the current user (or if it can reasonably be assumed that it was)." Not that that resolves the word order question in any way, there are just as many valid permutations; it just puts it in a location where the meaning of the sentence is clear way before you get there.

src/utils/memoize.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * Create a version of a function that returns results for repeated calls with
+ * the same argument from a cache, improving performance.
+ * @template T, R
+ * @param {function(T): R} func A function with one argument
+ * @returns {function(T): R} A memoized version of the function
```
AprilSylph: I get more accurate Intellisense with the following:


```suggestion
 * @template {function (T): R} Callback, T, R
 * @param {Callback} func A function with one argument
 * @returns {Callback} A memoized version of the function
```

Before:
> `const isAnimated: (arg0: any) => Promise<boolean>`

After:
> `const isAnimated: (sourceUrl: any) => Promise<boolean>`

src/utils/modals.js
```diff
@@ -63,6 +70,12 @@ export const showErrorModal = exception => {
   });
 };
 
+/**
+ * Create a version of a function that shows a takeover prompt to the user if it throws.
+ * @template T, R
+ * @param {function(...T): R} func A function
+ * @returns {function(...T): R} A version of the function with explicit error handling
```
AprilSylph: Same here:

```suggestion
 * @template {function (...T): R} Callback, T, R
 * @param {Callback} func A function
 * @returns {Callback} A version of the function with explicit error handling
```

Before:
> ```ts
> (alias) withModalOnError<{
>  currentTarget: any;
> } | undefined, Promise<void>>(func: (...arg0: ({
>  currentTarget: any;
> } | undefined)[]) => Promise<void>): (...arg0: ({
>  currentTarget: any;
> } | undefined)[]) => Promise<void>
> ```

After:
> ```ts
> (alias) withModalOnError<({ currentTarget: controlButton }: {
>  currentTarget: any;
> }) => Promise<void>, any, any>(func: ({ currentTarget: controlButton }: {
>  currentTarget: any;
> }) => Promise<void>): ({ currentTarget: controlButton }: {
>  currentTarget: any;
> }) => Promise<void>
> ```

src/utils/tumblr_helpers.js
```diff
@@ -68,8 +68,18 @@ export const isNpfCompatible = postData => {
   return isBlocksPostFormat || shouldOpenInLegacy === false;
 };
 
+/**
+ * Performs a "soft" navigation within Tumblr's single-page-application.
+ * @see https://github.com/tumblr/docs/blob/master/web-platform.md#navigate
+ * @param {string} location Path to navigate to
+ * @returns {void}
```
AprilSylph: `inject()` is always async, so this should actually be `Promise<void>`, right?

It doesn't look like the `tumblr.inject()` function actually waits for anything to be loaded before returning, either, which seems like a good detail to have.

```suggestion
 * @returns {Promise<void>} Resolves when the navigation has been requested; does _not_ wait for the target route chunk to load
```

marcustyphoon: Hmm, true. That, or we could curly-brace the function body so it's not returning a promise that's pointless to wait for (though we should still probably document why we did that).

Except, no, I just tested it: location.href doesn't change synchronously or in a queueMicrotask callback; it **does** change once you await the navigate call promise (though there is no visible change to the page at that point). So it's a meaningful promise and there's no reason to just eat it unconditionally.

AprilSylph (approved):

PR #2255, opened 2026-05-19
feat(Vanilla Audio, Vanilla Video): allow fine-tuning of default volume

### LINKED ISSUE #1797: Vanilla Audio/Vanilla Video default setting options
### Discussed in https://github.com/AprilSylph/XKit-Rewritten/discussions/1314

<sup>Originally posted by **Qo1697** October 24, 2023</sup>
The 'Vanilla Audio' and 'Vanilla Video' features both let you change the default volume that video/audio is set to, which is excellent.
However, the only options given are 25%/50%/75%/100%.
image

The smallest of these is still too loud for me on my current setup. Being able to enter a custom value there would be ideal, but if not I'd appreciate having 10% as an option on the dropdown.</div>

theIncandescentAnteater:
### Description

Fixes #1797 

Also adds "percent" input type, which takes only integers 0-100.

| Before | After|
|--------|--------|
| image | image |
| image | image |

### Testing steps

1. Load the modified addon
2. Enable Vanilla Audio
3. Set Default Volume
Expected behavior: Non-digit characters are not displayed. Numbers over 100 get rounded down to 100. When no digits are entered, the value should set to 0.
4. Find an audio post
Expected behavior: The volume slider should be at the entered percentage

2. Enable Vanilla Video
3. Set Default Volume
Expected behavior: Non-digit characters are not displayed. Numbers over 100 get rounded down to 100. When no digits are entered, the value should set to 0.
4. Find a video post
Expected behavior: The volume slider should be at the entered percentage

### DIFF
--- a/src/action/components/percent-preference/index.js
+++ b/src/action/components/percent-preference/index.js
@@ -1,0 +1,72 @@
+import { CustomElement, fetchStyleSheets } from '../index.js';
+
+const localName = 'percent-preference';
+
+const templateDocument = new DOMParser().parseFromString(`
+  <template id="${localName}">
+    <label for="percent"></label>
+    <input id="percent" type="text" size="28" spellcheck="false">
+  </template>
+`, 'text/html');
+
+const adoptedStyleSheets = await fetchStyleSheets([
+  '/lib/modern-normalize.css',
+  '/action/acorn.css',
+  './index.css',
+].map(import.meta.resolve));
+
+class PercentPreferenceElement extends CustomElement {
+  /** @type {string} */ featureName;
+  /** @type {string} */ preferenceName;
+
+  /** @type {HTMLInputElement} */ #inputElement;
+  /** @type {HTMLLabelElement} */ #labelElement;
+  /** @type {ReturnType<Window['setTimeout']>} */ #timeoutID;
+
+  constructor () {
+    super(templateDocument, adoptedStyleSheets);
+
+    this.#inputElement = this.shadowRoot.getElementById('percent');
+    this.#labelElement = this.shadowRoot.querySelector('label[for="percent"]');
+  }
+
+  /** @param {string} label Label displayed to the user to describe the preference. */
+  set label (label) { this.#labelElement.textContent = label; }
+  get label () { return this.#labelElement.textContent; }
+
+  /** @param {string} value The saved or default value of this preference. */
+  set value (value = '') { this.#inputElement.value = value; }
+  get value () { return this.#inputElement.value; }
+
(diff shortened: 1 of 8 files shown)


src/lib/modern-normalize.css
```diff
@@ -128,7 +128,8 @@ button,
 input,
 optgroup,
 select,
-textarea {
+textarea,
+percent {
```
AprilSylph: We can't modify vendored files, as doing so violates Firefox's addon review policies for source code.

(This change is a no-op anyway, since `percent` elements don't exist; your custom element type is `percent-preference`.)

```suggestion
textarea {
```

AprilSylph (commented):
Thanks for showing an interest in the project! We haven't had the pleasure of welcoming a new contributor in a while, so please enjoy your stay 😃 

I should be able to give this a proper review tomorrow.
For now, I'll just address this low-hanging fruit:

src/action/components/percent-preference/index.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+customElements.define(localName, PercentPreferenceElement);
+
+/**
+ * @typedef PercentPreferenceProps
+ * @property {string} featureName The feature's internal name (e.g. `"quick_tags"`).
+ * @property {string} preferenceName The preference's internal name (e.g. `"originalPostTag"`).
+ * @property {string} label The preference's label (e.g. `"Original post tag"`).
```
AprilSylph: The JSdoc for each preference component should ideally use an example of its own preference type:


```suggestion
 * @property {string} featureName The feature's internal name (e.g. `"vanilla_audio"`).
 * @property {string} preferenceName The preference's internal name (e.g. `"defaultVolume"`).
 * @property {string} label The preference's label (e.g. `"Default Volume"`).
```

AprilSylph (commented):
One more comment/suggestion to be addressed before merging.

There are further changes I would like to make to this before it hits users, but they're all a matter of my preference and not actually to do with the correctness of this PR, so I won't consider them blocking.

AprilSylph (approved):
Works as described; thanks for your contribution!

This PR will be listed in the patch notes of the next release, so you'll receive a GitHub mention notification when that goes out, and you'll be listed as a contributor on the release ([example](https://github.com/AprilSylph/XKit-Rewritten/releases/tag/v1.2.3)).

I'm not sure when the next release will happen, but I should probably make it soon-ish for the Panorama fixes... and then of course, there'll be an announcement on [@addons](https://www.tumblr.com/addons) when the patch is available for download on all browsers.

I hope the development/review process has been pleasant enough for you—thank you again!

PR #2257, opened 2026-05-19
feat(Panorama): Widen misc informative popups

marcustyphoon:
### Description

Panorama didn't widen this element (on "channel" pages). Now it does!

### Testing steps

Code review only is probably fine if you don't have one of these on any of your blogs.

### DIFF
--- a/src/features/panorama/index.js
+++ b/src/features/panorama/index.js
@@ -112,8 +112,8 @@ ${keyToCss('cell')}, ${postSelector}
   height: unset !important;
 }
 
-/* Fix ad containers */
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Random thought: maybe what I should be doing is imitating https://github.com/AprilSylph/Palettes-for-Tumblr/pull/233, systematically grabbing selectors that set `max-width: 556px` and overriding them if they're inside a timeline.

AprilSylph (approved):
Can't test this myself; assuming you did.

marcustyphoon:
Yep, the screenshot is of the widened element!

PR #2258, opened 2026-05-21
fix(Mutual Checker): hide posts from non-followed blogs in mutuals-only mode

### LINKED ISSUE #1209: Mutual Checker: Only-mutuals option may be unreliable
### Platform

MacOS 13.5 arm64

### Browser

Chrome

### Addon version

master

### Details

This `else if` statement is preceded by a bunch of early returns. If one of those early returns triggers, the else block will never occur. Mutual Checker thus doesn't hide posts from blogs you aren't following.

```js
const processPosts = function (postElements) {
    if (alreadyProcessed(postElement)) return;

    /* ... */

    if (!followingBlog) { return; }

    /* ... */
 
    if (isMutual) {
      postElement.classList.add(mutualsClass);
      postAttribution.prepend(createIcon(blogName));
    } else if (showOnlyMutuals) {
      postElement.classList.add(hiddenClass);
    }
  });
};
```

theIncandescentAnteater:
### Description

Fixes #1209 
Hides posts from non-mutuals (specifically posts without attribution, without a blog name, or that the user isn't following) that would've previously gone unfiltered

| Before | After |
|--------|--------|
| image | image | 

### Testing steps

1. Load the **un**modified addon
2. Enable Mutual Checker
3. Check only show posts from mutuals
4. Find post from non-mutual (may require following a tag to see "because you follow x") 
6. Load modified add on
7. See that the post is hidden on the dash

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -77,16 +77,19 @@ const addIcons = function (postElements) {
     if (alreadyProcessed(postElement)) return;
 
     const postAttribution = postElement.querySelector(postAttributionSelector);
-    if (postAttribution === null) { return; }
+    const blogName = postAttribution?.textContent.trim();
 
-    const blogName = postAttribution.textContent.trim();
-    if (!blogName) return;
+    if (postAttribution === null || !blogName) {
+      if (showOnlyMutuals) {
+        getTimelineItemWrapper(postElement)?.setAttribute(hiddenAttribute, '');
+      }
+      return;
+    }
 
     const followingBlog = await getIsFollowing(blogName, postElement);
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
Works as described. I do feel like this can be refactored to prevent repeating the attribute-setting line of code, but how exactly to do that is not obvious enough for me to simply leave a suggestion; I'll play with it in my editor.

AprilSylph:
Optional chaining on the definition of `blogName` is really the sauce here; by allowing it to be falsy and adjusting the code to accommodate that, we basically end up with the same steps in the non-mutuals scenario, but without any early-returns.

PR #2259, opened 2026-05-21
feat(UI): use `number` input type for percent-type preferences

AprilSylph:
<sub>cc @theIncandescentAnteater</sub>

### Description

- see https://github.com/AprilSylph/XKit-Rewritten/pull/2255#pullrequestreview-4326961276

Refactors the new `percent`-type preference to use number-type inputs. This adds increment/decrement buttons, <kbd>▲</kbd> and <kbd>▼</kbd> shortcuts, and allows us to offload validation to the user agent.

Invalid/empty values are no longer refused by the input, but are not saved; instead, the element is just put into a visibly invalid state. This results in slightly better UX when typos occur; typing a letter in between two digits of a valid number no longer erases the proceeding numbers, and <kbd>CTRL</kbd>+<kbd>Z</kbd>/<kbd>Command</kbd>+<kbd>Z</kbd> can now undo input clearing.

I assume using a number-type input is also better for mobile compatibility, since the browser should tell the OS to bring up a number-focused keyboard. That's just an assumption, though.

-|-|-|-
Screen Shot 2026-05-21 at 11 28 42 | Screen Shot 2026-05-21 at 11 29 09 | Screen Shot 2026-05-21 at 11 28 46 | Screen Shot 2026-05-21 at 11 29 13

https://github.com/user-attachments/assets/f664f7e8-513a-44e4-92f7-bbef03ceaf2e

This doesn't make the preference save as a number, which I think is fine. We could change this if we wanted to, by reading `this.#inputElement.valueAsNumber` instead of `this.#inputElement.value`, but existing installations have these preferences saved as strings already, and they're working just fine. Maybe it would be worth doing if we anticipated a need to use this preference type in more features.

Interestingly, we don't explicitly parse these values as numbers in Vanilla Audio/Vanilla Video... they just get cast as numbers when they get divided by 100. Apparently, leading zeroes cause no issues with this.

### Testing steps

1. Load the modified addon in a fresh profile
2. Open Vanilla Audio and Vanilla Video's preferences
    - **Expected result**: Both features' default volume is 100
3. Click one of the "Default Volume" preferences to focus it
4. Press <kbd>▲</kbd>/<kbd>▼</kbd>, or use the on-screen buttons, to adjust the value
    - **Expected result**: The value cannot go above 100
    - **Expected result**: The value cannot go below 0
5. Use your keyboard to type values, including invalid and empty values
    - **Expected result**: Values of 0-100 are accepted; the change is reflected in the Backup tab
    - **Expected result**: Numbers above 100 or below 0 are not accepted; the input border/outline turns red, and the change is not saved
    - **Expected result**: Non-number characters are not accepted
    - **Expected result**: An empty value is not accepted

### DIFF
--- a/docs/Chapter 4.1 - Feature metadata.md
+++ b/docs/Chapter 4.1 - Feature metadata.md
@@ -67,7 +67,7 @@ It is recommended to use camelCase for each preference name, so that the feature
 - Type: String
 - Required: Yes
 
-Type of preference. Supported values: `"checkbox"`, `"text"`, `"color"`, `"select"`, `"textarea"`, `"component"`, `"percent"`
+Type of preference. Supported values: `"checkbox"`, `"color"`, `"component"`, `"percent"`, `"select"`, `"text"`, `"textarea"`
 
 #### `"preferences"`: \<preference name\>: `"label"`
 - Type: String
@@ -97,10 +97,10 @@ Unused for other preference types.
 Default value of the preference to display to the user.
 
 If the preference `type` is `"checkbox"`, this value should be a boolean.  
-If the preference `type` is `"text"` or `"textarea"`, this value should be a string.  
-If the preference `type` is `"percent"`, this value should be a string representation of an integer between 0 and 100.  
 If the preference `type` is `"color"`, this value should either be a string representing a hexadecimal colour code (i.e. `"#1a2b3c"`) or an empty string.  
-If the preference `type` is `"select"`, this value should be a string that matches one of the `"options"` item's `"value"`.
+If the preference `type` is `"percent"`, this value should be an integer in the range of 0–100.  
+If the preference `type` is `"select"`, this value should be a string that matches one of the `"options"` item's `"value"`.  
+If the preference `type` is `"text"` or `"textarea"`, this value should be a string.
 
 #### `"preferences"`: \<preference name\>: `"inherit"`
 - Type: String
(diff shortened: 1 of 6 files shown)


marcustyphoon:
> I assume using a number-type input is also better for mobile compatibility, since the browser should tell the OS to bring up a number-focused keyboard.

It doesn't in iOS Safari, but there does in fact appear to be a way to fix this! https://catskull.net/ios-inputmode.html

marcustyphoon:
I'm not sure if we should keep the debounce? It can be a little strange if you backspace out the value 50, do nothing, and get a saved value of 5. This is quite minor, of course.

marcustyphoon:
Notably, this does let you set and save the string value `5e1`, which is amusing. As you noted, implicit casting makes this work fine. Maybe we want to be explicit about this one way or the other here to set a precedent for future uses?

marcustyphoon (approved):
Works as described!

AprilSylph:
> Notably, this does let you set and save the string value `5e1`, which is amusing. As you noted, implicit casting makes this work fine. Maybe we want to be explicit about this one way or the other here to set a precedent for future uses?

You know what, yeah, I actually really do want to normalise the value here.

AprilSylph:
6. Put in weird values (e.g. leading zeroes, scientific notation) that evaluate to valid integers
    - **Expected result**: The Backup tab shows the evaluated integer, even while the element shows your weird input
7. Close and reopen the control panel
    - **Expected result**: The input shows the stored number, without any trace of your weird input

marcustyphoon (approved):

PR #2260, opened 2026-05-21
fixes panorama on tagged pages and liked feed

### LINKED ISSUE #1785: Panorama: Nonfunctional on tagged pages; likes; community posts feed
### Platform

MacOS 14.7.1 arm64

### Browser

Chrome 135.0.7049.115

### Addon version

master branch as of #1291

### Details

As of the linked pull request (which will be in, presumably, v1.0.10 when it releases), the Panorama feature is back! However, as noted in the PR< it does not currently work, but probably should...

- on tagged pages (https://www.tumblr.com/tagged/xkit)
- on the "community posts" feed (https://www.tumblr.com/communities)
- on individual community posts
- on the likes page

theIncandescentAnteater:
### Description

i believe this fixes #1785. as far as i can see, the extension already works on the community feeds.

the classes on the tagged page differ from the dash, but the liked page doesn't have a main element at all

| Before | After |
|--------|--------|
| image | image |
| image | image | 

### Testing steps

1. Load the modified addon
2. Go to a [tagged page](https://www.tumblr.com/tagged/xkit)
3. See the wide post width
4. Go to a [the likes feed](https://www.tumblr.com/likes)
3. See the wide post width

### DIFF
--- a/src/features/panorama/index.js
+++ b/src/features/panorama/index.js
@@ -62,6 +62,13 @@ ${mainContentWrapper} > div > div:has(> div > ${keyToCss('timeline')}) {
 ${mainContentWrapper} > div > div > div:has(> ${keyToCss('timeline')}) {
   max-width: unset;
 }
+
+${keyToCss('grid')}:has(${keyToCss('layoutBody')}) {
+  grid-template-columns: 1fr auto;
+}
+${keyToCss('grid')} ${keyToCss('layoutBody')} {
+  max-width: var(${maxPostWidthVar});
+}
 `);
 mainStyleElement.media = `(min-width: ${widenDashMinWidth}px)`;
 
--- a/src/features/panorama/index.js
+++ b/src/features/panorama/index.js
@@ -51,6 +51,7 @@ ${mainPostColumn} > ${keyToCss('postColumn')} {
 
 ${mainContentWrapper}:has(> div > div > div > ${keyToCss('timeline')}) {
   flex-grow: 1;
+  max-width: calc(var(${maxPostWidthVar}) + ${sidebarOffset}px);
 }
 ${mainContentWrapper} > div:has(> div > div > ${keyToCss('timeline')}) {
   max-width: unset;
--- a/src/features/panorama/index.js
+++ b/src/features/panorama/index.js
@@ -63,11 +63,20 @@ ${mainContentWrapper} > div > div > div:has(> ${keyToCss('timeline')}) {
   max-width: unset;
 }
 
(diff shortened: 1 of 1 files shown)


marcustyphoon:
For clarity, the communities bullet point in the linked issue points specifically to https://www.tumblr.com/communities, the single "community posts" feed that contains posts from all of your communities, and individual community posts viewed alone! Panorama does indeed already work in individual community feeds.

marcustyphoon:
Setting a maximum post width (`maxPostWidthVar`) is working here on tagged pages, but not on the liked page!

theIncandescentAnteater:
thanks! i wasn't in any communities so i didn't even realize that was it's own feed

the max post width on the communities page is a little weird, but i couldn't figure out a better way. there very well may be one though.

this is what it looks like with a max width less than the full width (the grid column is still max width, so the sidebars are still pushed all the way out). 
image

panorama on communities feed (without restrictive max post width)
image

marcustyphoon:
Hm, yeah, I bet there's a way to adjust the size natively. Tumblr's grid-template-columns value is `556px minmax(280px,360px)` before we override it; surely we can make our override the same thing but with `maxPostWidthVar` inserted?

src/features/panorama/index.js
```diff
@@ -45,6 +45,30 @@ ${keyToCss('queueSettings')} {
...
+  max-width: unset;
+}
+${mainContentWrapper} > div > div:has(> div > ${keyToCss('timeline')}) {
+  max-width: unset;
+}
+${mainContentWrapper} > div > div > div:has(> ${keyToCss('timeline')}) {
+  max-width: unset;
+}
+
+${keyToCss('grid')}:has(${keyToCss('layoutBody')}) {
```
marcustyphoon:
```suggestion
${keyToCss('grid')}:has(> ${keyToCss('layoutBody')}) {
```

I try to be as specific as is reasonably possible with CSS selectors, particularly when using :has(). This will only save the browser rendering engine work when it finds a `layoutBody` element (not common) and only has to check its direct parent rather than all of them, so not actually a big deal in practice for performance. But I think the clarity is good!

theIncandescentAnteater:
overriding with `maxPostWidthVar` meant that posts were becoming wider than the screen. i'm sure there's a neat way to do it with a grid, but i just ended up adjusting it to mimic the flex layout of the other feeds. 

image

marcustyphoon:
Hm, would something like `grid-template-columns: minmax(556px, var(${maxPostWidthVar})) minmax(280px,360px);` work?

PR #2264, opened 2026-05-27
fix(Mutual Checker): do not hide posts from own blogs in mutuals-only mode

### LINKED ISSUE #2263: Mutual Checker "only show posts from mutuals" hides your own post (arguably intended)
### Platform

macOS 14.8.7 arm64

### Browser

Firefox 150

### Addon version

v1.3.3

### Details

Mutual checker's "Only show posts from mutuals on the dashboard" option was fixed in #2258 to do precisely what it says: it **only** shows posts from mutuals, which means it hides your own posts, since you're not mutuals with yourself. Realistically, people may not have wanted this.

Notably, if we consider this an equivalent to Show Originals (i.e. #1973), that feature has the option to show your own reblogs (though it's disabled by default... which I don't actually agree with but that I presumably agreed to at the time, but that's beside the point).

I will reply to the posts I've seen expressing confusion at their own posts being gone whether they think this feature ought to apply to your own posts or not. Whatever they say, though, I think this is at the very least not ideal?

**edit:** See, I think you can make the argument that your own posts **are** logically posts from mutuals, insofar as you do see your own posts on the dashboard, intrinsically do follow yourself, and intrinsically are thus your own mutual. The intent of the checkbox is probably to hide—well, to temporarily filter out, hence the logic behind #1973*—posts by those you have an asymmetrical relationship with. You don't have an asymmetrical relationship with yourself. By that argument it simply makes sense to exclude own posts from being hidden, with no change to the option label or additional option.

*if you wanted to _permanently_ not see posts by non mutuals, you would simply unfollow them; this feature only actually makes sense as a toggle

AprilSylph:
### Description

- fixes #2263

Mutual Checker should do nothing to one's own posts. The relationship you have with your own blogs is not a relationship between you and another party, which is what the feature deals with.

If you wanted to hide posts from non-mutuals _and_ your own posts, you would use Mutual Checker &rarr; "Only show posts from mutuals on the dashboard" in combination with Tweaks &rarr; "Hide my own posts on the dashboard".

### Testing steps

1. Load the modified addon
2. Open the Tumblr dashboard
3. Enable Mutual Checker &rarr; "Only show posts from mutuals on the dashboard"
    - **Expected result**: Posts from blogs you are not following are hidden
    - **Expected result**: Posts from blogs you are following, but not mutuals with, are hidden
    - **Expected result**: Posts from blogs you are mutuals with are not hidden, and have mutual icons
    - **Expected result**: Posts from your own blogs are not hidden, and do not have mutual icons

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -4,7 +4,7 @@ import { buildStyle, getTimelineItemWrapper, filterPostElements, getPopoverWrapp
 import { translate } from '../../utils/language_data.js';
 import { onNewPosts, onNewNotifications, pageModifications } from '../../utils/mutations.js';
 import { getPreferences } from '../../utils/preferences.js';
-import { blogData, notificationObject, timelineObject } from '../../utils/react_props.js';
+import { blogData, notificationObject, timelineObject, isMyPost } from '../../utils/react_props.js';
 import { buildSvg } from '../../utils/remixicon.js';
 import { followingTimelineSelector } from '../../utils/timeline_id.js';
 import { apiFetch } from '../../utils/tumblr_helpers.js';
@@ -75,6 +75,9 @@ const alreadyProcessed = postElement =>
 const addIcons = function (postElements) {
   filterPostElements(postElements, { includeFiltered: true }).forEach(async postElement => {
     if (alreadyProcessed(postElement)) return;
+
+    const postIsMine = await isMyPost(postElement);
+    if (postIsMine) return;
 
     const postAttribution = postElement.querySelector(postAttributionSelector);
     const blogName = postAttribution?.textContent.trim();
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Stating the concern from the linked comment more explicitly: should this affect posts on group blogs you're not part of that aren't written by you? Currently it doesn't.

(Of course, you do have the ability to follow/unfollow your own sideblogs, so one could technically make some argument that you aren't inherently mutuals with them and they shouldn't count here, but a) I really don't think anyone would do that; if you did choose to follow your own sideblogs it's because you want to see their posts, and b) that's not what the PR does; either way, we need a new util.)

AprilSylph:
I'm so silly. We already have a `blog.isMember` check in `getIsFollowing` with presumably the intention of triggering the same early return that #2258 removed; the correct fix would be to move that up and early-return on it again.

marcustyphoon (approved):

marcustyphoon:
I suppose that this later revision makes the second paragraph of the description not entirely clean. But I think that's fine.

PR #2268, opened 2026-06-03
feat: update meatballs util to work on ariakit menus

### LINKED ISSUE #2256: Meatballs util: Update for new-style menus
Tumblr has launched new-style meatball menus for blogs and ads, with plans to do the same with posts.

We should be able to use the ad variant as a reference to anticipate how the HTML structure will be on posts, and update the meatballs util to activate on them.

Styling updates will also be needed.

AprilSylph:
### Description

- resolves #2256

This isn't too complicated in JS, since we already listen for essentially every menu and then test if it's rendered within a post header or blog header in the React tree.

I'm very proud of the CSS here. Instead of trying to override the `.xkit-meatball-button` styles, I've opted to use a different classname when we add an item to Ariakit menus, which means there is no need to do any overriding of base styles.

### Screenshots

Desktop | Mobile
-|-
Screen Shot 2026-06-04 at 09 41 19 | Screen Shot 2026-06-04 at 09 41 34

https://github.com/user-attachments/assets/9cf1140c-3961-4552-a77b-df54ba693c07

### Testing steps

Well, there are no new-style post menus yet, and no live XKit features to test the blog version of the util with, so:
1. Pull this branch locally
2. Either merge in the branch for an unreleased feature which does add blog menu items, or modify an existing feature to add post menu items to blog headers instead
3. Load the locally-modified addon and enable at least one feature that adds post menu items, and one feature that adds blog menu items
4. Open Tumblr and open a post's meatball menu
    - **Expected result**: The enabled features which add post menu items continue to work as before, with no changes to styling or behaviour
5. Open a blog view's meatball menu
    - **Expected result**: The enabled features which add blog menu items do so, with the items matching the new menu style
6. Switch to the site's mobile layout, so that the new-style menu turns into a bottom sheet
    - **Expected result**: The enabled features which add blog menu items also activate here, with the items matching the bottom sheet menu style
    - **Expected result**: The bottom sheet's "Nevermind" button is not pushed out of view


marcustyphoon:
I think it could make sense to factor this to use (for both menu types) `aria-labelledby`, as it's technically more robust against a different menu type originating from a header—

—oh, that only works on new meatball menus. I was going to say "and it's more elegant," but I guess not. Oh well, anyway, as I was saying, I don't think we have to include that here even if we decide we want to do it.

AprilSylph:
> I think it could make sense to factor this to use (for both menu types) `aria-labelledby`, as it's technically more robust against a different menu type originating from a header—
> 
> —oh, that only works on new meatball menus. I was going to say "and it's more elegant," but I guess not. Oh well, anyway, as I was saying, I don't think we have to include that here even if we decide we want to do it.

Yeah, I did explore this at first, here's the main world code I came up with for that:

<details><summary>Code snippet</summary>

```js
export default function getMenuButton () {
  /** @type {HTMLElement} */
  const menuElement = this;

  const reactKey = Object.keys(menuElement).find(key => key.startsWith('__reactFiber'));
  let fiber = menuElement[reactKey];

  while (fiber !== null) {
    const { stateNode } = fiber;

    // Modern approach: Find the element which declares itself as controlling this element.
    if (stateNode instanceof HTMLElement && stateNode.hasAttribute('aria-controls')) {
      const ariaControlsIds = stateNode.getAttribute('aria-controls').split(' ');
      const ariaControlsElements = ariaControlsIds.map(document.getElementById);
      if (ariaControlsElements.includes(menuElement)) {
        return stateNode;
      }
    }

    // Legacy approach: Find the closest popover wrapper containing a button element.
    if (
      stateNode instanceof HTMLElement &&
      stateNode.dataset.testid === 'controlled-popover-wrapper' &&
      stateNode.querySelector('button') !== null
    ) {
      // For return type parity with the modern approach, return the button element.
      return stateNode.querySelector('button');
    }

    fiber = fiber.return;
  }
}
```

</details>

marcustyphoon:
Ah, too bad the aria attribute's on the button, not on the menu. Presumably the implementable version of what I was imagining would be something like this, but I feel like this is probably worse?

```js
  const closestHeader = await getClosestRenderedElement(meatballMenu, 'header');
  const controllingButton = document.querySelector(`[aria-controls="${meatballMenu.id || 'INVALID'}"]`);

  if (controllingButton?.matches(`${postSelector} *`)) {
    addTypedMeatballItems({
      meatballMenu,
      type: 'post',
      reactData: await timelineObject(meatballMenu),
      reactDataKey: '__timelineObjectData',
    });
  } else if (closestHeader?.matches(postHeaderSelector)) {
    addTypedMeatballItems({
      meatballMenu,
      type: 'post',
      reactData: await timelineObject(meatballMenu),
      reactDataKey: '__timelineObjectData',
      mode: 'legacy',
    });
```

marcustyphoon:
ughhhh do we have to hack some ariakit component so that the mobile view version doesn't get cut off by the bottom of the screen?

edit: muahahahahahaha, got it

marcustyphoon (approved):

PR #2269, opened 2026-06-03
fix: force update of bottom sheet container height

marcustyphoon:
Stacked PR on #2268.

### Description
 

This fixes an issue where the mobile layout of the parent PR suffered from an issue where Tumblr's code did not take the added meatball items into account when calculating the bottom sheet container height.

Apparently this is queried in a react ref callback. Thus, this calls the ref callback again.

Obviously I would love a less convoluted way to do this. Any ideas? Notably, resizing the window also seems to update it, but I have no idea why.

(To quickly find the relevant code, search `\w\(Math.min\(\w.clientHeight,` in the Firefox debugger tab.)

### Screenshots

Before | After
-|-
<video src="https://github.com/user-attachments/assets/09c58abf-6d49-4c6f-8ae3-41154a6e1521" > | <video src="https://github.com/user-attachments/assets/e01de12c-37dc-4e4a-a4ca-1006f2e30933" >

### Testing steps

See parent PR.

### DIFF
--- a/src/main_world/update_bottom_sheet_container_height.js
+++ b/src/main_world/update_bottom_sheet_container_height.js
@@ -1,0 +1,14 @@
+export default function updateBottomSheetContainerHeight () {
+  const container = this;
+  const reactKey = Object.keys(container).find(key => key.startsWith('__reactFiber'));
+  const fiber = container[reactKey];
+
+  /**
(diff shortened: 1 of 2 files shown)


AprilSylph (approved):
Oh, that's beautiful.

Issue #2243, opened 2026-05-09
AccessKit Disable GIFs: Remove/update "download immediately/download on hover" feature


marcustyphoon:
It appears that Tumblr has changed the GIF-displaying code to no longer display a stationary poster element, then hide it once the animated GIF it is for has been loaded. Instead, they're swapped (or something; I can't quite tell) using javascript. Presumably whoever made this change ensured that there will be no flickering effect on image load when this happens...? I hope?

This breaks the XKit code, added in #1853, which a) makes disabling most GIFs quite a bit faster and more efficient and b) adds a feature to not download them until they're hovered over. Disable GIFs still works, though (using our old canvas-based code path that was used before this change for webp images).

I don't currently have any ideas how to keep the "don't download" feature if Tumblr keeps this change, but I'm leaving it in the issue title because, like, idk, never say never.

I think it probably makes sense to not assign or helpwanted this issue and let the stale bot serve as an automatic remindme. We should at least see if the change is an a/b or a mistake or something (it hasn't been announced in changes yet).

marcustyphoon:
Slightly weird update: as noted in #2247, Tumblr now has an option to disable media autoplay that pauses GIFs until you click them... and what's interesting is that their version (currently) waits to download the animated version until you click it. So we technically have the capability of implementing this toggle again... **if and only if the user has their Tumblr set to disable media autoplay**. (Specifically, we'd simulate clicking the image either when we process it, or on first hover).

However, unless we can hack the tumblr frontend a bit, the preference toggle would do nothing (it would effectively always be "immediately") if the user had media autoplay enabled, as is the default (usually), and we don't have the preference pane area to explain this. So—to my disappointment—I don't currently have a practical idea about how to keep the feature.

~~Side note: I just submitted feedback to Tumblr about this, but their feature is missing a loading indicator when you've clicked to unpause a GIF and it hasn't finished downloading, and it can take quite a long time and just looks like the GIF is not animated in the meantime. If Tumblr doesn't fix this, and we do build #2247's compatibility feature, I think we would actually potentially want to simulate a knightriderloader (ugh).~~ This has been fixed!

PR #2270, opened 2026-06-03
feat(AccessKit Disable GIFs): Implement native tumblr media pausing compatibility

### LINKED ISSUE #2247: AccessKit Disable GIFs: Feature doesn't override Tumblr's GIF pausing
### Platform

macOS 14 arm64

### Browser

Edge 148.0.3967.54

### Addon version

v1.3.2

### Details

Tumblr has a new feature where, if you have "reduce motion" or the equivalent enabled in your operating system settings and you're using a web browser that exposes this property to web pages, it pauses GIFs until you click on them. _**edit:** This can now be toggled in tumblr settings; see post below._

AccessKit's Disable GIFs feature currently isn't aware of this, so the Tumblr feature and the XKit feature overlap: images will appear paused with the AccessKit Disable GIFs label, and then when you mouse over them they'll still be paused and have the Tumblr-native GIF label. Once you click, Tumblr makes them animated and AccessKit Disable GIFs pause-on-hover works as expected after that.

I think that ideally we would make AccessKit Disable GIFs replace the Tumblr version of the feature with its own when it's enabled. As noted in #2243, though, this is difficult because Tumblr now swaps out the animated and non-animated versions using javascript rather than having them both accessible so we can easily choose which one to display ourselves. As a less sophisticated implementation, we could potentially have AccessKit Disable GIFs just click the paused image, but I don't love that because it's not reversible, and we do try to make the off switches for our features fully functional.

I have to acknowledge that one could make the argument that, as Tumblr has now duplicated one of our features, we should simply remove our version. If the Tumblr version was a toggle in the Tumblr settings, I could see this argument (and I would then point out that I disagree with it; unpause-on-hover is what users want). However, because the current version of the Tumblr feature is triggered by doing something that a) nobody knows about, b) has side effects nobody wants, but most importantly c) literally doesn't work in some browsers, I think this argument is pretty dead on arrival.

marcustyphoon:
### Description

This makes the AccessKit Disable GIFs feature override the Tumblr "disable media autoplay" feature if it's enabled. Resolves #2247.

Currently this does not address #2243: the "download paused gifs [immediately/on hover]" setting remains inoperative, equivalent to selecting "immediately."

### Screenshots

https://github.com/user-attachments/assets/fe3a1038-90b3-4080-affd-0ed0942284a8

### Testing steps

- Set the "media auto-play" setting in https://www.tumblr.com/settings/dashboard to "never", or to "use system" with your operating system set to reduce motion.
- Open a page like https://www.tumblr.com/search/gif/recent. Enable AccessKit Disable GIFs and confirm that posts stay paused, but switch to the XKit GIF labels, and that they immediately animate when moused over.
- Open a page like https://www.tumblr.com/search/gif/recent with AccessKit Disable GIFs already enabled and confirm that posts stay paused, switching quickly to the XKit GIF labels, and that they immediately animate when moused over.
- Repeat the above with webp images, like on https://www.tumblr.com/tagged/webp.
- Enable sufficient throttling in browser dev tools to observe this. Open a page like https://www.tumblr.com/search/gif/recent, scroll down, and confirm that this is the precise sequence of events when a gif (not a webp, unfortunately) is loaded:
  - The gradient placeholder appears, with a loading icon in the top right.
  - The stationary image appears, with an AccessKit GIF label in the top right, with no loading icon visible beneath it. If the image is hovered during this stage, the AccessKit GIF label disappears and is replaced with a loading icon in the top right.
  - Fully loaded: if not hovered, the stationary image appears, with an AccessKit GIF label in the top right. If hovered, the animated image appears, with nothing in the top right.


src/features/accesskit/disable_gifs.js
marcustyphoon: Ah, interesting: this `isAnimated` check doesn't work in blog view now (but does work outside of it). Race condition, I guess. (The only difference I know of is that blog view images aren't loading=lazy, and I don't see why that would cause this).

I guess the correct behavior would be to... assume that all webp images that are clickable are animated? Hm, but currently we click the image in a separate part of the code. That's annoying.

marcustyphoon: Aha, subtlety when solving this: the third image in https://www.tumblr.com/uyallstars/808763769014681600/pok%C3%A8mon-the-movie-2000 has the url https://44.media.tumblr.com/e1e034a13feb0825b20e5c3acb2d3424/999f368947654671-e8/s540x810_f1/0f8607b0a0a85332b5aaa7e0c76beb87844aaf07.gifv when paused, but is a .webp when unpaused.

marcustyphoon: https://github.com/marcustyphoon/XKit-Rewritten/commit/48678c9e68d80ca4c2677dbd6f9c6a9ca0e57189 write code, they said. it'll be fun, they said.

marcustyphoon: Implemented, with the caveat that this inherits Tumblr's erroneous "treat it like it's animated" behavior on non-animated webp images when media autoplay is disabled. I just sent a support message about that.

marcustyphoon:
- [x] #2297

PR #2278, opened 2026-06-09
feat(Mutual Checker, Show Originals): Combined custom dashboard tab functionality

### LINKED ISSUE #1973: Mutual Checker, Show Originals: Combined custom dashboard tab functionality
### Concept

> One inserted tab bar. The util provides the "All Posts" tab, features can register additional tabs (inserted to the left). If there are no additional tabs registered, the util removes the tab bar and becomes inactive.
> 
> Screen Shot 2025-11-07 at 12 56 52
> 
> CSS would need to be adjusted to align text to vertical middle for better mobile compatibility.
> 
> We don't have to figure out how to make this scrollable when the maximum number of tabs is three. XKit Rewritten is a packaged add-on, so we don't need to build extensibility until we need to use that extensibility.
> 
> _Originally posted by @AprilSylph in https://github.com/AprilSylph/XKit-Rewritten/issues/1972#issuecomment-3502428036_
>

marcustyphoon:
### Description

Resolves #1973.

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -6,7 +6,8 @@ import { onNewPosts, onNewNotifications, pageModifications } from '../../utils/m
 import { getPreferences } from '../../utils/preferences.js';
 import { blogData, notificationObject, timelineObject } from '../../utils/react_props.js';
 import { buildSvg } from '../../utils/remixicon.js';
-import { followingTimelineSelector } from '../../utils/timeline_id.js';
+import { followingTimelineFilter } from '../../utils/timeline_id.js';
+import { timelineTabs } from '../../utils/timeline_tabs.js';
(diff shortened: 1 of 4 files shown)


marcustyphoon:
- [ ] decide how to save selected tabs
- [ ] use radio buttons? maybe a form?
- [ ] lifecycle bug: enabling/disabling any feature resets selected tabs
- [ ] update mutual checker option name

marcustyphoon:
I realized while working on this that if I imagine polling the entire (small) set of users that have both features enabled on purpose—or, creating a version of this UI that allows both filters to be controlled individually, shipping it, advertising it, getting a somewhat maximal (still pretty small) set of users that are interested in both features, then polling them—about whether one of the modes they want and would use regularly is "original posts from mutuals," I'm not confident that the majority of them would say no. Which seems like something I probably ought to believe if I'm working on this? :D

PR #2280, opened 2026-06-10
refactor: Use `toggleAttribute`/`hasAttribute` functions

marcustyphoon:
### Description

Hey, April, did you know that javascript... ah, #2139; yes you did.

This makes use of the `hasAttribute` and `toggleAttribute` functions that are more semantically correct, rather than `getAttribute` and `setAttribute` with values that are/comparisons to empty strings, in places where we use boolean data attributes.

### Testing steps

I have not extensively tested this at present.

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -257,7 +257,7 @@ const processRows = function (rowsElements) {
 };
 
 const processHoverableElements = elements =>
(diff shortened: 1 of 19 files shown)


AprilSylph (approved):
Not able to test:
- `src/features/anti_capitalism/index.js`
- `src/features/no_recommended/hide_recommended_posts.js`

Everything else confirmed working as expected.

PR #2285, opened 2026-06-12
fix(Tweaks): "Hide filtered posts entirely" breaking if endless scrolling is disabled / on multi-column views

### LINKED ISSUE #2284: Tweaks: "Hide filtered posts entirely" breaks if endless scrolling is disabled / on multi-column views
### Platform

macOS 14.8.7 arm64

### Browser

Firefox 150.0.3

### Addon version

v1.3.4

### Details

Filtered posts aren't hidden if endless scrolling is disabled or in timelines that are in the multi-column ("masonry") view mode. Sorry everyone! This is due to a bug in a change I made in #2132.

(I actually commented about that specific line change being kind of weird, but I thought it was fine and tested it. Turns out I didn't test it without endless scrolling on, which would have shown that this ultimately surfaces another bug: our `getTimelineItemWrapper` function isn't idempotent if one has endless scrolling disabled, and we're calling it twice.)

marcustyphoon:
### Description

This fixes filtered posts not being hidden if endless scrolling is disabled.

The cause for this was that (as I noted in #2132) the code was, to appear more semantically correct, calling `getTimelineItemWrapper` twice. I thought the utility function was idempotent (could be called repeatedly with the same result), and when the resulting code worked I didn't think about it further, but it actually wasn't idempotent with endless scrolling disabled.

This PR makes two changes, either of which fix the problem alone: a) `getTimelineItemWrapper` no longer crashes if called multiple times with endless scrolling disabled, and b) the filtered post hiding feature no longer calls it multiple times.

Resolves #2284.

### Testing steps

- Confirm that filtered posts are hidden both with and without endless scrolling enabled.
- Confirm that filtered posts are hidden on tagged pages in both the normal and masonry view modes. (I filter "a" to test this more easily.)

### DIFF
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -21,7 +21,7 @@ const targetWrapperSelector = keyToCss(
  * @returns {Element | null} The timeline item wrapper
  */
 export const getTimelineItemWrapper = element =>
-  (element.closest('[data-timeline-id] > div > div') ||
+  element.closest('[data-timeline-id] > div > div') ||
   element.closest(listTimelineObjectSelector);
 
 /**
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -9,7 +9,6 @@ export const blogViewSelector = '[style*="--blog-title-color"] *';
 export const notificationSelector = `:is(${keyToCss('notification')}[role="listitem"], ${keyToCss('activityItem')})`;
 
 const listTimelineObjectSelector = keyToCss('listTimelineObject');
-const cellSelector = keyToCss('cell');
 const targetWrapperSelector = keyToCss(
   'targetWrapper',
   'targetWrapperBlock',
@@ -22,10 +21,8 @@ const targetWrapperSelector = keyToCss(
  * @returns {Element | null} The timeline item wrapper
  */
 export const getTimelineItemWrapper = element =>
-  (element.closest('[data-timeline-id]') && element.closest(listTimelineObjectSelector)?.parentElement) ||
-  element.closest(cellSelector) ||
-  element.closest(listTimelineObjectSelector) ||
-  element;
+  (element.closest('[data-timeline-id] > div > div') ||
+  element.closest(listTimelineObjectSelector);
 
 /**
  * @param {Element} element Element within a popover wrapper
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -24,7 +24,8 @@ const targetWrapperSelector = keyToCss(
 export const getTimelineItemWrapper = element =>
   (element.closest('[data-timeline-id]') && element.closest(listTimelineObjectSelector)?.parentElement) ||
   element.closest(cellSelector) ||
-  element.closest(listTimelineObjectSelector);
+  element.closest(listTimelineObjectSelector) ||
+  element;
 
 /**
  * @param {Element} element Element within a popover wrapper
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -8,6 +8,8 @@ export const trailItemSelector = `${postSelector} ${keyToCss('reblog')}`;
 export const blogViewSelector = '[style*="--blog-title-color"] *';
 export const notificationSelector = `:is(${keyToCss('notification')}[role="listitem"], ${keyToCss('activityItem')})`;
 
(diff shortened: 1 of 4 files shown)


src/utils/interface.js
```diff
@@ -24,7 +24,8 @@ const targetWrapperSelector = keyToCss(
 export const getTimelineItemWrapper = element =>
   (element.closest('[data-timeline-id]') && element.closest(listTimelineObjectSelector)?.parentElement) ||
   element.closest(cellSelector) ||
```
AprilSylph: This could be made idempotent more directly? This way the function would continue to return `null` for elements which are not timeline items, which I think is the less confusing thing to do from a dev standpoint.

```suggestion
  (element.closest('[data-timeline-id] > div > div') ||
```

src/utils/interface.js
```diff
@@ -24,7 +24,8 @@ const targetWrapperSelector = keyToCss(
...
-  element.closest(listTimelineObjectSelector);
+  element.closest(listTimelineObjectSelector) ||
+  element;
```
AprilSylph:
```suggestion
  element.closest(listTimelineObjectSelector);
```

src/utils/interface.js
AprilSylph:
```suggestion
```

AprilSylph: Can't tell if this suggestion is completely broken, or if it's just displaying incorrectly...

marcustyphoon: Uhhhh I guess we can find out!

marcustyphoon:
Will that break on stuff with `data-timeline`?

AprilSylph:
Timeline V1s don't have cells, so the existing `.closest(listTimelineObjectSelector)` should suffice for them. The new code is literally "find the wrapper in Timeline V2; fall back to finding the wrapper in Timeline V1". Should double-check that nothing is broken on Queue pages to be sure of this.

marcustyphoon:
Tempted to do, like, `:is(${keyToCss('cell')}, :not([class]))`, but maybe that's excessive.

AprilSylph (dismissed):
Tested and working:
- [x] `hide_filtered_posts` on dashboard, endless scrolling enabled
- [x] `hide_filtered_posts` on dashboard, endless scrolling disabled
- [x] `hide_filtered_posts` on permalink view, endless scrolling enabled
- [x] `hide_filtered_posts` on permalink view, endless scrolling disabled
- [x] PostBlock on dashboard
- [x] PostBlock on permalink view
- [x] PostBlock on Queue

marcustyphoon:
Do we need to apply that margin change to the other features that hide posts?

AprilSylph:
Only PostBlock _needs_ the margin override, but only by coincidence—no other post-hiding feature targets elements found in the queue. If Tumblr decided to start showing filtered screens on your own posts, it would be needed for `hide_filtered_posts` as well.

TIL: Timeline V1s that are not the queue (e.g. https://www.tumblr.com/timeline/blog_subscriptions) are also using virtualisation now?? Seems we could simplify even further:

```js
export const getTimelineItemWrapper = element => element.closest(`${timelineSelector} > div > div`);
```

marcustyphoon:
Hm, I don't know how to be confident about this. I guess we have grid, masonry, queue, endless scrolling on and off for each... I think the code technically implies that `scrollContainer` is optional but I don't know when that would ever be used... tumblr patio is probably normal? Surely it's fine.

AprilSylph:
You could probably just revert dd54732fbbf72af05e97f46ff43fe449745866a2 to be honest, I might have jumped the gun with it.

ecdc466757a523708f7ad60f6c8120b6347b2891 is necessary, though.

marcustyphoon:
I mean, I think it should be fine.

marcustyphoon:
A post in grid mode (i.e. in your likes) that you've expanded to open does hit the wrong element with that, actually, yeah.

How about we leave interface.js as-is in this PR, make the other two changes, commit that, and move to another thread for the util simplification testing?

AprilSylph:
> How about we leave interface.js as-is in this PR, make the other two changes, commit that, and move to another thread for the util simplification testing?

Makes sense to me!

AprilSylph (approved):

PR #2289, opened 2026-06-13
feat(No Recommended): Split community-posts-in-search-results into separate option

marcustyphoon:
### Description

As discussed in #2286, this splits the "Hide posts from non-joined communities in For You and search results" option in No Recommended into the original "Hide posts from non-joined communities in For You" and a separate new "Hide posts from non-joined communities in search results."

As I said in that discussion, I think this is probably a net negative on "giving the majority of affected users what they desire." I do, though, ultimately not view that as the end goal here. XKit features should make sense, be internally consistent, not have unexpected effects, respect the design intent of the site, etc, *even when it conflicts with what some users desire* (I, for example, will always outright refuse to make the frequently-requested "hide the explore button" tweak because the explore tab is a legitimate part of the site whether or not a user currently ever wants to use it and that is the single place in the UI to access it, so logically it should be visible there).

I think it's ultimately thus the most logical to follow the suggestion in the discussion and split this preference, *not* making it inherit from the old one. This errs on the side of "users who want something hidden should have to go find the setting to hide that thing." XKit does nothing when you first enable it for a reason. We're not the developer of Tumblr Savior.

Blocked by #2288.

### Testing steps

Find a community post in a search result ([tumblr.com/search/evil%20science/recent](https://www.tumblr.com/search/evil%20science/recent) is usually pretty good). Enable the search-result-specific No Recommended feature and confirm that it is hidden.

### DIFF
--- a/src/features/no_recommended/feature.json
+++ b/src/features/no_recommended/feature.json
@@ -18,6 +18,11 @@
       "label": "Hide posts from non-joined communities in For You",
       "default": false
     },
+    "hide_search_community_posts": {
+      "type": "checkbox",
+      "label": "Hide posts from non-joined communities in search results",
+      "default": false
+    },
     "hide_answertime": {
       "type": "checkbox",
       "label": "Hide the answertime banner",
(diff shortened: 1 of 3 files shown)


AprilSylph (approved):
Both options working as they describe themselves!

PR #2291, opened 2026-06-18
refactor: improve element shorthand return types

AprilSylph:
### Description

TIL about [`@template`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template). With this, we can have accurate return types in the new element creation shorthand functions.

### Screenshots

N/A, dev improvement only

### Testing steps

1. Pull this branch locally
2. With VS Code, open a file that imports any element creation shorthand functions from `dom.js`
3. Hover the usage(s) of the shorthand functions
    - **Expected result**: The return type matches the exact DOM interface of the element type
    e.g. `(alias) form(props?: Properties | undefined, children?: Children): HTMLFormElement`
    e.g. `(alias) input(props?: Properties | undefined): HTMLInputElement`
4. Hover any variables that are defined using a shorthand function
    - **Expected result**: The type of the variable matches the exact DOM interface of the element type
    e.g. `const formElement: HTMLFormElement`
    e.g. `const commentInput: HTMLInputElement`

### DIFF
--- a/src/utils/dom.js
+++ b/src/utils/dom.js
@@ -50,41 +50,46 @@ export function element (tagName, properties = {}, children = []) {
   return element;
 }
 
-/** @typedef {(props?: Properties, children?: Children) => HTMLElement} HTMLShorthand */
-/** @typedef {(props?: Properties) => HTMLElement} VoidHTMLShorthand */
-/** @typedef {(props?: Properties, children?: Children) => SVGElement} SVGShorthand */
+/**
+ * @template {HTMLElement | SVGElement | MathMLElement} T
+ * @typedef {(props?: Properties, children?: Children) => T} E
+ */
+/**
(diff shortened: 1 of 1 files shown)


marcustyphoon:
I don't think it's common practice to use single letter variable names for what are E and V here, is it?

marcustyphoon:
(Not that we can't. I just want to clarify whether you're doing it because it's common to single-letter the generic type variable, or whether it's just to save horizontal space.)

AprilSylph:
It's just to save horizontal space, yeah. Is it too cute? It might be a bit too cute.

marcustyphoon:
I think I could go either way on it, honestly.

marcustyphoon (approved):

AprilSylph:
I'm biased towards cuteness.

PR #2298, opened 2026-06-29
fix(AccessKit Disable GIFs): stop pausing GIFs when option is disabled

### LINKED ISSUE #2297: AccessKit Disable GIFs: Feature breaks GIF animation after disabling if Tumblr autoplay is disabled
### Platform

MacOS 14.8.7 arm64

### Browser

Firefox 150.0.3 (aarch64)

### Addon version

v1.3.5

### Details

If you:
- turn off Tumblr media autoplay
- turn on AccessKit Disable GIFs
- look at a GIF (**edit:** specifically via soft navigation to a page with one, so that AccessKit starts processing it before it fully loads!)
- turn *off* AccessKit Disable GIFs
- click the GIF

...then the GIF mysteriously stays paused.

Why? Because this is a permanent event listener; every time a load event fires on the image element in question, we pause it again, and this is never removed if you disable the feature. As discussed in #2247 (which this is effectively a sub-issue of, but I consider it a separate bug), Tumblr now changes the `src` property on a single image element when you click it with media autoplay disabled, which will trigger this again.

https://github.com/AprilSylph/XKit-Rewritten/blob/15eda2c366b9c1e436498f6f510c54d70f4126f6/src/features/accesskit/disable_gifs.js#L209-L211

Fixable via either `gifElement.addEventListener('load', () => pauseGif(gifElement), { once: true });` (simple; perhaps technically susceptible to some kind of race condition) or with an AbortSignal (to be fully robust; probably seems unnecessary?)

AprilSylph:
### Description

- fixes #2297

I went with the `AbortController` method to limit unintended effects. I don't know why there would be multiple `load` events on one image element, but we're running callbacks on all of them currently with no downsides... other than the fact that they keep firing after the option is disabled, which is obviously not good.

### Screenshots

Before | After
-|-
<video src="https://github.com/user-attachments/assets/5458417c-6e23-4324-bbed-3c9fabef0ba1"></video> | <video src="https://github.com/user-attachments/assets/187efc48-0a03-4bf6-a5ff-520f2f698b6e"></video>

### Testing steps

1. Go to [Tumblr &rarr; Settings &rarr; Dashboard](https://www.tumblr.com/settings/dashboard) and set "Media auto-play" to "Never auto-play"
2. Load the modified addon, and enable AccessKit &rarr; "Pause GIFs until they are hovered over"
3. Reload your Tumblr tab
4. Go to Explore &rarr; More &rarr; GIFs
    - **Expected result**: Both AccessKit's and Tumblr's GIF labels are present on the GIFs in posts
    - **Expected result**: Hovering a GIF removes its AccessKit GIF label, but not its Tumblr GIF label
5. Disable AccessKit &rarr: "Pause GIFs until they are hovered over"
    - **Expected result**: AccessKit's GIF labels disappear, leaving only Tumblr's GIF labels
6. Click a Tumblr-paused GIF
    - **Expected result**:  🆕 The GIF starts animating

### DIFF
--- a/src/features/accesskit/disable_gifs.js
+++ b/src/features/accesskit/disable_gifs.js
@@ -5,6 +5,9 @@ import { memoize } from '../../utils/memoize.js';
 import { pageModifications } from '../../utils/mutations.js';
 import { getPreferences } from '../../utils/preferences.js';
 
+/** @type {AbortController}   */ let loadEventController;
+/** @type {"eager" | "lazy"}  */ let loadingMode;
+
 const canvasClass = 'xkit-paused-gif-placeholder';
 const pausedPosterAttribute = 'data-paused-gif-use-poster';
 const pausedBackgroundImageVar = '--xkit-paused-gif-background-image';
@@ -12,8 +15,6 @@ const hoverContainerAttribute = 'data-paused-gif-hover-container';
 const labelAttribute = 'data-paused-gif-label';
 const labelSizeAttribute = 'data-paused-gif-label-size';
 const containerClass = 'xkit-paused-gif-container';
-
-let loadingMode;
 
 const hovered = `:is(:hover, [${hoverContainerAttribute}]:hover *)`;
 const parentHovered = `:is(:hover > *, [${hoverContainerAttribute}]:hover *)`;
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Not sure if we **also** want to set `once: true`? It seems most [word that I'm forgetting that means two pieces of code do the same thing goes here] as the `pauseGif(gifElement);` path?

AprilSylph:
It would have parity. Yeah, you're right.

marcustyphoon:
"Isomorphic" may have been what I was thinking of? Though I don't know that that's technically an accurate description :D

marcustyphoon (approved):

AprilSylph:
Reverting the title change, just realised it no longer relates at all to the bug that this fixes!

---

EDIT: Didn't say it would be a clean revert. 😛

Issue #2308, opened 2026-07-04
Tag Replacer: Improve handling of posts with maximum tag count


marcustyphoon:
### Description

Unconfirmed, but a user post mentioned that Tag Replacer doesn't work correctly with posts with the maximum tag count. This makes sense, as we add the new tag before removing the old one, and I think we generally want to continue doing that; it's a better failure mode. We could, though, filter out posts that already have the maximum number of tags, batch them, and reverse the order for those batches.

(Not sure what the failure mode is, at the moment; if having a single post with the maximum tag count makes the process stop, that's obviously pretty bad.)

marcustyphoon:
Oh boy. The failure mode is that the old tag gets removed but the new tag doesn't get added. That's also quite bad.

marcustyphoon:
After #2309, the new behavior is that posts that would have too many tags after the addition step (which is still first) will be skipped. This is still improvable, such as by displaying what those posts are (copy from #1639), by making a single tag replace work on posts with 30 tags (if we're confident this isn't prone to server-side race conditions with the two mega-edit util API calls, or via a direct edit), etc.

PR #2309, opened 2026-07-04
fix(Tag Replacer): Skip posts with too many tags

marcustyphoon:
### Description

Quick fix for the worst part of #2308: avoid trying to add tags to posts that have too many tags, and which would thus silently have incorrect/incomplete changes performed to them.

### Testing steps

- Make a test post with a replaceable tag (which, interestingly, must be in the first 20 tags?) and a total of 30 tags. Pasteable:`#skdjhfkjs #29 other tags: #2 #3 #4 #5 #6 #7 #8 #9 #10 #11 #12 #13 #14 #15 #16 #17 #18 #19 #20 #21 #22 #23 #24 #25 #26 #27 #28 #29`
- Confirm that the tag can no longer be replaced with Tag Replacer.
- Confirm that the tag *can* be removed by Tag Replacer.
- todo: write more tests here

### DIFF
--- a/src/features/tag_replacer/index.js
+++ b/src/features/tag_replacer/index.js
@@ -5,6 +5,8 @@ import { addSidebarItem, removeSidebarItem } from '../../utils/sidebar.js';
 import { apiFetch } from '../../utils/tumblr_helpers.js';
 import { userBlogs } from '../../utils/user.js';
 
+const MAX_POST_TAGS = 30;
+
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
Used your example tags on a new post, added `#2` to an old post with no other tags, then tried to replace `#2` with `#30`. The new post was skipped, the operation on the old post succeeded, reported numbers in the modal reflected this.

PR #2315, opened 2026-07-16
chore: Set `allowScripts` package value for upcoming npm 12

marcustyphoon:
### Description

npm 12 (which, to my knowledge, will ship for the first time with some random non-major version of node, making ci runners switch to it spontaneously) has the breaking change that dependency postinstall scripts are blocked unless explicitly allowed. Besides the odd rollout, this is a good thing. npm-using projects should probably check if they need to configure postinstall scripts in advance.

This project doesn't need to allow any, so we don't have to do anything. Explicitly denying the only one we have cleans up the `npm warn install-scripts 1 package had install scripts blocked because they are not covered by allowScripts` warning every time you install, though.

Generated by running `npm rebuild` and, in this case, `npm deny-scripts --all`.

unrs-resolver honestly isn't that clear about why it has a postinstall, but it's to mitigate an old npm bug apparently, so whatever:
https://github.com/unrs/unrs-resolver/issues/85
https://github.com/npm/cli/issues/4828

### Screenshots

n/a

### Testing steps

- Delete node_modules. `npm install` and `npm run posttest`.
- If desired, do this with both npm@11 and npm@12 (`npm install -g npm`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,9 @@
     "remixicon": "^4.9.1",
     "sortablejs": "^1.15.7"
   },
+  "allowScripts": {
+    "unrs-resolver": false
+  },
   "webExt": {
     "sourceDir": "src/",
     "build": {


AprilSylph (approved):
`npm cit` running and exiting cleanly on node@24 with both npm@11 and npm@12 (via volta). ✅

marcustyphoon:
I can do this for the other addons repos, or you can, either way!

AprilSylph:
I'll leave it to you 👍

PR #2328, opened 2026-07-29
chore(Timeformat): Disable feature on unreleased UI elements (for now)

marcustyphoon:
### Description

This disables Timeformat on some UI elements that appear in the Tumblr code but that aren't released. (We can make layout fixes for and enable it in these locations later if and when they launch).

### Screenshots

n/a

### Testing steps

Confirm that Timeformat continues to work normally.


AprilSylph (approved):
Seems like the correct thing to do given that TimeFormat doesn't affect the current overlay version of messaging!

Issue #2341, opened 2026-08-13
Replace RemixIcon with Material Symbols/Icons


AprilSylph:
- relates to https://github.com/AprilSylph/XKit-Rewritten/pull/2083#pullrequestreview-3739001556

---

I think RemixIcon is getting a little long in the tooth. It's handy to have simple identifiers for icons, yes, but it also means we bundle a gazillion icons with the addon which we don't actually use.

I would rather we use Google's _Material Symbols_ and/or _Material Icons_ as bare SVGs, both in the control panel and for the Quick Tags and Trim Reblogs post control buttons, and drop RemixIcon completely.

For example, the current manifest for Trim Reblogs:

```json
{
  "title": "Trim Reblogs",
  "description": "Trim threads after posting or drafting",
  "icon": {
    "class_name": "ri-scissors-cut-line",
    "color": "white",
    "background_color": "#56bc8a"
  },
  "help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#trim-reblogs",
  "relatedTerms": [ "Condensed Reblogs", "Editable Reblogs", "Cut Posts", "RP Format", "Beta Editor" ]
}
```

And a proposed updated version:

```json
{
  "title": "Trim Reblogs",
  "description": "Trim threads after posting or drafting",
[...]

PR #2343, opened 2026-08-14
refactor(UI)!: use individual SVG files for feature icons

AprilSylph:
### Description

- see https://github.com/AprilSylph/XKit-Rewritten/issues/2341

Removes `src/lib/remixicon/`, committing feature icons as `icon.svg` files instead; `remixicon.symbol.svg` alone has been retained in order for `src/utils/remixicon.js` to continue to work without significant modification. This reduces the size of the built addon from 2.5 MB all the way down to 836 KB!

All `icon.svg` files have a link to their license and to their place of origin, the same as all the Acorn icons in the control panel.

Includes updates to our documentation to reflect these changes.

### Screenshots

Before | After
-|-
Screenshot 2026-08-14 at 3 50 22 pm | Screenshot 2026-08-14 at 3 48 58 pm

### Testing steps

Load the modified addon, and smoke-test the control panel. All features should render with the same icons as before, although their alignment might be slightly different; they should be centred properly now, if anything.

Enable the following features, and verify that they continue to correctly insert their icons on Tumblr:
- Mutual Checker
- Quote Replies
- Quick Tags
- Trim Reblogs


marcustyphoon:
Can be half that again if we use individual svgs for content script icons, too, I think.

AprilSylph:
That's the next PR :smirk:

marcustyphoon:
Hm... now that I think about it, the comment should probably include the icon path or id, huh. Kind of weird not to be able to look up which one it is.

ex:

```js
#!/usr/bin/env node

import fs from 'node:fs';
import path from 'node:path';

const license = fs
  .readFileSync('node_modules/remixicon/fonts/remixicon.symbol.svg', { encoding: 'utf8' })
  .match(/<!--.*-->/s)
  .at(0);

fs.globSync('src/features/*/feature.json').forEach(match => {
  const feature = JSON.parse(fs.readFileSync(match));
  const id = feature.icon?.class_name?.replace(/^ri-/, '');
  if (id) {
    const iconPath = fs.globSync(`node_modules/remixicon/icons/*/${id}.svg`).at(0);
    const svgText = fs.readFileSync(iconPath, { encoding: 'utf8' });
    fs.writeFileSync(
      path.join(path.dirname(match), 'icon.svg'),
      [license, '<!--', `Source: ${iconPath}`, '-->', svgText, ''].join('\n'),
    );
  }
});
```

marcustyphoon:
Man, dealing with script arguments is such a pain. Anyway, if desired, we could commit a script to facilitate new features or grabbing non-feature icons for other uses, a la:

```js
#!/usr/bin/env node

import fs from 'node:fs';
import path from 'node:path';
import util from 'node:util';

const { values: { id, target } } = util.parseArgs({
  options: {
    id: { type: 'string', short: 'i' },
    target: { type: 'string', short: 't' },
  },
  strict: true,
});
if (!id || !target) {
  console.log('Run this script with --id and --target! Examples:');
  console.log('dev/copy-remix-icon.js --id repeat-line --target src/features/quick_reblog');
  console.log('dev/copy-remix-icon.js --id repeat-line --target src/features/quick_reblog/repeat-line.svg');
  process.exit(1);
}

const iconPath = fs.globSync(`node_modules/remixicon/icons/*/${id}.svg`).at(0);
if (!iconPath) {
  console.log(`Could not find icon file with id ${id}!`);
  process.exit(1);
}

const targetIsFile = !!path.extname(target);
const targetFile = targetIsFile ? target : path.join(target, 'icon.svg');

const dirname = path.dirname(targetFile);
try {
  fs.accessSync(dirname);
} catch {
  console.log(`Could not access path ${dirname}!`);
  process.exit(1);
}

const license = fs
  .readFileSync('node_modules/remixicon/fonts/remixicon.symbol.svg', { encoding: 'utf8' })
  .match(/<!--.*-->/s)
  .at(0);
const svgText = fs.readFileSync(iconPath, { encoding: 'utf8' });

fs.writeFileSync(
  targetFile,
  [license, '<!--', `Source: ${iconPath}`, '-->', svgText, ''].join('\n'),
);
```

AprilSylph:
I'm not interested in ever committing any new RemixIcon icons; this PR is just for getting things under the hood closer to where I want them to be before any user-facing changes happen.

I have my own ideas about committing the icon name; should put that on the checklist...

marcustyphoon:
Well, as someone with four new feature PRs... :D But, yeah, obviously only relevant if there's an extended period between this and additional changes, and in any case I can of course run it without it being in the repo.

docs/Chapter 4.0 - Feature framework.md
```diff
@@ -2,6 +2,7 @@
 
 Each feature is comprised of a subfolder in `src/features/`, containing:
 - `feature.json` metadata file (required)
+- `icon.svg` icon file (optional)
```
marcustyphoon, line 5: I might not consider this optional unless we added a fallback that painted the background color on an empty icon. Sure, it doesn't crash the control panel, but.

AprilSylph: I _am_ trying to discontinue the background colours, is the thing.

I would count this as a documentation issue, though; should be explicit that no icon file means all `icon` metadata properties are ignored.

---

https://github.com/AprilSylph/XKit-Rewritten/blob/4f6bd6c9690ba55af0d878d50155f3b8cea8a8bd/docs/Chapter%204.1%20-%20Feature%20metadata.md?plain=1#L19-L23

marcustyphoon (approved):

AprilSylph:
I think it makes sense to denote this as a breaking change, since it's a change to the framework that requires features to update for it with regression as the only alternative. It's not a breaking change for end users, but I would say that end users aren't consumers of the framework at all; the framework is consumed by devs who write the features, and then those features are consumed by end users.

Basically: please have an exclamation mark to acknowledge and apologise that I've made more work for you.

marcustyphoon:
I think the transition to _different_ icons makes sense as a breaking! This one, eh, it's just moving some files about to produce the same result, no big deal.

I assume you intend to keep working on this set of refactors before shipping?

AprilSylph:
Yeah, there's no value in releasing this refactor by itself, but every commit to the main branch should still leave the repository in a good state (for both devs and for users). The reality is that new features need to be written differently now! (Even if only slightly.)

Who knows, I might feel like reviewing Find Inactives before I continue the refactor/switchover.

marcustyphoon:
> I might feel like reviewing Find Inactives before I continue the refactor

Oh, okay, gotcha, so not necessarily yes to the way I intended it! (i.e. "is your next commit on this likely to be tomorrow," that sort of thing.)

marcustyphoon:
Okay, in retrospect I guess if you were planning to work on it the way I was thinking, you likely wouldn't have merged this as a PR that didn't change remixicon.js. Well—whatever, anyway, for clarity, I assume this means I shouldn't wait for a full icon replacement to update feature PRs?

AprilSylph:
It's ultimately up to you to bet on what I'll do next, since nothing is ever certain :P but yeah, a full icon replacement is going to be a lot of work, probably more work than reviewing Find Inactives, so having that be up-to-date and ready to merge (approval notwithstanding) would be nice!

PR #2348, opened 2026-08-24
refactor: use new element shorthands in modules using `buildSvg()`

AprilSylph:
### Description

- relates to #2341

I'm trying to optimise `lib/remixicon.symbol.svg` away, which means drastically changing what `buildSvg` does. However, most of the usages of `buildSvg` are nested inside `dom()` calls, which I just can't abide. Not to mention, `buildSvg` uses `dom()` itself! Oh, the horror!

So, no more `dom()` and `buildSvg` in the same files. This is its own PR so that it can be tested separately from any actual changes to `buildSvg`.

### Screenshots

No visual changes expected.

### Testing steps

Load the modified addon, and smoke-test all the features affected:
- Mutual Checker (verify: mutuals icon, follower icon)
- Quick Tags (verify: post control button, post form button)
- Quote Replies
- Trim Reblogs

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -1,5 +1,5 @@
 import { keyToCss } from '../../utils/css_map.js';
-import { dom } from '../../utils/dom.js';
+import { path, svg, title } from '../../utils/dom.js';
 import { buildStyle, getTimelineItemWrapper, filterPostElements, getPopoverWrapper, notificationSelector } from '../../utils/interface.js';
 import { translate } from '../../utils/language_data.js';
 import { onNewPosts, onNewNotifications, pageModifications } from '../../utils/mutations.js';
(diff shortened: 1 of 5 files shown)


AprilSylph:
...I just realised I left the `xmlns` properties intact. Those aren't necessary anymore.

marcustyphoon:
Oh yeah, I guess "remove every use of the dom util" was one of the potential answers for that question I dm'd you about potential stuff you might want me to execute.

AprilSylph:
I still think the best way to go about removing `dom()` calls is "as needed". I wouldn't review a PR that only does it for its own sake, i.e. not as a precursor to something we actually want to do.

marcustyphoon (approved):

marcustyphoon:
Not sure if "E shorthands" is the most clear title; I had no idea what that referred to :D

AprilSylph:
Yeah, I was just trying to get the commit title down to a good character length, but then got lazy on the PR title just copying the commit title wholesale. Will update.

PR #2352, opened 2026-08-26
fix(No Recommended): Enable blog/community carousel hiding in masonry views

### LINKED ISSUE #2322: No Recommended: Broken recommended blog/tag/community hiding between posts in masonry views
### Platform

macOS 14.8.7 arm64

### Browser

Firefox 153.0

### Addon version

v1.3.6

### Details

In masonry views, recommended blog/tag/community elements between posts are displayed differently and No Recommended fails to hide them. ("Check out these communities" and "Check out these tags" are totally unaffected; I think the recommended blog carousel is hidden but its title is not.)

We should probably do a pass of checking extension functionality on masonry views in general. The caught up line thingy does seem to work, for example. Seen posts should probably be fine? Et cetera.

marcustyphoon:
### Description

In masonry views, recommended blog/tag/community elements between posts are displayed differently and No Recommended fails to hide them. This fixes two of the three.

This currently doesn't include tag carousels since... I can't find one. I wrote "Check out these tags" in the linked issue, though, so obviously they exist. Based on staring at a bunch of minified javascript, I'd believe you if you told me the correct selector for them is or includes `keyToCss('tagGrid')`, but also if you did not say that. So.

Related/mostly resolves: #2322.

### Screenshots

n/a

### Testing steps

- Confirm that the "Hide recommended blogs between posts" and "Hide recommended communities between posts" No Recommended options still works on normal timelines.
- Confirm that the "Hide recommended blogs between posts" and "Hide recommended communities between posts" No Recommended options work on masonry-view timelines.

### DIFF
--- a/src/features/no_recommended/hide_blog_carousels.js
+++ b/src/features/no_recommended/hide_blog_carousels.js
@@ -2,6 +2,7 @@ import { keyToCss } from '../../utils/css_map.js';
 import { buildStyle, getTimelineItemWrapper } from '../../utils/interface.js';
 import { pageModifications } from '../../utils/mutations.js';
 import { timelineObject } from '../../utils/react_props.js';
+import { timelineSelector } from '../../utils/timeline_id.js';
 
 const hiddenAttribute = 'data-no-recommended-blog-carousels-hidden';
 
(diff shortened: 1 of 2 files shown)


marcustyphoon:
I would say I really wasn't sure what decisions to make to make this as robust as possible? Like, should the masonry selector be keyed on `keyToCss('masonry')` instead of a timeline selector? Should we be checking the title text? I dunno, I don't have a good grasp on this.

PR #2354, opened 2026-08-26
fix(No Recommended): Enable tag carousel hiding in masonry views

marcustyphoon:
### Description

In masonry views, recommended blog/tag/community elements between posts are displayed differently and No Recommended fails to hide them. This fixes the final one.

See #2322.

For some reason, it seems like we changed the blog/community carousel hiding code, but not the tag carousel code? I don't remember why now.

### Screenshots

n/a

### Testing steps

- Confirm that the "Hide recommended tags between posts" No Recommended option still works on normal timelines.
- Confirm that the "Hide recommended tags between posts" No Recommended options work on masonry-view timelines.

The For You page is where I found one of these (eventually).


marcustyphoon:
> I would say I really wasn't sure what decisions to make to make this as robust as possible? Like, should the masonry selector be keyed on `keyToCss('masonry')` instead of a timeline selector? Should we be checking the title text? I dunno, I don't have a good grasp on this.

As I said in the prior PR.

PR #2355, opened 2026-08-26
docs: rewrite Chapter 4.1 - Feature metadata.md

AprilSylph:
### Description

- follows up #2347

Rewrites our feature manifest documentation to be much more complete and a lot more respectable.

The new structural format of the document, splitting properties and types into their own sections, is inspired by Chrome's developer documentation (e.g. https://developer.chrome.com/docs/extensions/reference/api/tabs).

The table format for describing items is inspired by MDN (e.g. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/description).

The schema was incredibly useful in putting this together, and the new structure of the document mirrors it very closely.

### Screenshots

N/A

### Testing steps

[Read the rewritten document](https://github.com/AprilSylph/XKit-Rewritten/blob/579e976fd99221b008a105bd33cef7d17b74a1a4/docs/Chapter%204.1%20-%20Feature%20metadata.md). Are any details from the previous version omitted? Have I made any grammatical errors, or written any sentences that are difficult to parse? Do all the anchors actually work!?

### DIFF
--- a/docs/Chapter 4.1 - Feature metadata.md
+++ b/docs/Chapter 4.1 - Feature metadata.md
@@ -40,7 +40,7 @@ If this property is omitted, no description is displayed for the feature.
 
 |                 |                                                                                       |
 | --------------- | ------------------------------------------------------------------------------------- |
-| **Type**        | `{ color: string; background_color: string; }`                                        |
+| **Type**        | `{ color?: string; background_color?: string; }`                                      |
 | **Mandatory**   | No                                                                                    |
 | **Description** | Properties for customising how the feature's `icon.svg` renders in the control panel. |
 | **Example**     | See [§ Properties → `"icon"` → Example](#example).                                    |
@@ -166,7 +166,7 @@ This value always a string; this documentation defines a `Label` type to avoid r
 
 |                 |                                                                                       |
 | --------------- | ------------------------------------------------------------------------------------- |
-| **Type**        | `string \| undefined`                                                                  |
+| **Type**        | `string`                                                                  |
 | **Description** | The storage key to inherit the value of, if the preference has not been set.          |
 | **Example**     | <pre lang="json">"inherit": "no_recommended.preferences.hide_recommended_blogs"</pre> |
 
@@ -201,11 +201,11 @@ The value type of `"default"` varies; keep reading for definitions of each prefe
 
 ### `CheckboxPreference`
 
-|                 |                                                                           |
-| --------------- | ------------------------------------------------------------------------- |
-| **Type**        | `{ type: "checkbox"; label: Label; default: boolean; inherit: Inherit; }` |
-| **Description** | A checkbox-type preference.                                               |
-| **Example**     | See [§ Types → `CheckboxPreference` → Example](#example-2).               |
+|                 |                                                                             |
+| --------------- | --------------------------------------------------------------------------- |
+| **Type**        | `{ type: "checkbox"; label: Label; default: boolean; inherit?: Inherit; }`  |
+| **Description** | A checkbox-type preference.                                                 |
+| **Example**     | See [§ Types → `CheckboxPreference` → Example](#example-2).                 |
 
 #### `"default"`
 
@@ -225,11 +225,11 @@ Must be a boolean. Omitting this property is not allowed.
 
 ### `ColorPreference`
 
-|                 |                                                                       |
-| --------------- | --------------------------------------------------------------------- |
-| **Type**        | `{ type: "color"; label: Label; default: string; inherit: Inherit; }` |
-| **Description** | A color-type preference.                                              |
-| **Example**     | See [§ Types → `ColorPreference` → Example](#example-3).              |
+|                 |                                                                         |
+| --------------- | ----------------------------------------------------------------------- |
+| **Type**        | `{ type: "color"; label: Label; default: string; inherit?: Inherit; }`  |
+| **Description** | A color-type preference.                                                |
+| **Example**     | See [§ Types → `ColorPreference` → Example](#example-3).                |
 
 #### `"default"`
 
@@ -279,11 +279,11 @@ For an example of a component preference module file, see [`src/features/postblo
 
 ### `PercentPreference`
 
-|                 |                                                                         |
-| --------------- | ----------------------------------------------------------------------- |
-| **Type**        | `{ type: "percent"; label: Label; default: number; inherit: Inherit; }` |
-| **Description** | A percent-type preference.                                              |
-| **Example**     | See [§ Types → `PercentPreference` → Example](#example-5).              |
+|                 |                                                                           |
+| --------------- | ------------------------------------------------------------------------- |
+| **Type**        | `{ type: "percent"; label: Label; default: number; inherit?: Inherit; }`  |
+| **Description** | A percent-type preference.                                                |
+| **Example**     | See [§ Types → `PercentPreference` → Example](#example-5).                |
 
 #### `"default"`
 
@@ -305,7 +305,7 @@ Must be an integer ranging from 0 to 100. Omitting this property is not allowed.
 
 |                 |                                                                                                                       |
 | --------------- | --------------------------------------------------------------------------------------------------------------------- |
-| **Type**        | `{ type: "select"; label: Label; options: { label: string; value: string; }[]; default: string; inherit: Inherit; }`  |
+| **Type**        | `{ type: "select"; label: Label; options: { label: string; value: string; }[]; default: string; inherit?: Inherit; }` |
 | **Description** | A select-type preference.                                                                                             |
 | **Example**     | See [§ Types → `SelectPreference` → Example](#example-6).                                                             |
 
@@ -340,7 +340,7 @@ Must be a string that matches the `"value"` of one of the defined options. Omitt
 
 |                 |                                                                       |
 | --------------- | --------------------------------------------------------------------- |
-| **Type**        | `{ type: "text"; label: Label; default: string; inherit: Inherit; }`  |
+| **Type**        | `{ type: "text"; label: Label; default: string; inherit?: Inherit; }` |
 | **Description** | A text-type preference.                                               |
 | **Example**     | See [§ Types → `TextPreference` → Example](#example-7).               |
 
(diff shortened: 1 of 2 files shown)


AprilSylph:
Hmm, maybe I should use my classic "pad sections with double newlines" technique here too. I'm kinda having a hard time parsing the difference between third-level and fourth-level sections without them...

AprilSylph:
Yeah, I think that's better.

God, the table of contents is _really_ nice now.

marcustyphoon:
This is an interesting one to review for me: there's of course no way you would have known this, but hilariously, the chrome extension developer docs just so happen to be the runaway winner for "API docs about whose authors I have muttered the most things that I don't *actually* mean and wouldn't actually say to them; I simply enjoy being dramatic about user interfaces which use so much vertical space and so little indentation or visual differentiation that if you scroll to an arbitrary location and squint, there's no way to tell what you're looking at."

Obviously, my saying that would have to imply that there are comparable API documentation sites whose layouts I think are superior, though, which... eh, nothing comes to mind. https://leafletjs.com/reference.html (which makes more use of table rows and inline type assertions) and https://docs.stripe.com/api/v2/core/accounts/create (which uses nested details/summary-type expandables and indentation for nested properties well in some cases) were some kind of interesting examples.

**edit:** Ah, https://expressjs.com/en/5x/api/express/ is decent also.

I tried squashing the layout to a degree, the results of which you can see [here](https://github.com/user-attachments/files/31514100/Chapter.4.1.-.Feature.metadata.md); I certainly found it hard to keep a professional look while staying within the bounds of gfm. 



Anyway: I don't think any of that is actually important! Let me give specific suggestions below.

marcustyphoon:
- I don't know that the "example" table rows are super necessary, as they all don't go particularly far!
- `inherit: Inherit;` in a typescript-style typedef implies that it's not optional.
- I guess Label and Inherit can't really be elegantly named as their full paths (e.g.`preferences.[preferenceName].label`/`Preference.label`). Maybe that could be listed... I dunno, somewhere? When I read them it's not clear to me what objects they would be part of, which isn't the end of the world, but.
- A trick I discovered while doing the above is that you can put links in a code block if you use \<code>\</code> instead of backticks; that would let you hyperlink Preference, Label and Inherit in the typedefs. (And the components of the union type, not that anyone would ever follow those links, but.)

AprilSylph:
> I don't know that the "example" table rows are super necessary, as they all don't go particularly far!

I was also questioning whether the majority of them should be included. I've deleted the ones that only linked to that section's "Example" subsection.

> `inherit: Inherit;` in a typescript-style typedef implies that it's not optional.

Fixed!

> I guess Label and Inherit can't really be elegantly named as their full paths (e.g.`preferences.[preferenceName].label`/`Preference.label`). Maybe that could be listed... I dunno, somewhere? When I read them it's not clear to me what objects they would be part of, which isn't the end of the world, but.

Yeah, they're not actually part of the `Preference` type, since `ComponentPreference` doesn't support them. I'm sticking to the schema and just counting them as their own types, which most of the other preference types then reference. This doesn't resolve the clarity issue, however...

> A trick I discovered while doing the above is that you can put links in a code block if you use \<code\>\<\/code\> instead of backticks; that would let you hyperlink Preference, Label and Inherit in the typedefs.

...I think this does!

> (And the components of the union type, not that anyone would ever follow those links, but.)

Well, now I gotta, for consistency. Done.

marcustyphoon:
Hm—I guess in theory all of the examples could be inlined into tables—not sure if that's better or worse.

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,382 @@
...
+
+|                 |                                                                                   |
+| --------------- | --------------------------------------------------------------------------------- |
+| **Type**        | `string`                                                                          |
+| **Description** | The preference's label to be displayed in the control panel.                      |
+| **Example**     | <pre lang="json">"label": "Hide recommended blogs in the blog view sidebar"</pre> |
+
+Most preference types require a label.
+
+This value always a string; this documentation defines a `Label` type to avoid re-describing the `"label"` property on each preference type.
```
marcustyphoon:
```suggestion
This value is always a string; this documentation defines a `Label` type to avoid re-describing the `"label"` property on each preference type.
```

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,382 @@
...
+### `"deprecated"`
+
+|                 |                                             |
+| --------------- | ------------------------------------------- |
+| **Type**        | `boolean`                                   |
+| **Mandatory**   | No                                          |
+| **Description** | Whether or not this feature is deprecated.  |
+| **Example**     | <pre lang="json">"deprecated": true</pre>   |
+
+Deprecated features are hidden on all installations by default, allowing it to be discontinued without being outright removed.
```
marcustyphoon: Seems like a singular/plural mismatch?

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,382 @@
...
+| **Type**        | `{ type: "component"; src: string; }`                       |
+| **Description** | A component-type preference. Allows rendering of custom UI. |
+
+This is the only preference type which does not support `"default"` or `"inherit"`.
+
+#### `"src"`
+
+A URL, relative to the `src/` directory, pointing to a module file for a Web Component to be rendered in the feature's preference list.
+
+Using a custom Web Component instead of any of the other provided preference types (or in conjunction with them) allows for complex management of that feature's data, such as creating tag bundles or listing blocked posts.
```
marcustyphoon:
```suggestion
Using a custom Web Component instead of any of the other provided preference types (or in conjunction with them) allows for complex management of a feature's data, such as creating tag bundles or listing blocked posts.
```

```suggestion
Using a custom Web Component instead of any of the other provided preference types (or in conjunction with them) for a feature allows for complex management of that feature's data, such as creating tag bundles or listing blocked posts.
```

AprilSylph:
> Hm—I guess in theory all of the examples could be inlined into tables—not sure if that's better or worse.

Worse, because GitHub doesn't support multiline code blocks inside of tables. You can put the `<br>` elements in there, and it'll display as multiline on other renderers, but GitHub just collapses them down into spaces.

---

Observe (check my markdown!):

|                 |                                                                                                             |
| --------------- | ----------------------------------------------------------------------------------------------------------- |
| **Type**        | <code>{ type: "checkbox"; label: [Label](#label); default: boolean; inherit?: [Inherit](#inherit); }</code> |
| **Description** | A checkbox-type preference. Renders in the control panel as a checkbox input followed by a text label.      |
| **Example** | <pre lang="json">"showTagSuggestions": {<br>  "type": "checkbox",<br>  "label": "Suggest tags from the post being reblogged",<br>  "default": false<br>}</pre> |

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,383 @@
...
+| **Description** | Whether or not this feature is deprecated.  |
+| **Example**     | <pre lang="json">"deprecated": true</pre>   |
 
-#### `"preferences"`: \<preference name\>: `"options"`
-- Type: Array
-- Required: Yes, if `type` is `"select"`
+Deprecated features are hidden on all installations by default, allowing features to be discontinued without being outright removed.
 
-For `"select"`-type preferences, an array of objects each with `"value"` and `"label"` properties. Unused for other preference types.
+Only installations with "special access" to a given deprecated feature will continue to be able to enable and run it.
```
marcustyphoon: Might be clearer as e.g. "Only installations which had a feature enabled at the time when it became deprecated will be able to continue seeing it in the feature list and to run it, a state known as 'special access.'" or etc.

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,383 @@
...
-- Required: No
 
-Object with 1 or more custom keys; each key is used internally as a preference's name.
+|                 |                                                                                                       |
+| --------------- | ----------------------------------------------------------------------------------------------------- |
+| **Type**        | <code>Record\<string, [Preference](#preference)\></code>                                              |
+| **Mandatory**   | No                                                                                                    |
+| **Description** | An object consisting of one or more preference definition objects, keyed by internal preference name. |
+
+It is recommended to use camelCase for preference names in most cases. This allows the feature code to use destructuring assignments when fetching preference values from storage, without needing to perform renames to obey the project's JavaScript style guide.
```
marcustyphoon: Or "assignment."

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,383 @@
...
+| **Example**     | <pre lang="json">"inherit": "no_recommended.preferences.hide_recommended_blogs"</pre> |
+
+Most preference types support inheriting their value from another storage key, if both of the following are true:
+
+- The inheriting preference (where `"inherit"` is defined) does not have any value saved in storage.
+- The inherited preference (which `"inherit"` points to) does have a value saved in storage.
+
+This is useful for splitting one option into multiple options with finer granularity, or moving preferences from one feature to another, without requiring user action if it can be avoided.
+
+The value type for each preference type is different, so it is not automatically possible for a preference to inherit from any preference. If a preference wants to inherit a stored value that is not compatible with its preference type, the feature code must account for this on a per-case basis.
```
marcustyphoon: Would this imply not using the `inherit` key?

docs/Chapter 4.1 - Feature metadata.md
AprilSylph: The subsequent subsection explains how special access is granted.

docs/Chapter 4.1 - Feature metadata.md
```diff
@@ -1,109 +1,383 @@
...
 
-URL which points to a usage guide or extended description for the feature.
+|                 |                                                                                                           |
+| --------------- | --------------------------------------------------------------------------------------------------------- |
+| **Type**        | `string`                                                                                                  |
+| **Mandatory**   | No                                                                                                        |
+| **Description** | URL pointing to the usage guide or extended description for this feature.                                 |
+| **Example**     | <pre lang="json">"help": "https://github.com/AprilSylph/XKit-Rewritten/wiki/Features#vanilla-audio"</pre> |
+
+If this property is omitted, the feature is marked as "New".
```
marcustyphoon: This could describe that it defines the target of a link/that a "new" label is rendered in place of the link.

docs/Chapter 4.1 - Feature metadata.md
AprilSylph: It's probably just easier to say that inheriting from different preference types is unsupported.

marcustyphoon: I honestly hadn't put any thought into whether cross-mode inheritance would crash the config tab code in any particular scenario. I'll bet there are some setups that are possible to do.

marcustyphoon (approved):

PR #2356, opened 2026-08-27
refactor: remove remixicon dependency

AprilSylph:
### Description

- relates to #2341 

Removes the RemixIcon dependency completely, with `remixicon.js` being renamed and transformed into simply `icons.js`—a utility which fetches icons from features' `icon.svg` files, rather than any huge central repository.

This shrinks the bundle size again, from 809 KB all the way down to 342 KB this time.

That is of course not the primary motivation for this PR; the real reason to do this is so we can use icons from other sources without having to introduce any new dependencies. Freedom!!

### Screenshots

The only expected visual difference is that I've updated the Quick Tags icon colour in the post form to use one of the new design system colours.

Before | After
-|-
Screen Shot 2026-08-27 at 11 26 10 | Screen Shot 2026-08-27 at 11 26 23
Screen Shot 2026-08-27 at 11 26 40 | Screen Shot 2026-08-27 at 11 26 49
Screen Shot 2026-08-27 at 11 27 21 | Screen Shot 2026-08-27 at 11 27 30

### Testing steps

Smoke-test every feature that previously relied on `buildSvg`, either directly or transitively:
- Quick Tags
  - ...in the post footer
  - ...in the new post form
- Quote Replies
- Trim Reblogs


marcustyphoon:
Looks pretty similar to what I was thinking, besides that I considered making the utility take a URL, and feeding it `import.meta.resolve('./icon.svg')` in each feature!

AprilSylph:
Interesting... it would certainly be more extensible that way, allowing features to have more than one icon at their disposal.

marcustyphoon:
Question: do we want to get cute and have icons.js replace hyphens with underscores, such that feature names would be considered canonically kebab case and feature **paths** are underscored?

Ah—but extension storage already has everything namespaced into underscore case, which would be confusing if we did that. Never mind.

marcustyphoon:
> Interesting... it would certainly be more extensible that way, allowing features to have more than one icon at their disposal.

Certainly if we ever need it for some niche reason we could make icons.svg have an internal function that takes a URL and a clean one that takes a feature name, and export them both. The question is whether a feature would plausibly want >1 control button/post action, in which case it would be better to plumb something all the way through.

edit: Oh, also, the id thing is a pain. You can css.escape() (part of?) the icon path and use it. Or use a randomly generated id, technically, if you use memoize, but I don't love having inspect element show you something totally unintelligible.

marcustyphoon (approved):
Works as-is, of course.

AprilSylph:
I think my stance is, as usual, "let's not do a bunch of work on a util which will yield a purely theoretical future benefit". We can build it if and when we need it.

PR #2363, opened 2026-08-31
fix(Icons util): Multiple duplicate symbol elements being created

### LINKED ISSUE #2358: Icons util: Multiple duplicate symbol elements created
### Platform

MacOS 14.8.8 arm64

### Browser

Firefox 154.0.1

### Addon version

master

### Details

Totally harmless, but there's technically a race condition here since the getElementById is not in the same microtask loop as creating the checked-for element, resulting in duplicate symbol elements.

https://github.com/AprilSylph/XKit-Rewritten/blob/2d561ce16756fed9cdd46d9e5a1de567b56ed3c8/src/utils/icons.js#L17-L24

Resolvable by, for example:

```js
const getSymbolId = memoize(featureName => {
  const symbolId = `${symbolsClassName}__${featureName}`;

  fetch(browser.runtime.getURL(`/features/${featureName}/icon.svg`))
    .then(iconResponse => iconResponse.text())
    .then(iconText => new DOMParser().parseFromString(iconText, 'image/svg+xml').firstElementChild)
    .then(iconElement => {
      iconElement.setAttribute('id', symbolId);
      symbols.append(iconElement);
    });

  return symbolId;
});

/**
 * @param {string} featureName The internal name of a feature calling this utility (e.g. `"quick_tags"`)
 * @returns {SVGSVGElement} An SVG element that renders the icon for the specified feature
 */
export const getIcon = featureName => svg({}, [use({ href: `#${getSymbolId(featureName)}` })]);
```

Something more inline would also work, with a slight adjustment to memoize so that it can be used on functions with more than one argument.

marcustyphoon:
### Description

As described in the linked issue, this fixes a small race condition and prevents multiple duplicate symbol elements from being created by the new version of the icons util.

Resolves #2358.

### Screenshots

n/a

### Testing steps

Smoke test icon insertion.

### DIFF
--- a/src/utils/icons.js
+++ b/src/utils/icons.js
@@ -1,4 +1,5 @@
 import { aside, svg, use } from './dom.js';
+import { memoize } from './memoize.js';
 
 const symbolsClassName = 'xkit-symbols';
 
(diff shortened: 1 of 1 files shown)


AprilSylph (approved):
Huh, somehow didn't realise that memoization means we don't even need the presence check anymore. Cool!

marcustyphoon:
Yep. This would break if the symbols element got removed from the dom, of course! But so would the original remix icon code.

PR #2366, opened 2026-08-31
refactor: `runFeature` → `runFeatures`

AprilSylph:
### Description

- developed from https://github.com/AprilSylph/XKit-Rewritten/pull/2359#issuecomment-5477447845

> ...huh, I guess there are no usages of `runFeature`/`destroyFeature` that aren't trying to run/destroy multiple features from an arbitrary array. So, on that front, making them expect an array argument makes a ton of sense.

This diff best viewed excluding whitespace-only changes.

### Screenshots

N/A

### Testing steps

1. Load the modified addon
2. Open a Tumblr tab
3. Enable a feature with its own preference-handling code (e.g. Classic Footer)
    - **Expected result**: The feature activates instantly
    - **Expected result**: Changes to the feature's preference are reflected instantly
4. Enable a feature with a static stylesheet (e.g. Classic Search)
    - **Expected result**: The feature activates instantly
    - **Expected result**: The feature's stylesheet is applied to the page
5. Enable a feature with a constructed style element (e.g. Sidebar Blogs)
    - **Expected result**: The feature activates instantly
    - **Expected result**: The feature's style element is applied to the page
5. Reload the tab
    - **Expected result**: All enabled features activate after pageload
6. Disable a feature
    - **Expected result**: The feature deactivates instantly
    - **Expected result**: The feature's accompanying stylesheet and/or style element are removed from the page

### DIFF
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -21,15 +21,12 @@
 
   const timestamp = Date.now(); // Prevent referencing outdated resources after Firefox extension update/restart
 
-  /** @type {(name: string) => Promise<FeatureModule>} */
-  const getFeature = name => import(browser.runtime.getURL(`/features/${name}/index.js`));
-
   /**
    * @param {string[]} names Internal feature names to resolve modules for.
    * @returns {Promise<[string, FeatureModule][]>} Entries of feature modules, keyed by name.
    */
   const getFeatures = names => Promise.all(
-    names.map(async (name) => [name, await getFeature(name)]),
+    names.map(async (name) => [name, await import(browser.runtime.getURL(`/features/${name}/index.js`))]),
   );
 
   const runFeatures = async function (names) {
--- a/src/content_scripts/main.js
+++ b/src/content_scripts/main.js
@@ -1,5 +1,14 @@
 'use strict';
 
+/**
+ * @typedef FeatureModule
+ * @property {() => Promise<void>}  main The function for running the feature when it is enabled.
+ * @property {() => Promise<void>}  clean The function for cleaning up the feature when it is disabled.
+ * @property {() => Promise<void>}  [onStorageChanged] Preference handling for the feature. If not provided, the feature is restarted when its preferences change.
+ * @property {boolean}              [stylesheet] Whether or not the feature has a static stylesheet at `index.css`. Optional.
+ * @property {HTMLStyleElement}     [styleElement] A style element constructed by the `buildStyle()` utility. Optional.
+ */
+
 {
   const MAX_BOOT_ATTEMPTS = 3600; // 60 seconds on 60Hz displays; 10 seconds on 360Hz displays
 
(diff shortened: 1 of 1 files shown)


src/content_scripts/main.js
```diff
@@ -12,68 +21,84 @@
...
+  /**
+   * @param {string[]} names Internal feature names to resolve modules for.
+   * @returns {Promise<[string, FeatureModule][]>} Entries of feature modules, keyed by name.
+   */
+  const getFeatures = names => Promise.all(
+    names.map(async (name) => [name, await getFeature(name)]),
+  );
+
+  const runFeatures = async function (names) {
+    (await getFeatures(names)).forEach(([name, module]) => {
```
marcustyphoon, line 33: `getFeatures(names).then(someVarName => someVarName.forEach(([name, module]) => {` is kind of awkward, but `for (const [name, module] of await getFeatures(names)) {` is pretty elegant, if you want to avoid the parens.

src/content_scripts/main.js
```diff
@@ -12,68 +21,84 @@
...
-    if (styleElement) {
-      styleElement.dataset.xkitFeature = name;
-      document.documentElement.append(styleElement);
-    }
+  /**
+   * @param {string[]} names Internal feature names to resolve modules for.
+   * @returns {Promise<[string, FeatureModule][]>} Entries of feature modules, keyed by name.
+   */
+  const getFeatures = names => Promise.all(
+    names.map(async (name) => [name, await getFeature(name)]),
```
marcustyphoon, line 32: This could, of course, be either a tuple or a named object. `module: await getFeature(name)` is a few more characters so normally I might avoid it, but it does read rather elegantly.

AprilSylph: Eh, I like entries just fine. Especially when they're typed properly!

src/content_scripts/main.js
AprilSylph: Am I crazy for liking `.forEach` more than `for`? I feel like `for` should be reserved for cases where `.forEach` won't cut it, like when we need to use `await` inside a loop (and we care about not executing the next item until we're done with this one). I really don't mind the parentheses at all.

marcustyphoon: I don't think so, no! Particularly in a case like this where plausibly if we did add an await somewhere in the body we would want the `forEach async` behavior rather than the for-of await. Or, perhaps more generally, using the non-async forEach construction means that you can't later add an await to the body without adding the keyword to the callback, which inherently gives you an opportunity to think about which async loop behavior you actually want.

AprilSylph: > if we did add an await somewhere in the body we would want the `forEach async` behavior rather than the for-of await

Nooooo, that would make all the callbacks run ~simultaneously, and then we'd have inconsistent timing again!

marcustyphoon: Only the parts after the await! The part of an async function before the first await is run synchronously (i.e. in order). So, like, if we fetched the stylesheets and constructed them (for some reason), and used an inline await instead of .then(), we'd be fine so long as the main() call was before that.

marcustyphoon (approved):
Reviewed as the diff to master (using my alias `!git checkout --no-overlay "$1" -- . #`), as this is more replacement than additive change.

marcustyphoon:
You know, I have no idea how github native stacks interact with merge type rules. Does this become two squash commits or one? Normally two, I believe.

AprilSylph:
Let's find out!

AprilSylph:
Yep, two squashed commits. It's just like merging them separately, except slower!

marcustyphoon:
Well, if you did have one pointing into the other like this and merged the first one normally, it would auto close the second one if you have branch deletion on, right? And if you didn't, you'd have to manually rebase the second one because you squash merged the first one.

If you wanted a single squash commit, merging the second one into the first one (not using github's stacks) would be how you would do that in any case, of course. In this case that could make sense, as the second reverts a bunch of the first. But if it's an additive situation I think the stack mechanics are helpful? (The single click rebase-both on web is also nice.)

PR #2369, opened 2026-09-02
perf: do not load jQuery in the control panel

AprilSylph:
### Description

We don't really rely jQuery in the control panel as of #2042, but we're still evaluating and using it for... one line of code.

Let's not do that. Vanilla JS will do just fine.

This isn't really a refactor since, if anything, it makes that one line of code technically slightly harder to read. But I'd still like to do this anyway, since it saves CPU cycles for _basically_ no cost to the developer experience.

### Screenshots

N/A

### Testing steps

Load the modified addon, and open the control panel.

Smoke-test the filter dropdown; it should correctly hide and unhide features when you select Enabled, Disabled, or All.

Smoke-test Quick Tags' tag bundle management; reordering tag bundles should still work as before. (TIL: Sortable uses jQuery if it finds that `$` is defined.)

Smoke-test Painter's preferences; all the colour preferences should still bring up colour pickers, and those colour pickers should still work as before. (Coloris doesn't hook into jQuery in any way, but jQuery was originally only added to the control panel to enable Spectrum colour-picking, which was only added for `color`-type preferences.)

### DIFF
--- a/src/action/popup.html
+++ b/src/action/popup.html
@@ -11,7 +11,6 @@
     <link rel="stylesheet" href="./acorn.css">
     <link rel="stylesheet" href="./popup.css">
     <script defer src="../browser_polyfill.js"></script>
-    <script defer src="../lib/jquery.slim.min.js"></script>
     <script defer src="./popup.js"></script>
     <script defer src="./render_backup.js"></script>
     <script type="module" src="./components/sponsor-progress/index.js"></script>
--- a/src/action/popup.js
+++ b/src/action/popup.js
@@ -81,7 +81,7 @@ document.getElementById('filter').addEventListener('input', event => {
 
   switch (event.currentTarget.value) {
     case 'all':
-      $('.filter-hidden').removeClass('filter-hidden');
+      featureElements.forEach(featureElement => featureElement.classList.remove('filter-hidden'));
       break;
     case 'enabled':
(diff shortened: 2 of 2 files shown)


marcustyphoon:
You know, I can't say I've ever had a rule for myself that a refactor can't add more code :D Or, I guess, make the code arguably worse in some way... but that's sort of hard to define, as if a PR made the code objectively worse, we wouldn't merge it.

marcustyphoon (approved):

marcustyphoon:
I tried to get cute with it and, imo, did not succeed.

```js
document.getElementById('filter').addEventListener('input', ({ currentTarget: { value } }) => {
  [...document.querySelectorAll('xkit-feature')].forEach(featureElement =>
    featureElement.classList.toggle(
      'filter-hidden',
      { all: false, enabled: featureElement.disabled, disabled: !featureElement.disabled }[value],
    ),
  );
  /* */
```

**edit:** Eh, not really better.

```js
document.getElementById('filter').addEventListener('input', (event) => {
  [...document.querySelectorAll('xkit-feature')].forEach(featureElement => {
    const shouldBeHidden = {
      all: false,
      enabled: featureElement.disabled,
      disabled: !featureElement.disabled,
    }[event.currentTarget.value];

    featureElement.classList.toggle('filter-hidden', shouldBeHidden);
    if (shouldBeHidden) {
      featureElement.shadowRoot.querySelector('details').open = false;
    }
  });
  checkForNoResults();
});
```

PR #2373, opened 2026-09-03
feat(utils): add custom cleanup utilities

AprilSylph:
### Description

I was thinking—hey, we managed to squash the extension bundle a ton just by removing RemixIcon, why not do the same with jQuery? We only use it for a select few things. Why not just write some helpers around `querySelectorAll` instead?

So, I did that, starting with just the `utils/` folder so as to limit the testing needed per approval. Replacing all our jQuery usages in one PR would be nuts.

The one jQuery usage that is not supplanted by a custom util is the usage of `$.param()` in the mega edit utility, which has instead been replaced with its vanilla equivalent (`(new URLSearchParams()).toString()`). These methods differ in that `$.param` can accept and properly convert object/array values to parameters, but we're not supplying `$.param` with anything except string values, so it doesn't actually matter.

Only now that I open this PR do I realise that jQuery Slim is only 56 KB uncompressed, so even if we go through with this, the extension bundle won't shrink that much and will likely have approximately zero performance impact on any machine.

...but it's about the principle of the thing! We don't need it, we shouldn't bundle it!

Oh, I guess removing it would resolve some `web-ext lint` warnings. That's always good.
(Remind me to try out another different colour picker library later...)

### Screenshots

N/A, no behavioural changes expected

### Testing steps

Load the modified addon on Firefox, and verify that the following are still removed and re-inserted on extension restart:
- Post control buttons (Quick Tags, Trim Reblogs)
- Hidden post controls (PostBlock, Tweaks)
- `<aside id="xkit-symbols">` (Quick Tags, Trim Reblogs)
- Meatball menu buttons (PostBlock, Mirror Posts, NotificationBlock)
- Post form control buttons (Quick Tags)
- Sidebar sections (Limit Checker, Tag Tracking+)

Additionally, on any browser:
- Hidden posts should still be unhidden when expanding them on a blog view permalink page
- Hidden posts should still be unhidden when disabling the respective feature/option that hid them
- Meatball menu buttons should still be removed when disabling the respective feature that added them
- Mass Deleter should still be able to delete posts
- Tag Replacer should still be able to edit posts

### DIFF
--- a/src/utils/cleanup.js
+++ b/src/utils/cleanup.js
@@ -1,0 +1,56 @@
+/**
+ * Remove elements that have the specified ID from the document.
+ * @param {string} id A string representing the ID of the element to remove.
+ */
+export function removeElementsById (id) {
+  [...document.querySelectorAll(`#${CSS.escape(id)}`)].forEach(element => element.remove());
+}
+
+/**
+ * Remove elements that have the specified class name from the document.
+ * @param {string} className A string representing the class name of the elements to remove.
+ */
+export function removeElementsByClassName (className) {
+  [...document.querySelectorAll(`.${CSS.escape(className)}`)].forEach(element => element.remove());
+}
+
+/**
+ * Remove elements that have the specified attribute from the document.
+ * @param {string} attribute A string representing the attribute name to match.
+ */
+export function removeElementsByAttribute (attribute) {
+  [...document.querySelectorAll(`[${CSS.escape(attribute)}]`)].forEach(element => element.remove());
+}
+
+/**
+ * Remove elements that match the given CSS selector from the document.
+ * @param {string} selector A string containing one or more selectors to match.
+ */
+export function removeElementsBySelector (selector) {
+  [...document.querySelectorAll(selector)].forEach(element => element.remove());
+}
+
+/**
+ * Remove child elements that have the specified attribute from a parent element.
+ * @param {Element} parentNode The element to remove children from.
+ * @param {string} attribute A string representing the attribute name to match.
+ */
+export function removeChildrenByAttribute (parentNode, attribute) {
+  [...parentNode.querySelectorAll(`:scope > [${CSS.escape(attribute)}]`)].forEach(element => element.remove());
+}
+
+/**
+ * Remove the specified class name from all elements in the document.
+ * @param {string} className A string representing the class name to remove.
+ */
+export function removeClassNameFromElements (className) {
+  [...document.querySelectorAll(`.${CSS.escape(className)}`)].forEach(element => element.classList.remove(className));
+}
+
+/**
+ * Remove the specified attribute from all elements in the document.
+ * @param {string} attribute A string representing the name of the attribute to remove.
+ */
+export function removeAttributeFromElements (attribute) {
+  [...document.querySelectorAll(`[${CSS.escape(attribute)}]`)].forEach(element => element.removeAttribute(attribute));
+}
--- a/src/utils/cleanup.js
+++ b/src/utils/cleanup.js
@@ -43,7 +43,7 @@ export function removeChildrenByAttribute (parentNode, attribute) {
  * Remove the specified class name from all elements in the document.
  * @param {string} className A string representing the class name to remove.
  */
-export function removeClassNameFromElements (className) {
+export function removeClassName (className) {
   [...document.querySelectorAll(`.${CSS.escape(className)}`)].forEach(element => element.classList.remove(className));
 }
 
@@ -51,6 +51,6 @@ export function removeClassNameFromElements (className) {
  * Remove the specified attribute from all elements in the document.
  * @param {string} attribute A string representing the name of the attribute to remove.
  */
-export function removeAttributeFromElements (attribute) {
+export function removeAttribute (attribute) {
   [...document.querySelectorAll(`[${CSS.escape(attribute)}]`)].forEach(element => element.removeAttribute(attribute));
 }
--- a/src/utils/mega_editor.js
+++ b/src/utils/mega_editor.js
@@ -49,6 +49,6 @@ export const megaEdit = async function (postIds, options) {
 
   return inject(
     '/main_world/post_request.js',
-    [`https://www.tumblr.com/${pathname}`, $.param(requestBody)],
+    [`https://www.tumblr.com/${pathname}`, (new URLSearchParams(requestBody)).toString()],
   );
 };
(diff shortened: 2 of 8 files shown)


marcustyphoon:
Ah, so, #697 :D

AprilSylph:
What was I doing 4 years ago that I didn't want to review-- [checks datestamp] oh. That's why.

Um, sorry.

marcustyphoon:
Aside: I was thinking it might be a good time for a release! Particularly if Mozilla winds up taking longer to review the icon changes, since they're technically additions. If this is the next project, it's both a removal (should review well) and not user facing (doesn't need to be shipped at any particular time).

marcustyphoon:
> What was I doing 4 years ago that I didn't want to review-- [checks datestamp] oh. That's why.
> 
> Um, sorry.

Oh no worries! I can't say I actually have a preference between the jquery and custom syntax.

AprilSylph:
Oh, I'm not even really done with the icons, this was just easier than the next (final?) step of that.

marcustyphoon:
Oh, right, yeah, I forgot about that.

AprilSylph:
> Anyway, we could adopt the multiple-arguments-are-allowed thing in the new PR if desired.

Hmm, I'd rather not honestly, it make things weird for `removeChildrenByAttribute`. `string, ...string[]` is an odd argument signature, and having it among a sea of `...string[]` signatures would make it even more confusing.

marcustyphoon:
Obviously removeChildrenByAttribute should be added to the Element prototype :3

(for 100% clarity this is a joke)

marcustyphoon:
Out of curiosity: JQuery removal is 355k -> 335k gzipped, about 5 percent.

Actually—with a grain of salt, I had a local LLM produce an unreadable shell oneliner—now I'm curious what we get if we just gzip everything individually and sort...

```
   36162     156326   ./lib/moment.js
   28153     119403   ./lib/sortable.esm.js
   19481      56032   ./lib/jquery.slim.min.js
   10113      42971   ./lib/coloris.js
    8771       8740   ./icons/128.png
    7441       7411   ./icons/96.png
    6029       5999   ./icons/48.png
    5921       5891   ./icons/64.png
    5376       7966   ./content_scripts/meatballs.css
    4664      16710   ./features/quick_reblog/index.js
```

We could cut like 10% by replacing moment, but Luxon's format strings are different (...why?), day.js's modularity [doesn't really seem like it's being helpful here if I'm honest](https://day.js.org/docs/en/plugin/advanced-format), and I care about shaving 10% off of a web extension package... precisely not at all, if the code isn't being made cleaner.

src/utils/cleanup.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Remove elements that have the specified ID from the document.
+ * @param {string} id A string representing the ID of the element to remove.
+ */
+export function removeElementsById (id) {
+  [...document.querySelectorAll(`#${CSS.escape(id)}`)].forEach(element => element.remove());
```
marcustyphoon, line 6: Reads a little funny, verbally, not to use getElementById here, but it wouldn't be robust that way, obviously.

src/utils/cleanup.js
```diff
@@ -0,0 +1,56 @@
...
+  [...document.querySelectorAll(`.${CSS.escape(className)}`)].forEach(element => element.classList.remove(className));
+}
+
+/**
+ * Remove the specified attribute from all elements in the document.
+ * @param {string} attribute A string representing the name of the attribute to remove.
+ */
+export function removeAttributeFromElements (attribute) {
+  [...document.querySelectorAll(`[${CSS.escape(attribute)}]`)].forEach(element => element.removeAttribute(attribute));
+}
```
marcustyphoon, line 56: While these names read to me like they should have a second argument (which elements?), I don't have a better idea.

src/utils/mega_editor.js
```diff
@@ -49,6 +49,6 @@ export const megaEdit = async function (postIds, options) {
 
   return inject(
     '/main_world/post_request.js',
-    [`https://www.tumblr.com/${pathname}`, $.param(requestBody)],
+    [`https://www.tumblr.com/${pathname}`, (new URLSearchParams(requestBody)).toString()],
```
marcustyphoon, line 52: This should be mentioned in the PR body, for reference!

Anyway, I thiiiiiiiink it works? It's definitely not directly equivalent, but we have no official docs about what Tumblr is supposed to support, right?

```js
const test = { string: 'a', array: [1, 2, 3], object: { hello: 'world' } };

// string=a&array[]=1&array[]=2&array[]=3&object[hello]=world
console.log('$.param(test)', decodeURIComponent($.param(test)));

// string=a&array=1,2,3&object=[object+Object]
console.log('new URLSearchParams(requestBody).toString()', decodeURIComponent(new URLSearchParams(test).toString()));
```

AprilSylph: We don't; this endpoint is completely undocumented.

We already convert all the values to string here, so we're not using the depth feature of `$.param` anyway:

https://github.com/AprilSylph/XKit-Rewritten/blob/8f9e7aa6940a829f87a8641ff0793d4568c99ec0/src/utils/mega_editor.js#L40-L44

src/utils/cleanup.js
AprilSylph: `remove*FromDocument` would probably be more accurate.

AprilSylph: No... that kinda implies removing stuff from the document element... hmm...

AprilSylph: `removeClassNameFromAll` / `removeAttributeFromAll`?

AprilSylph: ...`removeClassNameFromAttached` / `removeAttributeFromAttached`?

src/utils/mega_editor.js
marcustyphoon: Oh, whoops, I missed the `.join`s.

src/utils/cleanup.js
marcustyphoon: I assume we don't want to do straight `removeClassName`/`removeAttribute`... though I think it does read the clearest when I put it side by side with a bunch of `removeElementsByWhatever` calls. The difference in length makes it more obvious whether you used the wrong one.

AprilSylph: True!

marcustyphoon (approved):
Approved, whatever we do re: names.

PR #2374, opened 2026-09-04
docs: rewrite Chapter 4.2 - Feature modules.md

AprilSylph:
### Description

- relates to #2355

Gives the same treatment to the feature modules document as was given to the feature metadata document.

There's no schema to validate and cross-reference this time, but there's also only five exports to document, so I'm pretty sure I got everything right.

### Screenshots

N/A, documentation changes only

### Testing steps

Proofread [the new version of `docs/Chapter 4.2 - Feature modules.md`](https://github.com/AprilSylph/XKit-Rewritten/blob/aprilsylph/documentation/feature-modules-rewrite/docs/Chapter%204.2%20-%20Feature%20modules.md). Can you spot any grammatical errors, opportunities to clarify, or factual inaccuracies?

Also, look at that new table of contents... mmmm-mm. Delicious.

### DIFF
--- a/docs/Chapter 4.2 - Feature modules.md
+++ b/docs/Chapter 4.2 - Feature modules.md
@@ -1,33 +1,90 @@
 # Feature modules
 
-Modules may export any of the following:
+Every feature must have a module file to define its behaviour.
 
-## `main()`
-- Type: Async Function
-- Required: No
+Feature module files are expected to export certain named constants by the main content script:
 
-The main function of the feature. Will be called whenever the user enables the feature, even if the tab(s) XKit is running in is/are not focused. Will also be called upon pageload if the feature is enabled.
+<br>
 
-## `clean()`
-- Type: Async Function
-- Required: No
+## Methods
 
-The cleanup function of the feature. Called whenever the user disables the feature, also regardless of tab focus.
+### `main()`
 
-## `onStorageChanged()`
-- Type: Async Function
-- Required: No
+|                 |                                                                                               |
+|-----------------|-----------------------------------------------------------------------------------------------|
+| **Type**        | `() => Promise<void>`                                                                         |
+| **Mandatory**   | No                                                                                            |
+| **Description** | The main function of the feature. Everything this function does must be undone by `clean()`.  |
+| **Example**     | <pre lang="js">export const main = async () => addSidebarItem(sidebarOptions);</pre>          |
 
-The preference-handling code of the feature. Added as a `browser.storage.local.onChanged` listener when the feature is enabled, and removed when the feature is disabled. If the module does not export this function, the feature will be automatically restarted when its preferences are changed.
+When a Tumblr tab is loaded with XKit Rewritten enabled, each enabled feature is evaluated and runs its `main()` function.
 
-## `stylesheet`
-- Type: Boolean
-- Required: No
+When XKit Rewritten is running in any open Tumblr tabs, and the user enables an XKit Rewritten feature:
+1. If the feature has not run in any of those tabs previously, the feature module is evaluated in those tabs.
+2. The feature's `main()` function is then run in those tabs, even if those tabs are not active or focused.
 
-Whether the feature has a static stylesheet. If true, there should be a `.css` file of matching name in the same directory level. The stylesheet is automatically added and removed during the feature's lifecycle.
+<br>
 
-## `styleElement`
-- Type: HTMLStyleElement
-- Required: No
+### `clean()`
 
-An HTML `<style>` element containing computed and/or dynamic styles, as created by the `buildStyle` utility function. The element is automatically added to the document root and removed during the feature's lifecycle.
+|                 |                                                                                             |
+|-----------------|---------------------------------------------------------------------------------------------|
+| **Type**        | `() => Promise<void>`                                                                       |
+| **Mandatory**   | No                                                                                          |
+| **Description** | The cleanup function of the feature. This function must undo everything done by `main()`.   |
+| **Example**     | <pre lang="js">export const clean = async () => removeSidebarItem(sidebarOptions.id);</pre> |
+
+When a user disables an XKit Rewritten feature, the feature's `clean()` function is run in all Tumblr tabs where that feature's `main()` function has run, even if those tabs are not active or focused.
+
+<br>
+
+### `onStorageChanged()`
+
+|                 |                                                                                                                                 |
+|-----------------|---------------------------------------------------------------------------------------------------------------------------------|
+| **Type**        | <code>(changes: Record\<string, <a href="https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageChange">StorageChange</a>\>) => Promise\<void\></code>  |
+| **Mandatory**   | No                                                                                                                              |
+| **Description** | The preference-handling code of the feature. If not specified, the feature will be restarted when its preferences are changed.  |
+| **Example**     | <pre lang="js">export const onStorageChanged = async (changes) => Object.keys(changes).some(key => key.startsWith('panorama')) && main();</pre> |
+
+If provided, this function is added as a [`storage.StorageArea.onChanged`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged) listener on [`browser.storage.local`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/local).
+
+This listener fires on _all_ changes to `browser.storage.local`, even if the changes are to a different feature's preferences or to storage values that are not preferences at all. This is useful for reacting to changes in another feature's data (e.g.: Quick Reblog displaying new Quick Tags bundles as they are created), or applying changes made to the feature's own custom storage key across all open Tumblr tabs (e.g.: PostBlock hiding newly-hidden posts).
+
+If not provided, the feature's `clean()` and `main()` functions will be called sequentially whenever its preferences are changed.
+
+<br>
+
+## Properties
+
+### `stylesheet`
+
+|                 |                                                                                                     |
+|-----------------|-----------------------------------------------------------------------------------------------------|
+| **Type**        | `boolean`                                                                                           |
+| **Mandatory**   | No                                                                                                  |
+| **Description** | Whether the feature has a static stylesheet, found at `./index.css` in the feature directory if so. |
+| **Example**     | <pre lang="js">export const stylesheet = true;</pre>                                                |
+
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Link for my own clickies: https://github.com/AprilSylph/XKit-Rewritten/blob/aprilsylph/documentation/feature-modules-rewrite/docs/Chapter%204.2%20-%20Feature%20modules.md

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,90 @@
...
+| **Example**     | <pre lang="js">export const onStorageChanged = async (changes) => Object.keys(changes).some(key => key.startsWith('panorama')) && main();</pre> |
+
+If provided, this function is added as a [`storage.StorageArea.onChanged`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged) listener on [`browser.storage.local`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/local).
+
+This listener fires on _all_ changes to `browser.storage.local`, even if the changes are to a different feature's preferences or to storage values that are not preferences at all. This is useful for reacting to changes in another feature's data (e.g.: Quick Reblog displaying new Quick Tags bundles as they are created), or applying changes made to the feature's own custom storage key across all open Tumblr tabs (e.g.: PostBlock hiding newly-hidden posts).
+
+If not provided, the feature's `clean()` and `main()` functions will be called sequentially whenever its preferences are changed.
```
marcustyphoon, line 54: One thing that this doesn't make totally clear is that the _default_ behavior is to restart the feature, which is slow/expensive, and that supplying `onStorageChanged()` _replaces_ that behavior. Thus, it's useful if a feature can apply its preference changes in a more lightweight way or wants to opt out of running clean and creating UI flicker (examples: Classic Footer, Panorama), but means features that want to add additional preference handling code but also to keep the restart behavior must implement it (example: Vanilla Video).

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,90 @@
...
+| **Type**        | `HTMLStyleElement`                                                                            |
+| **Mandatory**   | No                                                                                            |
+| **Description** | A JavaScript pointer to a `<style class="xkit">` element created by the `buildStyle` utility. |
+| **Example**     | <pre lang="js">export const styleElement = buildStyle();</pre>                                |
+
+When a feature is run (i.e., its `main()` function is called), and that feature exports this constant, the exported `<style>` element is also added to the document root.
+
+When the same feature is disabled, its `<style>` element is also removed from the document.
+
+The benefit of using a `styleElement` over a static `stylesheet` is the ability to include CSS constructed at runtime, which is invaluable when considering the generated nature of Tumblr's CSS class names.
```
marcustyphoon, line 88: I would make this reference `keyToCss` explicitly. Also, link to https://github.com/tumblr/docs/blob/master/web-platform.md#getcssmap, why not. "The generated nature of Tumblr's CSS class names" means nothing by itself!

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,90 @@
...
+
+This element is never cloned, nor does it ever expire within XKit Rewritten's running lifecycle. Therefore, it is possible to create and export the pointer first, and then construct the CSS later. This is a useful pattern when the CSS the feature wants to construct varies based on the user's preferences.
```
marcustyphoon, line 90: Perhaps consider dropping the "pointer" language? We export a style element; it is possible to export the element first and construct and apply the CSS content later. Referencing a pointer makes me think there's more going on here.

marcustyphoon:
I do like the old "any of" language existing somewhere, and/or some new language that indicates that these are the constants that *can* be exported in order to implement the feature's desired functionality. Yes, they're all labelled as not required, but so is the icon field in the feature manifest; there's a difference between "the framework won't crash if you omit this, but your feature isn't complete without it" and "there is no expectation that you include this; use it if it's what you need." (For maximum clarity, this could link to a css-only feature.)

docs/Chapter 4.2 - Feature modules.md
AprilSylph: Restructured these paragraphs:
- **Before**: Basics &rarr; Details &rarr; Fallback
- **After**: Basics &rarr; Rationale &rarr; Details

https://github.com/AprilSylph/XKit-Rewritten/blob/c22741e9874e306c526e2a14d35da8749788cc13/docs/Chapter%204.2%20-%20Feature%20modules.md?plain=1#L50-L54

Honestly, it feels odd to not end on the fallback, but the rationale doesn't make sense without front-loading the default behaviour. And, is it really a "fallback" if most features rely on it? No, calling it a "default" makes much more sense.

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
...
-Modules may export any of the following:
+Every feature must have a module file to define what it is supposed to do when run on a Tumblr webpage.
 
-## `main()`
-- Type: Async Function
-- Required: No
+Feature module files are expected to export any of the named constants outlined in this document, as necessary to achieve the feature's intended behaviour.
 
-The main function of the feature. Will be called whenever the user enables the feature, even if the tab(s) XKit is running in is/are not focused. Will also be called upon pageload if the feature is enabled.
+While none of these constants are technically required, a feature is not complete without exporting at least one of `main()`, `stylesheet`, or `styleElement`.
```
marcustyphoon, line 7: Challenge accepted: write a feature that only uses `onStorageChanged()` :D

(No, that would be silly.)

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
 # Feature modules
 
-Modules may export any of the following:
+Every feature must have a module file to define what it is supposed to do when run on a Tumblr webpage.
```
marcustyphoon, line 3: Could also use the term "modifications" in some way!

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
...
+|-----------------|---------------------------------------------------------------------------------------------------------------------------------------|
+| **Type**        | <code>(changes: Record\<string, <a href="https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageChange">StorageChange</a>\>) => Promise\<void\></code>  |
+| **Mandatory**   | No                                                                                                                                    |
+| **Description** | The preference change-handling code of the feature. If not specified, the feature will be restarted when its preferences are changed. |
+| **Example**     | <pre lang="js">export const onStorageChanged = async (changes) => Object.keys(changes).some(key => key.startsWith('panorama')) && main();</pre> |
+
+If provided, this function is added as a [`storage.StorageArea.onChanged`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/onChanged) listener on [`browser.storage.local`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/local).
```
marcustyphoon, line 52: Only when the feature is enabled, I suppose.

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
...
+| **Example**     | <pre lang="js">export const styleElement = buildStyle();</pre>        |
+
+When a feature is run (i.e., its `main()` function is called), and that feature exports this constant, the exported `<style>` element is also added to the document root.
+
+When the same feature is disabled, this `<style>` element is also removed from the document.
+
+The benefit of using a `styleElement` over a static `stylesheet` is the ability to include CSS constructed at runtime, including (but not limited to) using the output of the [`keyToCss()`](../src/utils/css_map.js) utility to target Tumblr's own elements. This is necessary for both robustness and readability because [Tumblr uses compiled class names](https://github.com/tumblr/docs/blob/master/web-platform.md#getcssmap).
+
+This element is never cloned, nor does it ever expire within XKit Rewritten's running lifecycle. Therefore, it is possible to create and export an empty `<style>` element first, and then construct the CSS later. This is a useful pattern when the CSS the feature wants to construct varies based on the user's preferences.
```
marcustyphoon, line 92: Could be more specific about how:

```suggestion
This element is never cloned, nor does it ever expire within XKit Rewritten's running lifecycle. Therefore, it is possible to create and export an empty `<style>` element first, and then construct the CSS and set its textContent later. This is a useful pattern when the CSS the feature wants to construct varies based on the user's preferences.
```

Or something like that.

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
...
+<br>
+
+### `styleElement`
+
+|                 |                                                                       |
+|-----------------|-----------------------------------------------------------------------|
+| **Type**        | `HTMLStyleElement`                                                    |
+| **Mandatory**   | No                                                                    |
+| **Description** | A `<style class="xkit">` element created by the `buildStyle` utility. |
+| **Example**     | <pre lang="js">export const styleElement = buildStyle();</pre>        |
```
marcustyphoon, line 84: Kind of tempted to include a keyToCss call here. But it should be clear from looking at a feature that uses the export, which someone will realistically have to do anyway. If we were trying to target "the documentation on its own is enough to write a feature module," we would need to fully document all of the util exports in markdown, which... no thanks.

docs/Chapter 4.2 - Feature modules.md
```diff
@@ -1,33 +1,92 @@
...
+## Properties
+
+### `stylesheet`
+
+|                 |                                                                                                     |
+|-----------------|-----------------------------------------------------------------------------------------------------|
+| **Type**        | `boolean`                                                                                           |
+| **Mandatory**   | No                                                                                                  |
+| **Description** | Whether the feature has a static stylesheet, found at `./index.css` in the feature directory if so. |
+| **Example**     | <pre lang="js">export const stylesheet = true;</pre>                                                |
```
marcustyphoon, line 69: Dawns on me that this could be `export const stylesheet = './index.css';` if we were making everything more generic (see: https://github.com/AprilSylph/XKit-Rewritten/issues/2356#issuecomment-5438003312). We aren't, though, so.

docs/Chapter 4.2 - Feature modules.md
marcustyphoon, line 3: We call it a "module script" in chapter 3; it could be "module script file." (Or "javascript module file" or etc.)

docs/Chapter 4.2 - Feature modules.md
marcustyphoon: "set its textContent property to a constructed CSS value later" etc.

marcustyphoon (approved):

PR #2375, opened 2026-09-04
feat(Tag Tracking+): Reduce API load when multiple Tumblr tabs are open

### LINKED ISSUE #1813: Tag Tracking+: Reduce API load when multiple Tumblr tabs are open
### Concept

It would require a fairly clever use of extension storage, at least in all of the iterations I've thought of so far, but we should be able to share data between Tumblr tabs to prevent Tag Tracking+ from generating additional unnecessary API load when multiple tabs are open.

This is especially important if a user follows many, many tags. Opening multiple Tumblr tabs in succession can currently cause up to hundreds of API requests in quite rapid succession.

### LINKED ISSUE #2338: Tag Tracking+: Infinite loop if user tracks zero tags
### Platform

macOS 14.8.8 arm64

### Browser

Chromium 151.0.7922.77

### Addon version

v1.3.6

### Details

https://github.com/AprilSylph/XKit-Rewritten/blob/40a87e68356f5227d46c4d10301f8d004bcaff0a/src/features/tag_tracking_plus/index.js#L93

This makes an interval with zero delay if the user doesn't track any tags. It doesn't seem to cause a noticeable performance hit on my mac in practice, but also, neither did the one that was literally locking people out of using Tumblr, so...

marcustyphoon:
### Description

I think this is my first version of #1813 that I'm not aware of any bugs in.

This is a full(?) version of what I hoped to do to Tag Tracking+: an implementation which:
- does not perform more API fetches with many tabs open than with one tab open
- does not perform more API fetches when opening/refreshing many tabs than when opening/refreshing one tab
- does not perform as many API fetches if you open a tab, close it, open a new tab, close it, etc. as previously
- does not meaningfully change the user experience
- improves edge case behavior (repeatedly disabling/enabling the feature, only tracking one tag)

General description of the implementation:
- Unread counts are stored in storage, along with their last-updated timestamps, which is a single source of truth for every tab (besides the dom). A count never goes back to `...` and a sidebar never goes back to `data-loaded=false`; besides this, all tabs' sidebars are rendered using the same data (`updateSidebar` depends only on storage and is called only in `onStorageChanged`).
- As the storage is a cache—any entries older than 30 seconds are wiped on feature enable—I had the idea that we could exclude the storage entry from the backup tab export, but we of course don't have to do this. (Every tag tracked by any user that's been logged in with the feature enabled on the installation is already in there; we're adding basically no additional data to the export.)
- Only one tab may refresh at a time (per tracked tag list)*, implemented by each tab keeping a timestamp of when any other tab has refreshed and declining to refresh unless it's been longer than the desired interval (i.e. the actively refreshing tab must have been closed). This isn't 100% robust (closed-tab race condition chance; waking a computer from sleep), but it should fix itself quickly.
- Refresh behavior is also dependent entirely on storage contents (checked in real time), so it doesn't matter which tab does the refresh.
  - A new tab marks all of its tracked tags as "needs initial load" in `main`; if any tags need initial load, they're loaded in order by the refreshing tab at up to 2/sec.
  - If no tags need initial load, tracked tags are refreshed in least-up-to-date order at an arbitrary rate until (with few enough tracked tags) they're all fresher than an arbitrary ttl. (Not sure what we want to set these numbers to, but I think it makes sense to have both of them.)
- Stopping the refresh loop is rewritten so that it actually stops immediately (idk what to call the technique honestly).
- On initial load—i.e. when hard navigating to tumblr, refreshing the page, or enabling the feature—storage contents are wiped **unless** they're quite recent (30 seconds). This means all or a significant chunk of an initial load can be reused if you f5 during it, open a new tab during it, etc, but f5 still means "fetch new counts, please."
- _In a separate commit, because maybe this is getting too cute:_ when in show-all-counts mode, even though fresh cached counts are used during initial load instead of fetching, a "loaded" boolean field is used to make them not appear in the sidebar immediately (which looks kind of strange if they're preceded by some non-cached tags); instead, they become "loaded" and visible when they would otherwise have been fetched, preserving the "waterfall" sequential load sequence. Yes, this means the caching doesn't meaningfully reduce the load time in show-all-counts mode if you keep this commit.

*Not tested, but I think conceptually the e.g. "six Firefox container tabs logged into three accounts with different tracked tag lists" scenario should work correctly (at up to 3x API load vs a single tab, depending on tracked tag count and overlap). Tabs with the same tracked tag list should elect a single tab that's allowed to refresh.

Note: I don't love the errored-tag behavior here, but I'm not sure what the ideal behavior actually is. The main problem when deciding that is that in practice the main way I see the ⚠️ appear is when interrupting a fetch by closing a tab/putting the computer to sleep/disconnecting from the internet (probably shouldn't sync that to other tabs at all, and if one does, the tab should probably immediately retry the errored tag), but that doesn't necessarily demand the same behavior as "something is wrong with fetching this tag specifically" (in which case one should probably only reattempt it as part of the cycle... or at minimum shouldn't spam attempts every half a second at the expense of all other tags).

Draft because of that, because console.info() calls are not removed currently, and because maybe I can still think of a solution with fewer lines of code.

**edit:** Oh, and reminder to myself to consider using `document.hasFocus()` in some way. Also, potentially always run a 500ms loop and do nothing if the count mod 60 (or whatever) isn't zero in background-refresh mode, as this is more flexible.

### Screenshots

n/a (I could take a video of side by side tabs with devtools open, I suppose, but that would be a lot of work)

### Testing steps

tba

### DIFF
--- a/src/action/render_backup.js
+++ b/src/action/render_backup.js
@@ -20,8 +20,14 @@ const sleep = ms => new Promise(resolve => setTimeout(() => resolve(), ms));
 
 class UserInterrupt extends Error { name = 'UserInterrupt'; }
 
+const getCurrentStorage = () => browser.storage.local.get()
+  .then(storageLocal => {
+    const entries = Object.entries(storageLocal);
+    return Object.fromEntries(entries.filter(([key]) => !key.startsWith('caches.')));
+  });
+
 const onStorageChanged = async function () {
-  const storageLocal = await browser.storage.local.get();
+  const storageLocal = await getCurrentStorage();
   exportValueTextarea.value = JSON.stringify(storageLocal, null, 2);
 };
 
@@ -42,7 +48,7 @@ const localCopy = () => {
 };
 
 const localExport = async function () {
-  const storageLocal = await browser.storage.local.get();
+  const storageLocal = await getCurrentStorage();
   const stringifiedStorage = JSON.stringify(storageLocal, null, 2);
   const storageBlob = new Blob([stringifiedStorage], { type: 'application/json' });
   const blobUrl = URL.createObjectURL(storageBlob);
@@ -114,7 +120,7 @@ async function onImportSubmit (event) {
     importSuccessBar.hidden = true;
     importErrorBar.hidden = true;
 
-    const currentStorage = await browser.storage.local.get();
+    const currentStorage = await getCurrentStorage();
     const parsedStorage = JSON.parse(importText);
 
     if (Object.keys(currentStorage).length !== 0) {
--- a/src/features/tag_tracking_plus/index.js
+++ b/src/features/tag_tracking_plus/index.js
@@ -6,19 +6,52 @@ import { addSidebarItem, removeSidebarItem } from '../../utils/sidebar.js';
 import { tagTimelineFilter } from '../../utils/timeline_id.js';
 import { apiFetch, onClickNavigate } from '../../utils/tumblr_helpers.js';
 
-const storageKey = 'tag_tracking_plus.trackedTagTimestamps';
+const timestampsStorageKey = 'tag_tracking_plus.trackedTagTimestamps';
+/** @type {Record<string, number>} */
 let timestamps;
+
+const unreadCountsStorageKey = 'caches.tag_tracking_plus.unreadCounts';
+/** @type {Record<string, { unreadCountString: string, updated: number }>} */
+let unreadCounts;
 
 const excludeClass = 'xkit-tag-tracking-plus-done';
 const includeFiltered = true;
 
-let trackedTags;
-const unreadCounts = new Map();
+let trackedTags = [];
+let trackedTagsString;
 
 let sidebarItem;
 
+const BACKGROUND_REFRESH_INTERVAL = 10_000; // Minimum time between background refresh fetches (of any tag).
+const PER_TAG_BACKGROUND_REFRESH_INTERVAL = 120_000; // Minimum time between background refresh fetches of a specific tag.
+const INTIIAL_LOAD_INTERVAL = 500; // Minimum time between initial load fetches.
+
+const INITIAL_LOAD_STORED_COUNT_MAX_AGE = 30_000; // During initial load (i.e. reloading the page), load stored counts up to this age instead of fetching.
+
+const countIsStale = (tag, ttl) => unreadCounts[tag] && Date.now() - unreadCounts[tag].updated > ttl;
+
+// If multiple browser tabs are currently refreshing the same set of tracked tag(s), only allow one to refresh.
+let lastRefreshInAnotherTab = 0;
+const otherTabRefreshChannel = new BroadcastChannel('xkit-tag-tracking-plus-refresh-sync');
+otherTabRefreshChannel.addEventListener('message', event => {
+  if (event.data.trackedTagsString === trackedTagsString) {
+    lastRefreshInAnotherTab = Date.now();
+  }
+});
+const thisTabShouldRefresh = interval => {
+  const timeSinceRefresh = Date.now() - lastRefreshInAnotherTab;
+  if (timeSinceRefresh < interval * 1.5) {
+    console.info(`Tag Tracking+: skipping refresh; another tab refreshed ${timeSinceRefresh}ms ago`);
+    return false;
+  }
+  otherTabRefreshChannel.postMessage({ trackedTagsString });
+  return true;
+};
+
 const refreshCount = async function (tag) {
   if (!trackedTags.includes(tag)) return;
+
+  console.info(`Tag Tracking+: REFRESHING ${tag}`);
 
   let unreadCountString = '⚠️';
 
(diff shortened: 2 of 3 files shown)


marcustyphoon:
Thought: the norm in this codebase is to frown upon extracting things into utils if they're not imminently going to be used in more than one place. I do think, though, that—if sufficiently well-named—an extracted function is an effective form of encapsulation. "Variable X is only used as part of the creation of behavior Y and for no other purpose, which I can immediately tell because it's in a totally different file than the rest of this code, so the total number of interactions I need to hold in my head to follow along with this is less."

(I was going to add an additional thought that this could potentially be used for one of my thoughts about reducing the API load of Mutual Checker, but honestly, it can't, really, and we don't have any other features that I can think of that would really even consider it. So... eh.)

```diff
diff --git a/src/features/tag_tracking_plus/index.js b/src/features/tag_tracking_plus/index.js
index 652b68c0..5862e744 100644
--- a/src/features/tag_tracking_plus/index.js
+++ b/src/features/tag_tracking_plus/index.js
@@ -3,6 +3,7 @@ import { onNewPosts } from '../../utils/mutations.js';
 import { getPreferences } from '../../utils/preferences.js';
 import { timelineObject } from '../../utils/react_props.js';
 import { addSidebarItem, removeSidebarItem } from '../../utils/sidebar.js';
+import { createMultiTabRateLimitFunction } from '../../utils/tab_coordination.js';
 import { tagTimelineFilter } from '../../utils/timeline_id.js';
 import { apiFetch, onClickNavigate } from '../../utils/tumblr_helpers.js';
 
@@ -24,8 +25,7 @@ let unreadCounts;
 const excludeClass = 'xkit-tag-tracking-plus-done';
 const includeFiltered = true;
 
-let trackedTags = [];
-let trackedTagsString;
+let trackedTags;
 
 let sidebarItem;
 
@@ -37,23 +37,7 @@ const INITIAL_LOAD_STORED_COUNT_MAX_AGE = 30_000; // During initial load (i.e. r
 
 const countIsStale = (tag, ttl) => unreadCounts[tag] && Date.now() - unreadCounts[tag].updated > ttl;
 
-// If multiple browser tabs are currently refreshing the same set of tracked tag(s), only allow one to refresh.
-let lastRefreshInAnotherTab = 0;
-const otherTabRefreshChannel = new BroadcastChannel('xkit-tag-tracking-plus-refresh-sync');
-otherTabRefreshChannel.addEventListener('message', event => {
-  if (event.data.trackedTagsString === trackedTagsString) {
-    lastRefreshInAnotherTab = Date.now();
-  }
-});
-const thisTabShouldRefresh = interval => {
-  const timeSinceRefresh = Date.now() - lastRefreshInAnotherTab;
-  if (timeSinceRefresh < interval * 1.5) {
-    console.info(`Tag Tracking+: skipping refresh; another tab refreshed ${timeSinceRefresh}ms ago`);
-    return false;
-  }
-  otherTabRefreshChannel.postMessage({ trackedTagsString });
-  return true;
-};
+let thisTabShouldRefresh;
 
 const refreshCount = async function (tag) {
   if (!trackedTags.includes(tag)) return;
@@ -224,7 +208,6 @@ export const onStorageChanged = async (changes) => {
 export const main = async function () {
   const trackedTagsData = (await apiFetch('/v2/user/tags')) ?? {};
   trackedTags = trackedTagsData.response?.tags?.map(({ name }) => name) ?? [];
-  trackedTagsString = JSON.stringify(trackedTags);
 
   sidebarItem = addSidebarItem({
     id: 'tag-tracking-plus',
@@ -266,6 +249,8 @@ export const main = async function () {
   }
   await browser.storage.local.set({ [unreadCountsStorageKey]: unreadCounts });
 
+  thisTabShouldRefresh = createMultiTabRateLimitFunction(`tag-tracking-plus-${JSON.stringify(trackedTags)}`);
+
   onNewPosts.addListener(processPosts);
   startRefreshLoop();
 };
@@ -275,10 +260,6 @@ export const clean = async function () {
   onNewPosts.removeListener(processPosts);
 
   removeSidebarItem('tag-tracking-plus');
-
-  trackedTags = [];
-  trackedTagsString = undefined;
-  lastRefreshInAnotherTab = 0;
 };
 
 export const stylesheet = true;
diff --git a/src/utils/tab_coordination.js b/src/utils/tab_coordination.js
new file mode 100644
index 00000000..2e062fe4
--- /dev/null
+++ b/src/utils/tab_coordination.js
@@ -0,0 +1,35 @@
+/**
+ * Creates a function for rate limiting a task, with support for coordinating between browser tabs.
+ * The returned function should be called as a necessary condition for running a task. It will return true only if the task has not been run recently within a specified interval, and if the task has not been run in another browser tab running the same code within an increased interval.
+ * Thus, tasks contingent upon this function will only execute as quickly as the specified interval. The execution will typically be done by one ("driving") tab, usually the last one opened.
+ * @param {string} id Identifier for this rate limit instance
+ * @returns {(interval: number) => boolean} A function that returns true if the task should be run.
+ */
+export const createMultiTabRateLimitFunction = id => {
+  let lastRun = 0;
+  let lastRunInOtherTab = 0;
+
+  const channel = new BroadcastChannel(`xkit-multi-tab-rate-limit-${id}`);
+  channel.addEventListener('message', () => { lastRunInOtherTab = Date.now(); });
+
+  return function thisTabShouldRunTask (interval) {
+    const now = Date.now();
+    const timeSinceThisTabRun = now - lastRun;
+    const timeSinceOtherTabRun = now - lastRunInOtherTab;
+    if (timeSinceThisTabRun < interval * 0.9) {
+      console.info(
+        `XKit Rewritten: skipping ${id.slice(0, 17)} task; this tab ran it ${timeSinceThisTabRun}ms ago`,
+      );
+      return false;
+    }
+    if (timeSinceOtherTabRun < interval * 0.9 * 2) {
+      console.info(
+        `XKit Rewritten: skipping ${id.slice(0, 17)} task; another tab ran it ${timeSinceOtherTabRun}ms ago`,
+      );
+      return false;
+    }
+    channel.postMessage(true);
+    lastRun = Date.now();
+    return true;
+  };
+};
```

marcustyphoon:
https://github.com/AprilSylph/XKit-Rewritten/compare/master...AprilSylph:XKit-Rewritten:marcustyphoon/tag-tracking-deduplicate-storage-dev-13 Hm, maybe pure-storage is actually better?

**edit:** Aw. I briefly got excited that even without a background script, we could use session storage in content scripts on a progressive enhancement basis by calling `browser.storage.session.setAccessLevel({ accessLevel: 'TRUSTED_AND_UNTRUSTED_CONTEXTS' });` in the extension popup. I hadn't realized that that's completely unimplemented in Firefox.

PR #2377, opened 2026-09-06
refactor(Mutual Checker): Use hypothetical isFollowingYou react data

marcustyphoon:
### Description

Mutual Checker's code currently checks Tumblr's internal data to see if it can determine if someone follows you, falling back to an API request if it can't. Tumblr's internal blog data can be configured by Staff to include/not include various fields (the `fields[blogs]` query parameter in the API requests performed to populate the page) in various circumstances, so while we usually don't have to fetch here, sometimes we do.

Mutual Checker doesn't use the same possibly-API-fetch-saving code path for whether a user is following you, as we didn't observe that field being included anywhere. I assume it isn't. However, there's no particular reason it couldn't be. This PR copies the code from one to the other, so that if Staff does include that field for whatever reason, we automatically save the request. It's not more maintenance burden if it actually reduces the unique line of code count even if it increases the raw line of code count, right?

Obviously feel free to not spend the time reviewing this, April, unless you can think of a reason it would ever be useful.

Note that yes, I tested this with the timeline API request hacked to always include `?is_following_you`, and it indeed does make Mutual Checker function with no additional API requests. (Of course, this is not the same as "no API impact" if Tumblr's servers must do additional work to include that field.)

### Screenshots

n/a

### Testing steps

Smoke test Mutual Checker.

### DIFF
--- a/src/features/mutual_checker/index.js
+++ b/src/features/mutual_checker/index.js
@@ -83,7 +83,7 @@ const addIcons = function (postElements) {
       ? await getIsFollowing(blogName, postElement)
       : false;
     const isMutual = followingBlog
-      ? await getIsFollowingYou(blogName)
+      ? await getIsFollowingYou(blogName, postElement)
       : false;
 
     if (isMutual) {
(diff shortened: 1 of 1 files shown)


src/features/mutual_checker/index.js
```diff
@@ -116,7 +116,7 @@ const getIsFollowing = async (blogName, element) => {
       await blogData(element),
       (await timelineObject(element))?.blog,
       (await timelineObject(element))?.authorBlog,
-    ].find((data) => blogName === data?.name);
+    ].find((data) => blogName === data?.name && data.followed !== undefined);
```
marcustyphoon, line 119: Note that this is an additional change to getIsFollowing.

PR #2380, opened 2026-09-08
refactor: drop remaining `$.param` usages

AprilSylph:
### Description

- see also: #2373 &rarr; [`ab7f750`](https://github.com/AprilSylph/XKit-Rewritten/pull/2373/changes/ab7f750d5fac61a34917f47fd98e6326ed74cf74)

We're not using the object conversion features of `$.param`, so we don't need to use it. The vanilla `URLSearchParams` constructor will do just fine, instead.

Bonus: when we know the object is going to be cast to a string, we don't even need to call `.toString()` on it manually!

So, it's literally a drop-in replacement here.

...I did write some JSdoc as well. I love types...

### Screenshots

N/A

### Testing steps

Make sure that Mass Privater and Tag Replacer both still work.
Be sure to test Mass Privater both with and without tag criteria.

### DIFF
--- a/src/features/mass_privater/index.js
+++ b/src/features/mass_privater/index.js
@@ -154,6 +154,7 @@ const showPostsNotFound = ({ name }) =>
     buttons: [modalCompleteButton],
   });
 
+/** @type {(params: { uuid: string; name: string; tags: string[]; before: number; }) => Promise<void>} */
 const privatePosts = async ({ uuid, name, tags, before }) => {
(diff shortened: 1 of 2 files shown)


marcustyphoon:
Interesting that there's no way to do this without string concat. Like, I feel like I would retroactively go back and make it ```new URL(`/v2/blog/${uuid}/posts`, { tag, before, limit: 50 }, /* base */)```.

marcustyphoon (approved):

PR #2383, opened 2026-09-08
fix(Sidebar util): Improve sidebar item spacing

### LINKED ISSUE #2368: Improve sidebar item spacing
### Description

On most sidebars, there's a 38px gap between items, which our XKit sidebar items replicate and look good with. I want to change the margin used for that case from bottom to block, for Reasons™, but that's beside the point.

In the mobile nav drawer, we use 0.75em between items instead. I think this one should be applied to the top as well—oh, hey, changing that margin from bottom to block would do that. Simple enough.

On communities pages and https://www.tumblr.com/communities, though, the sidebar is flex with gap: 20px (and I have to imagine this sidebar component will eventually be used elsewhere). What's the right spacing there, and how do we want to implement it? 38px symmetrical gaps look kind of too much; 20px symmetrical gaps (e.g. "`display: contents` the sidebar element if its parent has a set `gap` value, don't put any margin on the sidebar items themselves, and use a flex gap and margin-block for the sidebar element in normal circumstances") look too small. A fixed 38px gap between items with a reduced outside gap looks weird too.

My thought so far is something vaguely like

```css
#xkit-sidebar.in-flex {
  display: contents; /* this would actually have to be done as "displayContentsUnlessDisabledAttr" */
}
#xkit-sidebar.in-flex .xkit-sidebar-item {
  margin-block: /* some arbitrary value that looks good like 10px */;
}

.xkit-sidebar-item {
  /* no margin */
}

#xkit-sidebar {
  --gap: 38px;
  margin-block: --gap;
  display: flex;
  gap: --gap;
}

nav #xkit-sidebar {
  --gap: 0.75em;
}
```

but maybe there's a smarter thing to do here.

marcustyphoon:
### Description

Hey, actually, you know what? Unconditional `display: contents` is kind of fine actually? (We can definitely choose not to do this, if we want the ability to put increased spacing around the block of sidebar elements. But this was the simplest solution.)

This makes sidebar item spacing symmetrical top/bottom in all of the locations that I know of where we can insert a sidebar, and makes sidebar items respect the `gap` value if inserted into a flex container. Please share your opinions on spacing when reviewing this.

Resolves #2368.

#### Should be improved:

| | Before | After |
| - | - | - |
| Community Page |  |  |
| Mobile Drawer|  |  |
| Tagged Page |  |  |
| Saved Posts Page |   |  |

#### Should be identical:

| | Before | After |
| - | - | - |
| Most Pages — under premium cta, hidden recommended blogs |  |  |
| Most Pages — under search bar |  |  |
| Most Pages — under recommended blogs |  |  |
| Inbox |  |  |

### Testing steps

- Confirm that the displayed spacing is applied in the specified locations. Try to find other places we insert sidebar items that I have forgotten; do they have correct-looking spacing?
- Confirm that sidebar items vanish when XKit Rewritten is disabled in Firefox.

### DIFF
--- a/src/content_scripts/sidebar.css
+++ b/src/content_scripts/sidebar.css
@@ -1,11 +1,15 @@
 .xkit-sidebar-item {
   width: 100%;
-  margin-bottom: 38px;
+  margin-block: 38px;
 
   font-family: var(--font-family);
   font-size: 1.125rem;
   font-weight: 400;
   line-height: 1.333;
+}
+
+.in-flex-container > .xkit-sidebar-item {
+  margin-block: 5px;
 }
 
 .xkit-sidebar-item > h1 {
@@ -81,7 +85,7 @@
 }
 
 nav .xkit-sidebar-item {
-  margin-bottom: 0.75em;
+  margin-block: 0.75em;
 }
 
 nav .xkit-sidebar-item > h1 {
--- a/src/content_scripts/sidebar.css
+++ b/src/content_scripts/sidebar.css
@@ -1,6 +1,6 @@
 .xkit-sidebar-item {
   width: 100%;
-  margin-block: 38px;
+  margin-bottom: 38px;
 
   font-family: var(--font-family);
   font-size: 1.125rem;
@@ -8,6 +8,9 @@
   line-height: 1.333;
 }
 
+.with-top-margin > .xkit-sidebar-item {
+  margin-top: 38px;
+}
 .in-flex-container > .xkit-sidebar-item {
   margin-block: 5px;
 }
--- a/src/utils/interface.js
+++ b/src/utils/interface.js
@@ -110,6 +110,7 @@ export const displayBlockUnlessDisabledAttr = 'data-xkit-display-block';
 export const displayInlineBlockUnlessDisabledAttr = 'data-xkit-display-inline-block';
 export const displayFlexUnlessDisabledAttr = 'data-xkit-display-flex';
 export const displayInlineFlexUnlessDisabledAttr = 'data-xkit-display-inline-flex';
+export const displayContentsUnlessDisabledAttr = 'data-xkit-display-contents';
 
 /**
  * This variable is set to "unset" in the src/content_scripts/interface.css static stylesheet.
@@ -140,6 +141,9 @@ document.documentElement.append(
     }
     [${displayInlineFlexUnlessDisabledAttr}] {
       display: var(--none-if-xkit-disabled, inline-flex);
+    }
+    [${displayContentsUnlessDisabledAttr}] {
+      display: var(--none-if-xkit-disabled, contents);
     }
   `),
 );
--- a/src/utils/sidebar.js
+++ b/src/utils/sidebar.js
@@ -87,7 +87,7 @@ const mrecContainerSelector = `${keyToCss('mrecContainer')} *`;
 // Sidebar on explore/search/hubs
 const desktopContainerSelector = `aside ${keyToCss('desktopContainer', 'summary')}`;
 // Sidebar on most other desktop pages
-const sidebarItemSelector = keyToCss('sidebarItem', 'sidebarContent');
+const sidebarItemSelector = keyToCss('sidebarItem', 'sidebarContent', 'sidebarScrollableContent');
 // Footer beneath sidebar area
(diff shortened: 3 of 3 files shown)


marcustyphoon:
Ah, wait, this will look different if you have Tumblr Premium. Crap.

marcustyphoon:
Wait... the sidebar utility inserts the items after the first sidebar element in most cases, which will be the ad-free cta if you don't have premium, but will be... I guess the blog recommendations if you do? Which will then be hidden if you choose the relevant option in No Recommended?

Okay, back up: where are these things even supposed to be?

marcustyphoon:
Normally I would wait for a discussion on this, but I'll probably be busy next week, and for "we should ship a Classic Search fix in a timely fashion, and if we're going to do that I may as well try to slip in other fixes relevant to potentially upcoming Tumblr feature pages" reasons (and, perhaps, contribution streak metagaming reasons, ssh) I guess I should probably start thinking through this myself.

Okay, so... the simplest way to explain the intended logic, I assume, is "x list of sibling elements should definitely be above the XKit sidebar, y list of sibling elements should definitely be below it, a sibling element we target but haven't thought of a positioning opinion about should default to [above/below? probably above], and it's undefined behavior if a should-be-below sibling is above a should-be-above sibling, but hopefully that doesn't happen."

(...Hey, wait a second, #2382 could plausibly just be adding the sticky sidebar ad to the list of target options. Anyway.)

Right now, we check the first sibling element and pick whether to be above or below it. That would differ from what I said above if the existing elements are `above-above-below-below`, say.

More concretely, if you don't have premium, currently the dashboard has `premium cta-xkit sidebar-recommended blogs-radar`, making recommended blogs in practice a "below," but if you do, it's `recommended blogs-xkit sidebar-radar`, making it an "above." We could decide it's supposed to be below (add a siblingCandidates filter for it), decide it's supposed to be above (either exclude the ad-free cta or refactor to actually iterate through the candidates and implement something like the above quote), or decide that, actually, the discrepancy is intentional.

And then relevant to this PR: if we decide it's supposed to be **below**, we could put in an override to make 0 top margin on `.first-sidebar-element > .xkit-sidebar-item` (where `first-sidebar-element` is applied to the sidebar if it's `:first-child, ${keyToCss('searchShadow')} + *`), or we could not put top margin on sidebar elements unless [some condition] (or not at all, outside of the sidebar/flex layouts... but, as I mentioned, potentially upcoming Tumblr feature pages). If we decide it's supposed to be **above** or **keep the current behavior**, then... only the second option works; we can't tell if we're supposed to be visually the first sidebar item because No Recommended can hide the recommended blogs element.

Well, that's... multiple options. I'll have to think about which is the most robust and most clear.

AprilSylph:
I'm also busy next week and I'm too tired to discuss anything today.

marcustyphoon:
Cool, yeah. I'm also happy to not ship this one in a timely fashion, too; it's just spacing. I'll keep rubber ducking it, though.

- I guess the simplest way to do the quoted logic is `const lastAboveElement = siblingCandidates.filter(/* */).at(-1); lastAboveElement ? lastAboveElement.after(sidebarItems) : siblingCandidates.at(0).before(sidebarItems);`. That's not too bad. 

- Hmm, would it be helpful to add `keyToCss(searchShadow)` as a sibling candidate... no, not really.

- Does it make sense to say all potentially conditional elements must be below? I don't think so. Certainly the premium cta looks way better as above. The recommended blogs thingy... I personally think looks better as below? I think my rule of thumb design wise is that the "heaviest", most permanent-feeling elements should be on top. Framed elements, in particular. The recommended blogs element used on the dash isn't framed and has those dismiss x buttons, making it feel... secondary? [word that starts with sub- that I'm not thinking of?] to the XKit sidebar elements, whose title underlines connect to the vertical line to the left. Contrast the recommended blogs element in the blog view sidebar, which would definitely feel like an "above" if we put sidebar elements there.
  - Okay, no, the visual weight argument doesn't necessarily hold on explore pages, where I don't really want to put the XKit items below followed tags and recommended blogs. Ehhhh.

marcustyphoon:
For simplicity, I'll table this and make this PR functional without the ordering change, and file an issue to think more about ordering rules.

Edit: Done, and also added a lot more screenshots.

PR #2385, opened 2026-09-09
refactor(Tag Tracking+): Improve loop edge cases, misc cleanup

### LINKED ISSUE #2338: Tag Tracking+: Infinite loop if user tracks zero tags

marcustyphoon:
### Description

Note: every change in this PR is included in/superseded by #2375. It only makes sense to review this if not imminently merging that one (which: extremely reasonable position)... or, in theory, to reduce the diff of that PR, but actually this implementation differs more from it than I hoped, so it doesn't really do that.

This:
- Rewrites the Tag Tracking+ refresh loop so that it can actually be immediately stopped instead of finishing its loop unconditionally, firing up to 100 extra API requests over time if the feature is disabled. Now: because we use `onStorageChanged` in Tag Tracking+, this doesn't happen on preference changes, so it only happens on manual feature disable; yes this is _entirely_ pointless. There might be edge cases where this rewrite also improves loop timing consistency, as it no longer has a setInterval call that's up to 50 minutes in duration, but I can't think of a situation where the old code would actually misbehave. Putting the device to sleep? Nah, should be fine.
- Stops doing anything after adding the sidebar if there are 0 tracked tags. Resolves #2338.
- Renames storageKey to timestampsStorageKey (and types the storage object).

### Screenshots

n/a

### Testing steps

- Optionally, un-revert the test commit.
- Enable Tag Tracking+ with browser dev tools open. Disable it during the initial load waterfall and confirm that API fetches immediately stop.
- Enable Tag Tracking+ with browser dev tools open. Disable it after the initial load waterfall during regular background refreshing and confirm that API fetches immediately stop.
- Smoke test Tag Tracking+.

### DIFF
--- a/src/features/tag_tracking_plus/index.js
+++ b/src/features/tag_tracking_plus/index.js
@@ -83,7 +83,7 @@ const updateSidebarStatus = () => {
 
 let currentRefreshLoop;
 const startRefreshLoop = async () => {
-  const thisRefreshLoop = Date.now();
+  const thisRefreshLoop = Symbol('loop identifier');
   currentRefreshLoop = thisRefreshLoop;
 
   for (const tag of trackedTags) {
--- a/src/features/tag_tracking_plus/index.js
+++ b/src/features/tag_tracking_plus/index.js
@@ -21,6 +21,8 @@ let sidebarItem;
 const refreshCount = async function (tag) {
   if (!trackedTags.includes(tag)) return;
 
+  console.info(`Tag Tracking+: REFRESHING ${tag}`);
+
   let unreadCountString = '⚠️';
 
   try {
@@ -95,7 +97,7 @@ const startRefreshLoop = async () => {
       if (currentRefreshLoop !== thisRefreshLoop) return;
       await Promise.all([
         refreshCount(tag),
-        new Promise(resolve => setTimeout(resolve, 30000)),
+        new Promise(resolve => setTimeout(resolve, 5000)),
       ]);
     }
   }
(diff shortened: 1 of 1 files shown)


src/features/tag_tracking_plus/index.js
```diff
@@ -80,18 +81,26 @@ const updateSidebarStatus = () => {
   }
 };
 
-const refreshAllCounts = async (isFirstRun = false) => {
+let currentRefreshLoop;
+const startRefreshLoop = async () => {
+  const thisRefreshLoop = Date.now();
```
marcustyphoon: This could be a number of things. `= Symbol();` might be the cutest/most semantic, now that I think of it.

Anyway, if anyone knows of a better way to make an interruptible loop like this, I'm all ears.

marcustyphoon: You know what, yeah, I'm doing `Symbol()`. Not the most common piece of javascript, but I don't mind informing a hypothetical new developer of its existence, and it most correctly represents the intent: "value that is unique, and whose purpose is exclusively being unique."

Issue #2156, opened 2026-03-17
Quick Reblog: Already-reblogged checkbox not faded on disabled reblog buttons


marcustyphoon:
### Platform

macOS 14.8.4 arm64

### Browser

Firefox 148.0

### Addon version

master branch

### Details

If you're reblogged a post with Quick Reblog that now has reblogs disabled, the reblog button for that post will be green at a 50% opacity. The checkmark, however, will still be 100% opacity because Tumblr's css is `.jpQn0[aria-disabled="true"] svg  { opacity: .5; }`, targeting the svg rather than the containing button.

PR #2391, opened 2026-09-12
feat(Classic Footer): Always grey out reblog buttons on posts that cannot be reblogged

### LINKED ISSUE #2152: Classic Footer: Potentially grey out reblog buttons on posts that cannot be reblogged?
### Description

With the new Tumblr footer design, a post that cannot be reblogged but that has reblogs must have a clickable reblog button, as that's the only reasonable/primary way for a user to get to the list of reblogs. It will produce a menu that says "This post is not available to be reblogged. [See reblogs]"

This doesn't really make sense with the old paradigm, which Classic Footer reproduces. With the "Turn reblog buttons back into links" option enabled, it _definitely_ doesn't make sense; the reblog button cannot be used to get to the list of reblogs anyway, so why should it on these few posts? Even with that option disabled, it _probably_ doesn't make sense; a Classic Footer user would be more likely to click the note count and then switch to the reblogs tab than they would to click the reblog button and select "See reblogs," right? (This one is a little bit more ambiguous.)

Classic Footer could disable and grey out the reblog button in these cases, communicating more information to the user and matching what I would think most users would expect and the way things used to work in the footer version it's simulating. I think it probably makes the most sense to do this unconditionally?

marcustyphoon:
### Description

> With the new Tumblr footer design, a post that cannot be reblogged but that has reblogs must have a clickable reblog button, as that's the only reasonable/primary way for a user to get to the list of reblogs. It will produce a menu that says "This post is not available to be reblogged. [See reblogs]"

This greys out the reblog button if Classic Footer is enabled and the post in question cannot be reblogged, even if there are reblog notes on it and Tumblr's code does not do this, as discussed in the linked issue.

This doesn't actually prevent the user from clicking the button, even though it applies the "not-allowed" cursor, because I don't actually see any reason to prevent the click. The popup is informative.

Resolves #2152.

Best viewed with whitespace changes hidden.

### Testing steps

Find, or create, a post ([ex](https://www.tumblr.com/blowery/827561533264494592)) with reblogs disabled, but which has reblogs. Enable Classic Footer and confirm that the reblog button appears disabled, including when hovered, but its tooltip still appears. Compare with a post with disabled reblogs with no reblogs ([ex](https://www.tumblr.com/transienttest/827588714668703744/no-reblogs)).

### DIFF
--- a/src/features/classic_footer/index.js
+++ b/src/features/classic_footer/index.js
@@ -7,6 +7,7 @@ import { getPreferences } from '../../utils/preferences.js';
 import { timelineObject, trailItem } from '../../utils/react_props.js';
 
 const activeAttribute = 'data-classic-footer';
+const forceDisableReblogAttribute = 'data-classic-footer-force-disable-reblog';
 const noteCountClass = 'xkit-classic-footer-note-count';
 const modernStyleClass = 'xkit-classic-footer-ds-look';
 const reblogLinkClass = 'xkit-classic-footer-reblog-link';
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Ah, hm, this greys out alreadyReblogged reblog buttons. I guess we probably don't want that. Would have to just to the `pointer-events: none` thing then. (Semi-relevant: #2156.)

Then again, I dunno. Maybe it's cool looking?

PR #2392, opened 2026-09-14
fix(Classic Search): use bare pathname for default mode navigation

### LINKED ISSUE #2386: Classic Search: Feature broken in non-new-tab mode
### Platform

MacOS

### Browser

Firefox 155.0.1

### Addon version

master branch

### Details

Typing something into the search box with Classic Search enabled and pressing enter navigates to a broken page.

AprilSylph:
### Description

- fixes #2386 

`navigate()` can't handle protocol-relative URLs anymore, so let's drop that in the default (same-tab) case.

For the new tab case, using the current `location.origin` as a base makes more sense than hard-coding `//www.tumblr.com`. Or, at least, it's more semantic?

I also added URI component encoding to the tag string to eliminate the possibility of redirects when using spaces or special characters.

### Screenshots

Before | After
-|-
Screenshot 2026-09-14 at 10 51 21 am | Screenshot 2026-09-14 at 10 51 39 am

### Testing steps

1. Load the modified addon as a temporary extension
2. Enable Classic Search, leaving its sole option disabled
3. Open a Tumblr tab and perform a search
    - **Expected result**: The tag hub for your search term is opened in the same Redpop session
4. Enable Classic Search &rarr; "Open tagged pages in a new tab"
5. Perform a Tumblr search again
    - **Expected result**: The tag hub for your search term is opened in a new tab


src/features/classic_search/index.js
```diff
@@ -18,12 +18,12 @@ const replaceSearchForm = function ([searchFormElement]) {
     event.preventDefault();
 
     const query = event.target.querySelector('input').value;
-    const address = `//www.tumblr.com/tagged/${query}?sort=recent`;
+    const pathname = `/tagged/${encodeURIComponent(query)}?sort=recent`;
 
     if (newTab) {
-      window.open(address);
+      window.open(`${location.origin}${pathname}`);
```
marcustyphoon, line 24: Tempted to `new URL()` here.

AprilSylph: I did think about that, but it would be longer:

```js
// Current
window.open(`${location.origin}${pathname}`);

// Implicit casting to string
window.open(`${new URL(pathname, location.origin)}`);

// Explicit casting to string
window.open((new URL(pathname, location.origin)).href);
```

marcustyphoon: True.

(For the record, I don't think you need the double—triple?—parens in that last one. I can see the clarity argument, though I think it's kind of a mixed bag.)

marcustyphoon (approved):

PR #2393, opened 2026-09-14
fix(Classic Search): remove static loading throbber

### LINKED ISSUE #2163: Classic Search: Search bar can get stuck with loading animation
### Platform

macOS 14.8.4 arm64

### Browser

Firefox 148.0

### Addon version

master branch

### Details

It's not the easiest thing to reproduce consistently, but with Classic Search enabled and the correct timing of an interaction with the search bar shortly after loading a tumblr tab, the Classic Search-affected search bar will have a three-element loading animation forever, since we copy it out of the real search bar element and it'll never get removed.

AprilSylph:
### Description

- fixes #2163 

Removes the parent of the `[role="progressbar"]` element in the Classic Search form clone before it's added to the page, since we don't attempt to utilise it in any way.

### Screenshots

Before | After
-|-
Screenshot 2026-09-14 at 11-04-04 test search on Tumblr | Screenshot 2026-09-14 at 11-04-24 test search on Tumblr

### Testing steps

1. Load the modified addon
2. Enable Classic Search
3. Disable Classic Search &rarr; "Open tagged pages in a new tab"
4. Open a Tumblr tab, and search for something too quickly for the typeahead to load
    - **Expected result**: The tag hub for your search term is opened in the same Redpop app session
    - **Expected result**: The search bar _does not_ gain an unmoving knight rider loader element


marcustyphoon:
I can't remember my precise set of rules about css selector cost, but that doesn't look like the cheapest rule in the world. Any reason not to just delete it in js?

AprilSylph:
I guess not?

marcustyphoon (approved):

PR #2398, opened 2026-09-22
feat: recreate icon in Figma

AprilSylph:
### Description

Recreates the extension icon in Figma, which I think is a little more professional than GIMP.

Re-exporting the icon certainly results in fewer weird resizing artifacts than before...

Removes now-unused GIMP assets.

### Screenshots

N/A

### Testing steps

Play spot-the-difference with the image diffs. At a normal viewing distance, you should fail to spot any; zooming in, the new versions should look better.


marcustyphoon:
They can be smaller, if we want.

marcustyphoon (approved):

